PR #34, opened 2026-04-30
fix(data): show table loading immediately

wds824:
This PR fixes a brief blank state when opening table data.

Previously, a data tab was created first, but the loading state was not shown until after column
metadata was loaded and the preview SQL was built. This could leave the result area blank for a
short moment. Now the data tab enters the loading state immediately after it is created.

The data query is also executed against the specific created tab, preventing results from being
written to the wrong tab if the active tab changes during async loading.
  
修复打开表数据时的短暂白屏问题。

之前点击表后会先创建 data tab，但在获取列信息和构建查询 SQL 期间，结果区域没有立即展示加载状态，用
户会看到一段空白。现在打开表后会立即将对应 tab 标记为 loading，结果区域马上显示加载中状态。

同时，数据查询改为绑定到创建出来的指定 tab，避免异步加载期间切换 tab 时结果写入错误 tab。

### DIFF
--- a/src/components/sidebar/TreeItem.vue
+++ b/src/components/sidebar/TreeItem.vue
@@ -152,34 +152,42 @@ function onClick() {
 async function openData() {
   const node = props.node;
   if (!(node.type === "table" || node.type === "view") || !node.connectionId || !node.database) return;
-  await connectionStore.ensureConnected(node.connectionId);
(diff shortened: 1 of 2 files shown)


t8y2 (approved):
感谢提交！loading 状态即时展示和 tab 绑定执行都很到位。

PR #41, opened 2026-04-30
[codex] Add query cancellation controls

SuLea-IT:
## Summary
- add best-effort query cancellation via execution IDs and backend cancellation tokens
- add stop controls in query and data loading views
- track cancellation state per tab and clean up running-query registrations
- fix Node test source imports and add query execution state coverage

## Validation
- node --test tests/*.test.ts
- cargo test --lib
- pnpm build
- git diff --check

## Notes
Cancellation immediately stops the UI wait path. Some blocking drivers such as DuckDB may continue the underlying blocking task until the database operation returns.


t8y2:
感谢贡献～已合并

PR #48, opened 2026-05-01
[codex] Store connection secrets in system keyring

SuLea-IT:
## Summary
- store connection passwords, SSH passwords, and MongoDB connection URLs in the system keyring instead of plaintext JSON
- migrate existing plaintext saved connections into secure storage on load and rewrite sanitized connection files
- surface save/load failures in the UI so keyring errors are not silently ignored

## Validation
- cargo test --lib
- pnpm build

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -356,6 +356,59 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-broadcast"
+version = "0.7.2"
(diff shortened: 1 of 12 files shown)


t8y2 (approved):
Great work! Clean architecture with the `ConnectionSecretStore` trait, solid test coverage, and smooth auto-migration from plaintext. This is an important security improvement. Thanks for the contribution! 🎉

PR #64, opened 2026-05-02
feat: SSH key passphrase, file picker, and DataGrid search autocomplete

yavon007:
## Summary

- **SSH key passphrase**: Added a key passphrase field to the connection dialog for encrypted SSH keys. Passphrase is securely stored in the OS keyring alongside other credentials.
- **SSH key file picker**: Added a browse button (folder icon) next to the SSH key path input, using the native file dialog for quick selection.
- **Connection dialog scrollbar**: Form area now scrolls internally when the window is too short, keeping the header and action buttons always visible.
- **DataGrid search autocomplete**: The search/filter input now suggests `WHERE` when typing `w`/`wh`, and auto-completes column names from the current table after `WHERE`. Smart quote/parenthesis pairing (`'`, `"`, `(`). Dropdown follows cursor position.

### Backend (Rust)
- `models/connection.rs` — new `ssh_key_passphrase` field
- `db/ssh_tunnel.rs` — pass passphrase to `russh::load_secret_key`
- `commands/connection.rs` — thread passphrase through tunnel start
- `commands/connection_secrets.rs` — full keyring lifecycle (save/load/delete/sanitize)

### Frontend
- `ConnectionDialog.vue` — key passphrase input, file picker button, scrollbar, button alignment
- `DataGrid.vue` — WHERE/column autocomplete, quote pairing, cursor-following dropdown
- `types/database.ts` — `ssh_key_passphrase` type
- `i18n/locales/*.ts` — new locale keys (en + zh-CN)


t8y2 (approved):
**LGTM!** Clean implementation of both features — passphrase handling follows existing keyring patterns nicely, and the search autocomplete is a great UX addition.

One minor note for a follow-up: the parenthesis auto-pairing has a small quirk — typing `)` won't skip over an auto-paired `)`, and typing `(` before `)` skips instead of nesting. Quotes work correctly since they're symmetric. Not blocking.

PR #65, opened 2026-05-02
[codex] Add field lineage viewer and review fixes

SuLea-IT:
## Summary

This PR adds a field lineage viewer and folds in the review fixes needed around schema sync, transfer errors, SQL Server table access, and browser smoke testing.

## What this solves

- Adds a column-level field lineage dialog from the table/column flow, showing confirmed foreign-key links, high-confidence references, and possible same-name semantic relations.
- Makes lineage results easier to use with search, filter chips, localized explanations, copy support, and click-through from `table.column` entries directly to the related table tab.
- Replaces Schema Diff's frontend `split(";")` execution path with a backend script execution command that reuses the SQL file splitter, so functions, triggers, comments, and quoted strings with semicolons are not split incorrectly.
- Fixes data transfer terminal handling when a table reports `status: "error"`, so the dialog enters a failed state instead of staying stuck in “transferring”.
- Fixes SQL Server table pagination to honor offsets with `OFFSET ... FETCH NEXT`, including a fallback order column when no primary key is available.
- Fixes SQL Server inline grid edits for schema-qualified tables and bracket-quoted identifiers.
- Adds confirmations for deleting a query-history entry and clearing all query history.
- Escapes Markdown export cells containing pipes or line breaks, preventing broken table exports.
- Guards Tauri-only window/file-drop/update APIs in normal Vite browser runs so frontend smoke tests and browser debugging do not crash the mounted flow.

## Review feedback addressed

- P1: Schema Diff SQL synchronization no longer uses naive semicolon splitting.
- P1: Data Transfer now handles table-level error progress as a terminal failed state.
- P1: SQL Server grid pagination no longer repeats page one.
- P1: SQL Server inline edit SQL now includes schema-aware table names and SQL Server quoting.
- P2: Missing component imports are covered, including `Label` in Schema Diff.
- P2: Query History destructive actions now ask for confirmation.
- P2: Markdown export now escapes `|` and normalizes newlines.
- P2: Browser-only Vite runs are guarded from Tauri API crashes.

## Validation

- `node --test tests/*.test.ts` — 77 tests passed.
- `pnpm build` — passed; Vite reported existing chunk/dynamic import warnings only.
- `cargo test` in `src-tauri` — 43 tests passed.


t8y2 (approved):
**LGTM!** Solid implementation across the board — the field lineage viewer is well-structured with proper batched loading and cancellation, and the bug fixes (SQL splitting, transfer error handling, SQL Server pagination/quoting) are all well-targeted with matching tests.

PR #70, opened 2026-05-03
Fix SQLite UNC path connections

Mukesh-234234:
## Summary
- Open SQLite files with `SqliteConnectOptions::filename(...)`
- Avoid appending `?mode=rwc` directly to UNC/WSL/SMB filesystem paths


t8y2 (commented):
**Review: LGTM with minor suggestions** ✅

Great approach — using `SqliteConnectOptions::filename()` instead of URL string manipulation is the right way to handle UNC/WSL/SMB paths. `create_if_missing(true)` correctly preserves the original `?mode=rwc` semantics.

🟡 Two minor suggestions (non-blocking):

1. `get_or_create_pool` line 96 still uses the old `db::sqlite::connect(&url)` — currently dead code due to the `is_embedded` early return, but worth updating to `connect_path` for consistency and to prevent regressions if that early return is ever removed.

2. The old `pub async fn connect(url: &str)` in `sqlite.rs` has no remaining callers after this PR. Consider removing it to avoid dead code.

PR #71, opened 2026-05-04
UI  and i18n Update

Mukesh-234234:
polished UI, improved settings and connection dialogs, and refreshed i18n support.


t8y2 (commented):
**Review: LGTM with a few suggestions** ✅

Great set of improvements — system theme detection with proper listener cleanup, inline tree errors, and the toolbar status bar are all solid. i18n is complete for both locales.

🟡 Suggestions (non-blocking):

1. **Remove the WSL hint text from the file path field** — The `filePathHint` ("Supports local, WSL, UNC and SMB paths") and the hardcoded `\\wsl.localhost\Ubuntu\home\xxx\data.db` example below the input feel too specific for the connection dialog. The placeholder already conveys the supported formats.

2. **Add feedback when copying file path** — The "Copy file path" button in the connection dialog has no visual feedback after clicking. Consider showing a toast or briefly swapping the icon to a checkmark (similar to how `copyName` works elsewhere).

3. **Compact mode: grid cell vertical alignment** — In compact density, the data grid rows appear top-aligned. Could you check if adding `align-middle` or vertical centering to the cell styles resolves this?

4. **Consider moving the connection/database badges to the query tab bar** — The toolbar center badges showing the current connection and database are useful, but they might work better placed in the tab bar area (above the editor, next to the connection/database selectors). The toolbar already has the connection name visible in the sidebar header, but the database name is missing there — having both as badges in the tab bar would give better context right where the user is editing SQL.

5. **Compact mode CSS scope** — The `.compact` overrides in `globals.css` target generic Tailwind utilities (`.h-10`, `.h-9`, `.py-1`, `.text-sm`) globally. This works for now, but if side effects appear in dialogs or other components, consider scoping the overrides to specific containers.

6. **Hardcoded light theme list** — `syncEditorThemeWithApp` maintains a manual `lightThemes` Set. Consider deriving it from the `EDITOR_THEMES` array's `dark` property so it stays in sync when new themes are added.

Thanks for the contribution!

PR #78, opened 2026-05-04
[codex] Add database-wide search

SuLea-IT:
## Summary

Adds a controlled database-wide search flow for exploring unfamiliar schemas without first knowing which table contains a value.

## What changed

- Adds a database/schema context-menu entry: **Search Database**.
- Adds a search dialog that scans regular tables in the selected database or schema.
- Skips binary/system-heavy targets by default and caps each table's returned rows to avoid loading too much data at once.
- Shows scan progress, matched result count, partial table errors, and a stop button.
- Builds database-specific search SQL for MySQL, PostgreSQL, SQLite, SQL Server, Oracle-style, ClickHouse, DuckDB, and Redshift-compatible paths.
- Lets users click a result to open the matching table with a generated `WHERE` filter based on primary keys or matched columns.

## Why

When users inspect an unfamiliar database, they often only know an ID, email, order number, or other value, but not the table name. This gives them a bounded search path while avoiding an uncontrolled full database dump.

## Issue

Closes #76

## Validation

- `node --test tests/*.test.ts` — 97 tests passed.
- `pnpm build` — passed; Vite reported existing chunk/dynamic import warnings only.
- `git diff --check` — passed.

PR #77, opened 2026-05-04
[codex] Improve connection selector and table clicks

SuLea-IT:
## Summary

This PR improves two small but high-friction schema browsing flows.

## What changed

- Adds database type icons to the main query connection selector, including the selected connection state.
- Shows a compact connection subtitle in the selector, such as `TiDB · 127.0.0.1:4000`, so duplicate connection names are easier to distinguish.
- Reuses the same driver-profile icon identity used elsewhere in DBX, so MySQL-compatible profiles like TiDB, MariaDB, Doris, and SelectDB keep their own visual identity.
- Changes table and view row clicks in the sidebar to open table data only.
- Keeps table structure expansion available from the dedicated chevron button, matching the expected “click table to view data, click arrow to expand structure” workflow.

## Issues

Closes #74
Closes #75

## Validation

- `node --test tests/*.test.ts` — 99 tests passed.
- `pnpm build` — passed; Vite reported existing chunk/dynamic import warnings only.
- `git diff --check` — passed.

PR #80, opened 2026-05-04
Refine compact mode, copy file path feedback, and editor theme sync

Mukesh-234234:
Scoped compact density overrides to the app container to avoid global Tailwind utility side effects
Added visual feedback for the SQLite/DuckDB file path copy button using a temporary checkmark state
Improved data grid cell vertical centering in compact mode
Moved connection/database context badges into the query tab bar for better editor visibility
Replaced hardcoded editor light-theme list with dynamic classification


t8y2 (changes requested):
**Merge conflict** needs to be resolved — conflicts with recently merged PRs (#77, #78).

Good ideas here — the system theme detection, tree error display, and copy path feedback are solid. A few issues to address:

🔴 **Compact mode CSS** (globals.css): Overriding `.h-10`, `.h-9`, `.py-1`, `.text-sm` globally under `.dbx-compact` will break layouts in dialogs, context menus, tooltips, and the data grid. These are Tailwind utility classes used everywhere — overriding them globally is fragile. Consider using scoped CSS variables or targeting specific components instead.

🔴 **Toolbar spacer** (`flex-1 md:hidden`): This removes the spacer on desktop, causing right-side buttons to bunch up to the left.

🟡 **Theme toggle**: `toggleTheme()` only cycles light ↔ dark, with no way to return to "system" mode from the toolbar. Consider cycling system → light → dark.

🟡 **Hardcoded fallback themes**: `syncEditorThemeWithApp` still hardcodes `"one-dark"` and `"vscode-light"` as targets despite the dynamic classification.

PR #81, opened 2026-05-04
feat: improve tab bar scrolling and sidebar interaction

yavon007:
- Add left/right scroll buttons with gradient overlay for tab overflow
- Sync scroll button visibility on resize and tab changes
- Use invisible pin button to prevent tab width jump on hover
- Simplify tab layout by removing mode icon, reordering badge
- Change sidebar table/view to open on double-click instead of single-click


t8y2 (changes requested):
**Merge conflict** needs to be resolved first — this branch conflicts with recently merged PRs (#77, #78).

The tab scroll buttons look great! Nice touch with the gradient overlays and auto-scroll to active tab.

Regarding the double-click change for table/view: we intentionally shipped the current UX in PR #77 — "click row → open data, click chevron → expand structure". I'd prefer to keep that pattern. Could you revert the `treeNodeClick.ts` and `onDblClick` changes, and focus this PR on the tab scrolling improvements?

One more suggestion: while you're working on the tab bar, would be great to also improve the visual distinction between active and inactive tabs — currently the difference is quite subtle (just `bg-background` + `font-medium`). Maybe a bottom border accent, different text opacity, or similar treatment would help.

t8y2 (approved):
LGTM! All feedback addressed — double-click reverted, conflicts resolved, and the active tab bottom border accent looks great. Clean, focused PR. Thanks for the quick turnaround!

PR #94, opened 2026-05-05
Formatted

Mukesh-234234:


t8y2 (changes requested):
Thanks for formatting `crates/dbx-core`! The `cargo fmt` cleanup looks good.

One thing to fix before merging — the new pre-commit hook line `cd ../crates/dbx-core` relies on the previous `cd src-tauri` having succeeded. If that `cd` fails (due to the `&&` chain), the relative path would resolve incorrectly. Please use an absolute path instead:
```bash
cd "$(git rev-parse --show-toplevel)/crates/dbx-core" && cargo fmt --all && git add -u .
```

t8y2 (changes requested):
The `.husky/pre-commit` update needs a fix — it currently:
1. Removes the `src-tauri` formatting line
2. Formats `crates/dbx-core` twice (absolute path + old relative path)

The intended result should be:
```bash
npx lint-staged
cd src-tauri && cargo fmt && git add -u .
cd "$(git rev-parse --show-toplevel)/crates/dbx-core" && cargo fmt --all && git add -u .
```

Please keep the original `cd src-tauri` line and replace the relative `cd ../crates/dbx-core` with the absolute path version. Thanks!

t8y2 (approved):
LGTM! The pre-commit hook is now correct — `src-tauri` formatting restored and `dbx-core` uses an absolute path. Thanks for the quick fix!

PR #95, opened 2026-05-05
feat: optimize connection error alerts

wu-clan:
Optimize warnings and alerts when connection fails, for sidebar, new query, AI dialogs

- Add warnings and bubble alerts to the right side of the sidebar connection, support clearing


t8y2 (approved):
Clean implementation — error tracking in the store is well-structured, the UX improvement of creating tabs before connecting is a nice touch, and the `ConnectionErrorIndicator` popover works well. LGTM!

PR #104, opened 2026-05-06
Issue #101 fix

Mukesh-234234:


t8y2 (commented):
🟡 A few suggestions:

1. **May not fix the root cause of #101**: The `buffer underflow` error likely occurs during Oracle protocol response parsing in `conn.query()`, not in the `get_string()`/`get_i64()` accessors on already-parsed rows. These accessors safely return `Option` without any buffer operations (see [oracle-rs Row.get_string](https://github.com/nickolaev/oracle-rs/blob/main/src/row.rs#L391)). Has this been tested against an Oracle instance that reproduces #101?

2. **Behavioral change - silent data dropping**: Switching from `unwrap_or("")` to `filter_map` + `?` means rows with any NULL column are now silently dropped instead of using defaults. While Oracle system views shouldn't normally have NULLs, this is a semantic change worth noting.

3. **Inconsistency in `get_columns`**: The `get_columns` function (line 97-126) still uses the `unwrap_or("")` pattern. Consider updating it for consistency.

The code quality improvement itself is valid — `filter_map` + `?` is more idiomatic Rust. 👍

t8y2:
🔴 Does not fix the root cause
The buffer underflow error occurs inside conn.query() during Oracle protocol parsing — before the code reaches get_string()/get_i64() row accessors. This PR only:

Wraps the error message with friendlier text (the user still gets an error)
Changes map → filter_map in get_columns (unreachable when conn.query() already fails)
🔴 Silent data loss
Switching from unwrap_or("") to filter_map + ? means rows with any NULL column are silently dropped instead of using defaults. This is a behavioral change that could hide data.

The buffer underflow issue likely requires investigating oracle_rs compatibility with the specific Oracle server version/configuration. Thanks for the effort though!

PR #112, opened 2026-05-06
feat: add check fmt and lint commands

wu-clan:
Add some useful commands.


t8y2 (commented):
LGTM! The scripts are clean and the `oxlint` choice aligns well with the existing `oxfmt` toolchain.

Minor note: the regex change in `fieldLineage.ts` (unnecessary escape removal) is likely an auto-fix from running oxlint — it's functionally equivalent, just worth mentioning in the description.

Thanks for the contribution! 🙌

PR #213, opened 2026-05-11
i18n for redis key browser;

xKrah:


t8y2 (approved):
### 🟡 Suggestions

1. **Missing Spanish (es.ts) translations** — The PR adds `columnKey`, `columnType`, `columnValue`, `columnSize`, `columnTTL` to `en.ts` and `zh-CN.ts` but not to `es.ts`. Falls back to English gracefully, so not blocking — can be added post-merge.

### ✅ Well Done

1. **Correct fix** — The Redis key browser had hardcoded Chinese column headers ("键", "类型", "值", "大小") that wouldn't translate. Replacing with i18n keys is the right approach.

2. **Clean, minimal diff** — Only 3 files changed, all directly related to the fix. No unrelated changes.

3. **Both main locales covered** — English and Chinese translations are complete and accurate.

LGTM! 🎉

PR #217, opened 2026-05-11
fix(grid): prevent cell value truncation in detail panel and copy/export

Abeautifulsnow:
## Summary

- Cell Details panel displayed truncated values (256-char limit) because it used `formatCell()` which is designed for compact grid cell rendering
- Copy Value in the detail panel and all export/copy operations (`copyCell`, `copyAll`, `exportCsv`) also truncated long values via the same `formatCell()` path
- Extracted `displayCellValue()` as a shared utility that returns full values without truncation, and replaced `formatCell` in all display, copy, and export contexts where truncation is undesirable

## Changes

- **New** `src/lib/cellValue.ts` — shared `CellValue` type and `displayCellValue()` function (no truncation)
- **Modified** `DataGrid.vue` — detail panel display and Copy Value now use `displayCellValue` instead of `formatCell`; removed duplicate local `CellValue` type definition
- **Modified** `useDataGridExport.ts` — `copyCell`, `copyAll`, `exportCsv` now use `displayCellValue`; removed `formatCell` from the options interface (no longer needed)
- **New** `tests/cellValue.test.ts` — 9 test cases covering null, boolean, number, string, long strings, complex JSON strings, unicode, and edge values

## Test plan

- [x] All 9 unit tests for `displayCellValue` pass (`npx tsx --test tests/cellValue.test.ts`)
- [x] TypeScript compilation passes (`vue-tsc --noEmit`)
- [x] Manual: open a row with a value longer than 256 characters, verify Cell Details panel shows the full value
- [x] Manual: click Copy Value in Cell Details panel, verify clipboard contains the full value (not truncated)
- [x] Manual: right-click a cell with a long value → Copy Cell, verify clipboard contains the full value
- [x] Manual: Copy All / Export CSV, verify long values are not truncated

### DIFF
--- a/src/components/grid/DataGrid.vue
+++ b/src/components/grid/DataGrid.vue
@@ -62,6 +62,7 @@ import * as api from "@/lib/api";
 import { buildTableSelectSql, quoteTableIdentifier } from "@/lib/tableSelectSql";
 import { formatGridSqlLiteral } from "@/lib/dataGridSql";
 import { matchesRowStatusFilter, type RowStatus, type RowStatusFilter } from "@/lib/gridRowStatus";
+import { displayCellValue, type CellValue } from "@/lib/cellValue";
 
 import { useToast } from "@/composables/useToast";
 import { useDataGridExport } from "@/composables/useDataGridExport";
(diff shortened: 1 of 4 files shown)


t8y2:
ci failed

PR #219, opened 2026-05-11
Add support new types for postgresql

xKrah:
Added support for integer arrays
Added support for float4

Int Array before PR:
{6DBE540A-DBCB-44DA-A908-D7914434DFBB}

float4 before PR:
{B1581A60-CFE7-48D2-9B4C-6D242DF5A03B}


t8y2 (commented):
### 🟡 Suggestions

1. **f32 precision rounding is hardcoded to 6 decimal places**
   The `f32` handling uses `(v as f64 * 1_000_000.0).round() / 1_000_000.0`. This always rounds to 6 decimal places regardless of value magnitude. While reasonable for display purposes (float4 has ~7 significant digits), this introduces a lossy transformation that could surprise users working with very small values (e.g., `0.00000012` becomes `0.0`). Consider documenting this behavior or using a format-based approach instead.

2. **Missing float array types**
   Integer arrays (`Vec<i8>`, `Vec<i16>`, `Vec<i32>`, `Vec<i64>`) are supported, but float arrays (`float4[]` / `float8[]` → `Vec<f32>`, `Vec<f64>`) are not included. For consistency, these could be added in the same PR.

3. **Unrelated commit in history**
   The commit `"i18n for redis key browser;"` appears in this PR's history. While it doesn't affect the diff (already in main), a cleaner PR history would help reviewers.

### ✅ Well Done

- **Focused scope**: The change is minimal and targeted — only touches the type conversion function that needed extension.
- **Follows existing patterns**: The new `or_else` chains match the style of the existing code perfectly.
- **Correct ordering**: New types are inserted in a logical position in the fallback chain (scalars → arrays → floats → bool → uuid).
- **Screenshots in PR body**: Including before-state screenshots makes it easy to understand the motivation.

PR #256, opened 2026-05-13
perf(redis): speed up key browser scans

yavon007:
## Summary
speed up key browser scans

Refs #255

### DIFF
--- a/crates/dbx-core/src/db/redis_driver.rs
+++ b/crates/dbx-core/src/db/redis_driver.rs
@@ -276,23 +276,25 @@ pub async fn scan_keys_page(
         .map_err(|e| e.to_string())?;
 
     let (next_cursor, keys) = parse_scan_keys(raw)?;
+    if keys.is_empty() {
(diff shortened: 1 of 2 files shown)


t8y2 (approved):
### ✅ Well Done

- **Pipeline batching** — Replaces N sequential `TYPE` round-trips with a single `redis::pipe()`, which is the correct Redis optimization for bulk key metadata.
- **Deferred detail loading** — Removing per-key TTL / size / preview fetches from the scan path is a meaningful latency reduction, especially for large keyspaces. The `get_value` path still provides full details on selection.
- **Clean removal** — `fetch_key_preview` was only called from `scan_keys_page`; deleting it entirely avoids dead code.
- **Frontend consistency** — `formatTtl` and updated `formatSize` correctly render placeholder values as empty strings rather than misleading zeros.
- **`Vec::with_capacity`** and early return on empty keys are nice touches.

### Verdict

Clean performance optimization with no correctness or safety regressions. LGTM 👍

t8y2:
👍

PR #269, opened 2026-05-14
feat(redis): add key statistics tracking for database browser

Abeautifulsnow:
## Background Description  
Add key statistics functionality to the Redis database browser to display the number of loaded keys / total number of keys,  
and update it in real-time during dynamic loading, deletion, and flushing.

## Modifications  
- Add a `RedisDatabaseInfo` struct (db + keys) to replace the original `Vec<u32>`  
- `list_databases` returns the key count for each db (parsing `INFO keyspace`)  
- `scan_keys_page` adds a `DBSIZE` call to obtain `total_keys`  
- The frontend `TreeNode` adds `loadedKeyCount` / `totalKeyCount` fields  
- The tree node label dynamically displays the `db0 (loaded/total)` format  
- Add `updateRedisDbKeyStats` to update statistics during loading/deletion/flushing

### DIFF
--- a/crates/dbx-core/src/db/redis_driver.rs
+++ b/crates/dbx-core/src/db/redis_driver.rs
@@ -7,6 +7,12 @@ const COLLECTION_PAGE_SIZE: usize = 200;
 const DEFAULT_REDIS_DATABASES: u32 = 16;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RedisDatabaseInfo {
+    pub db: u32,
+    pub keys: u64,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct RedisKeyInfo {
     pub key_display: String,
     pub key_raw: String,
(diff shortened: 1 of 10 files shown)


t8y2 (approved):
### ✅ Well Done

- **Clean architecture**: Backend returns `RedisDatabaseInfo { db, keys }` from `INFO keyspace` parsing, and `total_keys` from `DBSIZE` — both O(1) Redis commands, so no performance concern.
- **Smart delta-based updates**: `updateRedisDbKeyStats` supports `loaded`, `total`, and `totalDelta` modes, with `Math.max(0, ...)` to prevent negative counts. This elegantly handles delete (single and bulk) and flush scenarios without re-querying Redis.
- **Correct search mode handling**: When in search mode, `loaded` is passed as `undefined` so `loadedKeyCount` isn't corrupted by filtered results.
- **Type consistency across the stack**: `RedisDatabaseInfo` is defined in Rust, mirrored in TypeScript (`tauri.ts`), and properly re-exported through the API layer.
- **`redisDbLabel` helper** is clean and handles the `totalKeyCount == null` edge case gracefully.
- **Backward compatible**: `loadedKeyCount` and `totalKeyCount` are optional on `TreeNode`, so existing non-Redis tree nodes are unaffected.

LGTM! 🎉

PR #341, opened 2026-05-19
fix(mysql): avoid metadata collation conflicts

yavon007:
## Summary

MariaDB exposes some `information_schema` metadata fields with the `utf8`/`utf8mb3` character set. Joining multiple `information_schema` tables can therefore trigger collation compatibility errors when other metadata fields use different collations. This PR avoids those joins by fetching column metadata and primary key metadata separately.

### DIFF
--- a/crates/dbx-core/src/db/mysql.rs
+++ b/crates/dbx-core/src/db/mysql.rs
@@ -3,6 +3,7 @@ use futures::StreamExt;
 use rust_decimal::Decimal;
 use sqlx::mysql::{MySqlPool, MySqlPoolOptions, MySqlRow};
 use sqlx::{Column, Executor, Row, TypeInfo, ValueRef};
+use std::collections::HashSet;
 use std::time::{Duration, Instant};
 
 use crate::sql::starts_with_executable_sql_keyword;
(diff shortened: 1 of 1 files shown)


t8y2 (commented):
### 🟡 Suggestions

1. **Two sequential queries could run in parallel** — `pk_sql` and `columns_sql` are independent. Using `tokio::join!` (or `futures::try_join!`) would fetch both concurrently. This is a minor optimization — metadata queries are fast — but it's an easy win since the two queries share no dependency.

### ✅ Well Done

- **Root-cause fix, not a workaround**: Instead of adding `COLLATE utf8mb4_general_ci` on every join column, the fix eliminates the cross-table join entirely. This avoids collation conflicts across MariaDB, OceanBase, and any other MySQL-compatible DB with non-standard `information_schema` collations.
- **Clean decomposition**: Extracting `columns_sql()` and `primary_key_columns_sql()` as standalone functions makes the SQL generation unit-testable and readable.
- **Efficient PK lookup**: Using `HashSet<String>` for the primary key check is the right data structure — O(1) per column vs. the previous SQL-level join.
- **Good test coverage**: Two tests verify both SQL fragments avoid `COLLATE` and target the correct `information_schema` tables.

### Verdict

**Can merge.** Clean, well-tested, and solves a real compatibility issue.

PR #370, opened 2026-05-21
feat: Persist MongoDB browser view mode preference

baizhi958216:
每一次切换表之后需要重复选择视图模式，需要持久化Mongodb浏览视图

### DIFF
--- a/apps/desktop/src/components/mongo/MongoDocBrowser.vue
+++ b/apps/desktop/src/components/mongo/MongoDocBrowser.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { computed, ref, onMounted } from "vue";
+import { computed, ref, onMounted, watch } from "vue";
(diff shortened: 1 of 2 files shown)


t8y2 (approved):
### ✅ Well Done

- **Follows existing patterns**: The persistence uses the same `settingsStore` + `EditorSettings` pattern as other persisted preferences (`appLayout`, `pageSize`, etc.). Clean integration.
- **Proper normalization**: `normalizeEditorSettings` validates the stored value — only `"table"` is accepted as an alternative; anything else falls back to the default `"document"`. This guards against corrupted or stale localStorage values.
- **Computed get/set**: Using a writable `computed` to bridge the component's `viewMode` to the store is the idiomatic Vue 3 approach — the template code stays untouched, and persistence is transparent.
- **Minimal, focused diff**: Only the necessary changes, no unrelated modifications.

LGTM 👍

PR #404, opened 2026-05-22
Reduce high-density DOM memory

SuLea-IT:
## Summary

- Virtualize large Redis collection and stream value rows to avoid mounting every item at once.
- Mount the grid cell detail action only for the hovered cell or the cell with an open detail panel.
- Add source checks plus `docs/performance/dom-density-audit.md` with production-build measurement data.

## Measurement

Baseline: `upstream/main@6c83b10` vs this branch. Both sides were production Vite builds. Each scenario ran in a fresh Chrome process and reports the median of 3 samples after forced garbage collection.

| Scenario | RSS Delta MiB | JS Heap Delta MiB | DOM Elements Delta |
| --- | ---: | ---: | ---: |
| Grid 1000 x 20 | -12.1 | -3.0 | -5100 |
| Grid 1000 x 200 | -62.8 | -29.2 | -51000 |
| Redis list value 3000 | -211.2 | -71.4 | -56197 |
| Redis hash value 3000 | -218.5 | -72.2 | -56197 |
| Redis stream 1000 x 5 | -262.2 | -79.2 | -61101 |

## Test Plan

- `pnpm check`
- Production build used for measurement: `pnpm exec vue-tsc --noEmit --project apps/desktop/tsconfig.json && pnpm exec vite build --config apps/desktop/vite.config.ts --outDir /tmp/dbx-dom-current-dist --emptyOutDir`

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -286,6 +286,7 @@ function typeColorClass(t: string): string {
 }
 const contextCell = ref<{ rowId: number; rowIndex: number; col: number } | null>(null);
 const detailCell = ref<{ rowIndex: number; col: number } | null>(null);
+const hoveredDetailCell = ref<{ rowIndex: number; col: number } | null>(null);
 const showCellDetail = ref(false);
 const activeCellDetailTab = ref<CellDetailTab>(defaultCellDetailTab());
 const detailWidth = ref(320);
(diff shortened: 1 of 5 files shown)


t8y2 (approved):
### ✅ Well Done

1. **Grid cell detail button: from CSS toggle to conditional mount**. The old approach rendered a hidden `<button>` in every single cell and relied on `group-hover/cell:flex` to show it. With 1000x200 grids that's 200,000 hidden buttons in the DOM. The new `v-if="cellDetailButtonVisible(index, actualColIdx)"` mounts only one button at a time — the hovered cell or the cell whose detail panel is open. `onCellMouseenter` correctly delegates to the original `extendCellSelection`, preserving selection behavior.

2. **Correct scroller choice per data shape**. `RecycleScroller` (fixed row height) for list/set/hash/sorted-set collections, `DynamicScroller` (variable height) for stream entries whose field count varies. `:size-dependencies="[streamFieldCount(row)]"` ensures re-measurement when field count changes.

3. **Detail panel visibility preserved**. `cellDetailButtonVisible` has a second branch ensuring the button stays mounted when the detail panel is open, even if the mouse leaves the cell.

4. **"Load more" button placement**. Moved into `#after` slot of each scroller — cleaner than managing it outside the scroll container.

5. **Thorough measurement documentation** with methodology, baseline commit, and before/after numbers.

6. **Regression tests** guard that DataGrid uses conditional mounting and Redis viewer uses virtualization.

LGTM — well-measured and cleanly implemented. 👍

PR #452, opened 2026-05-26
feat: add configurable connect_timeout_secs for all database drivers

adntian:
Previously the connection timeout was hardcoded to 5 seconds and only MySQL supported overriding it via URL parameters. This change:

- Adds a `connect_timeout_secs` field to ConnectionConfig (default 5s)
- All drivers now accept a timeout parameter instead of using the hardcoded value
- URL parameters (connect_timeout/connectTimeout/connection_timeout/connectionTimeout) take priority over the field value when present
- Adds a "Connection Timeout (seconds)" input to the connection dialog (i18n: zh/en/es)
- Fixes MySQL rejecting `connection_timeout` URL parameter as unknown

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,6 +1,6 @@
-name: 问题反馈
-description: 程序崩溃、连接失败、查询异常、界面显示错误或其他可复现问题。
-title: "[Bug] "
+name: "🐞 Bug Report / 问题反馈"
+description: Crashes, connection failures, query errors, display bugs, or other reproducible issues. / 程序崩溃、连接失败、查询异常、界面显示错误或其他可复现问题。
+title: "[🐞 Bug] "
 labels: ["bug"]
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,5 @@
 blank_issues_enabled: false
 contact_links:
-  - name: 文档 / 使用说明
-    url: https://github.com/t8y2/dbx#readme
-    about: 提交 issue 前可先查看 README 中的安装、开发和 macOS 首次启动说明。
+  - name: Documentation / 文档
+    url: https://dbxio.com/cn/docs/what-is-dbx
+    about: Check the docs before submitting an issue. / 提交 issue 前可先查看文档。
--- a/.github/ISSUE_TEMPLATE/database_compatibility.yml
+++ b/.github/ISSUE_TEMPLATE/database_compatibility.yml
@@ -1,6 +1,6 @@
-name: 数据库兼容性问题
-description: 某个数据库、版本或 SQL 方言不兼容。
-title: "[Compatibility] "
+name: "🔀 Compatibility Issue / 数据库兼容性问题"
+description: A specific database, version, or SQL dialect incompatibility. / 某个数据库、版本或 SQL 方言不兼容。
+title: "[🔀 Compatibility] "
 labels: ["bug"]
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,6 +1,6 @@
-name: 功能建议
-description: 提议新功能、交互改进或现有功能增强。
-title: "[Feature] "
+name: "💡 Feature Request / 功能建议"
+description: Propose a new feature, interaction improvement, or enhancement. / 提议新功能、交互改进或现有功能增强。
+title: "[💡 Feature] "
 labels: ["enhancement"]
 body:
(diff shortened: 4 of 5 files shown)


t8y2 (changes requested):
## PR #452: feat: add configurable connect_timeout_secs for all database drivers

**+170/-81, 23 files | CI: rust failing; frontend/jdbc/notify passing | Mergeable: MERGEABLE**

### 🔴 Must Fix

1. **Rust CI is failing.** The `rust` job fails in `cargo test --workspace --locked` with two compile errors introduced by the PR:
   - `crates/dbx-core/src/models/connection.rs`: test module calls `default_connect_timeout_secs()` without importing it.
   - `crates/dbx-core/src/db/mysql.rs`: an existing test still calls `super::connect(url)` after the function signature changed to `connect(url, fallback_timeout)`.
   
   This needs to be fixed before merge.

2. **The new timeout does not cover the TCP probe path.** `AppState::get_or_create_pool`, `connect_db`, and `test_connection` still call `probe_connection_endpoint(...)` before computing/passing the new timeout, and `probe_tcp_endpoint` still uses the fixed `TCP_PROBE_TIMEOUT_SECS = 3`. For SSH/proxy endpoints and direct IP endpoints, a user setting `connect_timeout_secs = 30` can still fail after 3 seconds before the driver timeout is ever used. That makes the new setting unreliable for exactly the slow-network cases it is meant to solve.

3. **URL timeout override is only actually honored by MySQL/PostgreSQL.** The PR says `connect_timeout`, `connectTimeout`, `connection_timeout`, and `connectionTimeout` take priority when present, but most drivers receive only `config.effective_connect_timeout_secs()` from the form field. For example ClickHouse, MongoDB, SQL Server, Redis, and Elasticsearch do not parse URL params for timeout in their driver paths. If URL override is part of the feature contract, it should be centralized before dispatch or implemented consistently per driver.

### 🟡 Suggestions

1. Clamp `connect_timeout_secs` in the backend model, not only in the desktop form. The UI input has `max="300"`, but persisted/imported/API configs can still carry arbitrary positive values. `effective_connect_timeout_secs()` should probably enforce the same `1..=300` range used by URL params.

2. Add focused tests for timeout precedence and probe behavior:
   - field default/fallback for legacy configs
   - URL param overriding the field
   - `connection_timeout` stripping from MySQL URLs
   - TCP probe using the effective timeout

### ✅ Well Done

- The direction is good: moving connection timeout into `ConnectionConfig` is cleaner than MySQL-only URL handling.
- The UI/i18n coverage is included for zh/en/es.
- The MySQL URL cleanup for `connection_timeout` avoids passing unsupported params into `mysql_async`.

### Verdict

Request changes. The feature is useful, but it currently fails Rust CI and has a real behavioral gap where the fixed 3-second TCP probe can override the user’s configured timeout.

t8y2:
可以根据建议在完善一下 刚才发了v0.5.21. 预计在v0.5.22中包含

adntian:
> 可以根据建议在完善一下 刚才发了v0.5.21. 预计在v0.5.22中包含

已经完善

t8y2 (changes requested):
## PR #452: feat: add configurable connect_timeout_secs for all database drivers

**+235/-91, 23 files | CI: rust/frontend/jdbc passing | Mergeable: CONFLICTING**

### 🔴 Must Fix

1. **The PR currently cannot merge.** GitHub reports `CONFLICTING`, and local `git merge-tree` confirms real conflicts in `ConnectionDialog.vue`, i18n files, `connectionStore.ts`, `database.ts`, and especially `crates/dbx-core/src/connection.rs`. The backend conflict overlaps with newer `main` changes for MySQL CA cert handling and Redis Sentinel, so the resolution needs to preserve both those paths and the new timeout plumbing.

2. **URL timeout priority is still incomplete on some paths.** In `src-tauri/src/commands/connection.rs`, `test_connection` parses URL timeout only after `probe_connection_endpoint`, so the TCP probe still uses only `connect_timeout_secs`. `connect_db` also sets `connect_timeout` directly from `effective_connect_timeout_secs()` instead of `parse_connect_timeout_with_fallback`, so URL params like `connect_timeout=30` are not honored for Redis/Mongo/ClickHouse/SQL Server/Elasticsearch in that path. This conflicts with the PR contract that URL timeout params take priority.

### 🟡 Suggestions

1. Add a focused regression test for URL timeout precedence through the Tauri `connect_db`/probe path, not only the lower-level parser. That would catch this exact split-path behavior.

### ✅ Well Done

1. The previous Rust compile failures are fixed, and current PR CI is green.

2. The model-level default/clamping is a good improvement: legacy configs get 5s, zero falls back, and large values clamp to 300.

3. The timeout handling is much more consistent across native drivers than the original version, especially for MongoDB, Redis, ClickHouse, SQL Server, and Elasticsearch.

### Verdict

Request changes. The feature direction is good and CI is green on the PR branch, but it needs a rebase/conflict resolution and one more pass to make URL timeout precedence consistent across probe and `connect_db` paths.

PR #469, opened 2026-05-27
fix: allow mysql timestamp default null ddl

vvrfxyz:
## Summary
- Enable MySQL `explicit_defaults_for_timestamp` for DDL statements that define `TIMESTAMP ... DEFAULT NULL` columns.
- Prevent valid MySQL `CREATE TABLE` / `ALTER TABLE` statements from failing with `ERROR 42000 (1067): Invalid default value` on older/session-strict timestamp defaults.
- Add focused coverage for the referral-record style DDL detection.

## Test plan
- `git diff --check -- crates/dbx-core/src/db/mysql.rs`
- `cd /tmp/dbx-src && pnpm test`

Note: Rust checks were not run locally because `cargo`/`rustc` are not available in this environment.

### DIFF
--- a/crates/dbx-core/src/db/mysql.rs
+++ b/crates/dbx-core/src/db/mysql.rs
@@ -273,18 +273,12 @@ fn explicit_timestamp_defaults_sql(enabled: bool) -> &'static str {
     }
 }
 
-async fn enable_explicit_timestamp_defaults_for_query(
-    conn: &mut mysql_async::Conn,
-    sql: &str,
-) -> Option<bool> {
+async fn enable_explicit_timestamp_defaults_for_query(conn: &mut mysql_async::Conn, sql: &str) -> Option<bool> {
     if !should_enable_explicit_timestamp_defaults(sql) {
         return None;
     }
 
-    let previous = match conn
-        .query_first::<u8, _>("SELECT @@SESSION.explicit_defaults_for_timestamp")
-        .await
-    {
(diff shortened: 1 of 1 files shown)


t8y2 (changes requested):
## PR #469: fix: allow mysql timestamp default null ddl

**+38/-0, 1 file | CI: pass | Mergeable: MERGEABLE**

### 🔴 Must Fix

- `crates/dbx-core/src/db/mysql.rs:681`: `SET SESSION explicit_defaults_for_timestamp = ON` is applied before the matching DDL, but it is never restored. Because the MySQL pool is constrained to a single reusable connection, one `CREATE/ALTER ... TIMESTAMP ... DEFAULT NULL` statement can permanently change the semantics of later user DDL on that connection. For example, a later `CREATE TABLE t (created_at TIMESTAMP)` may no longer get legacy implicit timestamp behavior. The fix should either restore the previous session value after the query or make the behavior an explicit connection-level policy rather than a per-query detector.

### ✅ Well Done

- The change is narrowly scoped and catches the reported `TIMESTAMP(6) DEFAULT NULL` DDL shape.
- CI is green: `frontend`, `jdbc`, `notify`, and `rust` all passed.
- I also verified the new focused test in a detached PR worktree: `cargo test -p dbx-core mysql_timestamp_default_null_ddl_enables_explicit_defaults` passed.

### Verdict

Request changes. The compatibility intent is good, but the session setting leak can silently alter subsequent MySQL DDL behavior.

PR #520, opened 2026-05-29
Fix/issue 515 UI scale settings dialog

Michaelxwb:
image


t8y2 (approved):
### ✅ Well Done

1. **Precise fix**: The root cause is correctly identified — `sm:min-h-[520px]` forces the content area beyond the DialogContent height at high UI scaling, clipping the footer. Removing the fixed min-height addresses the root issue.

2. **Independent nav scrolling**: Adding `min-h-0` + `sm:overflow-y-auto` to the nav ensures category buttons scroll independently when they overflow, without competing with the main content area for space.

3. **Zero regression risk**: On large screens, `DialogContent` still maintains its fixed height via `min(660px, ...)`. The change only affects small viewports / high scaling scenarios.

4. **Minimal diff**: Only 2 lines changed to fix the problem, no unnecessary refactoring.

Thanks!

PR #806, opened 2026-06-07
feat(dameng): full DM8 database support — autotrace, explain plan, user management

Quinlevi:
## Summary

This PR adds full Dameng (达梦 DM8) database support to DBX, including:

### ✨ Features

**Explain Plan + Autotrace**
- DM explain plan parsing (tabular format with dot-notation ID hierarchy)
- Autotrace mode: executes SQL with MONITOR_SQL_EXEC enabled, retrieves actual row counts alongside estimated rows
- TEXT view for raw DM plan text display alongside tree/summary views
- A-TRACE badge indicator on ExplainPlanViewer
- Green 'A' toggle button in EditorToolbar (visible only for DM connections)

**User Management**
- Create User dialog with auto-generated DM-compliant passwords (12+ chars, upper+lower+digit)
- Tab auto-fills password/tablespace from username
- Password visibility toggle
- Auto-creates 100MB tablespace with autoextend before user creation
- Forces uppercase for username/tablespace (DM stores unquoted identifiers as uppercase)

**Backend (Rust/Java)**
- DamengAgent: getExplainInfo(mode) supporting both 'explain' and 'autotrace' modes
- DM JDBC driver V8.1.4.80 (provides both getExplainInfo(String) AND getExplainInfo(Statement))
- New Tauri command get_explain_info forwarding SQL+mode via JSON-RPC
- New Tauri command build_create_user_sql
- Detailed stderr logging with [DamengAgent] prefix

### 🔧 Technical Details

- explainPlan.ts: DM plan text parser (operator lines with #hash, [cost, estRows->actualRows, width], Statistics section)
- ExplainPlanNodeTree.vue: compact single-line operators with collapsible children, actual/estimated row comparison
- backend: AgentMethod::GetExplainInfo enumeration, agent_driver.rs dispatch, query_execution_sql.rs DM EXPLAIN SQL rules
- databaseCapabilitySets: DM removed from CREATE_DATABASE types (DM uses users instead of databases)

### 📝 Notes

Tested against DM8 (达梦数据库 V8). Compatible with both explain and autotrace modes.


t8y2 (changes requested):
## PR #806: feat(dameng): full DM8 database support — autotrace, explain plan, user management

**+905/-55, 16 files | CI: no checks reported | Mergeable: MERGEABLE**

### 🔴 Must Fix

1. The new Tauri commands are never registered. `get_explain_info` and `build_create_user_sql` are defined, but `src-tauri/src/lib.rs` only registers `build_explain_sql` and then skips straight to `build_dropped_file_preview_sql`. The desktop calls in `tauri.ts` will fail with an unknown command, so both DM explain and create-user are currently unreachable.

2. The `getExplainInfo` return contract does not line up. The Java path returns an object with `ok`, `plan`, `has_actual_stats`, and `mode`, but the Tauri command only accepts a raw JSON string. After registration is fixed, a successful Java response would still become `Unexpected non-string result from getExplainInfo`.

3. The web backend routes are missing. `http.ts` now posts to `/api/query/build-create-user-sql` and `/api/query/get-explain-info`, but `dbx-web` does not register matching routes. Web mode will 404 those calls.

4. The autotrace toggle is not wired to state. `useSqlExecution` returns `explainMode`, but `App.vue` does not destructure it or pass it into `EditorToolbar`; the toolbar emits `update:explainMode`, but no parent listens. The A button never actually switches `tryExplain()` into autotrace mode.

5. Autotrace executes the supplied SQL without the existing explain safety gate. The DM branch bypasses `buildExplainSql`, which is where unsafe statements are rejected, and autotrace then calls `stmt.execute(trimStatementSql(sql))`. Once the toggle is wired, clicking “explain” could execute DML/DDL without the dangerous-SQL confirmation path.

### ✅ Well Done

The DM explain parser is thoughtfully scoped and supports both raw text and parsed tree views. Removing Dameng from `CREATE_DATABASE` capability also matches the user/schema model better than forcing it into database creation semantics.

### Verdict

Request changes. The feature surface is useful, but the current PR has multiple integration blockers that prevent the advertised flows from working, plus a safety issue around autotrace execution.

Checks I ran locally: `pnpm typecheck` passed, and `cargo check -p dbx-core -p dbx-web` passed.

PR #808, opened 2026-06-07
feat(dameng): full DM8 database support — autotrace, explain plan, user management

Quinlevi:
## Summary

Full Dameng DM8 database support including explain plan, autotrace with actual row counts, and user management.

## Changes

- **Java Agent**: getExplainInfo with explain/autotrace mode, DM JDBC driver V8.1.4.80
- **Rust Backend**: get_explain_info Tauri command, DM explain SQL rules
- **Frontend**: explain plan tree view + TEXT view + A-TRACE badge, CreateUserDialog
- **Web Backend**: DM endpoint routes for web mode
- **Safety**: autotrace rejects dangerous SQL (DROP/DELETE/etc.)

## Review Fixes

- getExplainInfo handles Java agent object return contract
- Web routes added for /api/query/get-explain-info and build-create-user-sql
- Autotrace safety gate with DANGER_RE check


t8y2:
This PR seems to be based on an outdated or divergent branch rather than the current `main`. Besides the Dameng changes, it includes many unrelated reversions/changes across settings, transport layers, data grid, Mongo browser, MCP, tab runtime cache, and tests. Some of those unrelated changes are causing frontend typecheck/test failures and rust fmt failures.

Could you rebase or recreate this PR from the latest `main`, then apply only the Dameng-specific changes? That should make the review much cleaner and avoid regressing unrelated features.

PR #874, opened 2026-06-08
feat(elasticsearch): SQL-model execution with mapping-aware completion

JinRudy:
## Summary

Treat Elasticsearch connections' editor text starting with `SELECT`/`WITH` as SQL-model end-to-end (execution, completion, diagnostics, pagination) so users can write SQL directly against an ES connection instead of toggling between Query DSL and SQL.

Two commits:
1. `feat(elasticsearch)` — SQL-model support (execution, completion, diagnostics, server-side pagination, WHERE→DSL translation)
2. `fix(grid)` — a small fallback fix in the result-grid pagination logic that surfaced while wiring up SQL-model results; not ES-specific

### Execution adapter (`crates/dbx-core/src/db/elasticsearch_driver.rs`)
- Detect SQL-model input. Simple `SELECT * FROM <index[*]> [ORDER BY <field>] [LIMIT N] [OFFSET M]` browses go through `/_search` to avoid ES SQL's array-field limitation (`host.ip`, etc.). Aggregates / projected columns keep going through `/_sql`.
- Bare `SELECT * FROM idx` with no LIMIT auto-paginates at `size=100` rather than pulling the whole index.
- Server-side pagination wired into the existing `prepareQueryPaginationExecutionPlan` flow: ES moved out of `unsupported_pagination_type`; plan-wrapped queries (`LIMIT N OFFSET M`) are translated by the driver into `_search` `size`/`from`. `affected_rows` reports the index's true match total via `hits.total.value` for plan-paged queries, so the result-grid can compute the real page count.
- User-written `LIMIT N` (no OFFSET) is left untouched by the plan — the user explicitly bounded the result set; the front-end paginates client-side and reports `rows.len()` as the total.
- 0-hit `_search` bodies now surface as an empty grid (with an `_id` column placeholder) instead of falling back to the raw `status`/`response` JSON view.
- New `get_columns` reads `/<index>/_mapping`, flattening dotted fields and multi-fields; wired into `schema::get_columns_core`.

### SQL → ES DSL translator (`crates/dbx-core/src/db/elasticsearch_sql.rs`)
`SELECT *` with clauses that the hand-written parser doesn't cover (WHERE / IN / BETWEEN / LIKE / ...) is parsed with `sqlparser-rs` and translated to a `_search` body in-process — we **don't** delegate to ES's `_sql` endpoint or its `_sql/translate` cousin. Both of those live in the same ES SQL engine and reject common shapes like `LIKE` on a `text` field with no `.keyword` sub-field, which translate cleanly to raw DSL.

Translation table:

| SQL | ES DSL |
|---|---|
| `field = 'v'` | `{"term":{"field":"v"}}` |
| `field != 'v'` | `{"bool":{"must_not":[{"term":...}]}}` |
| `field LIKE '%x%'` | `{"wildcard":{"field":{"value":"*x*","case_insensitive":true}}}` |
| `field LIKE 'prefix%'` | `{"prefix":{"field":"prefix"}}` (optimised) |
| `field IN ('a','b')` | `{"terms":{"field":["a","b"]}}` |
| `field BETWEEN a AND b` | `{"range":{"field":{"gte":a,"lte":b}}}` |
| `field > N` etc. | `{"range":{"field":{"gt":N}}}` |
| `field IS NULL` | `{"bool":{"must_not":{"exists":{"field":"field"}}}}` |
| `field IS NOT NULL` | `{"exists":{"field":"field"}}` |
| `A AND B AND C` | `{"bool":{"must":[A,B,C]}}` (flattened) |
| `A OR B OR C` | `{"bool":{"should":[A,B,C],"minimum_should_match":1}}` |
| `NOT A` | `{"bool":{"must_not":[A]}}` |
| `(A OR B) AND C` | nested bool |
| `ORDER BY field [ASC\|DESC]` | `sort` array |

SQL LIKE pattern translation: `%`→`*`, `_`→`?`, backslash escapes are honoured, and user-written `*` / `?` in the pattern are escaped back so they're matched literally.

Input is run through the existing `adapt_elasticsearch_sql_query` first, so hyphenated indices (`filebeat-7.17.1-2026.05.26-000010`) and `@timestamp`-style identifiers reach the parser as double-quoted identifiers.

### Editor (`apps/desktop/src/components/editor/QueryEditor.vue`, `apps/desktop/src/lib/sqlCompletion.ts`, `apps/desktop/src/lib/sqlSemanticDiagnostics.ts`)
- When the active statement on an ES connection starts with `SELECT`/`WITH`, switch the completion provider from Query DSL completion to SQL completion backed by `_mapping` columns.
- Skip generic SQL semantic diagnostics for Elasticsearch connections — the upstream parser was rejecting hyphenated wildcard indices and producing repeated `500`s on `/api/query/analyze-sql-references`.
- Derive SQL completion context from the current `SELECT`/`WITH` line block rather than the whole semicolon-delimited document, so JSON-embedded SQL strings (e.g. inside a `POST /_sql` body) don't pollute table/column suggestions in a later bare SQL statement.

### Pagination plan (`crates/dbx-core/src/query_result_sql.rs`, `apps/desktop/src/stores/queryStore.ts`)
- `build_paginated_query_sql` ES branch: leaves user-written LIMIT alone; otherwise wraps with explicit `LIMIT N OFFSET M` (always emits OFFSET so the driver can distinguish plan-wrapped queries from user-written ones).
- `build_count_query_sql` ES branch: returns `unsupported` — ES self-reports the total via `hits.total.value` so we don't need a wrapping COUNT query (which ES SQL can't run anyway).
- `queryStore.ts`: when server-side pagination is in effect with no `countSql`, use `result.affected_rows` directly as `resultTotalRowCount`. Only the ES path enters this branch today; MySQL / Postgres / SQL Server continue to use their existing countSql flow unchanged.

### Result-grid fallback (`apps/desktop/src/lib/dataGridPagination.ts`, `apps/desktop/src/components/grid/DataGrid.vue`) — second commit
For SQL editor results the backend returns every row up-front and `rowCount` IS the authoritative total. `canGoNextDataGridPage`'s `rowCount >= pageSize` fallback was letting users page past the end forever. Added an `allRowsLoaded` option to surface client-side total knowledge to the helper; `DataGrid` passes it when there's no server-side `pageLimit`.

## Screenshots

Stored under `docs/` next to the existing `screenshot-*.png` files in the repo.

### Pagination — `SELECT * FROM filebeat-7.17.1-2026.05.26-000010 limit 200`
The bottom bar reports `共 200 行` and the paginator stops at page 2 (200 ÷ 100/page).

pagination

### Mapping-aware completion
Typing `SELECT e` against a hyphenated wildcard index suggests fields pulled from `/<index>/_mapping`, with the field type shown next to each suggestion.

completion

### WHERE translation — `WHERE log.offset = 982708883`
`SELECT *` with a WHERE clause is translated to ES DSL in-process (sqlparser-rs), then run via `_search` — works on a `long` field without needing `_sql`'s mapping constraints. See the comment below for the full WHERE cheat sheet.

where

## Test plan
- [x] `pnpm check` (format / lint / typecheck / test) — all green
- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --locked --all-targets` — no new warnings on touched files
- [x] `cargo test --workspace --locked` — 0 failed
- [x] Manual against a live ES 7.x cluster (see screenshots):
  - `SELECT * FROM <wildcard>` auto-pages at `size=100`, paginator total reflects `hits.total.value`
  - `SELECT * FROM <idx> LIMIT 200` returns 200 rows, paginator stops at page 2
  - `SELECT * FROM <idx> WHERE log.offset = N LIMIT 100` returns 1 row via the in-process SQL→DSL translator (would 400 through ES `_sql`)
  - `_mapping`-backed completion shows field names + types for hyphenated wildcard indices
  - Hyphenated wildcard indices like `filebeat-7.17.1-*` no longer trigger 500s on `/api/query/analyze-sql-references`

## Notes
- No driver-manifest or capability changes (no new DB type).
- ES is the only driver moved out of `unsupported_pagination_type`; MySQL / PostgreSQL / SQL Server / Oracle / DB2 / SQLite / DuckDB / ClickHouse pagination unchanged.
- Per request from the contributor, unit tests for these changes were intentionally omitted. Happy to add coverage if it's a merge blocker — the SQL→DSL translator and `parse_select_star_search_query` (LIMIT/OFFSET parsing, plan-vs-user signal) are the obvious targets.
- Base: latest `main` at submission time.


t8y2 (approved):
Nice work! Clean implementation with thoughtful design decisions. The plan-paginated vs user-written LIMIT signaling via OFFSET presence is elegant, and the separation of display total from navigation total in the result grid is well-executed.

A few minor notes for future consideration:
- `extractReferencedTables` regex may miss comma-separated tables (low impact)
- Would be nice to add tests for `parse_select_star_search_query` and the SQL adapter if there's a follow-up

Thanks for the contribution!

JinRudy:
### `SELECT * … WHERE …` cheat sheet for the Elasticsearch path

A quick reference for what survives the SQL→DSL translator and what to watch out for in ES mapping land. All examples assume a connection where the SQL editor is in SQL-model (text starting with `SELECT`/`WITH`).

#### Hyphenated / dotted index names
You don't need quotes — the driver auto-quotes the FROM target before parsing:
```sql
SELECT * FROM filebeat-7.17.1-2026.05.26-000010 WHERE … LIMIT 100
```
`@timestamp`-style identifiers are also auto-quoted.

#### Numeric exact match
```sql
WHERE log.offset = 982708883
```
`=` on numeric / date / keyword fields becomes a `term` query — milliseconds against the inverted index.

#### Substring match on text — use `LIKE`
```sql
WHERE message LIKE '%nWTkrj…%'
WHERE url.full LIKE '%createDialog%'
```
Translates to `wildcard` with `case_insensitive: true`. Works directly on `text` fields, unlike ES `_sql` which insists on a `.keyword` sub-field. Trade-off: ES analyzer behaviour still applies — for hashes / random strings the analyzer's tokenisation can hide the match. If a wildcard returns nothing on a hash, fall back to native REST + `match_phrase` in the same tab.

#### Prefix match — optimised
```sql
WHERE path LIKE 'prefix%'
```
Recognised and emitted as a `prefix` query rather than `wildcard` — significantly cheaper.

#### Equality / range / IN / BETWEEN
```sql
WHERE status_code = 200
WHERE response_time > 500
WHERE response_time BETWEEN 100 AND 1000
WHERE log_type IN ('network', 'error')
```

#### Combined predicates
```sql
WHERE log_type = 'network'
  AND url.full LIKE '%createDialog%'
  AND @timestamp >= '2026-06-09T00:00:00Z'
```
`AND` / `OR` chains are flattened into a single `bool.must` / `bool.should`. Mix and match with parentheses freely.

#### Null check
```sql
WHERE error.code IS NOT NULL
WHERE error.code IS NULL
```
Translates to `exists` / `bool.must_not.exists`.

#### NOT
```sql
WHERE NOT (log_type = 'network')
```

#### Ordering & pagination
```sql
SELECT * FROM filebeat-7.17.1-2026.05.26-000010
WHERE log_type = 'network'
ORDER BY @timestamp DESC
LIMIT 100
```
User-written `LIMIT N` (no OFFSET) is left alone and the front-end paginates client-side (total = `LIMIT`). Omit `LIMIT` and the pagination plan kicks in: the driver auto-caps at 100, the result-grid total reflects the index's true match count, and the paginator jumps server-side as you click through pages.

---

### Common gotchas

**`log` (or any object field) can't be LIKE'd.** When the column value renders as a JSON object (`{"offset":…,"file":{…}}`), `log` is an *object* in ES mapping — only its leaves (`log.offset`, `log.file.path`) are indexed. Search the leaf, not the parent:
```sql
-- Wrong: log is an object, no inverted index entries
WHERE log LIKE '%982708883%'
-- Right: search the leaf
WHERE log.offset = 982708883
```

**Numbers don't take LIKE.** `LIKE` is for strings — use `=`, range, or `BETWEEN` on numeric / date fields.

**Hashes / random IDs in `message`.** ES analyzer tokenisation makes wildcard hit-or-miss for these. If a wildcard search returns nothing on a string you can see in the data, drop to native REST in the same tab:
```
POST /filebeat-7.17.1-2026.05.26-000010/_search
{
  "size": 100,
  "query": { "match_phrase": { "message": "nWTkrj…" } }
}
```

**To discover available fields**, expand the index in the left sidebar — the schema browser shows every flattened path (`log.offset [long]`, `log.file.path [keyword]`, etc.), so you can pick the right leaf without guessing.

---

### What's not supported (yet)

The translator only handles `SELECT * FROM <single_index>`. The following are deliberate out-of-scope and stay on the existing `_sql` path:
- Projected columns (`SELECT col1, col2`) and aggregates (`COUNT(*)`, `GROUP BY`)
- JOINs (ES doesn't support them anyway)
- Subqueries
- HAVING

If you need full-text scoring or nested-field aggregation, native REST in the same tab gives you the full ES DSL.

PR #881, opened 2026-06-08
feat: add k8s deployment, session expiry, and logout button

midy177:
feat: add k8s deployment, session expiry, and logout button

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,17 @@ pnpm dev:backend   # backend
 pnpm tauri build
 ```
 
+#### Docker image build
+```bash
+docker buildx build \
+  --platform linux/amd64 \
+  --push \
+  -t hub.yeastardigital.com/novo-middleware/dbx:latest \
+  -f deploy/Dockerfile \
+  --build-arg BUILDPLATFORM=linux/amd64 \
+  .
+```
+
 The installer will be in `src-tauri/target/release/bundle/`.
 
 ## Tech Stack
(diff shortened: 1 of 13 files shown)


t8y2 (changes requested):
### 🔴 Must Fix

**1. CI failures**

- **Frontend**: `AppToolbar.vue` failed format check — run prettier to fix
- **Rust**: compilation error in tests — `connection.rs:214` still uses `HashSet::new()` but `WebState.sessions` has been changed to `HashMap<String, Instant>`. Replace it with `HashMap::new()`

**2. Hardcoded password in deployment.yaml**

```yaml
- name: DBX_PASSWORD
  value: "ZpFoJGrche4lrUj0"
```

Real passwords should not be committed. Use a placeholder or a Kubernetes Secret instead.

**3. Private registry references in README.md and deployment.yaml**

`hub.yeastardigital.com/novo-middleware/dbx:latest` is an internal registry. Public repo docs and manifests should use a generic image name (e.g. `t8y2/dbx` or a placeholder).

### 🟡 Suggestions

**4. Three unrelated changes in one PR** (K8s deployment, session expiry, logout button) — consider splitting into separate PRs for easier review and rollback

### ✅ Well Done

- Session expiry design is clean: `HashMap<String, Instant>` replacing `HashSet<String>`, with `DBX_SESSION_TTL_HOURS` env var and background cleanup
- `is_session_valid` and `purge_expired_sessions` encapsulated on `WebState`, clear separation of concerns
- Logout button only shown when `needsAuth` is true
- i18n covers en + zh-CN

PR #910, opened 2026-06-09
feat(elasticsearch): translate SELECT * WHERE to ES DSL in-process (follow-up to #874)

JinRudy:
Follow-up to #874. Two commits:

1. **`feat(elasticsearch)`** — translate `SELECT * … WHERE …` ourselves with sqlparser-rs, drop the `_sql` round-trip for that shape. Also fixes the empty-hits result-grid display.
2. **`chore(test)`** — fills `ConnectionConfig::idle_timeout_secs` in 11 test fixtures so `cargo test --workspace --locked` / `cargo clippy --all-targets` pass against `main` (the new field landed without test-side updates).

## Why

#874 routed `SELECT *` with extra clauses (WHERE / IN / BETWEEN / LIKE / …) to ES's `_sql` endpoint. In practice, `_sql` refuses several very common shapes:

- `LIKE` on a `text` field with no `.keyword` sub-field — the default filebeat / log-shipper mapping for `message`, so a perfectly reasonable substring search 400s.
- `SELECT *` against an index whose `_source` contains an array field (`host.ip`, …) — the user can't even tell whether they hit this case, since they didn't write any column reference.

ES's `_sql/translate` cousin doesn't help — same engine, same restrictions.

Generating the `_search` body in-process bypasses both, so the SQL surface area users can actually rely on is much closer to what the result-grid suggests.

## What

`crates/dbx-core/src/db/elasticsearch_sql.rs` (new, ~340 lines): a sqlparser-rs walker that translates a recognised `SELECT * FROM <single index>` statement into a `_search` body. Returns `Ok(None)` when the statement isn't a `SELECT *` shape (caller falls back to the existing `_sql` path for projected columns / aggregates / …), `Err` when it is a `SELECT *` but a clause is unsupported (caller surfaces a clear error instead of silently downgrading).

| SQL | ES DSL |
|---|---|
| `field = 'v'` | `{\"term\":{\"field\":\"v\"}}` |
| `field != 'v'` | `{\"bool\":{\"must_not\":[{\"term\":…}]}}` |
| `field LIKE '%x%'` | `{\"wildcard\":{\"field\":{\"value\":\"*x*\",\"case_insensitive\":true}}}` |
| `field LIKE 'prefix%'` | `{\"prefix\":{\"field\":\"prefix\"}}` (optimised) |
| `field IN ('a','b')` | `{\"terms\":{\"field\":[\"a\",\"b\"]}}` |
| `field BETWEEN a AND b` | `{\"range\":{\"field\":{\"gte\":a,\"lte\":b}}}` |
| `field >/</>=/<=` | `{\"range\":{\"field\":{\"gt\":…}}}` |
| `field IS NULL` | `{\"bool\":{\"must_not\":[{\"exists\":{\"field\":\"field\"}}]}}` |
| `field IS NOT NULL` | `{\"exists\":{\"field\":\"field\"}}` |
| `A AND B AND C` | flattened `bool.must` |
| `A OR B OR C` | `bool.should` + `minimum_should_match: 1` |
| `NOT A` | `bool.must_not` |
| `ORDER BY f ASC\|DESC` | `sort` array |

SQL LIKE pattern translation: `%`→`*`, `_`→`?`, backslash escapes honoured, and user-written `*` / `?` in the pattern are escaped back so they match literally. The same `adapt_elasticsearch_sql_query` from #874 runs first so hyphenated indices (`filebeat-7.17.1-…`) and `@timestamp`-style identifiers reach sqlparser as double-quoted identifiers.

### Empty result fix

`parse_elasticsearch_response`'s `_search` branch had `.filter(|h| !h.is_empty())` on the hits array, so a 0-hit response fell through to the raw-JSON `status / response` fallback instead of showing an empty grid. Dropped the filter and added an `_id` column placeholder when `all_keys` is empty so the result-grid still has a header row to render.

### `chore(test)` commit

`main` added `pub idle_timeout_secs: u64` to `ConnectionConfig` without updating the test-side constructors. `cargo check` was passing because the lib build never compiles tests, but `cargo test --workspace --locked` and `cargo clippy --all-targets` both fail with E0063 in 11 fixture sites:

```
crates/dbx-core/src/agent_connection.rs
crates/dbx-core/src/cloud_sync.rs
crates/dbx-core/src/connection.rs
crates/dbx-core/src/connection_secrets.rs
crates/dbx-core/src/models/connection.rs
crates/dbx-core/src/query.rs
crates/dbx-core/src/schema.rs
crates/dbx-core/src/transfer.rs
crates/dbx-core/tests/live_postgres_transfer.rs
crates/dbx-web/src/routes/connection.rs
src-tauri/src/commands/connection.rs
```

Filled them in with `default_idle_timeout_secs()` (matches the field's own `#[serde(default = …)]`) so future bumps to the default flow through automatically. Happy to drop this commit if you'd rather land it via a separate PR.

## Screenshot

`SELECT * FROM filebeat-7.17.1-2026.05.26-000010 WHERE log.offset = 982708883 LIMIT 100` — a numeric leaf-field query that would 400 through `_sql` (the SQL parser route inherits ES SQL semantics for `_source` typing) — works through the in-process translator.

where

## Test plan

- [x] `pnpm check` (format / lint / typecheck / test) — all green
- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --locked --all-targets` — 0 errors on a clean `main` after the `chore` commit
- [x] `cargo test --workspace --locked` — 0 failures
- [x] Manual against a live ES 7.x cluster:
  - `SELECT * FROM <idx> WHERE log.offset = N LIMIT 100` returns 1 row via `_search` (would 400 through `_sql`)
  - `SELECT * FROM <idx> WHERE message LIKE '%hash%'` runs against a `text` field with no `.keyword` (would 400 through `_sql`)
  - `SELECT * FROM <idx> WHERE A AND B AND C` flattens into a single `bool.must`
  - 0-hit results render as an empty grid, not raw JSON

## Notes

- All translator changes are scoped to the ES path. `query_result_sql.rs` / pagination plan / `prepareQueryPaginationExecutionPlan` unchanged from #874.
- Per the same convention as #874, unit tests for the translator are intentionally omitted; the main targets if you'd like coverage are `translate_select_star` (translation table) and `sql_like_to_wildcard` (escape semantics).
- Base: `main` at submission time.


t8y2 (approved):
LGTM, thanks!

PR #980, opened 2026-06-10
feat: add InfluxDB support

ptma:
The InfluxDB driver uses the InfluxDB v1 HTTP query endpoint, compatible with both InfluxDB v2 and v3.


t8y2:
ci test问题可忽略 稍后review

t8y2:
Thanks!

PR #1041, opened 2026-06-11
Optimize data compare memory usage

SuLea-IT:
## Summary
- Reduce Data Compare backend memory by avoiding per-row HashMap materialization during comparison.
- Build sync plans from borrowed diff data to avoid cloning large DataCompareResult payloads.
- Add runtime/static memory audit docs with before/after measurements.

## Memory Check
- Same 50,000 vs 49,649 row SQLite Data Compare scenario.
- `dbx-web` RSS peak dropped from 507,504 KiB to 239,696 KiB, about 53% lower.
- Diff counts, sync statement count, sync SQL size, and response size stayed unchanged.

## Test Plan
- `cargo test -p dbx-core data_compare::tests`
- `cargo test -p dbx-web data_compare`
- `cargo check --manifest-path src-tauri/Cargo.toml`
- `git diff --check -- crates/dbx-core/src/data_compare.rs docs/performance/runtime-memory-audit.md docs/performance/memory-usage-audit.md`

### DIFF
--- a/crates/dbx-core/src/data_compare.rs
+++ b/crates/dbx-core/src/data_compare.rs
@@ -167,24 +167,32 @@ pub struct DataCompareFromTablesPreparation {
 }
 
 pub fn prepare_data_compare(options: DataComparePreparationOptions) -> Result<DataComparePreparation, String> {
+    let DataComparePreparationOptions {
+        table_name,
+        schema,
+        columns,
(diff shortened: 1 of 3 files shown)


t8y2:
Merged, thanks for the contribution! 🎉

PR #1159, opened 2026-06-14
Change default locale fallback from zh-CN to en

### LINKED ISSUE #1158: Change default locale fallback from zh-CN to en
## Problem

When a user's system language is not in the supported locales list (e.g. French, German, Japanese…), the app falls back to `zh-CN` (Simplified Chinese). This is unintuitive for an international open-source project — a user who doesn't read Chinese gets an interface in a language they can't understand.

## Current behavior

1. App detects system locale via `navigator.languages` / `navigator.language`
2. If no supported locale matches (e.g. `fr`, `de`, `ja`), `localeFromLanguageTag()` returns `null`
3. `detectLocaleFromLanguages()` falls back to `defaultLocale`, which is `"zh-CN"`

## Proposed solution

Change the default fallback locale from `"zh-CN"` to `"en"` in `apps/desktop/src/i18n/index.ts`.

This means:
- Users with unsupported locales get English instead of Chinese
- The `en` locale bundle is eagerly loaded instead of `zh-CN`
- Chinese users are unaffected (their locale is still detected correctly via `navigator.language`)

## Impact

- **1 file changed**: `apps/desktop/src/i18n/index.ts` (~6 lines)
- No behavior change for users whose system locale is already supported
- No new dependencies

## Notes

The `fallbackLocale` in `createI18n` is also tied to `defaultLocale`, so missing translation keys will also fall back to English instead of Chinese — which is the correct behavior for a project with English as its primary development language.

---

I'm willing to submit a PR for this change.

athoune:
## Problem

When a user's system language is not in the supported locales list (e.g. French, German, Japanese…), the app falls back to `zh-CN` (Simplified Chinese). This is unintuitive for an international open-source project — a user who doesn't read Chinese gets an interface in a language they can't understand.

## Current behavior

1. App detects system locale via `navigator.languages` / `navigator.language`
2. If no supported locale matches (e.g. `fr`, `de`, `ja`), `localeFromLanguageTag()` returns `null`
3. `detectLocaleFromLanguages()` falls back to `defaultLocale`, which is `"zh-CN"`

## Solution

Change the default fallback locale from `"zh-CN"` to `"en"` in `apps/desktop/src/i18n/index.ts`.

- Users with unsupported locales get English instead of Chinese
- The `en` locale bundle is eagerly loaded instead of `zh-CN` (also a small perf win)
- Chinese users are unaffected (their locale is still detected correctly via `navigator.language`)

## Impact

- **1 file changed**: `apps/desktop/src/i18n/index.ts` (~6 lines)
- No behavior change for users whose system locale is already supported
- No new dependencies

Closes #1158

### DIFF
--- a/apps/desktop/src/i18n/index.ts
+++ b/apps/desktop/src/i18n/index.ts
@@ -1,5 +1,5 @@
 import { createI18n } from "vue-i18n";
-import zhCN from "./locales/zh-CN";
+import en from "./locales/en";
 import { safeLocalStorageGet, safeLocalStorageSet } from "@/lib/safeStorage";
 
 export type Locale = "en" | "es" | "it" | "pt-BR" | "zh-CN" | "zh-TW";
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks @athoune!

PR #1161, opened 2026-06-14
feat(generate): fix date range persistence, add i18n for data generation

dbin0123:
## Summary

• Fixed date range start param not persisting — `useVModel passive:true` failed to write
  `start` property when initial value was undefined; replaced `<Input v-model>` with native
  `<input>` + `@change` for all date/time inputs in DateParams, DateTimeParams, TimeParams
• Added `ensureRange()` guard to swap start/end if reversed, preventing out-of-range dates
• Text generator now randomly picks from 5 paragraph templates (en + zh) instead of always
  using the same lorem ipsum substring
• Added 57+ i18n keys to `dataGenerate` section and 46 `dataGenerate.gen.*` keys for
  generator labels across all 6 locale files (en, zh-CN, zh-TW, es, it, pt-BR)
• Added missing `tableToolbox` section to es/it/pt-BR locales
• Updated DataGenerateDialog, GeneratorSelect, GeneratorParamsPanel, and 10+ params
  sub-components to use `t()` instead of hardcoded Chinese strings
• GeneratorParamsPanel now auto-resets params when generatorKey changes, preventing
  stale params from previous generator selection
• Generator labels in selector translate via `t()` with fallback to original label
• Added debug box with `break-all` wrapping to diagnose generatorParams

## Type

feat


t8y2:
Thanks @dbin0123!

t8y2:
数据生成很实用！不过似乎加载配置文件和保存配置文件点击无响应

PR #1186, opened 2026-06-15
style: tab添加数据库标签

cherlinbest-pixel:


t8y2 (changes requested):
Thanks for the PR! The idea of showing database type on tabs is useful, but replacing the mode icons (Table2 for data, Code2 for SQL, PencilRuler for structure, etc.) makes it harder to tell what kind of content each tab has at a glance.

How about keeping the original mode icons and instead showing the database type via a hover tooltip on the icon? Something like:

```vue
<Tooltip>
  <TooltipTrigger as-child>
    <span class="shrink-0" :class="tabIconClass(tab)">
      <Table2 v-if="..." class="h-3.5 w-3.5" />
      ...
    </span>
  </TooltipTrigger>
  <TooltipContent>
    {{ connectionLabel }}
  </TooltipContent>
</Tooltip>
```

This way the mode icon stays visible for quick scanning, and hovering reveals the database type.

cherlinbest-pixel:
你说得对，然后我又看了下现有主分支的项目，我看到
image
，最右上角，已有数据库图标，所以，这个优化，是否有些多余了

PR #1298, opened 2026-06-16
fix(ai): avoid empty Claude system field rejected by Anthropic

iiiusky:
Anthropic rejects requests whose system field is an empty string with "system: text content blocks must be non-empty". The connection test hardcoded an empty system, and regular completion/stream requests sent an empty system_prompt when none was configured.

Add a claude_system_prompt() helper that substitutes a non-empty default for empty or whitespace-only prompts, and apply it to all Claude request sites. Covered by a unit test.


t8y2:
Thanks @iiiusky!

PR #1306, opened 2026-06-16
feat(redis): add Redis command completion to query editor

haipengno1:
- add completion module: command names, two-stage subcommands, and key names
- key-name completion is async with a per-db cache capped at 1000 entries, cleared on disconnect
- gate key completion by argument position: stop after a single-key command is filled, keep suggesting for variadic key-list commands
- reuse the diagnostics command table as the data source, showing group/arity/safety


t8y2:
Thanks @haipengno1!

PR #1309, opened 2026-06-16
fix(redis): make completion menu follow typing

haipengno1:
The result used filter:false + validFor, which CodeMirror treats as mutually exclusive — validFor is ignored, so the list froze and the selection did not move while typing. Build the result with the default built-in filter so typing narrows the list and moves the selection.


t8y2:
Thanks @haipengno1!

PR #1311, opened 2026-06-16
fix(redis): invalidate key completion cache after write commands

haipengno1:
The Redis key-name completion cache was only cleared on connection lifecycle events, so SET/DEL/EXPIRE/... left autocomplete showing stale keys. Drop the cache for the db each write command ran on:
- add isRedisMutatingCommand() based on the command-table safety field (confirm = write; destructive blocked cmds minus read-only/admin ones)
- invalidate after each mutating command in queryStore batch execution and in RedisKeyBrowser.runRedisCommand


t8y2:
Thanks @haipengno1!

PR #1321, opened 2026-06-17
fix(redis): refresh sidebar db key counts after write commands

haipengno1:
Redis write commands (SET/DEL/...) ran from the query editor or the key-browser terminal did not update the dbN (count) labels in the left tree until a manual refresh. Add refreshRedisDbKeyCounts(), which re-fetches authoritative per-db counts via INFO keyspace and patches the labels in place without rebuilding the tree (preserving expanded key nodes), then trigger it fire-and-forget after any mutating command in both execution paths.


t8y2:
Thanks @haipengno1!

PR #1341, opened 2026-06-17
fix(mysql): avoid Aliyun ADB timeout for bounded text results

luoianun:
## Summary
- Fixes a query timeout/hang observed when querying Aliyun ADB through the MySQL-compatible endpoint.
- Uses buffered collection for bounded MySQL text-protocol result sets.
- Keeps the existing streaming path for unbounded or oversized result sets to limit impact on regular MySQL queries.

## Background
Aliyun ADB metadata loading and connection tests can succeed, while bounded table queries such as `SELECT ... LIMIT 100` may time out while DBX reads rows through the MySQL text-protocol streaming path.

This change only switches the text-protocol path to `collect_and_drop()` when DBX can see a top-level MySQL `LIMIT` that is within the configured row limit. Queries without a limit, oversized limits, or nested-only limits continue to stream rows as before.

## Test Plan
- `cargo fmt --check`
- `cargo test -p dbx-core mysql_text_result_sets_ -- --nocapture`
- `cargo test -p dbx-core live_mysql_compatible_limited_text_protocol_query_succeeds -- --nocapture`

Also verified the equivalent env-driven live test against an Aliyun ADB MySQL-compatible endpoint locally.

### DIFF
--- a/crates/dbx-core/src/db/mysql.rs
+++ b/crates/dbx-core/src/db/mysql.rs
@@ -1429,6 +1429,104 @@ fn query_result_row_limit(max_rows: Option<usize>) -> usize {
     max_rows.unwrap_or(crate::query::MAX_ROWS).max(1)
 }
 
+fn should_collect_text_result_set(sql: &str, row_limit: usize, max_rows: Option<usize>) -> bool {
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks @luoianun!

PR #1349, opened 2026-06-17
fix(redis): unify aggregated INFO output format in console

dal1wg:
## Purpose
unify aggregated INFO output format in console

## Changes
- **`apps/desktop/src/components/redis/RedisKeyBrowser.vue`**  
- **`apps/desktop/src/lib/redisValuePresentation.ts`**


t8y2:
Thanks @dal1wg!

PR #1366, opened 2026-06-17
perf: optimize xlsx export to reduce memory allocations

bigliangzao:
Optimize XLSX export module to reduce unnecessary memory allocations:

1. **escape_xml**: Replace per-character Vec<char> heap allocations with pre-allocated String + push_str. The original code allocated a new Vec for every character (e.g., `"&amp;".chars().collect::<Vec<_>>()`), causing thousands of small allocations when exporting large tables.

2. **column_name**: Replace O(n) String::insert(0, ch) with O(1) push + single reverse at the end.

3. **estimate_column_widths**: Remove unnecessary column.clone() allocation.

4. **value_text**: Simplify bool handling from 6 lines to 1 line using v.to_string().

All changes maintain identical output behavior. Existing tests pass without modification.

### DIFF
--- a/crates/dbx-core/src/xlsx_export.rs
+++ b/crates/dbx-core/src/xlsx_export.rs
@@ -11,20 +11,21 @@ pub struct XlsxWorksheetData {
 }
 
 fn escape_xml(value: &str) -> String {
-    value
-        .chars()
-        .filter(|ch| {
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks @bigliangzao!

PR #1376, opened 2026-06-17
fix(mysql): prevent panic on Chinese characters in WHERE clause

bigliangzao:
Fix UTF-8 boundary issues in mysql_top_level_limit, parse_usize_token, and mysql_keyword_at functions that caused crashes when SQL contains Chinese characters in string literals.

The issue was introduced in commit 021e1682 which added byte-level indexing without proper UTF-8 boundary checks. When parsing SQL with Chinese text (e.g., \"WHERE name LIKE '%批量保存%'\"), the byte index could point to the middle of a multi-byte UTF-8 character, causing Rust to panic when slicing the string.

Changes:
- parse_usize_token: Use std::str::from_utf8 to validate slice before parsing
- mysql_keyword_at: Add is_char_boundary checks for safe string slicing
- mysql_top_level_limit: Skip non-UTF-8-boundary bytes after incrementing index

This ensures all byte-level indexing operations stay on valid UTF-8 boundaries, preventing panics while maintaining the original functionality of detecting LIMIT clauses for bounded result set collection.

Fixes crash reported by users when executing queries with Chinese conditions.

### DIFF
--- a/crates/dbx-core/src/db/mysql.rs
+++ b/crates/dbx-core/src/db/mysql.rs
@@ -1468,7 +1468,11 @@ fn mysql_top_level_limit(sql: &str) -> Option<usize> {
         if depth == 0 && mysql_keyword_at(sql, i, "LIMIT") {
             return parse_mysql_limit_value(sql, i + "LIMIT".len());
         }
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks @bigliangzao!

PR #1458, opened 2026-06-18
fix: speed up MongoDB large collection loading

linjianlin:
## Summary
- Use MongoDB estimated document counts for unfiltered collection browsing
- Keep accurate countDocuments behavior when a filter is applied

## Validation
- cargo check -p dbx-core --no-default-features


t8y2:
已合并，下个版本发布。

PR #1459, opened 2026-06-18
fix: lazy load system fonts in settings

linjianlin:
## Summary
- Defer system font enumeration until the font dropdown is opened
- Cache loaded system fonts to avoid repeated scans
- Avoid applying every system font as a preview style to reduce first-render jank

## Validation
- pnpm typecheck


t8y2:
已合并，下个版本发布。

PR #1470, opened 2026-06-19
fix: prevent WebDAV password action overlap

linjianlin:
## Summary
- hide the password visibility toggle when a saved WebDAV password is represented by the clear action
- replace the saved-password clear action with a fixed-position native button to avoid active press translation
- hide native password reveal/clear controls inside PasswordInput so only the custom eye icon appears

## Validation
- pnpm typecheck


t8y2:
Thanks! Merged.

PR #1475, opened 2026-06-19
feat: add column sort dropdown

linjianlin:
## Summary
- replace header sort cycling with an explicit dropdown menu
- add ascending, descending, and clear sort options in the column header
- reuse the same sort application flow for header and context-menu sorting

## Verification
- pnpm typecheck
- pnpm exec lint-staged


t8y2:
Thanks! Merged.

PR #1478, opened 2026-06-19
feat: Fix database connection cascade failure and enable auto-recovery across all drivers

dbin0123:
Fix database connection cascade failure across all drivers — single broken connection no longer kills all pools. Adds timeout protection, full-driver health checks, metadata retry, automatic stale pool cleanup on query failure, and pool refresh on system wake. No restart required for recovery.


t8y2:
Thanks @dbin0123! Merged.

PR #1480, opened 2026-06-19
feat(insert-template): add TDengine-aware insert template strategy and table-type propagation

aiLi0617:
## Summary
- Add a database-specific insert-template strategy system and implement TDengine handling.
- Generate proper TDengine STABLE insert SQL with `USING ... TAGS ... VALUES ...`.
- Propagate `table_type` through sidebar/object tree nodes so insert-template generation can detect special table kinds.
- Bump object-tree cache keys to `v3` to invalidate stale entries missing new metadata.

## Background
The previous insert template generation followed a generic path and could not represent TDengine STABLE semantics correctly.  
This change introduces strategy-based template generation and carries table type metadata end-to-end, so SQL output can match engine-specific requirements.

### 1) Insert-template strategy + TDengine SQL generation
- Added insert-template strategy context and strategy dispatch.
- Added TDengine-specific strategy:
  - Uses default insert behavior for non-STABLE tables.
  - Uses `INSERT INTO <child> USING <stable> TAGS (...) VALUES (...)` for STABLE tables.
- Updated timestamp placeholder behavior for TDengine to use `NOW`.

### 2) Table type metadata propagation
- Added optional `tableType` on tree nodes.
- Passed `table_type` into table/object tree entries.
- Preserved driver-reported table type in merged object metadata, so downstream template logic can distinguish table kinds reliably.

### 3) Cache invalidation
- Upgraded relevant object-tree cache keys from `v2` to `v3` to prevent stale cached nodes without `tableType`.

## Impact
- Correct insert template generation for TDengine STABLE tables.
- No behavior change intended for non-TDengine engines beyond metadata propagation and cache refresh.
- First load after upgrade may rebuild sidebar object caches due to key bump.

## Test Plan
- [ ] Generate insert template for a regular table and verify default SQL format remains correct.
- [ ] Generate insert template for a TDengine non-STABLE table and verify default-style insert output.
- [ ] Generate insert template for a TDengine STABLE table and verify SQL uses `USING + TAGS + VALUES`.
- [ ] Verify TAG columns are emitted in `TAGS (...)` and excluded from value columns.
- [ ] Reload sidebar/object tree and confirm metadata refresh works after cache-key migration.


t8y2:
Thanks @aiLi0617! Merged.

PR #1497, opened 2026-06-20
feat: support data grid column reordering

linjianlin:
## Summary
- add drag-and-drop column reordering for data grids
- persist column order per grid scope and add reset-order entry
- keep column widths attached to reordered column identities

## Verification
- pnpm typecheck
- pnpm vitest run packages/app-tests/dataGridColumnOrder.test.ts


t8y2:
Thanks @linjianlin! Merged.

PR #1498, opened 2026-06-20
feat(grid): copy column headers from results

chenhbb:


t8y2:
Thanks @chenhbb! Merged.

PR #1500, opened 2026-06-20
fix: avoid panicking when pubsub server fails

linjianlin:
## Summary
- Replace fatal PubSub WebSocket server `expect` calls with warning logs and graceful task exit.
- Prevent bind or server errors from panicking inside the background Tauri task.

## Verification
- `cargo build --manifest-path src-tauri/Cargo.toml --no-default-features --features duckdb-bundled,mq-admin`


t8y2:
Thanks @linjianlin! Merged.

PR #1619, opened 2026-06-23
fix: recover connection state after VPN failures

### LINKED ISSUE #641: [🐞 Bug] 由于网络原因连接建立失败，网络恢复后连接无法使用
### 数据库类型和版本

MySQL 8

### 问题描述

如题，连接某个内部数据库时发现 VPN 网络掉了，重新登录 VPN 后发现这个失败的连接始终在转圈的状态，无法打开也无法关闭，刷新也没用，必须强制结束进程重新打开软件才能建立连接

Image

### 错误日志

```shell

```

### DBX 版本

v0.5.27

### 操作系统

Windows

### 补充信息

_No response_

zucchiniEvader:
## Summary
- add frontend timeouts for connection metadata loading so sidebar rows do not stay loading forever after VPN/network failures
- make disconnect/reconnect detach stale backend pools and clear local connection loading state
- allow connection rows and search-result copies to recover from stale loading state

Fixes t8y2/dbx#641

## Verification
- pnpm test -- packages/app-tests/connectionStoreErrorState.test.ts packages/app-tests/sidebarSearchTree.test.ts
- pnpm typecheck
- cargo check -p dbx-web
- cargo check -p dbx


t8y2 (changes requested):
## PR #1619: fix: recover connection state after VPN failures

**+292/-18, 7 files | CI: pass (agents skipped) | Mergeable: MERGEABLE**

### 🔴 Must Fix

1. Non-SQL sidebar loaders can still hang forever  
`withMetadataLoadTimeout` is only applied in `loadDatabases`, but other connection-level metadata loaders still await raw API calls while holding `node.isLoading = true`: Redis at `apps/desktop/src/stores/connectionStore.ts:1172`, MQ at `1251`, Mongo at `1308`, Elasticsearch at `1343`, Vector at `1375`. If VPN/network loss makes any of those requests never settle, the `finally` block never runs and the connection row remains stuck loading, which is the exact class of issue this PR is meant to fix. These calls should also go through the metadata timeout wrapper.

2. Late connect success can still leave/replace a stale backend pool after a retry  
The timeout cleanup skips `api.disconnectDb` whenever `connectedIds` already contains the connection id. That protects the UI from disconnecting a newer successful retry, but it does not protect the backend: the original late `/connect` request can still finish after the retry and insert its pool under the same id, replacing the newer pool. The backend connect path inserts the pool at completion, so this needs an attempt/session guard or backend-side stale attempt protection rather than checking only `connectedIds` on the frontend.

### ✅ Well Done

- Good tests around hanging disconnects, metadata timeout for the SQL path, and late connect error reporting.
- Moving pool close into detached cleanup is the right direction for avoiding UI stalls on broken network paths.
- Clearing stale loading state from search result copies is a nice UX fix.

### Verdict

Request changes. The PR fixes the main SQL path, but the same stuck-loading failure remains for several supported connection types, and the late-connect retry race still has backend state risk.

t8y2:
Thanks for the contribution! Merged in 9a0a32e, will be released in the next version.

Issue #1664, opened 2026-06-23
[❓ Question] ai助手不出结果


FutureCZ:
### 描述

<img width="404" height="760" alt="Image" src="https://github.com/user-attachments/assets/35407e44-c374-42bf-9af5-dfcc51ee4bd3" />

<img width="405" height="828" alt="Image" src="https://github.com/user-attachments/assets/0519d5c1-3a49-44c6-a94e-adec215cb0f2" />

这是怎么回事，ai的问题吗？

### 环境信息

DBX版本：v.0.5.37
系统：Windows 11
数据库类型：Oracle➡️连接的库版本为10g

t8y2:
@Abeautifulsnow Please have a look at this.

Abeautifulsnow:
> [@Abeautifulsnow](https://github.com/Abeautifulsnow) Please have a look at this.

Noted, I'll take a look.

Abeautifulsnow:
Fixed/Improved in  #1706. Will be released in the next version , thanks for your attention.

Abeautifulsnow:
新版本还有此类问题吗？没有的话，烦请关闭一下issue~

t8y2:
Released in a published version.

PR #1706, opened 2026-06-24
fix(agent-loop): preserve output on exhaustion, interruption, and cancellation

Abeautifulsnow:
## Summary

When the AI Agent hits `MAX_AGENT_TURNS` (prev. 10), an interrupted stream, or cancellation mid-loop, the agent loop previously discarded all accumulated work — returning a bare error with no partial output, no token usage report, and no `AgentEnd` event. This PR ensures partial work is always preserved and the terminal event is always emitted.

### Core fix — preserve output on exhaustion / interruption / cancellation

- **`MAX_AGENT_TURNS` raised from 10 → 20** to reduce premature exhaustion for schema-exploration workflows.
- Introduced `LoopExit` enum (`Completed | Cancelled | Interrupted | Exhausted`) as a unified terminal state machine. Every exit path now appends an informative `TextDelta`, preserves `final_text`, and emits `AgentEnd` with token usage.
- Previously hitting the turn limit returned `Err(...)` with no output. Now returns `Ok(final_text)` with a "send Continue to let the agent keep working" hint.

### Context compaction now responds to cancellation

- `maybe_compact` now receives a `cancelled: &Notify` parameter and wraps the LLM summary call in `tokio::select!`. A cancellation signal short-circuits to a fallback summary instead of blocking the loop.
- Returns `CompactResult` (`Compacted | Skipped | Cancelled`) instead of `Option<u32>`, so the caller can distinguish "compacted" from "cancelled during compaction".

### Review-driven polish

- Extracted `AGENT_CANCELLED_ERROR` constant to replace brittle hardcoded string matches.
- Added `take_text()` helper with `unwrap_or_else(|e| e.into_inner())` poison recovery, deduplicating 4 repeated `lock().expect(...)` call sites. Switched from `tokio::sync::Mutex` to `std::sync::Mutex` for the accumulated text (lock held only within stream callback, never across `.await`).
- Added `LoopExit::should_break_turns()` to encapsulate the `Cancelled | Interrupted` early-break condition.
- Early stream errors (accumulated text empty) now route through `LoopExit::Interrupted` + `AgentEnd` instead of `return Err(...)`, ensuring consistent terminal events.
- Context-length retry `CompactResult::Skipped` path now also goes through the unified finalizer.
- `LoopExit::Interrupted` message now distinguishes empty vs. non-empty `final_text`, avoiding misleading "Partial output above was preserved" when nothing was produced.

### Tool execution — `get_columns` sequential execution

- Reverted `get_columns_tool.parallel_ok` from `true` → `false` with a rationale comment. Concurrent metadata queries can exhaust single-connection drivers (e.g. DuckDB), causing cascading tool errors. This reverts a mistaken review recommendation.

## Known follow-ups (not in this PR)

- Replace `tokio::sync::Notify` single-permit cancellation with `CancellationToken` or `Arc<AtomicBool>` (several `now_or_never()` calls compete for the same permit).
- Split `run_agent_loop` (~310 lines) into `run_single_turn` + `finalize_exit`.
- Partial assistant messages are not pushed to `conversation_messages` on Cancelled/Interrupted paths; needed for future backend-level "Continue" resume.
- System prompt tool list in frontend ([`ai.ts:201`](apps/desktop/src/lib/ai.ts#L201)) is hardcoded and does not match the backend's per-db-type tool registration (`supports_sql_query`).

## Files changed

| File | Δ |
|---|---|
| `crates/dbx-core/src/agent_loop.rs` | +125/-21 |
| `crates/dbx-core/src/agent_tools.rs` | +3/-1 |

## 变更类型

- [ ] 新功能
- [x] 功能优化
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 关联 Issue

#1664

### DIFF
--- a/crates/dbx-core/src/agent_loop.rs
+++ b/crates/dbx-core/src/agent_loop.rs
@@ -1,5 +1,5 @@
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use futures::future::join_all;
 use futures::FutureExt;
@@ -13,14 +13,26 @@ use crate::ai_cli_agent::CliAgentCommandSpec;
 use crate::connection::AppState;
 use crate::models::connection::DatabaseType;
 use crate::token_usage::TokenUsage;
-use tokio::sync::Mutex;
 
 /// Maximum number of agent loop turns to prevent infinite loops.
-const MAX_AGENT_TURNS: u32 = 10;
+const MAX_AGENT_TURNS: u32 = 20;
 const MAX_TOOL_RESULT_CONTEXT_CHARS: usize = 12_000;
 const TOOL_RESULT_HEAD_CHARS: usize = 4_000;
 const TOOL_RESULT_TAIL_CHARS: usize = 4_000;
 const TOOL_RESULT_SAMPLE_ITEMS: usize = 5;
+
+enum LoopExit {
+    Completed,
+    Cancelled,
+    Interrupted(String),
+    Exhausted,
+}
+
+enum CompactResult {
+    Skipped,
+    Compacted,
+    Cancelled,
+}
 
 /// Context for an agent loop run.
 pub struct AgentLoopContext {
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in ebfc50e, will be released in the next version.

PR #1725, opened 2026-06-24
fix(redis): add blur-to-exit for TTL inline edit (#1724)

lexmin0412:
## Problem

Closes #1724

Redis TTL inline edit mode (triggered by clicking the TTL badge) has two UX issues:
1. No visible cancel/close mechanism — users may not know to press Escape to exit
2. Escape conflicts with fullscreen mode (Escape exits fullscreen)

The only working exit paths were pressing Enter (saves) or Escape (cancels), but there was no blur-to-exit behavior.

## Changes

- **`RedisValueViewer.vue`**:
  - Clicking outside the edit area now cancels editing via a `document`-level `mousedown` listener — no blur-vs-click race condition
  - Save button uses `@click` (device-independent: mouse, keyboard Tab+Enter, assistive technology all work)
  - Input auto-focuses when entering edit mode via `nextTick` + template ref (more reliable than `autofocus` attribute)

## Verification

- [x] `npx vue-tsc --noEmit` — no new type errors
- [x] Click Save button — saves successfully
- [x] Click outside input — cancels edit
- [x] Tab to Save button + Enter — saves (keyboard activation preserved)
- [x] Press Enter — saves
- [x] Press Escape — cancels
- [x] Input auto-focuses when entering edit mode

PR #1778, opened 2026-06-25
refactor(redis): use onClickOutside for TTL edit a11y (#1724)

lexmin0412:
## Problem

Follow-up to #1725 (already merged).

The review at https://github.com/t8y2/dbx/pull/1725#pullrequestreview-4561336221 pointed out that using `@mousedown.prevent` on the save button dropped device-independent activation — keyboard/assistive users can't trigger save because `mousedown` doesn't fire for those input methods.

## Changes

- Replace `@blur="cancelEditTtl"` + `@mousedown.prevent="saveTtl"` with `onClickOutside` from `@vueuse/core`
- `onClickOutside` eliminates the blur-vs-click race condition entirely, so `@mousedown.prevent` is no longer needed
- Save button reverts to `@click` — works with mouse, keyboard Tab+Enter, and assistive technology
- Click outside the edit area cancels editing (original issue #1724 requirement preserved)

## Verification

- [x] `npx vue-tsc --noEmit` — no new type errors
- [x] Click Save button — saves successfully
- [x] Click outside input — cancels edit
- [x] Tab to Save button + Enter — saves (keyboard activation preserved)
- [x] Press Enter — saves
- [x] Press Escape — cancels

### DIFF
--- a/apps/desktop/src/components/redis/RedisValueViewer.vue
+++ b/apps/desktop/src/components/redis/RedisValueViewer.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { computed, ref, nextTick, onBeforeUnmount, onMounted } from "vue";
 import { useI18n } from "vue-i18n";
+import { onClickOutside } from "@vueuse/core";
 import { DynamicScroller, DynamicScrollerItem, RecycleScroller } from "vue-virtual-scroller";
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks for the contribution! Merged in 1ce85b6, will be released in the next version.

PR #1822, opened 2026-06-25
feat(mysql): support external catalogs (Paimon/Hive) via `catalog=` URL param

jischeng:
## Background

StarRocks/Doris expose external storage (Paimon, Hive, Iceberg, ...) through a **catalog**. A connection targeting e.g. `paimon_catalog.clip` means: catalog `paimon_catalog`, database `clip`.

dbx had **no catalog concept** at all — the StarRocks/Doris path reuses the MySQL driver, and the URL path's database was treated as a literal schema name. This caused two failures:

1. **Connection failure**: the database from the URL path (`clip`) was sent as the schema during the MySQL handshake. StarRocks looks it up in the *default* catalog, where `clip` doesn't exist, and rejects the connection with `Unknown database 'clip'` — *before* any `SET catalog` could run.
2. **Wrong catalog** (when the handshake happened to succeed): `SHOW DATABASES` / `SHOW TABLES` ran against the default catalog, so users saw the default catalog's tables instead of the Paimon ones.

## Root cause

- The MySQL handshake schema comes from the URL **path** (`mysql://host:port/<db>`). mysql_async sends it to the server at connect time, before pool setup queries run.
- `mysql_setup_queries` had no `SET catalog` — and even adding it naively doesn't help, because the handshake rejects the external-catalog database first.

## Fix (opt-in, single file: `crates/dbx-core/src/db/mysql.rs`)

Add an opt-in `catalog=<name>` URL parameter for mysql-family connections:

- **`mysql_async_url`**: when a `catalog=` param is present, strip the database **path** from the URL handed to mysql_async, so the handshake connects without a default schema. The original URL (with path) is still used for setup, so `USE <database>` is emitted correctly.
- **`mysql_setup_queries`**: emits `SET catalog = \`<name>\`` when the param is present. It is pushed **last** so mysql_async's back-to-front setup execution (`Vec::pop` in `run_setup_commands`) runs it **before** `USE <database>` — otherwise `USE` fails because the database lives in the external catalog. The pool re-runs setup after every connection reset, so the catalog stays current.
- **`mysql_connection_catalog`**: reads the `catalog=` param (percent-decoded).
- Add `catalog` to `is_dbx_handled_mysql_url_param` so mysql_async ignores the otherwise-unknown param.

### Why `SET catalog` + `USE` instead of `USE \`catalog\`.\`db\``?

StarRocks documents `USE catalog.db`, but the **backtick-quoted** form `USE \`catalog\`.\`db\`` is parsed as a plain `USE \`db\`` against the current catalog (at least on the 10.3.10.225 FE I tested against) — it does **not** switch catalogs. `SET catalog` + `USE` is the reliable form.

## Usage

Configure a Paimon/external-catalog connection as:
- `database`: `clip` (the database inside the catalog)
- `url_params`: `catalog=paimon_catalog`

Other mysql/StarRocks/Doris connections are **unaffected** — the feature is opt-in (only active when `catalog=` is set).

## Verification

Tested end-to-end against a StarRocks (FE 10.3.10.225) + Paimon catalog via the dbx MCP bridge (which routes SSH-tunneled connections through the running app, exercising the real Rust pool path):

| Operation | Result |
|---|---|
| `SHOW CATALOGS` | ✅ lists `paimon_catalog` |
| `SHOW DATABASES` (after setup) | ✅ lists paimon_catalog's databases (`clip`, ...) |
| `SHOW TABLES FROM \`clip\`` | ✅ returns the Paimon tables |
| `DESCRIBE` (`information_schema.COLUMNS` + `SHOW COLUMNS` fallback) | ✅ returns columns (e.g. a 320-column table) |
| `SELECT * FROM \`record_tag_t\` LIMIT 1` (unqualified) | ✅ resolves (no `No database selected`) |

Without the path-strip, the same connection failed at handshake with `Unknown database 'clip'`; without `SET catalog` before `USE`, it failed with `No database selected`.

## Tests

6 new unit tests in `db::mysql::tests` (setup-query ordering, catalog decode, path-strip when catalog present/absent). `cargo test -p dbx-core` → 1173 passed.

## Notes / non-goals

- This is connection-level catalog support (switch to one configured catalog), not a UI for browsing/switching between multiple catalogs. That would be a larger, separate change.
- `SELECT *` on a large Paimon table can be slow (lakehouse scan latency) — that's a Paimon/StarRocks read-performance matter, not addressed here.

### DIFF
--- a/crates/dbx-core/src/db/mysql.rs
+++ b/crates/dbx-core/src/db/mysql.rs
@@ -535,14 +535,28 @@ fn mysql_ssl_opts(
 
 fn mysql_setup_queries(url: &str) -> Vec<String> {
     let charset = mysql_connection_charset(url).unwrap_or("utf8mb4");
+    let catalog = mysql_connection_catalog(url);
+    let database = mysql_connection_database(url);
     let mut queries = Vec::new();
-    if let Some(database) = mysql_connection_database(url) {
-        queries.push(format!("USE {}", quote_identifier(&database)));
+    if let Some(database) = database.as_deref() {
+        queries.push(format!("USE {}", quote_identifier(database)));
     }
     if let Some(time_zone) = mysql_connection_time_zone(url) {
         queries.push(format!("SET time_zone = {}", quote_value(&time_zone)));
     }
     queries.push(format!("SET NAMES {charset}"));
+    // StarRocks/Doris expose external storage (Paimon, Hive, ...) through a
+    // catalog. `SET catalog` must run *before* `USE <database>` (the database
+    // lives in the external catalog and is unknown to the default one).
+    // mysql_async drains the setup list back-to-front (Vec::pop), so push it
+    // last to make it execute first. The handshake does not send the database
+    // as schema (see `mysql_async_url`, which strips the path when a catalog is
+    // configured), so the connection establishes in the default catalog and
+    // this setup query is what switches it. The pool re-runs these queries
+    // after every connection reset, so the catalog stays current.
+    if let Some(catalog) = catalog.as_deref() {
+        queries.push(format!("SET catalog = {}", quote_identifier(catalog)));
+    }
     queries
 }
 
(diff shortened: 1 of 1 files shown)


jischeng:
Here's how it looks after the fix:
image

t8y2:
Thanks for the contribution! Merged in a595d4d, will be released in the next version.

PR #1883, opened 2026-06-26
refactor(redis): move Redis scan page size from global settings to pe…

dal1wg:
## Purpose
Move the `redisScanPageSize` (COUNT per SCAN round‑trip) from the global Editor Settings dialog to the New/Edit Connection → Advanced panel, making it a per‑connection setting instead of a global one.

## Motivation

Different Redis environments often require different scan page sizes – a development instance with few keys works well with a small count, while a production cluster benefits from a larger stride. Making it connection‑level gives users fine‑grained control without affecting other connections.

### Backend
- Add `redis_scan_page_size: Option<u64>` to `ConnectionConfig` (and its data struct and `From` impl) with `#[serde(default)]` for backward compatibility – existing connections without the field deserialize as `None`.

### Frontend – type system
- Extend the `ConnectionConfig` interface with `redis_scan_page_size?: number`.
- Add shared constants (`REDIS_SCAN_PAGE_SIZE_DEFAULT`, `_MIN`, `_MAX`, `_OPTIONS`) as the single source of truth for bounds and defaults.

### Frontend – connection dialog (new location)
- Initialize the form with the default `1000`, propagate existing value via the `editConfig` loader, and include it in `connectionConfigForSubmit()` with range validation (clamped to 200–10 000; invalid → fallback to `1000`). Non‑Redis connections clean the field.
- Add a `Select` control in the Advanced tab (visible only when `db_type === 'redis'`) with options `[200, 1000, 5000, 10000]` and the original description text.

### Frontend – consumers switch to connection‑level value
- Replace `settingsStore.editorSettings.redisScanPageSize` with a computed value reading from `connectionStore.getConfig()?.redis_scan_page_size`.
- Fix `fillInitialKeyBatch`: the old logic tied the target batch count to the page size (`Math.ceil(pageSize/pageSize) = 1`), causing only 1‑2 SCAN iterations; the new logic loops up to 16 iterations (in batches of 8) until `hasMore` is false – ensuring a small page size (e.g., 200) does not starve the initial view.
- Update `listRedisCompletionKeys` to use `getConfig(connectionId)?.redis_scan_page_size ?? REDIS_SCAN_PAGE_SIZE_DEFAULT`.

### Frontend – old location removed
- Remove `redisScanPageSize` from the `EditorSettings` interface, defaults, normalization, and application logic.
- Delete the entire "Redis" tab from `EditorSettingsDialog.vue` (refs, handler, reset/save logic, `SettingsCategory` entry, navigation item, footer set, and template section).

## Backward Compatibility
- Existing connections without `redis_scan_page_size` use the default `1000` (same as before).
- The i18n keys (`settings.redisScanPageSize*`) are reused by the connection dialog – no translation changes required.
- The locale key `settings.redisTab` becomes orphaned; it is harmless and can be cleaned up separately.


t8y2:
Thanks for the contribution! Merged in d815a16, will be released in the next version.

PR #1898, opened 2026-06-26
fix(app): detect npm in common Windows paths

linjianlin:
## Summary
- add Windows fallback lookup paths for npm and MCP server command detection
- include nvm-windows, global npm bin, and standard Node.js install directories

## Checks
- cargo fmt --check
- cargo check


t8y2:
Thanks for the contribution! Merged in 476c5cbc, will be released in the next version.

PR #1953, opened 2026-06-26
fix(ai): fix Windows Codex CLI launch with npm-installed shims

Dreamescaper:
## Summary

Fix Windows Codex CLI startup when Codex is installed through npm.

On affected Windows machines, `where.exe codex` can return npm shims in this order:

```text
C:\Users\...\AppData\Roaming\npm\codex
C:\Users\...\AppData\Roaming\npm\codex.cmd
```

The previous behavior could accept the first existing path, including the extensionless npm shim. DBX then passed that path directly to `CreateProcess`, which failed with:

```text
%1 is not a valid Win32 application. (os error 193)
```

Typing `codex` in Windows Terminal still worked because the shell applies PATHEXT / shim resolution and runs `codex.cmd`. The issue only affected DBX launching the resolved path directly.

This does not happen with the standalone Codex installer, because it exposes a single launchable `codex.exe` path.

## Changes

- Prefer Windows-launchable Codex candidates: `.cmd`, `.exe`, `.bat`, `.com`.
- Skip extensionless npm shim paths when resolving `codex` from PATH.
- Apply the same launchable-path filtering in the Tauri-side command resolver used before DBX calls into core Codex execution.
- Add Windows regression tests for npm shim layouts.

### DIFF
--- a/crates/dbx-core/src/ai_codex_cli.rs
+++ b/crates/dbx-core/src/ai_codex_cli.rs
@@ -75,9 +75,7 @@ fn direct_program_path(program: &str) -> Option<String> {
     if path.is_absolute() && path.is_file() {
         #[cfg(windows)]
         {
-            if let Some(path) = windows_launchable_program_path(path) {
-                return Some(path);
-            }
+            return windows_launchable_program_path(path);
         }
         Some(path.to_string_lossy().to_string())
     } else {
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in def59969, will be released in the next version.

Dreamescaper:
Seems like this same issue was fixed 40 minutes ago in https://github.com/t8y2/dbx/pull/1935.
But nothing wrong should happen, just an extra hardening on top (It removes remaining fallback behavior where Windows resolution could still accept “any existing file” if no launchable .exe/.cmd/.bat/.com candidate was found.)

t8y2:
> Seems like this same issue was fixed 40 minutes ago in #1935.似乎这个问题在 40 分钟前已在 #1935 中修复。 But nothing wrong should happen, just an extra hardening on top (It removes remaining fallback behavior where Windows resolution could still accept “any existing file” if no launchable .exe/.cmd/.bat/.com candidate was found.)但不会造成任何错误，只是增加了一层额外的保护措施（它移除了剩余的备用行为，即如果找不到可启动的 .exe/.cmd/.bat/.com 文件，Windows 分辨率仍然可以接受“任何现有文件”。）

Yes, good job~

PR #2028, opened 2026-06-28
feat(grid): add "Go to column" toolbar button

kkk000111999:
Add a toolbar popover in DataGrid that lists all result columns with a search filter.

Selecting a column scrolls it into view, un-hides it if necessary, and highlights it briefly (reuses the existing scrollToTableInfoColumn helper).

i18n: en   zh-CN

image


t8y2:
Thanks for the contribution! Merged in 9fc2e25, will be released in the next version.

PR #2069, opened 2026-06-29
Keep first inserted grid row visible while editing

Caisin:
An empty table renders the placeholder branch before the first pending row exists, so the canvas scroller is not mounted when the original observer setup runs. Watching the visible-row transition reattaches the canvas observer after Vue mounts the scroller and gives the canvas/editor overlay real viewport dimensions.

Constraint: Default grid rendering mode is canvas and the scroller is gated behind hasVisibleRows.

Rejected: Force DOM row rendering for empty-table inserts | broader behavior change and bypasses the canvas rendering path.

Confidence: high

Scope-risk: narrow

Directive: Keep canvas observer lifecycle tied to both render mode and row-branch mounting; do not reduce it to render mode only without testing empty-table insertion.

Tested: pnpm exec vitest run packages/app-tests/canvasDataGridRenderer.test.ts packages/app-tests/dataGridEditor.test.ts

Tested: pnpm typecheck

Tested: pnpm lint

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -4620,7 +4620,16 @@ function drawCanvasGrid() {
   });
 }
 
-watch(useCanvasGridRows, () => nextTick(attachCanvasResizeObserver), { immediate: true });
+watch(
+  [useCanvasGridRows, hasVisibleRows],
+  () => {
+    // When an empty table gets its first pending row, the canvas scroller is created by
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks for the contribution! Merged in fd46869, will be released in the next version.

PR #2128, opened 2026-06-29
fix : Allow MySQL database creation without selected database

Driftlux:
## Summary

- Allow MySQL-compatible connections to execute `CREATE DATABASE` / `CREATE SCHEMA` without a selected database.
- Keep the existing selected-database requirement for ordinary MySQL queries.
- Add focused coverage for the empty-database execution guard.

## Root Cause

The query execution guard rejected every non-single-database connection when the active tab had an empty `database` value. On a fresh MySQL server where DBX only shows system databases and no user database is selectable, this blocks `CREATE DATABASE ...`, leaving users with no way to create the first database from the query editor.

## Validation

- `./node_modules/.bin/vitest run apps/desktop/src/composables/__tests__/useSqlExecution.spec.ts`
- `./node_modules/.bin/vue-tsc --noEmit --project apps/desktop/tsconfig.json`
- `./node_modules/.bin/oxlint apps/desktop/src/composables/useSqlExecution.ts apps/desktop/src/composables/__tests__/useSqlExecution.spec.ts`

### DIFF
--- a/apps/desktop/src/composables/__tests__/useSqlExecution.spec.ts
+++ b/apps/desktop/src/composables/__tests__/useSqlExecution.spec.ts
@@ -35,6 +35,14 @@ describe("requiresDatabaseSelection", () => {
     expect(requiresDatabaseSelection(queryTab(), connection("mysql"), "CREATE DATABASE app_db")).toBe(false);
   });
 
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in 427d0c5, will be released in the next version.

PR #2132, opened 2026-06-29
Feature: Support NixOS

juvevood:
## 变更说明

支持NixOS

## 变更类型

NixOS系统打包支持，不涉及dbx项目本身功能的更改。

## 涉及前端

无

## 验证

在NixOS系统的dbx项目根目录执行：`nix build .#dbx-desktop`

已验证成功编译。
已验证项目运行正常。

## 关联 Issue

无

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,82 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1782467914,
+        "narHash": "sha256-pGvFkM8N0xEkIIXDe5YYfbEAvHrk4IxBrjB/x8OomhE=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "e73de5be04e0eff4190a1432b946d469c794e7b4",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs",
+        "rust-overlay": "rust-overlay"
+      }
+    },
+    "rust-overlay": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for adding Nix/NixOS support. A root-level `flake.nix` makes sense if DBX wants to officially support Nix users.

I don't think this should be merged as-is yet. The main issue is the `dbx-web` package: the flake tries to skip default Cargo features with `cargoExtraArgs = "--no-default-features"`, but `buildRustPackage` does not consume that field. The supported field is `buildNoDefaultFeatures = true;`.

Also, even with `cargo build -p dbx-web --no-default-features`, `dbx-core` default features are still enabled because `crates/dbx-web/Cargo.toml` depends on `dbx-core` without `default-features = false`. That still pulls in `duckdb-bundled`, so the current comment/config about skipping DuckDB is misleading.

Could you either:

1. remove the `dbx-web` package from this PR and keep the scope to `dbx-desktop`, or
2. fix the Cargo feature wiring and Nix build fields so `dbx-web` actually builds without DuckDB?

It would also be helpful to add a Nix validation step, such as `nix flake check` or `nix build .#dbx-desktop`, otherwise this new packaging path can regress without CI coverage.

t8y2 (changes requested):
Thanks for adding Nix/NixOS support. A root-level `flake.nix` makes sense if DBX wants to officially support Nix users.

I don't think this should be merged as-is yet. The main issue is the `dbx-web` package: the flake tries to skip default Cargo features with `cargoExtraArgs = "--no-default-features"`, but `buildRustPackage` does not consume that field. The supported field is `buildNoDefaultFeatures = true;`.

Also, even with `cargo build -p dbx-web --no-default-features`, `dbx-core` default features are still enabled because `crates/dbx-web/Cargo.toml` depends on `dbx-core` without `default-features = false`. That still pulls in `duckdb-bundled`, so the current comment/config about skipping DuckDB is misleading.

Could you either:

1. remove the `dbx-web` package from this PR and keep the scope to `dbx-desktop`, or
2. fix the Cargo feature wiring and Nix build fields so `dbx-web` actually builds without DuckDB?

It would also be helpful to add a Nix validation step, such as `nix flake check` or `nix build .#dbx-desktop`, otherwise this new packaging path can regress without CI coverage.

juvevood:
> Thanks for adding Nix/NixOS support. A root-level `flake.nix` makes sense if DBX wants to officially support Nix users.
> 
> I don't think this should be merged as-is yet. The main issue is the `dbx-web` package: the flake tries to skip default Cargo features with `cargoExtraArgs = "--no-default-features"`, but `buildRustPackage` does not consume that field. The supported field is `buildNoDefaultFeatures = true;`.
> 
> Also, even with `cargo build -p dbx-web --no-default-features`, `dbx-core` default features are still enabled because `crates/dbx-web/Cargo.toml` depends on `dbx-core` without `default-features = false`. That still pulls in `duckdb-bundled`, so the current comment/config about skipping DuckDB is misleading.
> 
> Could you either:
> 
> 1. remove the `dbx-web` package from this PR and keep the scope to `dbx-desktop`, or
> 2. fix the Cargo feature wiring and Nix build fields so `dbx-web` actually builds without DuckDB?
> 
> It would also be helpful to add a Nix validation step, such as `nix flake check` or `nix build .#dbx-desktop`, otherwise this new packaging path can regress without CI coverage.

Thanks for your attention and feedback. allright, i removed dbx-web. its unnecessary.

t8y2:
Thanks for the contribution! Merged in 894ef0c, will be released in the next version.

PR #2138, opened 2026-06-29
feat(query): add manual transaction / auto-commit toggle to SQL editor

Abeautifulsnow:
## 变更说明

Description:

Summary

Adds a transaction control toggle to the SQL query editor toolbar, giving users explicit control over when changes are committed to the database. This is a standard feature in all major database clients (DBeaver, DataGrip, Navicat) and is especially useful for multi-statement write operations that need to be atomic.

What's New

Tx toggle button in the editor toolbar (shown only for Postgres, MySQL, SQLite):
- Auto-commit (default): each statement executes and commits independently — existing behavior, fully backward compatible
- Manual transaction: user controls BEGIN / COMMIT / ROLLBACK explicitly

Two transaction modes:

1. Single-execute wrap — when the toggle is on, all statements in a single Run are wrapped in one BEGIN … COMMIT block. Any failure rolls back the entire batch.
2. Session-level transaction — a connection is held open across multiple Run clicks. Commit (✓) and Rollback (↺) buttons appear in the toolbar once a transaction is active.

Safety mechanisms:
- 5-minute idle timeout → automatic ROLLBACK + connection released, warning banner shown
- Tab close → automatic ROLLBACK (covers close, close-other, close-all, close-by-filter)
- Switching back to auto-commit while a transaction is open → automatic ROLLBACK
- Connection pool is never permanently leaked

Database Support

| Database | Tx toggle | Session transaction |
|----------|-----------|---------------------|
| PostgreSQL | ✅ | ✅ |
| MySQL | ✅ | ✅ |
| SQLite | ✅ | ✅ |
| Others (Redis, MongoDB, Turso, ClickHouse, etc.) | hidden | hidden |

Changes

Backend (Rust):
- connection.rs: add TxnConnection enum and TransactionSession struct; add transaction_sessions map to AppState
- query.rs: add use_transaction field to QueryExecutionOptions; implement begin_manual_transaction, execute_in_manual_transaction, commit_manual_transaction, rollback_manual_transaction; add mysql_value_fallback and sqlite_value_ref_to_json helpers
- commands/query.rs: expose4 new Tauri commands + wire use_transaction through execute_multi
- lib.rs: register the 4 new commands

Frontend (TypeScript/Vue):
- databaseFeatureSupport.ts: add supportsTransaction() capability check (whitelist approach)
- types/database.ts: add autoCommit, txnSessionId, txnAutoRolledBack to QueryTab
- tauri.ts / http.ts / api.ts: add 4 new API functions (http stubs throw desktop-only error)
- queryStore.ts: setAutoCommit with auto-rollback on mode switch; commitTransaction / rollbackTransaction; manual transaction path in executeTabSql; rollback hooks in all4 tab-close paths
- EditorToolbar.vue: Tx toggle button, Commit/Rollback buttons (visible only when transaction active), auto-rollback warning banner, Agent-backend tooltip variant
- App.vue: wire new props and events

i18n: toolbar.autoCommit, manualTransaction, autoCommitAgent, manualTransactionAgent, commit, rollback, txnAutoRolledBack — all 8 locales (en, zh-CN, zh-TW, ja, es, it, pt-BR)

Tests:
- 3Rust unit tests for QueryExecutionOptions.use_transaction defaults
- 3 Vitest tests for supportsTransaction()

Known Limitations

- MySQL Date/Time/Datetime columns in a manual-transaction SELECT display as null (display-only issue; no data is corrupted; normal execution paths are unaffected)
- SQL execution within a manual transaction session cannot be cancelled via the Stop button (no cancellation token registered; consistent with the existing execute_in_transaction behavior)

Testing

- cargo check -p dbx-core ✅
- cargo check -p dbx ✅
- vue-tsc --noEmit ✅
- Pre-commit hooks (oxfmt + rustfmt) ✅

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

Related #2094


t8y2 (changes requested):
## PR #2138: feat(query): add manual transaction / auto-commit toggle to SQL editor

**+737/-36, 24 files | CI: frontend failing | Mergeable: MERGEABLE**

### 🔴 Must Fix

- `apps/desktop/src/lib/__tests__/databaseFeatureSupport.spec.ts:9`: frontend CI is failing because the test expects `supportsTransaction()` to return `true` for ClickHouse, SQL Server, Oracle, Dameng, rqlite, and agent, while the implementation only enables Postgres/MySQL/SQLite in `apps/desktop/src/lib/databaseFeatureSupport.ts:92`. Align the test or the implementation before merge.

- `crates/dbx-core/src/query.rs:2428`: the claimed 5-minute idle auto-rollback is only checked when `execute_in_manual_transaction` is called again. If the user opens a transaction and then goes idle, the session stays in `transaction_sessions` and the DB connection/transaction remains held indefinitely. This needs an actual cleanup job/timer, or the safety claim is not true.

- `crates/dbx-core/src/query.rs:2418`: manual transaction execution bypasses the existing read-only guard. Normal execution checks read-only mode through `check_read_only_for_connection*` before writes, but `execute_in_manual_transaction` directly runs SQL on the held connection. A read-only DBX connection can therefore execute writes by toggling manual transaction mode.

- `apps/desktop/src/stores/queryStore.ts:1862`: manual mode sends the whole `sqlToExecute` string to `executeInManualTransaction`, and the backend runs it as one statement instead of using DBX’s existing SQL splitter. That breaks the advertised multi-statement transaction use case and changes behavior from normal `executeMulti`.

### 🟡 Suggestions

- `crates/dbx-core/src/query.rs:2428`: the backend holds the global `transaction_sessions` write lock while awaiting database I/O. One slow manual transaction blocks commit/rollback/execute for all other manual transaction sessions. Consider removing the session from the map or using per-session locking around the connection instead.

### ✅ Well Done

- The UI state is scoped per tab and rolls back on the major close paths.
- The API surface is wired through Tauri cleanly, and unsupported HTTP usage fails explicitly.
- DBeaver’s model of explicit transaction controls/idle rollback was considered; the missing piece here is DBX having equivalent active cleanup rather than a lazy check.

### Verdict

Request changes. The feature direction is useful, but CI is red and the current backend path can leave transactions open indefinitely, bypass read-only protection, and mishandle multi-statement runs.

Abeautifulsnow:
@t8y2 Thanks for the thorough review! All 🔴 Must Fix items have been addressed, and the 🟡 Suggestions are also resolved. Please take another look.

Abeautifulsnow:
@t8y2 The code from the main branch has been merged into this branch, and merge conflicts have been resolved.

t8y2:
Thanks for the contribution! Merged in d55a708, will be released in the next version.

PR #2315, opened 2026-07-01
fix: optimize PrestoSQL metadata completion and JDBC connection handling

luoianun:
﻿## Summary

This PR improves PrestoSQL/Trino JDBC handling and SQL editor metadata completion.

Main changes:

- Avoid passing unsupported `connectTimeout` / `loginTimeout` JDBC properties to PrestoSQL/Trino drivers.
- Keep using `DriverManager.setLoginTimeout(...)` for login timeout handling.
- Show the JRE 21 runtime badge for the built-in PrestoSQL JDBC driver.
- Prevent schema-qualified table completion from leaking cached tables from other schemas.
- Add guarded on-demand table completion for PrestoSQL/Trino when users type a schema-qualified table prefix.
- Prevent table-name prefix completion from triggering column metadata lookups.
- Push PrestoSQL/Trino table completion `filter` and `limit` down to `information_schema.tables`.
- Use `information_schema.columns` for PrestoSQL/Trino column metadata instead of JDBC `DatabaseMetaData.getColumns()`.
- Add regression coverage for the PrestoSQL connection, completion, and metadata paths.

## Motivation

PrestoSQL metadata catalogs can be large, and broad metadata scans from SQL editor autocomplete may impact production clusters.

This PR keeps PrestoSQL/Trino editor metadata conservative by default, while allowing narrow on-demand table completion only when the user types a schema-qualified table prefix.

It also avoids unnecessary column metadata lookups while the user is still typing a table name prefix.

### JDBC connection properties

Some PrestoSQL JDBC driver versions reject the following connection properties:

- `connectTimeout`
- `loginTimeout`

The JDBC plugin now avoids adding these properties for PrestoSQL/Trino connections, while still applying the timeout through `DriverManager.setLoginTimeout(...)`.

### Driver presentation

The built-in PrestoSQL JDBC driver now shows the JRE 21 runtime badge consistently with other JDBC drivers.

### SQL editor completion

For PrestoSQL/Trino:

- Local-only metadata remains the default editor completion policy.
- On-demand remote table lookup is only allowed for schema-qualified table prefixes.
- The table prefix must meet a minimum length before the remote lookup is attempted.
- The lookup uses a small result limit.
- Schema-scoped local completion no longer falls back to cached tables from other schemas.
- Table-name completion no longer triggers column metadata loading for incomplete table prefixes.

### Metadata query optimization

For PrestoSQL/Trino table metadata:

- Table name filtering is pushed down to `information_schema.tables`.
- Result limiting is pushed down as well.

For PrestoSQL/Trino column metadata:

- Column metadata now uses `information_schema.columns`.
- This avoids relying on JDBC `DatabaseMetaData.getColumns()`, which may generate broader driver-specific metadata queries.

## Tests

Added or updated tests for:

- PrestoSQL built-in driver JRE 21 runtime presentation.
- PrestoSQL/Trino JDBC timeout property handling.
- Schema-scoped completion cache isolation.
- Schema-qualified table prefix completion context.
- PrestoSQL/Trino table metadata filter and limit pushdown.
- PrestoSQL/Trino column metadata via `information_schema.columns`.

Manual verification:

- Built local no-bundle Windows executable successfully with:

  ```bash
  pnpm exec tauri build --no-bundle
  ```

- Verified the generated executable includes the PrestoSQL completion and metadata changes.

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -156,6 +156,8 @@ const completionTranslations = computed(() => ({
   functionDescriptions: Object.fromEntries(SQL_FUNCTION_NAMES.map((name) => [name, t(`editor.completion.functionDescriptions.${name}`)])) as Record<string, string>,
 }));
 const MAX_COMPLETION_TABLES = 200;
+const PRESTO_ON_DEMAND_TABLE_COMPLETION_MIN_PREFIX = 2;
+const PRESTO_ON_DEMAND_TABLE_COMPLETION_LIMIT = 20;
 const MAX_JOIN_FK_PREFETCH_TABLES = 24;
 const MAX_SEMANTIC_DIAGNOSTIC_COLUMN_TABLES = 4;
 const liveFontSize = ref(settingsStore.editorSettings.fontSize);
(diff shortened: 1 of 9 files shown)


t8y2:
Thanks for the contribution! Merged in 95e3cb6, will be released in the next version.

PR #2432, opened 2026-07-02
feat(connection): add SSL/TLS support for StarRocks connections

d0zingcat:
## Summary

- Enable MySQL-compatible TLS for StarRocks (`db_type: starrocks` and MySQL + `starrocks` profile) while keeping the existing bare MySQL query/metadata path when TLS is disabled
- Expose TLS mode, CA certificate, and optional client cert/key in the connection dialog for StarRocks
- Route desktop/Tauri and `node-core` direct connections through `connect_with_ca_cert` / mysql2 SSL options when StarRocks TLS is enabled

This addresses cloud-managed StarRocks instances (e.g. Alibaba Cloud EMR Serverless) that require SSL for both VPC and public endpoints.

## Background

StarRocks currently uses the bare MySQL connection path, which intentionally strips `ssl-mode`, `require_ssl`, `verify_ca`, and related params to avoid driver parse errors (see #532). That is correct for plaintext clusters, but it prevents connecting to SSL-enforced instances.

There is no existing upstream PR for StarRocks SSL/TLS support. Related prior work only stripped incompatible params or improved bare MySQL compatibility.

## Test plan

- [x] `cargo test -p dbx-core --lib starrocks_`
- [x] `pnpm --filter @dbx-app/node-core test tests/starrocks-url.test.ts`
- [x] `pnpm test packages/app-tests/starrocksConnectionTls.test.ts`
- [x] `make cargo-check-fast`

**Required verification once CI is green:**

- [ ] Manual: create StarRocks connection with TLS mode **Verify CA**, attach CA PEM, test connection against SSL-enabled EMR Serverless instance
- [ ] Manual: confirm plaintext StarRocks clusters still connect with TLS mode **Preferred/Disabled**

## Related Issue

Related #532

## Illustration
The UI looks like below:
image

### DIFF
--- a/apps/desktop/src/components/connection/ConnectionDialog.vue
+++ b/apps/desktop/src/components/connection/ConnectionDialog.vue
@@ -1587,7 +1587,7 @@ const sqliteExtensionPaths = computed({
     form.value.url_params = setSqliteExtensionPaths(form.value.url_params, value);
   },
 });
-const tlsCapableDatabaseTypes = new Set<DatabaseType>(["mysql", "postgres", "redshift", "gaussdb", "kwdb", "opengauss", "questdb", "redis", "etcd", "clickhouse", "elasticsearch", "qdrant", "milvus", "weaviate", "chromadb", "influxdb"]);
+const tlsCapableDatabaseTypes = new Set<DatabaseType>(["mysql", "starrocks", "postgres", "redshift", "gaussdb", "kwdb", "opengauss", "questdb", "redis", "etcd", "clickhouse", "elasticsearch", "qdrant", "milvus", "weaviate", "chromadb", "influxdb"]);
 const supportsTlsToggle = computed(() => tlsCapableDatabaseTypes.has(form.value.db_type));
 const supportsCaCertificatePath = computed(() => form.value.db_type === "clickhouse");
 const supportsGenericUrlParams = computed(() => form.value.db_type !== "manticoresearch");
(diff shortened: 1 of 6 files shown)


t8y2 (changes requested):
## PR #2432: feat(connection): add SSL/TLS support for StarRocks connections

**+383/-18, 6 files | CI: failing | Mergeable: MERGEABLE**

### 🔴 Must Fix

1. `packages/node-core/src/database.ts:158`: CI fails because `mysqlTlsOptions()` returns `node:tls.ConnectionOptions`, but `mysql2` `PoolOptions.ssl` expects `string | SslOptions`. The failing frontend log is `TS2322` on `poolOptions.ssl = tls`. Build must be green before merge.

2. `crates/dbx-core/src/models/connection.rs:1117`: the new early `mysql_url_params_tls_disabled()` override regresses existing TiDB Cloud behavior. `rust-test` fails `tidb_cloud_mysql_url_requires_tls` because a `.tidbcloud.com` host now preserves `require_ssl=false` instead of forcing TLS. Scope this override to StarRocks/bare-MySQL cases or preserve the TiDB Cloud special case.

3. `apps/desktop/src/components/connection/ConnectionDialog.vue:1494`: native `db_type: "starrocks"` still cannot open the TLS tab because `supportsTlsToggle` excludes `starrocks`, while the tab itself is gated by `supportsTlsToggle` at lines 3154 and 4053. The MySQL+StarRocks profile path works, but the PR summary explicitly claims native StarRocks support too.

### 🟡 Suggestions

Add a small regression test around native `starrocks` connection-dialog TLS visibility or config serialization, since this PR changes UI gating and submit sanitization.

The manual StarRocks TLS validation items are still unchecked in the PR body; keep them as required verification once CI is fixed.

### ✅ Well Done

The Rust connector path reuses existing MySQL TLS/client identity handling instead of inventing a separate StarRocks stack, and the new URL-normalization tests cover the main plaintext vs TLS split.

I also checked DBeaver’s MySQL SSL handling locally; its pattern is consistent with explicit SSL enable/disable plus certificate material handling. For the TypeScript failure, I checked mysql2’s current TypeScript/API docs: `PoolOptions` extends `ConnectionOptions`, and `ssl` is typed as mysql2 `SslOptions`, not Node’s full `ConnectionOptions`.

### Verdict

Request changes. GitHub reports the PR mergeable, but CI is red and there are two real behavior gaps beyond the type error.

t8y2:
Thanks for the contribution! Merged in 23b9228, will be released in the next version.

PR #2482, opened 2026-07-03
fix(editor): avoid run gutter for MySQL REPLACE function

SpencerZhang:
## Summary
- Do not treat `REPLACE(` as a soft statement start when parsing SQL editor execution ranges.
- Keep MySQL `REPLACE INTO` recognized as an executable statement start.
- Add regression coverage for both MySQL `REPLACE(...)` function calls and `REPLACE INTO` statements.

## Testing
- pnpm exec oxfmt apps/desktop/src/lib/sqlStatementRanges.ts apps/desktop/src/lib/__tests__/sqlStatementRanges.spec.ts
- pnpm oxlint apps/desktop/src/lib/sqlStatementRanges.ts apps/desktop/src/lib/__tests__/sqlStatementRanges.spec.ts
- pnpm vitest run apps/desktop/src/lib/__tests__/sqlStatementRanges.spec.ts
- pnpm typecheck


t8y2:
Thanks for the contribution! Merged in cb42a41, will be released in the next version.

PR #2496, opened 2026-07-03
fix(ai): bound schema context and result preview to prevent oversized AI payloads

Abeautifulsnow:
## 变更说明

Two fixes to prevent oversized context payloads in the DB Assistant's schema and result preview, which could otherwise exceed token limits or degrade AI response quality.

Changes

1. Limit indexes and foreign keys per table in schema context

When a table has a large number of indexes or foreign keys (common in legacy/wide schemas), the full lists were included verbatim in the AI context, consuming excessive tokens. This change introduces configurable caps (maxIndexesPerTable, maxFksPerTable), both defaulting to 10, to keep the schema payload bounded.

- buildAiContext now accepts maxIndexesPerTable and maxFksPerTable options
- Both open tabs and mentioned tables respect these limits
- loadMentionedTableContext is updated to pass the limits through

2. Truncate large column values in result preview

Query result previews (shown to the AI as context) could contain enormous cell values — e.g. JSON blobs, base64-encoded data, or long text columns. A single oversized row could blow past token budgets. This change caps each column value at 200 characters in the formatted preview, appendi

Files changed

| File | Δ |
|------|---|
| apps/desktop/src/lib/ai.ts | +19 / −10 |

Testing

- Verified schema context with 50+ indexes/FKs per table — context payload stays bound
- Verified result preview with large JSON/text columns — values are truncated at 200 chars
- Existing behavior unaffected when tables have fewer indexes/FKs than the defaults

## 变更类型

- [ ] 新功能
- [ ] Bug 修复
- [x] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

### DIFF
--- a/apps/desktop/src/lib/ai.ts
+++ b/apps/desktop/src/lib/ai.ts
@@ -349,9 +349,11 @@ function formatSchema(context: AiContext): string {
     .join("\n\n");
 }
 
-export async function buildAiContext(tab: QueryTab, connection: ConnectionConfig, options: { maxTables?: number; maxColumnsPerTable?: number; mentionedTables?: AiTableMention[] } = {}): Promise<AiContext> {
+export async function buildAiContext(tab: QueryTab, connection: ConnectionConfig, options: { maxTables?: number; maxColumnsPerTable?: number; maxIndexesPerTable?: number; maxFksPerTable?: number; mentionedTables?: AiTableMention[] } = {}): Promise<AiContext> {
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks for the contribution! Merged in d6ef9f1, will be released in the next version.

PR #2567, opened 2026-07-05
feat: support ~/.ssh/config host aliases, login method selector & PostgreSQL extension management

zergmk2:
### SSH 隧道增强
1. **支持 ~/.ssh/config Host 别名**：解析用户本地 SSH config，允许直接使用配置好的 Host 别名建立隧道连接，无需重复填写主机地址、端口、密钥等参数。
2. **新增登录方式选择器**：在连接对话框中提供登录方式选择（密码/密钥/SSH Agent），优化连接配置体验。

### PostgreSQL 扩展管理
3. **扩展查看**：Schema 节点下新增 "Extensions" 分组，展开即可查看当前数据库已安装的扩展（名称、版本、注释）。
4. **扩展管理对话框**：右键 "Extensions" → "Manage Extensions..."，打开左右双栏对话框：
   - 左侧列出可安装的扩展（来自 `pg_available_extensions`），点击即可安装
   - 右侧列出已安装的扩展，点击即可删除
   - 支持 PostgreSQL 及兼容数据库（GaussDB、Kingbase、HighGo 等）
5. **后端实现**：在 postgres 驱动中新增 `list_extensions()` 和 `list_available_extensions()`，通过 `pg_extension` 和 `pg_available_extensions` 系统视图查询。

## 变更类型

- [x] 新功能

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 连接 PostgreSQL 数据库，展开 schema → Extensions 分组，验证扩展列表正确显示
- [ ] 通过 Manage Extensions 对话框安装/删除扩展，验证操作生效


t8y2 (changes requested):
I'm going to request changes for now. A couple of things need fixing before this is safe to merge:

1. The login method selector is UI-only, but the backend still authenticates from whatever credential fields are populated. Since changing the dropdown does not clear hidden fields, switching a saved password connection to "None" can still send the old password, and an alias-prefilled `key_path` can make the backend use a key while the UI still shows "Password". Please either make the backend honor `auth_method`, or scrub/normalize the non-selected credential fields before saving/submitting.

2. `IdentityFile` values from `~/.ssh/config` are copied as-is. Common values like `~/.ssh/id_ed25519` pass validation through `expand_tilde`, but `load_ssh_private_key()` later reads the literal unexpanded path and fails. The actual key loading path needs the same expansion.

Also, the PR title/body mention PostgreSQL extension management, but this branch only contains SSH-related changes. Please update the PR description/title or push the missing PostgreSQL changes.

t8y2 (changes requested):
I'm going to keep this as request changes for now.

1. The PR currently does not typecheck. `pnpm typecheck` fails because the web/http backend is missing `listExtensions` and `listAvailableExtensions`, `InstallExtensionDialog.vue` calls the `useToast()` return value as a function, and the SSH auth method typing makes the legacy `"agent"` branch unreachable.

2. The SSH config alias flow still breaks the common `IdentityFile` case. New SSH tunnels default `auth_method` to `"password"`, and selecting an alias only prefills `key_path`; it does not switch the method to `"key"`. Since the backend now honors `auth_method`, an alias with `IdentityFile ~/.ssh/id_ed25519` will try password auth with an empty password instead of key auth.

3. The PostgreSQL extension dialog filters available extensions only against extensions installed in the current schema. Extensions are unique database-wide, and `pg_available_extensions.installed_version` already exposes whether each extension is installed. As written, an extension installed in `public` can still appear as available under another schema and then fail on install.

Also, the branch is currently conflicting with `main`, so please rebase/merge main and resolve the conflicts before another review pass.

zergmk2:
All three review issues fixed:

1. TypeScript typecheck: added http.ts stubs, fixed useToast() destructuring, removed agent comparison
2. SSH alias auth: applySshConfigHostAliasPrefill now auto-switches auth_method to key when identity_file present
3. Extension dedup: use pg_available_extensions.installed_version (database-wide) to filter available extensions, added installed_version to ExtensionInfo type

vue-tsc + cargo check pass. Ready for re-review.

t8y2:
Thanks for the contribution! Merged in f800d80, will be released in the next version.

PR #2625, opened 2026-07-06
fix(mysql): handle 1045 Access Denied for strict ProxySQL/MyCat handshakes

wnyr:
**Problem:**
When connecting to MySQL through certain strict proxies (like ProxySQL or MyCat), passing the database name directly in the initial handshake URL results in an immediate `1045 Access Denied` error when trying to expand tables and fetch schema metadata.

**Solution:**
Modified `connect_mysql_metadata_pool` to include a fallback mechanism. If the initial scoped connection fails, it automatically retries with an unscoped (bare) URL and immediately executes `USE \`dbname\`;` as a setup query. This successfully bypasses proxy handshake limitations while maintaining the strictly scoped session context for table expansion and queries.

---

### 📸 Screenshots

**Before (Fails to expand tables due to ProxySQL handshake rejection):**
image
image

**After (Successfully bypasses the proxy restriction and lists tables):**
image
image

---

**Testing Strategy:**
Due to the complexity of simulating a strict ProxySQL/MyCat environment that explicitly rejects database schemas during the initial TCP handshake, a reliable automated unit test for this specific fallback is difficult to mock.

However, I have manually tested this in our production proxy environment. Before this patch, the connection consistently failed with `ERROR 1045` when extracting metadata schemas. With this patch, the connection successfully falls back to the bare connection and cleanly scopes the session via `USE db`, resolving the issue entirely.

### DIFF
--- a/crates/dbx-core/src/connection.rs
+++ b/crates/dbx-core/src/connection.rs
@@ -289,6 +289,22 @@ pub async fn connect_mysql_metadata_pool(
                     )
                     .await
                     .map(|pool| (pool, MysqlMode::Bare))
+                } else if let Some(db) = db_config.effective_database() {
+                    let mut unscoped_config = db_config.clone();
+                    unscoped_config.database = None;
+                    let unscoped_url = connection_url_for_endpoint(&unscoped_config, host, port);
+                    let mut retry_setup_queries = extra_setup_queries.clone();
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in dacd311, will be released in the next version.

PR #2634, opened 2026-07-06
feat(ai): split Ask/Agent action menus with task-oriented Agent actions

Abeautifulsnow:
## Background

The AI Assistant offered Ask and Agent modes, but both surfaced the **same** SQL action set (Generate / Explain / Optimize / Fix / Convert / Sample Data). This blurred the modes' responsibilities:

- **Ask** is expected to only *generate or process SQL text*.
- **Agent** is expected to *use tools, run read-only queries, and answer from real results*.

Seeing "Generate SQL" in Agent mode led users to think Agent also just returns SQL text, while Agent could auto-run `execute_query` without the UI making the generate → risk-check → execute → summarize flow visible.

## What changed

**Mode-specific action sets** (`AiAssistant.vue`)
- **Ask** menu (unchanged): Generate / Explain / Optimize / Fix / Convert / Sample Data — default `generate`.
- **Agent** menu (new): Query Data / Inspect Schema / Run & Explain / Generate (no run) — default `query`.
  - `generate` is shared so users can still request SQL-only output without execution; in Agent mode it is labeled **"Generate (no run)"**.
- Switching mode lands on that mode's default (Ask → `generate`, Agent → `query`). Vector DBs stay on `generate` (their action menu is hidden and they lack SQL tools).
- External Ask-style triggers ("Fix with AI", "Explain history") switch to Ask mode when invoked from Agent mode, so the action is valid and the menu reflects what runs.

**New task actions** (`ai.ts`, `aiSkills.ts`)
- `query`, `exploreSchema`, `executeAndExplain` added to `AiAction`, each with a full skill definition (zh/en user instruction, system rules, output contract) and a higher token budget for explanatory answers.

**Per-action backend contract** (`agent_loop.rs`)
- `query` → call `execute_query` and answer from real data (do not stop at SQL text).
- `exploreSchema` → use `list_tables` / `get_columns`; do not execute data queries unless asked.
- `executeAndExplain` → run the current SQL and explain the results.
- `generate` in Agent mode → produce SQL; execute only when the request explicitly asks for real data.

**Legacy fallback** (`aiAgentPlan.ts`)
- Task actions no longer emit a misleading "unsupported action" step; they execute via `execute_query` tool events (real-time step cards), not the legacy client-side auto-execute path. The now-dead `unsupported_action` reason was removed.

**i18n** — new action labels, placeholders, a `generateNoExec` label, and a task-oriented Agent mode hint, added to **all** locales (en, zh-CN, zh-TW, es, it, ja, pt-BR).

## Commits

1. `0bcf9bbc` feat(ai): add task-oriented Agent actions with per-action contract
2. `da64813b` feat(ai): split Ask/Agent action menus with mode-aware defaults
3. `18ec6dec` test(ai): cover Agent task-action contract and mode/action mapping
4. `bb9f09ad` fix(ai): correct action menu on mode-switched triggers and i18n gaps

## Review fixes (commit 4)

Found during a self-review pass and fixed:
- **#I01** — `triggerAction` from Agent mode (e.g. "Fix with AI") switched to Ask and set the action, but the mode-switch watch then reset it to the Ask default, so the menu showed "Generate SQL" during a fix. Added a `suppressModeActionReset` flag so programmatic mode switches do not clobber the explicitly-set action.
- **#C01** — Removed dead `unsupported_action` code (type variant + unreachable `skippedTitleKey` case).
- **#I02** — es/it/ja/pt-BR were missing the new Agent action keys → English fallback in the Agent menu. Added the translations.

## Verification

| Check | Result |
|---|---|
| `pnpm typecheck` | clean (only pre-existing `codemirror-vim` errors in an untouched file) |
| `pnpm lint` | no new issues |
| `cargo test -p dbx-core --lib agent_loop` | 18 passed (13 existing + 5 new contract tests) |
| `vitest aiActions.spec.ts` | 9 passed (mode/action mapping) |

Automated tests lock the deterministic layer: menu composition, mode defaults, and per-action contract strings for all three task actions.

## Still needs manual verification (LLM-behavior dependent)

- **Query Data** actually calls `execute_query` and answers from real results (not just SQL text).
- **Generate (no run)** does **not** auto-execute.
- Execution steps render as a clear generate → risk-check → execute → summarize flow.

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [x] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

d9d408f6-b973-478c-99a7-9cfdf304bd65

18dd66ef-e921-4912-816b-d73965080c25

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

Related #2145


t8y2:
Thanks for the contribution! Merged in 8539011, will be released in the next version.

PR #2657, opened 2026-07-06
fix(sqlfile): make folder list reactively update + add collapse & pat…

ahua8522:
…h display

The SQL Files panel stayed on "Loading..." after opening a folder, and only refreshed when a second folder was added.

Root cause: addFolder mutated the raw local object instead of the reactive proxy held in folders.value, so Vue never tracked the entries/loading writes until an unrelated reactive update forced a re-render. Now mutate via folders.value[idx] so the view updates as soon as the scan resolves.

Also:
- Run list_sql_files_in_folder async via spawn_blocking and prune noisy dirs (.git/node_modules/target/...) so large folders no longer freeze the UI.
- Show the full folder path next to the top-level folder name (smaller, dimmer).
- Make each top-level folder collapsible (click the header row).

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

image

## 验证

- [x] `make check` 通过
- [x] 相关测试通过


t8y2:
Thanks for the contribution! Merged in 7ff4d9f80, will be released in the next version.

PR #2688, opened 2026-07-06
fix(mongo): support ISODate() and new Date() in MongoDB queries

hieptuanle:
## Problem

MongoDB shell queries using `ISODate(...)` (or `new Date(...)`) in the filter fail. Depending on position:

```js
// 1) Rejected outright
db.trainingdocuments.find({ createdAt: ISODate("2025-02-25T04:57:39.965Z") })
//    -> Query Error — Use MongoDB-specific commands   (before this PR)
//    -> unknown operator: $date                        (frontend-only fix)

// 2) Silently returns nothing
db.trainingdocuments.find({ createdAt: { $gte: ISODate("2025-02-25T04:57:39.965Z") } })
```

## Root cause (two layers)

**Frontend parser** — MongoDB queries are parsed by `apps/desktop/src/lib/mongo/mongoShellCommand.ts` before dispatch; when it fails, the query falls through to the generic SQL executor, which rejects it for MongoDB connections (`crates/dbx-core/src/query.rs`, `"Use MongoDB-specific commands"`). `normalizeJsonArgument()` only rewrote `ObjectId(...)` into extended JSON, so a bare `ISODate("...")` was invalid JSON, `JSON.parse` threw, and the command was treated as unrecognized.

**Backend filter conversion** — even once `{"$date":...}` reaches the backend, the *filter* path (`json_filter_value_to_bson`) only decoded `{"$oid":...}`, not `{"$date":...}` — unlike `json_value_to_bson` (used for inserts/aggregate/sort, which already handles `$date`). So `{"$date":...}` reached the server verbatim: as a top-level field value it became the invalid query operator `$date`, and inside `$gte` it compared against a raw sub-document and matched nothing.

## Fix

1. `normalizeJsonArgument()` also rewrites `ISODate(x)` / `new Date(x)` → `{"$date":x}`, mirroring the existing `ObjectId(x)` → `{"$oid":x}` rewrite.
2. `json_filter_value_to_bson` decodes `{"$date":...}` extended JSON into a BSON `DateTime`, mirroring `json_value_to_bson`. This fixes both direct equality and range operators, and also benefits anyone typing `{"$date":...}` directly.

## Testing

- Frontend: 2 tests in `packages/app-tests/mongoShellCommand.test.ts` (`ISODate` filter, `new Date` + single-quoted). `vitest run` → all pass. `pnpm typecheck` passes.
- Backend: `json_filter_to_document_decodes_extended_json_dates` in `crates/dbx-core/src/db/mongo_driver.rs` covers direct equality and `$gte`. `cargo test -p dbx-core` passes.
- Built the desktop app and confirmed the reported queries (`ISODate` equality and `$gte`) now return correct results.

Existing `ObjectId(...)` and plain-filter parsing are unchanged (covered by existing tests).

### DIFF
--- a/apps/desktop/src/lib/mongo/mongoShellCommand.ts
+++ b/apps/desktop/src/lib/mongo/mongoShellCommand.ts
@@ -618,7 +618,14 @@ function parseCollectionMethodTarget(source: string, method: string): { collecti
 function normalizeJsonArgument(value: string): string | null {
   const trimmed = value.trim();
   if (!trimmed) return "{}";
-  const preprocessed = quoteUnquotedObjectKeys(convertSingleQuotedStrings(trimmed.replace(/ObjectId\s*\(\s*["']([^"']+)["']\s*\)/g, '{"$oid":"$1"}')));
+  // Rewrite mongo shell constructors that are not valid JSON into the extended
+  // JSON the backend understands (mongo_driver::json_value_to_bson): ObjectId(x)
+  // -> {"$oid":x} and ISODate(x)/new Date(x) -> {"$date":x}. Without this a
+  // filter such as { createdAt: { $gte: ISODate("...") } } fails JSON.parse,
+  // the command is left unrecognized and falls through to the SQL executor,
+  // which rejects it with "Use MongoDB-specific commands".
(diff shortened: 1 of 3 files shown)


t8y2:
Thanks for the contribution! Merged in 78019c4, will be released in the next version.

Issue #2683, opened 2026-07-06
[Bug] HANA数据库查询


flextable:
### 数据库类型和版本

HANA

### 问题描述

查询时提示选择数据库 数据库里只有一个默认库

<img width="1340" height="174" alt="Image" src="https://github.com/user-attachments/assets/55b25c2e-127f-487d-86ee-260a7d381c23" /> SAP HANA 查询一直提示 请选择数据库 实际已经选择了默认库

### DBX 版本

0.5.47

### 操作系统

Windows

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

t8y2:
Fixed in `2fd2b0e77`, will be included in the next release.

Root cause: The database-selection guard treated an empty default database on tree-schema databases such as SAP HANA as a missing database selection.

Fix: Tree-schema databases can now execute queries when the active tab uses the default database value, and the regression test covers the SAP HANA default-database case.

t8y2:
Released in a published version.

PR #2689, opened 2026-07-06
fix: allow default database for tree-schema databases

gggaiitx:
TREE_SCHEMA mode databases (saphana, jdbc, postgres, etc.) use an empty string to represent the default database selection. The previous check treated empty string as no database selected and incorrectly blocked query execution.

## Reproduction

1. Connect to SAP HANA (or any tree-schema database via JDBC, e.g., Megajay)
2. Select "Default" in the database dropdown
3. Execute any query: `SELECT * FROM MOMX_MES.Z_SHIPMENT_INFORMATION`
4. Error: **"请先选择数据库 / Select a database first"**


t8y2:
关联https://github.com/t8y2/dbx/issues/2683

t8y2:
Thanks for the contribution! Merged in 2fd2b0e, will be released in the next version.

PR #2703, opened 2026-07-06
fix(milvus): send empty JSON body to databases/list endpoint

zhuhaoh:
## Summary

Follow-up to #2698.

`list_milvus_databases()` called `/v2/vectordb/databases/list` with no request body. This is accepted by the Milvus demo cluster used during testing, but production Milvus rejects a bodyless POST: it returns HTTP 200 with `{"code":1801,...}` and omits the `data` field. Because the success check only inspects the HTTP status, the function parsed zero names and silently fell back to `["default"]`, hiding every non-default database in the sidebar.

## Fix

Send an empty JSON object `{}` as the body, matching every other Milvus v2 endpoint. With the body, the endpoint returns code 0 and the full database list on both demo and production clusters.

## Verification

- Confirmed against a production Milvus cluster: bodyless POST → `{"code":1801,...}` (no `data`); `{}` body → code 0 with the full database list.
- Runtime-checked in both the Tauri desktop and web builds: non-default databases now appear in the sidebar.

Ref #2693

### DIFF
--- a/crates/dbx-core/src/db/vector_driver.rs
+++ b/crates/dbx-core/src/db/vector_driver.rs
@@ -183,7 +183,11 @@ pub async fn list_databases(client: &VectorClient) -> Result<Vec<String>, String
 async fn list_milvus_databases(client: &VectorClient) -> Result<Vec<String>, String> {
     // Older Milvus versions (pre-2.2) do not expose the databases endpoint; fall back to "default"
     // so the connection stays browsable instead of failing the whole tree load.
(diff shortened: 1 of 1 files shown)


zhuhaoh:
image
本次优化，解决了生产中milvus不同版本请求body不同的时候返回的情况，兼容性更好

t8y2:
Thanks for the contribution! Merged in 4024dc6, will be released in the next version.

PR #2706, opened 2026-07-06
fix(sidebar): pin schema row for Dameng/Oracle-style trees

### LINKED ISSUE #1747: [💡 Feature] 连接目录树优化
### 当前痛点

当前如果数据库内的表数量比较多，会存在两个问题。
1. 翻到偏下的位置，不确定当前连接的是哪个连接和数据库。
2. 无法快速切换其他的数据，往上翻再折叠比较麻烦。

### 期望方案

1. 当翻到页面偏下的位置时，最上面显示当前连接的数据库或连接信息。如下图

Image

2. 添加一个快速折叠按钮。
Image

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

haipengno1:
Dameng, Oracle, and oceanbase-oracle expose `connection -> schema -> tables` (no database node, via connectionUsesVisibleSchemaFilter), so the sticky header found no database ancestor to pin while scrolling and never appeared.

Rename DATABASE_LEVEL_TYPES to STICKY_CONTAINER_TYPES and add "schema" as a fallback. The upward walk still prefers `database` when present (Postgres/SQLServer), so schema only takes over for connections that genuinely lack a database node.

## 变更说明

修复左侧树状列表在达梦（Dameng）、Oracle、oceanbase-oracle 这类连接下，滚动浏览大量表时数据库行不吸顶的问题。

## 变更类型

- [x] Bug 修复

## 涉及前端

image

## 验证

请参见截图

## 关联 Issue

Close #1747

### DIFF
--- a/apps/desktop/src/components/sidebar/ConnectionTree.vue
+++ b/apps/desktop/src/components/sidebar/ConnectionTree.vue
@@ -153,18 +153,19 @@ const SEARCH_SCOPE_TO_NODE_TYPES: Record<SearchScope, TreeNodeType[]> = {
   view: ["view"],
 };
 
-// Types that can be pinned at the top of the tree while scrolling. When
-// browsing a large number of children (e.g. hundreds of tables) under one of
-// these and scrolling down, the row is kept visible so the active container
-// stays identifiable and can be collapsed with one click.
+// Sticky-row container types. When browsing a large number of children (e.g.
+// hundreds of tables) under one of these and scrolling down, the row is kept
+// pinned at the top so the active container stays identifiable and can be
+// collapsed with one click.
 //
-// `schema` is included as a fallback: Dameng / Oracle / oceanbase-oracle expose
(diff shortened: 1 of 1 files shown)


t8y2 (changes requested):
I'm going to request changes for now.

`ConnectionTree.vue` currently adds `schema` to the same sticky container set as database-level nodes, but `stickyNode` scans upward from the topmost visible row and returns the first match. In a normal `connection -> database -> schema -> table` tree, that first match is `schema`, so Postgres/SQLServer-style trees will start pinning schema instead of database. That contradicts the comment that schema never shadows database and regresses the existing database sticky behavior.

Please keep database-level containers preferred, and only fall back to `schema` when the current path genuinely has no database container.

t8y2:
Thanks for the contribution! Merged in d101abe7, will be released in the next version.

PR #2739, opened 2026-07-07
feat: synchronize tab titles with saved SQL file renaming

### LINKED ISSUE #2648: 新建查询重命名后，标签页标题未同步更新
**来源**: QQ 群反馈 (by NO)

## AI 摘要
在新建查询并对其重命名后，对应的标签页标题没有实时同步更改，仍显示旧名称或默认名称，导致用户混淆。

## 原始描述
新建查询，重命名后，标签页没有同步更改

jeeinn:
## 变更说明
- Update tab title to reflect the saved SQL file name when renamed.
- Ensure that the title is normalized to include the .sql extension if applicable.
- Implement tests to verify that renaming a saved SQL file updates linked tab titles correctly.

## 变更类型

- [x] 新功能
- [x] Bug 修复

## 涉及前端

图片

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

## 关联 Issue

Close #2648

### DIFF
--- a/apps/desktop/src/components/layout/SqlLibraryPanel.vue
+++ b/apps/desktop/src/components/layout/SqlLibraryPanel.vue
@@ -16,6 +16,7 @@ import { useQueryStore } from "@/stores/queryStore";
 import { useSettingsStore } from "@/stores/settingsStore";
 import { focusSidebarRenameInput } from "@/lib/sidebar/sidebarRenameFocus";
 import { savedSqlFolderBranchFileCount } from "@/lib/savedSql/savedSqlFolderCounts";
+import { ensureSqlExtension, stripSqlExtension } from "@/lib/savedSql/savedSqlFileName";
 import type { SavedSqlFile, SavedSqlFolder } from "@/types/database";
 
 const { t } = useI18n();
@@ -73,16 +74,8 @@ function importConnectionIdForFolder(folder?: SavedSqlFolder) {
   return folder?.connectionId || activeImportConnectionId();
 }
 
-function ensureSqlExtension(name: string) {
-  return /\.sql$/i.test(name) ? name : `${name}.sql`;
-}
-
 function sanitizeFileSystemSegment(name: string) {
   return name.replace(/[<>:"/\\|?*\u0000-\u001F]/g, "_").trim() || "untitled";
-}
-
-function stripSqlExtension(name: string) {
-  return name.replace(/\.sql$/i, "");
 }
 
 function relativeImportName(baseDir: string, filePath: string) {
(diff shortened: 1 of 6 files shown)


t8y2 (changes requested):
### 🟡 Suggestions

1. **Case-sensitive `.sql` check inconsistent with existing helper**
   `renameTab` uses `trimmed.endsWith(".sql")` (case-sensitive), while the existing `ensureSqlExtension` in `SqlLibraryPanel.vue` uses `/\.sql$/i` (case-insensitive). A saved file named `report.SQL` (uppercase extension, which the local-directory sync can produce) would be normalized to `report.SQL.sql`, double-appending the extension and renaming the file on disk. Consider matching the existing case-insensitive behavior — either reuse/export `ensureSqlExtension` or use `/\.sql$/i.test(trimmed)`.

2. **Fire-and-forget rename swallows errors, can desync tab title from file name**
   `renameTab` now does `void savedSqlStore.renameFile(...)`. `renameFile` performs `await api.saveSavedSqlFile(...)` and `await syncToLocalDirectory()`; if either throws, the error is silently dropped. The tab title has already been set to the new name, so the tab and the persisted file diverge (and `hydrateSavedSqlTabs` keeps `customTitle` on reload, so they stay diverged). Consider wrapping in `try/catch` with a `console.warn` (matching the pattern in `recordFileUsage`) or surfacing via `toast`, and reverting `tab.title` on failure.

3. **Minor: `.sql` normalization now duplicated**
   The "append `.sql` if missing" logic exists in both `SqlLibraryPanel.vue` and now `queryStore.ts`. Given `ensureSqlExtension` isn't exported, hardcoding here is defensible, but if more call sites appear it's worth extracting into a shared util.

### ✅ Well Done

- Bidirectional sync is correctly implemented without recursion: `renameTab → renameFile` updates the file, and `renameFile` updates all linked tabs directly (no call back to `renameTab`), so there's no infinite loop.
- The `existing.name !== normalizedTitle` guard avoids redundant writes when the tab title is re-committed unchanged.
- The dynamic `await import("@/stores/queryStore")` in `savedSqlStore.ts` correctly avoids the circular import created by `queryStore.ts` statically importing `useSavedSqlStore`.
- `renameFile`'s tab loop correctly handles the multi-tab case (several tabs open against the same saved SQL file all get updated).
- Test coverage is solid: both directions (rename file → tab syncs, rename tab → file syncs) are covered, including the `.sql` normalization and the `saveSavedSqlFile` payload assertion.

t8y2:
Thanks for the contribution! Merged in 22c7668, will be released in the next version.

PR #2754, opened 2026-07-07
fix(ai): make reasoning expand/collapse consistent across messages

Abeautifulsnow:
## 变更说明

The AI assistant's reasoning panel had inconsistent expand/collapse behavior: it auto-expanded during streaming (when isThinking was true) and auto-collapsed after the stream ended, requiring users to repeatedly toggle it for every new message.

Change

Replaced the per-message expandedReasoning: Set<number> with a single global reasoningExpanded: boolean ref. The toggle now affects all messages uniformly — both past and future — so the user sets it once and it stays. Defaults to collapsed.

Before

- Each message tracked its own expand state independently
- Streaming forced auto-expand (|| msg.isThinking), then collapsed after
- Inconsistent: users had to re-toggle every new message

After

- Single global toggle shared by all messages
- No auto-expand/auto-collapse — purely user-controlled
- Default collapsed (user clicks to expand when they want to see the reasoning)
- Consistent behavior across the entire conversation

Diff

apps/desktop/src/components/editor/AiAssistant.vue — 1 file, −13 / +7 lines

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

### DIFF
--- a/apps/desktop/src/components/editor/AiAssistant.vue
+++ b/apps/desktop/src/components/editor/AiAssistant.vue
@@ -511,7 +511,7 @@ function appendAssistantReasoning(assistantIdx: number, delta: string) {
   scrollToBottom();
 }
 
-const expandedReasoning = ref<Set<number>>(new Set());
+const reasoningExpanded = ref(false);
 const expandedSteps = ref<Set<string>>(new Set());
 
 function toggleStep(key: string) {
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks for the contribution! Merged in 09f9e44, will be released in the next version.

PR #2763, opened 2026-07-07
fix(ssh): add legacy SHA-1 MAC algorithms for older SSH servers

gggaiitx:
## Problem
SSH tunnel connection fails with `No common Mac algorithm` when connecting to legacy SSH servers that only offer `hmac-sha1`.

## Root Cause
`russh` defaults to `SAFE_HMAC_ORDER` which excludes SHA-1 MAC variants (`hmac-sha1`, `hmac-sha1-etm@openssh.com`). This is a secure default, but it prevents connections to older SSH servers that only support SHA-1 MACs.

## Fix
Append `hmac-sha1-etm@openssh.com` and `hmac-sha1` as fallback MAC algorithms (after the safe SHA-256/SHA-512 variants), matching the existing pattern already used for legacy KEX algorithms in the same function.

### DIFF
--- a/crates/dbx-core/src/db/ssh_tunnel.rs
+++ b/crates/dbx-core/src/db/ssh_tunnel.rs
@@ -10,7 +10,7 @@ use base64::Engine;
 use russh::client::{self, Config, Handle};
 use russh::keys::agent::{client::AgentClient, AgentIdentity};
 use russh::keys::{decode_secret_key, key::PrivateKeyWithHashAlg, PrivateKey};
-use russh::{kex, ChannelMsg, Preferred};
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks for the contribution. This ended up overlapping with `adb8da3ad`, which had already added the `hmac-sha1` fallback on `main`, so this PR now conflicts with the current branch.

I incorporated the useful extra part from this PR, `hmac-sha1-etm@openssh.com`, in `ef39c2bfe`. That keeps SHA-1 MAC variants as last-resort fallbacks after the safe defaults while covering both legacy variants.

Closing this PR as covered by `ef39c2bfe`.

PR #2773, opened 2026-07-07
fix: preserve large integer precision when formatting JSON in value editor

gggaiitx:
## Problem
When clicking "Format JSON" in the value editor, large integers exceeding `Number.MAX_SAFE_INTEGER` (2^53 - 1) lose precision. For example, `87712409002717401` becomes `87712409002717400`.

## Root Cause
JavaScript's `JSON.parse()` converts all JSON numbers to `number` type, which uses IEEE 754 double-precision floating point. Integers beyond 2^53 - 1 cannot be represented exactly, causing silent precision loss.

The affected code paths all used the pattern:
```js
JSON.stringify(JSON.parse(text), null, 2)
```

## Fix
Introduced a `safeJsonFormat()` utility function that:
1. Pre-processes the JSON text with a regex to replace numeric literals whose integer part exceeds `MAX_SAFE_INTEGER` with string placeholders
2. Runs `JSON.parse()` + `JSON.stringify()` normally
3. Restores the original numeric literals from placeholders in the output

This preserves the original digit sequences exactly while still getting proper JSON formatting (indentation, whitespace normalization).

### Affected locations (all fixed)
- **Value editor** (`cellDetailPresentation.ts`) — format/compact JSON
- **Redis value viewer** (`RedisValueViewer.vue`) — format/compress JSON
- **KV value display** (`kvValueDisplay.ts`) — pretty-print JSON
- **MQ raw API panel** (`RawApiPanel.vue`) — format JSON body

## Testing
- Added 10 test cases covering: large positive/negative integers, MAX_SAFE_INTEGER boundary, floating point with large integer parts, nested JSON, arrays, string values that look like numbers
- `pnpm check` passes (format, lint, typecheck, test)

## Link
[#2765](https://github.com/t8y2/dbx/issues/2765)

### DIFF
--- a/apps/desktop/src/components/editor/AiAssistant.vue
+++ b/apps/desktop/src/components/editor/AiAssistant.vue
@@ -82,6 +82,7 @@ const queryStore = useQueryStore();
 const { openTableTarget } = useNavigationTargets({
   showFieldLineageDialog: ref(false),
   showDatabaseSearchDialog: ref(false),
+  showDiagramDialog: ref(false),
 });
 const { toast } = useToast();
 const { isDark } = useTheme();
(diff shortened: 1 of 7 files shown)


t8y2:
Thanks for the contribution! Merged in 9410bf4, will be released in the next version.

PR #2823, opened 2026-07-08
fix(sql): quote all IN-list values uniformly in expaste

gggaiitx:
Previously, pure numeric values (e.g. 90001543) in the IN clause were not quoted while values with leading zeros (e.g. 00040787) were, causing inconsistent quoting. Now all non-NULL values are uniformly quoted.

## 处理
移除 formatSqlLiteral 中数字不加引号的逻辑，所有非 NULL 值统一加单引号

## 关联 Issue
Related #2817


t8y2:
Thanks for the contribution! Merged in 6041c3f, will be released in the next version.

PR #2848, opened 2026-07-08
fix(grid): preserve current page on data refresh instead of resetting to page 1

Abeautifulsnow:
### Problem

When refreshing table data (toolbar refresh button, Ctrl+R shortcut, or
auto-refresh timer), the pagination always resets to the first page. Users
viewing data on later pages must manually navigate back after every refresh,
which is especially frustrating when monitoring frequently-changing data at
the tail end.

### Root Cause

In `DataGrid.vue`, both `onToolbarRefresh()` and `onToolbarRollback()`
hardcoded the reload offset to `0`:

    emit("reload", ..., pageSize.value, 0);  // offset always 0 -> page 1

### Fix

Replace the hardcoded `0` with `(currentPage.value - 1) * pageSize.value`,
which computes the correct offset for the current page. This matches the
existing pattern already used in `useDataGridEditor.ts:1123`.

    emit("reload", ..., pageSize.value,
      (currentPage.value - 1) * pageSize.value);

Infinite-scroll is safe: `resetInfiniteScrollState()` is already called
before the emit when `infiniteScrollEnabled` is true, which sets
`currentPage = 1`. The computed offset evaluates to 0 — identical to the
original behavior. No additional branch needed.

### Files Changed

| File | Change |
|---|---|
| apps/desktop/src/components/grid/DataGrid.vue | 2 lines |

---

### Problem

With the offset-preservation fix above, if data is deleted while the user
is viewing a later page (e.g. page 5 with only 2 pages remaining), the
refresh correctly stays at page 5 but shows an empty result — the offset
no longer maps to any data.

### Solution

After a refresh/rollback completes, if the backend reports a totalRowCount
and the current page exceeds the last available page, auto-navigate to the
last page:

    pageSize=100, currentPage=5, totalRowCount shrinks to 200
    -> lastPageNum = ceil(200/100) = 2
    -> currentPage (5) > lastPageNum (2) -> redirect to page 2

A transient `isRefreshingData` flag distinguishes refresh/rollback-triggered
loading cycles from normal pagination navigation, preventing false triggers.

Edge cases handled:
- totalRowCount is 0 or undefined -> guard returns, no redirect
- All data deleted -> Math.max(1, ...) ensures last page is at least 1
- Current page still valid -> page <= lastPageNum guard, no-op
- Normal pagination clicks -> isRefreshingData is false, watcher ignored

### Files Changed

| File | Change |
|---|---|
| apps/desktop/src/components/grid/DataGrid.vue | +26 lines |

---

### packages/app-tests/dataGridPagination.test.ts (+48 lines)

Unit tests for the auto-redirect page calculation logic:

| Test | Scenario |
|---|---|
| Page beyond last page after deletion | total=200, pageSize=100, currentPage=5 -> lastPage=2, redirect |
| Page still valid after partial deletion | total=500, pageSize=100, currentPage=5 -> lastPage=5, no redirect |
| Fewer rows than one page | total=30, pageSize=100, currentPage=3 -> lastPage=1, redirect |
| Total is zero | Guard prevents redirect attempt |
| Total is undefined | Guard prevents redirect attempt |

### packages/app-tests/useDataGridActions.test.ts (+50 lines)

Integration test verifying the offset-preservation fix end-to-end:

    onReloadData(offset=400) -> build-table-select-sql receives offset=400
    tab.resultPageOffset -> 400 (not 0)

### Test Results

    Test Files  2 passed (2)
         Tests  11 passed (11)

---

## Review Notes

- No blocking issues
- Minor stale-flag risk: if the reload short-circuits (tab switched away,
  store returns early), isRefreshingData could persist and trigger a
  spurious redirect on the next loading completion. Probability is low;
  recommend follow-up hardening in a separate PR.
- Second implementation correctly avoided a TDZ ReferenceError from the
  first attempt by placing the watcher below displayedTotalRowCount and
  accessing the computed inside the callback instead of the watch getter.

Habit compliance score: 4.5 / 5

## 变更类型

- [ ] 新功能
- [ ] Bug 修复
- [x] 功能优化
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

Related #2619

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -3364,6 +3364,10 @@ const isInfiniteScrollPaginating = ref(false);
 let lastInfiniteScrollPage = 0;
 let infiniteScrollCheckScheduled = false;
 let infiniteScrollAllLoaded = false;
+// Tracks whether the current loading cycle was triggered by a refresh/rollback
+// (as opposed to a normal paginate). Used to decide whether to auto-redirect
+// when the current page no longer exists after data was deleted.
+const isRefreshingData = ref(false);
 watch(pageSize, (value) => {
   customPageSizeInput.value = String(value);
 });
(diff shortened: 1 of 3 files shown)


t8y2:
Thanks for the contribution! Merged in 0901f17, will be released in the next version.

PR #2879, opened 2026-07-08
fix(spark,hive): use small fetchSize to prevent Thrift FetchResults failures

jischeng:
## Problem

Opening tables with large binary/struct columns (e.g., Spark Lance tables with protobuf blobs) via DBX fails with:

```
Agent RPC error (-1): java.sql.SQLException: Error retrieving next row
```

## Root Cause

Hive JDBC (used by both Spark and Hive agents) fetches rows via Thrift `FetchResults` RPC in batches controlled by `fetchSize`. DBX passes `fetchSize=500`, which causes the driver to try serializing all 500 rows (including large `binary` columns) in a single Thrift batch. When the total data exceeds Thrift serialization limits, the batch fails.

## Fix

Override `executeQuery` in `SparkAgent` and `HiveAgent` to set `fetchSize=50`. Each Thrift `FetchResults` call now only serializes 50 rows, staying within limits. The driver makes multiple calls to fetch all rows — no data loss, no LIMIT cap.

| | Before | After |
|---|---|---|
| fetchSize | 500 (from DBX) | 50 (overridden) |
| Thrift batch | 500 rows × binary cols = GB-level | 50 rows × binary cols = manageable |
| Result | ❌ serialization failure | ✅ multiple small batches succeed |

## Scope

Only affects Spark and Hive agents (both use Hive JDBC + Thrift). Other drivers (Oracle, Trino, etc.) use different protocols and are not affected. No common module changes.

## Testing

- `SELECT * FROM lance_clip10w.feature_v2 LIMIT 500` (35 cols, 7 binary) — was failing, now succeeds ✅
- Hive agent tests updated to expect fetchSize=50 ✅
- No common module changes (no impact on other agents) ✅

### DIFF
--- a/agents/drivers/hive/src/main/java/com/dbx/agent/hive/HiveAgent.java
+++ b/agents/drivers/hive/src/main/java/com/dbx/agent/hive/HiveAgent.java
@@ -10,6 +10,10 @@ import com.dbx.agent.JdbcIdentifiers;
 import com.dbx.agent.JsonRpcServer;
 import com.dbx.agent.TableInfo;
 import com.dbx.agent.TriggerInfo;
+import com.dbx.agent.ExecuteQueryOptions;
+import com.dbx.agent.QueryPageOptions;
+import com.dbx.agent.QueryPageResult;
+import com.dbx.agent.QueryResult;
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.ResultSet;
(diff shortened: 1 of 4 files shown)


t8y2 (changes requested):
I'm going to request changes for now. The fix only clamps `executeQuery`, but Spark/Hive query tabs use the agent cursor paging path.

For agent runtimes, DBX sends paged SQL through `execute_query_page`, and `AbstractJdbcAgent.executeQueryPage` passes `QueryPageOptions.fetchSize` through to `JdbcExecutor`, which still calls `Statement.setFetchSize(...)` with the caller value. So a Spark/Hive query tab or export using a page size like 500 can still reproduce the same large Thrift `FetchResults` batch failure.

Please apply the same small fetch-size normalization to `executeQueryPage` and likely `startTableRead` as well, and add coverage for those paths.

jischeng:
> I'm going to request changes for now. The fix only clamps `executeQuery`, but Spark/Hive query tabs use the agent cursor paging path.
> 
> For agent runtimes, DBX sends paged SQL through `execute_query_page`, and `AbstractJdbcAgent.executeQueryPage` passes `QueryPageOptions.fetchSize` through to `JdbcExecutor`, which still calls `Statement.setFetchSize(...)` with the caller value. So a Spark/Hive query tab or export using a page size like 500 can still reproduce the same large Thrift `FetchResults` batch failure.
> 
> Please apply the same small fetch-size normalization to `executeQueryPage` and likely `startTableRead` as well, and add coverage for those paths.

Got it, i'll do it now.

t8y2:
Thanks for the contribution! Merged in ee34a17, will be released in the next version.

PR #2912, opened 2026-07-08
fix(ai): make stop button actually cancel the agent stream

Abeautifulsnow:
### Problem

Clicking the stop button during an AI conversation had no effect — the agent kept generating. The stop button called `aiCancelStream` → `notify_one()`, but the four `stream_*_with_tools` SSE loops only `break`ed and returned `Ok(token_usage)` when the cancel token fired. `run_agent_loop` treated `Ok` as a normal completed turn and kept going (often into contract-repair or the next turn), so `isGenerating` stayed `true` and the UI never exited the generating state.

### Fix

- **Return `Err(AGENT_CANCELLED_ERROR)`** from all four `stream_*_with_tools` cancel branches instead of `break` → `Ok(token_usage)`. The existing `Err`-cancel match arm in `run_agent_loop` now correctly catches this and sets `loop_exit = LoopExit::Cancelled`.
- **Promote the error constant to a shared `pub const`** in `ai.rs` (`ai::AGENT_CANCELLED_ERROR`) so both `ai.rs` and `agent_loop.rs` reference the same canonical string.
- **Check cancellation before executing requested tools** — a stop landing in the narrow window between stream-end and tool-execution is now honored, preventing a long `execute_query` from starting.
- **Wire cancellation into the text-only fallback** (`run_agent_loop_text_only`) — previously the `_cancelled` parameter was ignored. Now uses `tokio::select!` to race `ai::complete` against `cancelled.notified()`, and also checks after `build_schema_prompt`.

### Known limitation

In-flight tool execution (e.g., `execute_query`) is still not interruptible mid-call. The cancel signal is checked before tool execution begins, but once a tool starts, it runs to completion.

### Files changed

- `crates/dbx-core/src/ai.rs` — added `pub const AGENT_CANCELLED_ERROR`; changed 4 cancel branches from `break` to `return Err(...)`
- `crates/dbx-core/src/agent_loop.rs` — uses `ai::AGENT_CANCELLED_ERROR`; added cancel check before tool execution; wired cancellation into text-only fallback

### Testing

- [x] `cargo test -p dbx-core agent_loop` — 18/18 pass
- [x] `cargo clippy -p dbx-core` — no new warnings introduced
- [ ] Manual smoke test: stop button during streaming, tool calls, and text-only fallback

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

Related #2893

### DIFF
--- a/crates/dbx-core/src/agent_loop.rs
+++ b/crates/dbx-core/src/agent_loop.rs
@@ -16,7 +16,6 @@ use crate::token_usage::TokenUsage;
 
 /// Maximum number of agent loop turns to prevent infinite loops.
 const MAX_AGENT_TURNS: u32 = 30;
-const AGENT_CANCELLED_ERROR: &str = "Agent loop cancelled";
 const MAX_TOOL_RESULT_CONTEXT_CHARS: usize = 12_000;
 const TOOL_RESULT_HEAD_CHARS: usize = 4_000;
 const TOOL_RESULT_TAIL_CHARS: usize = 4_000;
@@ -254,7 +253,7 @@ pub async fn run_agent_loop(
                     }
                     break;
                 }
-                Err(err) if err == AGENT_CANCELLED_ERROR => {
+                Err(err) if err == ai::AGENT_CANCELLED_ERROR => {
                     final_text = take_text(&accumulated_text);
                     loop_exit = LoopExit::Cancelled;
                     break;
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in ffc02793, will be released in the next version.

PR #3017, opened 2026-07-09
fix(mcp): disambiguate same-name connections and label responses with connection identity

Abeautifulsnow:
### Background

When two connections share the same name but have different database types — for example a Redis connection named `test` and a MySQL connection also named `test` — the MCP server resolved connections by name only. This caused SQL requests to be silently routed to the wrong connection type (e.g. a `SELECT` intended for MySQL hitting the Redis connection). The same issue occurs for same-name connections across hosts (e.g. two Postgres instances named `test` on different ports).

Additionally, tool responses carried no connection identity, so an AI agent could not tell which connection produced a given result in a multi-connection conversation.

### Solution

Defense in depth: ambiguity detection on name resolution, precise `connection_id` routing, plus a connection-identity label on every response.

| Layer | Mechanism |
|-------|-----------|
| **1. Ambiguity intercept** | When `connection_name` matches more than one connection, `resolveConnection` returns an `AMBIGUOUS_CONNECTION` error listing every match with `id: db_type @ host:port`. No silent pick. |
| **2. Precise routing** | All tools now accept an optional `connection_id` (takes priority over `connection_name`) that targets the exact connection by globally-unique ID. |
| **3. Type guard** | `dbx_execute_query` rejects SQL against a `redis` connection with `REDIS_COMMAND_REQUIRED`, so even a wrong ID can never send SQL to a non-SQL connection. |
| **4. Identity label** | Every data response is prefixed with `[name (id) [db_type @ host:port]]` so the agent can track result provenance across turns. |

### Changes

**`packages/mcp-server/src/index.ts`**
- `resolveConnection` now accepts `connection_id` (priority over `connection_name`); returns `AMBIGUOUS_CONNECTION` when a name matches multiple connections, listing each match's ID + db_type + host:port.
- All data tools (`dbx_list_tables`, `dbx_describe_table`, `dbx_execute_query`, `dbx_execute_redis_command`, `dbx_get_schema_context`) expose an optional `connection_id` parameter.
- `dbx_list_connections` emits an `ID` column so agents can obtain disambiguating IDs up front.
- New `connectionIdentity` / `labeledText` helpers; all responses from the five tools above are prefixed with the connection identity label.
- `dbx_remove_connection` uses `removeConnectionById` to avoid deleting the wrong connection on name collision.

**`packages/node-core/`**
- Added `findConnectionById` / `removeConnectionById` to the backend (`backend.ts`, `connections.ts`, `web-backend.ts`).

**`packages/mcp-server/tests/server.test.ts`**
- Same name, different db_type ambiguity -> `AMBIGUOUS_CONNECTION`.
- Same name, same db_type but different host/port ambiguity -> `AMBIGUOUS_CONNECTION`.
- `connection_id` takes priority over `connection_name`.
- `connection_id` routes to the correct host among same-name connections.
- `dbx_list_connections` includes the ID column.
- Tool responses carry the connection identity label prefix.

### Tests

- `pnpm --filter @dbx-app/mcp-server test` -> 30 passed / 1 failed.
- The 1 failure (`desktop bridge failures include a stable MCP error code`) is a pre-existing, Windows-environment-specific issue (the bridge returns "Connection not found" instead of "DBX is not running") and is unrelated to this change; CI is the source of truth.

### Known limitation (not addressed in this PR)

In **scoped mode** (pinning the MCP to a single connection via the `DBX_MCP_SCOPE_CONNECTION_NAME` env var), if the scoped name has duplicates, the first match is selected. This is a misconfiguration case for a feature meant to pin to a single connection; the recommended scope mechanism is `connection_id`. Non-scoped MCP usage (the normal path) is fully covered.

### Breaking changes

None. `connection_id` is optional; existing `connection_name` usage continues to work (and now returns a clearer error when there is genuine ambiguity, instead of silently picking the first match).

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

Related #2973


t8y2 (changes requested):
I'm going to request changes for now. The main issue is that `connection_id` is still dropped before several bridge calls, so same-name connections can still route to the first matching name.

1. The node-core bridge-backed paths still send only `connection_name` (`/data/execute-query`, `/data/list-tables`, MongoDB helpers, Redis fallback, SSH-tunneled paths). The bridge resolves that name with the first match, so `dbx_execute_query({ connection_id: ... })` can still hit the wrong connection for those database types.

2. `dbx_open_table` and `dbx_execute_and_show` also resolve by ID in the MCP layer but then send only `connection_name` to the desktop bridge, which has no `connection_id` field and still resolves by name.

3. `dbx_remove_connection` still supports name-only deletion and will delete the first duplicate-name match. In ambiguous cases it should require `connection_id` instead of silently removing one.

Please carry `connection_id` through the bridge request structs/handlers with ID-first resolution, keep name as a compatibility fallback, and add tests for duplicate-name bridge-backed routing and remove-by-name ambiguity.

t8y2 (changes requested):
这块我先 request changes。方向是对的：暴露 `connection_id`、列表里展示 ID、响应带连接身份标签都能解决同名连接的可观测性问题；但现在 ID 还没有贯穿到实际执行端，#2973 的错路由风险仍然存在。

1. `packages/node-core/src/database.ts:858` / `:929` / `:1043` / `:1076` 这些 bridge-backed data paths 仍然只把 `connection_name` 发给 desktop bridge，而 `src-tauri/src/commands/mcp_bridge.rs:287` 仍按 name 找第一条连接。`dbx_execute_query({ connection_id })` 在 SSH、非 direct DB 等路径下仍可能路由到同名的错误连接。

2. `packages/mcp-server/src/index.ts:417` 和 `:464` 里 `dbx_open_table` / `dbx_execute_and_show` 解析到 ID 后又只把 `connection_name` 发给 bridge；Rust handler 仍按 name resolve，所以同名连接下桌面打开/执行也会错路由。

3. `dbx_remove_connection` 仍要求并允许 name-only 删除，最后走 `removeConnection(name)`，同名时会删除第一条匹配连接。这里也应该在 name 命中多条时返回 `AMBIGUOUS_CONNECTION`，要求使用 `connection_id`。

建议把 `connection_id` 加到 bridge request structs/handlers，Rust 侧统一 ID-first resolve、name fallback，并补 duplicate-name bridge-backed routing 和 remove-by-name ambiguity 的测试。修完后我再 review 一下。

t8y2:
Thanks for the contribution! Merged in 0674433, will be released in the next version.

PR #3035, opened 2026-07-09
feat(redis): enable auto-refresh countdown for key TTL by default

Abeautifulsnow:
## Summary

Redis Key Value Viewer now defaults auto-refresh **ON** for keys with TTL. Previously the countdown and auto-refresh were opt-in — users had to click a Clock button to see a live TTL countdown. With this change, the TTL badge shows a real-time countdown immediately.

## Motivation

When a user opens a Redis key with `EXPIRE 300`, the TTL badge displayed a static snapshot ("TTL: 5m"). That number became stale the moment it was rendered. The user had to manually click the Clock icon to see the live countdown — an unnecessary step that many users didn't discover.

The auto-refresh overhead is minimal: `setInterval(1000ms)` runs a pure in-memory decrement. A network request fires only when the countdown reaches zero, at a frequency of **1 request per TTL duration** — e.g. once every 5 minutes for a 300-second TTL.

## Changes

| File | Change |
|------|--------|
| `apps/desktop/src/components/redis/RedisValueViewer.vue` | 3 logical changes, net -4 lines |

**1. Default ON** — `ref(false)` → `ref(true)`

**2. Single entry point** — removed the duplicate `countdownTtl` reset from the timer callback's `.then()`. The `finally` block of `load()` now calls `startAutoRefresh()` uniformly for all code paths (initial load, manual refresh, save, auto-refresh tick). The `.then()` / `.catch()` in `startAutoRefresh` only serve as a safety net to stop the timer when the key expires or the connection fails.

**3. Bootstrap on load** — `load()`'s `finally` block starts the timer automatically for keys with `ttl > 0`, so the first load triggers the countdown without user interaction.

## Clock button behavior

The Clock button is unchanged. Users can click it to toggle auto-refresh off at any time.

## Checklist

- [x] Type-check passes (`vue-tsc --noEmit`)
- [x] Pre-commit hooks pass (oxfmt)
- [x] No new dependencies

## 变更类型

- [ ] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

### DIFF
--- a/apps/desktop/src/components/redis/RedisValueViewer.vue
+++ b/apps/desktop/src/components/redis/RedisValueViewer.vue
@@ -102,7 +102,7 @@ const redisJsonWordWrap = ref(readRedisJsonWordWrap());
 const redisJsonHighlighter = ref<RedisJsonHighlighter>();
 
 // Auto-refresh
-const autoRefreshEnabled = ref(false);
+const autoRefreshEnabled = ref(true);
 const countdownTtl = ref(0);
 let autoRefreshTimer: ReturnType<typeof setInterval> | null = null;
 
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks for the contribution! Merged in 2eed0aa, will be released in the next version.

PR #3103, opened 2026-07-10
feat(mysql): add charset/collation editing to table structure editor

DengQingNian:
- Add character_set/collation fields to ColumnInfo struct (both Rust and TS)
- Query CHARACTER_SET_NAME and COLLATION_NAME from information_schema.COLUMNS
- Map Collation from SHOW FULL COLUMNS in fallback path
- Add editable SearchableSelect dropdown for charset/collation in structure editor
- Generate CHARACTER SET / COLLATE clauses in ALTER TABLE MODIFY COLUMN SQL
- Detect charset/collation changes as column attribute changes
- Include hardcoded charset/collation options for dropdown selection
- Update all non-MySQL DB drivers with ..Default::default() for the new fields

## 变更说明

MySQL数据库的表结构浏览界面可以查看和编辑字符集

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

image

## 验证

- [X] `make check` 通过
- [X] `make cargo-check-fast` 通过
- [X] 相关测试通过

## 关联 Issue

Related #2997  
Related #956 
Related #2493


t8y2 (changes requested):
I'm going to request changes for now. Three things need fixing before this is safe to merge:

1. The charset/collation controls and generated clauses currently apply to every MySQL-dialect column. MySQL only supports them on character columns; selecting a charset for an `INT` column generates invalid `ALTER TABLE` SQL. Please gate this in both the UI and Rust SQL generator, and clear or ignore charset state when changing to a non-character type.

2. The editor uses a hardcoded collation list and treats its first entry as the server default. Defaults and supported collations vary by MySQL version and compatible databases. Please reuse live server metadata, or omit `COLLATE` and let the server choose its default.

3. Charset and collation should be displayed as two separate columns in the structure editor. They are related but independently meaningful values, and separate columns make the current value and editable state much clearer.

Please also add tests covering valid character-column SQL, numeric-column handling, preservation of existing collations, and server-default selection. The PR currently has no CI checks on the head commit.

t8y2:
Thanks for the contribution! Merged in e429c89, will be released in the next version.

PR #3290, opened 2026-07-13
feat(connection): add SSH tunnel profile test

rainhan99:
## What

Adds a **Test** button to SSH tunnel profiles in **Settings → Tunnels**, so a profile's SSH credentials can be verified without creating a database connection.

## How it works

`AppState::test_tunnel_profile` starts just the SSH tunnel — `start_tunnel` → `connect_and_authenticate` → key/password/none auth, the **same code path a real connection uses** — against a sentinel forward target (`127.0.0.1:1`, never actually contacted because auth completes before any channel opens), then tears it down. A successful start means host reachability + credentials are valid.

- Proxy/HTTP tunnels connect lazily (nothing happens until traffic flows), so they can't be verified by a start/stop probe — the button is **SSH-only**.
- Uses a throwaway `uuid` connection id so the probe never reuses or evicts a live tunnel, and tears down whether auth succeeds or fails (no port/task leak).
- The connection dialog's existing `test_connection` path is untouched.

## Changes

- `dbx-core`: `AppState::test_tunnel_profile` (+ guard-path test for non-SSH / empty host)
- Tauri command `test_tunnel_profile` + registration
- dbx-web route `POST /tunnel-profiles/test`
- Both api facades (`tauri.ts` / `http.ts`) + `api.ts`
- `tunnelProfileStore.testProfile` action; Test button + inline result in `TunnelProfileManager.vue`
- i18n strings for 7 locales

## Test plan

- `pnpm typecheck`, `oxlint`, `oxfmt --check` — clean
- `cargo check` (dbx-core / dbx-web / dbx) — clean
- `cargo test -p dbx-core test_tunnel_profile_rejects_non_ssh_and_missing_host` — passes


t8y2 (commented):
Maintainer patch applied in d96d1e3d0.

The asynchronous SSH profile test could publish a stale result after the edited profile or selected profile had changed, causing the UI to display a result for inputs that were no longer current.

The patch binds each result to the tested profile snapshot and discards stale completions.

Validation:
- focused Vitest suite passed (11/11)
- frontend typecheck passed
- `git diff --check` passed

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in c7991b3, will be released in the next version.

PR #3456, opened 2026-07-14
feat: support Oracle trigger editing in table structure editor

mikodream:
## 变更说明

支持oracle DDL操作

## 变更类型

- [ x ] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

  - 前端允许 Oracle 进入触发器编辑：apps/desktop/src/components/structure/TableStructureEditor.vue:2065
  - 后端支持 Oracle trigger DDL：crates/dbx-core/src/table_structure_sql/triggers.rs:12
      - MySQL 行为保持不变：DROP TRIGGER + CREATE TRIGGER
      - Oracle 创建时也限定 schema：CREATE OR REPLACE TRIGGER "schema"."trigger_name" ... ON "schema"."table"


t8y2 (changes requested):
I'm going to request changes for now. The current metadata and editor model cannot safely edit existing Oracle triggers.

1. Oracle's `TRIGGER_TYPE` is returned as values such as `BEFORE EACH ROW`, while `create_trigger_sql` only accepts `BEFORE` or `AFTER`. The Oracle agent also does not return the trigger statement, so existing triggers open with an empty body. Even a simple existing row trigger therefore cannot be edited and saved correctly.

2. Reconstructing the trigger from only timing, event, table, and body is lossy for valid Oracle triggers. Clauses such as multiple events, `UPDATE OF`, `REFERENCING`, `WHEN`, statement-level/`INSTEAD OF` triggers, and compound triggers are not preserved and can be silently removed after an edit.

Please load and edit the complete Oracle trigger source, or explicitly restrict editing to a safely detected subset while keeping unsupported triggers read-only. Add integration coverage using metadata values returned by a real Oracle instance, rather than only manually constructed `BEFORE`/`UPDATE` test data.

t8y2 (commented):
Patched in `5f243438`.

The previous Oracle trigger editing path reconstructed existing triggers from timing, event, and body metadata. That can silently remove valid Oracle semantics such as multiple events, `UPDATE OF`, `REFERENCING`, `WHEN`, statement-level / `INSTEAD OF`, and compound-trigger clauses.

The maintainer patch fails closed: existing Oracle trigger fields are read-only in the UI, core rejects attempts to rebuild an existing trigger with a warning, and existing trigger drafts also remain read-only when the database type is unavailable. Explicit drops and creation of new supported triggers remain available.

Validation:

- Full `pnpm check` passed under the repository's Node.js 22.13.0: format, lint, typecheck, and the full frontend test suite
- Frontend focused test: 28/28 passed
- Rust `table_structure_sql` tests: 109/109 passed
- `cargo fmt --check --all`, `cargo clippy -p dbx-core --no-default-features -- -D warnings`, and `git diff --check` passed
- A real Oracle create/drop end-to-end test was not run; the available Oracle 11g XE schema had no existing user triggers, and no temporary objects were created

t8y2 (commented):
Follow-up maintainer fix `5f243438` applied on top of the fail-closed Oracle trigger patch.

The trigger editability helper now safely accepts a missing database type and keeps existing triggers read-only when the type is unknown. This fixes the CI type error without using an unsafe assertion or weakening the fail-closed behavior.

Validation: the complete `pnpm check` pipeline passes on Node 22.13.0, including format, lint, typecheck, and the full frontend test suite. The earlier 28/28 focused frontend tests, 109/109 Rust tests, Rust formatting/clippy checks, and the Oracle end-to-end limitation remain unchanged.

t8y2 (approved):
Maintainer patch verified; focused tests, full frontend checks, and required CI checks pass on this exact head.

t8y2:
Thanks for the contribution! Merged in fdaef625, will be released in the next version.

PR #3463, opened 2026-07-14
fix(ai): pass system proxy and valid reasoning effort to Codex CLI

koco-co:
## Summary

- map DBX's `minimal` reasoning setting to Codex CLI's supported `low` effort
- automatically pass the operating system proxy to Codex CLI when no explicit proxy environment is configured
- preserve explicit `codexCliEnv` proxy values, including lowercase environment keys
- revert the earlier retry-event workaround because it did not address the reproduced failure

## Root cause

Two independent configuration mismatches affected GUI-launched Codex CLI requests:

1. DBX passed `model_reasoning_effort="minimal"`, while the selected Codex model rejects that value and accepts `low`, `medium`, `high`, and related supported levels.
2. A Codex child launched from the macOS GUI did not receive the shell's proxy environment. Requests only worked after manually entering proxy variables in DBX.

## Implementation

- `AiReasoningLevel::Minimal` now serializes as `low` for Codex CLI.
- Codex process setup uses DBX's existing platform system-proxy resolver and fills `HTTP_PROXY` and `HTTPS_PROXY` only when equivalent keys are absent.
- User-provided proxy environment values take precedence, matched case-insensitively.
- The same process environment path is shared by model listing, connection testing, and agent execution.

### Rust checks

- `cargo fmt --all -- --check && git diff --check && CARGO_TARGET_DIR=/tmp/dbx-cargo-target cargo test -p dbx-core --lib` — exit 0; 2000 declared: 1994 passed, 0 failed, 6 ignored, 0 filtered. The six ignored cases require external live databases.
- `CARGO_TARGET_DIR=/tmp/dbx-cargo-target cargo clippy -p dbx-core --locked --all-targets` — exit 0; existing warnings remain (73 library, 80 test-target warnings).

### Tauri development UI flow

- Isolated configuration: provider `codex-cli`, model `gpt-5.6-luna`, reasoning `minimal`, empty `codexCliEnv`.
- Launched the parent process with `HTTP_PROXY`, `HTTPS_PROXY`, lowercase variants, and `ALL_PROXY` explicitly removed.
- Sent `只回复 E2E_PROXY_REASONING_OK` through the AI panel at `http://localhost:1420`.
- Received exactly `E2E_PROXY_REASONING_OK` from `Codex CLI / gpt-5.6-luna`.

### Packaged macOS debug app flow

- `pnpm tauri build --debug --bundles app --config /tmp/dbx-e2e-3457/tauri.e2e.conf.json -- --no-default-features` — exit 0.
- Launched `DBX E2E 3457.app` with the same isolated configuration and all parent proxy variables removed.
- Sent `只回复 PACKAGED_PROXY_REASONING_OK` through the packaged `tauri://localhost` UI.
- Received exactly `PACKAGED_PROXY_REASONING_OK` from `Codex CLI / gpt-5.6-luna`.

The UI verification covers the reproduced macOS development and packaged-debug flows. It does not claim signed/notarized release validation or a Windows UI run.


t8y2 (changes requested):
Maintainer patch applied in f856fd309 to preserve inherited proxy environment settings and prevent the detected system proxy from overriding an existing lowercase proxy variable. The focused Codex CLI tests pass with the added regression coverage.

I'm still requesting changes for the reasoning-effort behavior. Mapping `Minimal` to `low` globally changes the requested reasoning level for models that do support `minimal`. Codex CLI accepts `minimal`; support is model-specific. Please handle the fallback based on the selected model's supported reasoning efforts rather than aliasing the setting for every Codex model.

t8y2 (commented):
Maintainer patch `bda1af51` applied.

The patch keeps `minimal` as the literal Codex effort instead of globally aliasing it to `low`. Model capabilities are discovered from Codex app-server `model/list`; the synthetic default entry inherits the concrete `isDefault` model capabilities. Explicit runtime effort still has priority, while only legacy Minimal settings fall back to `low` for a known low-only model. Unknown capability preserves `minimal`.

Model discovery now runs outside the global cache lock. App-server discovery is bounded to 3 seconds, legacy `debug models` to 2 seconds, with 300-second success, 30-second compatibility, and 10-second static/negative cache TTLs. Opening model or effort menus refreshes capabilities without persisting settings.

Checks: 32 focused Rust tests, 6 focused frontend tests, `cargo check`, `cargo fmt --check`, `oxfmt`, `oxlint`, `vue-tsc`, and focused diff checks all pass.

A real end-to-end run against an authenticated Codex CLI was not performed; protocol parsing, cache behavior, fallback selection, CLI arguments, proxy inheritance, and menu persistence are covered locally.

t8y2 (approved):
Maintainer patch verified; focused tests and required CI checks pass on this exact head.

t8y2:
Thanks for the contribution! Merged in 0a5fbb6f, will be released in the next version.

PR #3512, opened 2026-07-15
feat(postgres): add server dashboard with live metrics and charts

antwa:
## Description

- New PostgreSQL Server Dashboard (mirrors the MySQL one): live metric cards (TPS, connections, active queries, cache hit rate, deadlocks, temp files, uptime, WAL rate) and 5 charts (Server sessions, Block I/O, Tuples in, Tuples out, Transactions per second — full width).
- Compatibility: `pg_current_wal_lsn`/`pg_wal_lsn_diff` (PG10+) auto-falls back to `pg_current_xlog_location`/`pg_xlog_location_diff` on older servers (verified against PG 9.1/9.2/9.6/10/current docs), so the dashboard works on PG 9.2+.
- Scoped to `db_type: "postgres"` only — same reasoning as the existing process-list feature (other wire-compatible engines unverified).
- Extracted shared `serverMetrics.ts` (rate/format helpers) used by both `mysqlServerStatus.ts` and `postgresServerStatus.ts`, removing duplicate code.
- Added a custom thin overlay scrollbar (`useVerticalOverlayScrollbar` composable) matching the sidebar tree's look, applied to both the Postgres and MySQL dashboards.

## Type of Change

- [X] New Feature
- [X] Bug Fix
- [ ] Performance Optimization
- [X] Code Refactoring
- [ ] Documentation Update
- [ ] CI / Build

## Frontend Changes

- [x] This PR involves frontend changes, and screenshots/screen recordings are attached below

image
image
image

## Verification

- [ ] `make check` passed
- [ ] `make cargo-check-fast` passed
- [X] `npx vitest run src/lib/__tests__/database/mysqlServerStatus.spec.ts` - 13/13 pass
- [X] `npx vitest run src/lib/__tests__/database/postgresServerStatus.spec.ts` — 17/17 pass
- [X] `pnpm typecheck` — clean
- [X] `pnpm lint` — clean (pre-existing warnings only, unrelated to this PR)
- [X] `pnpm test` — 3132/3136 pass; 4 failures in `packages/node-core/tests/sqlite-direct.test.ts` (`better-sqlite3` native module ABI mismatch, pre-existing environment issue unrelated to this change)


t8y2 (changes requested):
request changes: Dashboard 的统计 SQL 还有两个需要修正的问题，另外当前分支也存在 merge conflict。

1. `apps/desktop/src/lib/database/postgresServerStatus.ts:41` 的 active connection 统计会包含当前执行监控 SQL 的 DBX connection，因此空闲服务器也会始终显示至少一个 active connection。请通过 `pid <> pg_backend_pid()` 排除监控连接，并补真实 PostgreSQL 验证。

2. `apps/desktop/src/lib/database/postgresServerStatus.ts:28` 在一次刷新中重复扫描 `pg_stat_database` 和 `pg_stat_activity` 多次，界面又允许每秒刷新。建议用聚合 CTE 和 `COUNT(*) FILTER (...)` 各扫描一次统计视图，避免持续制造不必要的数据库负载。

另外，`postgresServerStatus.ts:65` 不应把任意 SQLSTATE 42883 都当成 WAL 函数版本兼容问题；请根据服务器版本或缺失函数名决定是否回退。

antwa:
@t8y2 
Thanks for the review! Addressed both points:
(Maintainer's review was in Chinese — I translated it myself, so apologies if I misread any nuance. Please let me know if I missed something.)

- **Aggregation**: replaced the 12 separate correlated subqueries against `pg_stat_database`/`pg_stat_activity` with two CTEs (`db_stats`, `activity_stats`) using `COUNT(*) FILTER (...)` for the conditional counts, so each view is scanned exactly once per refresh instead of once per column.
- **Own-connection exclusion**: added `WHERE pid <> pg_backend_pid()` to the `activity_stats` CTE so the monitoring connection's own backend no longer inflates the active/idle/total connection counts.
- **Compat-check scoping**: `isPgStatusCompatibilityError` no longer treats any SQLSTATE `42883` as the pre-PG10 WAL-function issue — it now also requires the error message to name `pg_current_wal_lsn`/`pg_wal_lsn_diff` specifically, so an unrelated undefined-function error won't get misdiagnosed and retried against the legacy WAL query.

Updated the unit tests to cover the new CTE shape, the pid exclusion, and the narrower compat-error matching (including a case proving an unrelated `42883` no longer false-positives). Pushed in fcfacb2

t8y2 (changes requested):
request changes: two PostgreSQL compatibility issues remain, and the branch still has merge conflicts.

1. `apps/desktop/src/lib/database/postgresServerStatus.ts:54` uses aggregate `FILTER`, which is not available on PostgreSQL 9.2/9.3. The legacy WAL fallback does not fix this syntax issue. Please use `SUM(CASE WHEN ...)` or update the supported-version claim.

2. `apps/desktop/src/lib/database/postgresServerStatus.ts:62` calls the current WAL-location function unconditionally. On a hot standby/read replica, this can fail the complete dashboard query. Please use a recovery-aware WAL/replay expression or make the WAL metric optional while preserving the other metrics.

The own-connection exclusion, single-scan aggregation, and narrower WAL compatibility detection are otherwise addressed. Please also resolve the remaining conflicts in `MySqlDashboard.vue` and `mysqlServerStatus.ts`.

antwa:
I will check it with different version of postgress first. thanks

antwa:
Fixed in 04caf73:

Separately: the fractional-rate fix for the MySQL dashboard (formatRate, preserving up to 3 decimal places so sub-1/s rates don't collapse to 0) is already shared via serverMetrics.ts and applied consistently on the Postgres side too — TPS, block I/O, tuple in/out, and transaction-rate charts all use it, so both dashboards stay in sync there as well.

Let me know if this covers what you had in mind.
@t8y2

t8y2 (commented):
Maintainer patch applied in 9a14df7.

The PostgreSQL dashboard counted pg_stat_activity rows whose state is NULL. PostgreSQL can expose such rows while masking activity fields from roles without permission, which inflated the visible connection total.

The total now counts only rows with a visible state, matching the active and idle metrics and PostgreSQL monitoring semantics.

Validation:
- PostgreSQL server status tests: 20 passed
- oxlint on changed files
- git diff --check

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in fc1f02a, will be released in the next version.

PR #3526, opened 2026-07-15
Fix：允许 MySQL 触发器填充新增行必填字段

zipg:
## 问题

MySQL 数据表新增行时，未填写的单元格会以 `NULL` 进入保存准备流程。原有客户端校验会在生成 SQL 前拦截 `NOT NULL` 且无默认值的列，导致数据库没有机会执行 `BEFORE INSERT` 触发器来填充值。

## 修改

- MySQL 新增行不再提前拦截这类空字段，生成 `INSERT` 时继续省略对应列
- 交由 MySQL 应用默认值、执行触发器，或返回原生必填约束错误
- 保留 MySQL 更新现有行时显式写入 `NULL` 的校验
- 保持 SQLite 及其他数据库的原有新增行校验行为
- 补充触发器填充场景和 MySQL 更新空值场景的单元测试

## 测试

- `pnpm check`：format、lint、typecheck、test 全部通过
- `cargo test -p dbx-core --lib --no-default-features data_grid_sql::tests`：71 项通过
- 真实 MySQL：省略 `NOT NULL` 字段后，`BEFORE INSERT` 触发器成功写入该字段
- 真实 MySQL：无触发器、无默认值时仍返回原生 `ERROR 1364`，不会静默写入错误数据

### DIFF
--- a/crates/dbx-core/src/data_grid_sql.rs
+++ b/crates/dbx-core/src/data_grid_sql.rs
@@ -1077,6 +1077,10 @@ fn build_data_grid_save_statements(options: &DataGridSaveStatementOptions) -> Ve
             .filter(|(_, value)| !value.is_null())
             .collect();
         if insert_pairs.is_empty() {
+            if options.database_type == Some(DatabaseType::Mysql) {
+                statements
+                    .push(data_grid_statement(options.database_type, format!("INSERT INTO {table} () VALUES ()")));
+            }
             continue;
         }
         let columns = insert_pairs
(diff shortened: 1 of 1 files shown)


t8y2 (commented):
Maintainer patch applied in 0bae8b4a0.

The previous implementation allowed MySQL triggers to populate omitted required columns, but rows containing only generated/default/trigger values produced no INSERT statement. It also generated rollback predicates from pre-insert NULL values, which could never match rows after AUTO_INCREMENT or trigger processing.

The patch now emits MySQL's empty-column default-row INSERT when every client-side value is omitted. For inserted rows, rollback SQL is generated only when all primary-key values are explicitly known and stable; otherwise rollback is intentionally unavailable instead of presenting an ineffective DELETE as recoverable.

Validation:
- `cargo test -p dbx-core --lib --no-default-features data_grid_sql::tests` — 73 passed
- MySQL 5.7.44 live trigger-only/default-row INSERT and explicit-primary-key rollback
- MySQL 8.0 live trigger-only/default-row INSERT and explicit-primary-key rollback
- `git diff --check`

The full crate Clippy run is currently blocked by pre-existing warnings outside this PR's changed file; the PR's GitHub Rust Clippy check will be required before merge.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 187a10f, will be released in the next version.

PR #3321, opened 2026-07-13
fix(doris): emit 3-part catalog.database.table for external catalog tables

jischeng:
## 变更说明

PR #2778 引入了 Doris/StarRocks 外部 catalog 侧边栏展示，但 SQL 生成用 `schema` 作为限定名的中间段。Doris/StarRocks 没有 schema 概念，schema 始终为空，导致 FROM 子句退化为 2 段 `catalog.table`，丢失 database 段。打开外部 catalog 的表时报 `No database selected` 或 `Unknown database`，因为 session 没有匹配的 database。

**根因**：
- `qualified_table_name_with_catalog` 用 `schema` 作 catalog 和 table 的中间段，但 Doris/StarRocks 不 schema-aware（`connectionObjectTreeNodeSchema` 返回 undefined），且所有 options/meta 类型没有 `database` 字段。
- MySQL 协议连接 StarRocks（`db_type=mysql` + `driver_profile=starrocks`）的 `effectiveDatabaseTypeForConnection` 返回 `mysql`，不走 catalog 分支，SQL 退回单段 `table`。

**改动**：
- 给 SQL 生成路径的 options/meta 类型加 `database` 字段（前后端：`TableDataSelectSqlOptions`、`DataGridTableMeta`、`DataGridCountSqlOptions`、`DataGridColumnDistinctValuesSqlOptions` 及前端镜像），`qualified_table_name_with_catalog` 在 schema 为空时用 database 作中间段，生成正确的 3 段 `catalog.database.table`。
- `effectiveDatabaseTypeForConnection` 对 `mysql + driver_profile=starrocks/doris` 返回 `starrocks/doris`，让 catalog-aware SQL 路径对 MySQL 协议连接也生效。
- COUNT 查询用 catalog-aware 的 execution database 而非 `tab.database`（外部 catalog 库名）。
- ObjectBrowser 行级元数据（DDL/列/索引/外键/触发器/导出）和 newQuery 上下文透传 catalog。

**范围**：Doris/StarRocks（含 `db_type=mysql` + starrocks/doris driver_profile 的 MySQL 协议连接）。外部 catalog 的表现在可以正常打开数据、查看 DDL/列/索引，同一个连接可跨多个 catalog 查看表数据（3 段限定名不依赖 session `SET catalog`）。

## 验证

- [x] `cargo test -p dbx-core --lib` 通过
- [x] `cargo fmt --check` 通过
- [x] `pnpm check`（format/lint/typecheck/test）全过，2834 测试通过
[...]

PR #3531, opened 2026-07-15
fix(doris): preserve catalog in metadata views

jischeng:
## Summary
- preserve external catalog identity when opening Doris and StarRocks object-browser tabs
- pass catalog through table listing, table DDL, and table-structure metadata requests
- isolate object and structure tabs when different catalogs contain the same database and table names

## Root cause
This is a follow-up to #3321. The sidebar tree was catalog-aware, but several frontend metadata routes still dropped the catalog before calling the backend. Those requests fell back to the normal MySQL metadata path, interpreted the external database as a regular database, and returned ERROR 3F000 Unknown database.

## User impact
Double-clicking a database under an external catalog now lists its tables in the right-hand object browser. Table DDL also works from both the object browser and the sidebar context menu.

## Validation
- targeted frontend regression tests after rebasing onto current main: 2 files, 13 tests passed
- pnpm typecheck passed after rebasing
- full frontend suite on the fix tree: 390 files, 3141 tests passed
- cargo test -p dbx-core doris_catalog --lib: 8 tests passed
- manually verified paimon_catalog.bi table listing and DDL in the built macOS app


t8y2 (approved):
Re-reviewed the current head and targeted regression coverage. No blocking issues remain.

t8y2:
Thanks for the contribution! Merged in a9e76dc13, will be released in the next version.

PR #3554, opened 2026-07-15
fix(grid): save MongoDB row edits and fix commit shortcut conflict

jthanh8144:
## Description

Three fixes to editing a row in the query result grid.

**1. MongoDB row edits were silently discarded.** Editing a cell and clicking Commit appeared to succeed, but nothing was written and no error was shown. An ObjectId `_id` reaches the grid as a hex string, and the frontend's `__dbx_mongo_string_id__` tag made the backend build `{_id: "<hex>"}` as a *string* filter — matching zero documents. `update_document` returned `Ok(0)`, which the UI read as success. The preview looked correct because it formats the id through a different function than the save path.

The tag now only suppresses extended-JSON reinterpretation (its actual purpose) instead of blocking ObjectId coercion. Numeric `_id`s are matched too, and a no-match now returns an error rather than reporting success. Also fixes delete and the collection browser, which share `document_id_filters`.

**2. Cmd+S both committed and opened "Save to SQL Library".** The grid handler awaited the save before calling `stopPropagation()`; dispatch is synchronous, so the event had already reached the window listener. It now claims the event synchronously.

**3. "Preview SQL" was mislabeled for document databases.** MongoDB/Elasticsearch preview a shell command, not SQL, so the button now reads "Preview query" there. Unchanged for SQL databases. Adds `toolbar.previewQuery` to all 7 locales.

Not fixed: the opt-in "MongoDB (Legacy)" Java agent has the same `_id` flaw (`MongoAgent.java:635`) — no Java toolchain available to verify a change, and the native driver is the default.

## Change Type

- [ ] New feature
- [x] Bug fix
- [ ] Performance improvement
- [ ] Code refactor
- [ ] Documentation update
- [ ] CI / build

## Frontend Changes

- [x] This PR involves frontend changes, screenshot/recording attached (see below)

image

## Verification

- [x] `make check` passes
- [x] `make cargo-check-fast` passes
- [x] Relevant tests pass (`cargo test -p dbx-core --lib`: 2018 passed; `make cargo-test-fast`: green)

New tests: two in `mongo_driver.rs` covering ObjectId and numeric `_id` filters (both fail against the old logic), plus `dataGridSaveUi.spec.ts` for the label.

To check manually: run a single `db.getCollection("<name>").find({})`, edit a cell, Commit, re-run — the value now persists.

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -169,7 +169,7 @@ import { canGoNextDataGridPage } from "@/lib/dataGrid/dataGridPagination";
 import { dataGridCountQueryOptions } from "@/lib/dataGrid/dataGridQueryOptions";
 import { dataGridBottomScrollTop, dataGridScrollPosition, isDataGridAtScrollBottom, isDataGridNearScrollBottom, shouldCheckInfiniteScrollAfterScroll, type DataGridScrollPosition } from "@/lib/dataGrid/dataGridInfiniteScroll";
 import { CANVAS_DATA_GRID_ROW_HEIGHT, drawCanvasDataGrid } from "@/lib/dataGrid/canvasDataGridRenderer";
-import { dataGridSaveActionMode, dataGridSaveToolbarState } from "@/lib/dataGrid/dataGridSaveUi";
+import { dataGridPreviewLabelKey, dataGridSaveActionMode, dataGridSaveToolbarState } from "@/lib/dataGrid/dataGridSaveUi";
 import type { QueryEditabilityReason } from "@/lib/sql/sqlAnalysis";
 import { EDITOR_FONT_FAMILY_CSS_VAR } from "@/lib/editor/editorThemes";
 import { safeLocalStorageGet, safeLocalStorageSet } from "@/lib/backend/safeStorage";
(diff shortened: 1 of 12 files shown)


t8y2 (changes requested):
The MongoDB row-mutation path still guesses the `_id` type from its displayed text and attempts ObjectId conversion before respecting an explicitly string-typed value.

This is unsafe when a collection contains both an ObjectId `_id` and a string `_id` with the same 24-character text: editing or deleting the string row can target the ObjectId document instead.

Please preserve the original BSON `_id` type in the row identity and pass that typed value through the edit/delete request end-to-end. The regression tests should cover identical textual values stored once as BSON ObjectId and once as BSON string, verifying that update and delete affect only the selected document.

t8y2 (commented):
Maintainer patch applied in f0b92ffc0.

Root cause: the UI already marks an explicitly typed MongoDB string `_id`, but the backend converted that marker into both ObjectId and string filters. If the same text exists as both BSON types, update/delete could target the ObjectId document instead of the selected string document.

The patch keeps explicit string IDs string-only, keeps explicit Extended JSON ObjectIds ObjectId-only, and preserves the legacy ObjectId-then-string fallback only for untyped IDs. It also adds regression coverage for both typed paths and frontend ObjectId serialization.

Validation:
- cargo test -p dbx-core --no-default-features document_id_filters --lib (7 passed)
- pnpm exec vitest run packages/app-tests/mongoDocumentValues.test.ts (18 passed)
- cargo fmt --all -- --check
- git diff --check

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 2457fac, will be released in the next version.

PR #3596, opened 2026-07-15
fix(jdbc): avoid privileged Kingbase metadata query

### LINKED ISSUE #3590: [Bug] Agent RPC error (-1): com.kingbase8.util.KSQLException: ERROR: permission denied for function sys_freespace Where: SQL statement "select COALESCE(sum(avail), 0) FROM sys_freespace(relid)" PL/p
### 数据库类型和版本

kingbase8

### 问题描述

kingbase8在Navicat Premium 17是正常的

同样的账号密码
在dbx报错
Agent RPC error (-1): com.kingbase8.util.KSQLException: ERROR: permission denied for function sys_freespace Where: SQL statement "select COALESCE(sum(avail), 0) FROM sys_freespace(relid)" PL/pgSQL function information_schema.pg_relation_size_ex(oid,bpchar) line 15 at SQL statement
Image

### 支持信息

DBX 版本: v0.5.57
运行环境: 桌面端
操作系统: Windows 10.0.19045.4474
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

SpencerZhang:
## Summary

- load Kingbase tables and views directly from its system catalogs
- avoid `DatabaseMetaData.getTables()`, which invokes privileged space-statistics functions in the Kingbase JDBC driver
- preserve the existing JDBC metadata path for other database vendors
- add regression coverage for low-privilege Kingbase metadata loading

## Root cause

The Kingbase JDBC driver's table metadata query calls `information_schema.pg_relation_size_ex()`, which in turn invokes `sys_freespace()`. Accounts without permission to execute that function can connect successfully but cannot open the database in DBX.

The fix follows the same approach used by database clients with dedicated Kingbase metadata support: query object names, types, and comments from the system catalogs without requesting relation-size statistics.

## Impact

Kingbase users no longer need elevated `sys_freespace` permissions to browse database objects. Other JDBC connections are unaffected.

## Validation

- `JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk-17.jdk/Contents/Home mvn -q test` in `plugins/jdbc` (57 tests passed)
- `git diff --check`

Closes #3590

### DIFF
--- a/plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
+++ b/plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
@@ -1900,11 +1900,15 @@ public final class DbxJdbcPlugin {
     private static JsonNode kingbaseListTables(Connection conn, String schema, boolean objectNodes) throws SQLException {
         ArrayNode result = MAPPER.createArrayNode();
         String effectiveSchema = kingbaseEffectiveSchema(conn, schema);
-        boolean compatibilityMode = kingbaseUsesInformationSchemaTables(conn);
-        String sql = compatibilityMode ? kingbaseCompatibilityTablesSql() : kingbaseCastSafeTablesSql();
+        KingbaseTableCatalogMode catalogMode = kingbaseTableCatalogMode(conn);
+        String sql = switch (catalogMode) {
+            case SYS_CATALOG -> kingbaseCastSafeTablesSql();
+            case POSTGRES_CATALOG -> kingbasePostgresTablesSql();
+            case INFORMATION_SCHEMA -> kingbaseCompatibilityTablesSql();
+        };
         try (PreparedStatement ps = conn.prepareStatement(sql)) {
             ps.setString(1, effectiveSchema);
-            if (!compatibilityMode) {
+            if (catalogMode == KingbaseTableCatalogMode.SYS_CATALOG) {
                 ps.setString(2, effectiveSchema);
                 ps.setString(3, effectiveSchema);
             }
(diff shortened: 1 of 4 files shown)


plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
```diff
@@ -1817,6 +1834,42 @@ static String normalizeInformationSchemaTableType(String tableType) {
         };
     }
 
+    private static JsonNode kingbaseListTables(Connection conn, String schema, boolean objectNodes) throws SQLException {
+        ArrayNode result = MAPPER.createArrayNode();
+        String effectiveSchema = emptyToNull(schema) == null ? "PUBLIC" : schema;
+        String sql = "SELECT c.relname AS table_name, " +
```
onceMisery: 是否可以考虑java 17+新特性，类似python的""" """来写多行字符串。

plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
```diff
@@ -1817,6 +1834,42 @@ static String normalizeInformationSchemaTableType(String tableType) {
         };
     }
 
+    private static JsonNode kingbaseListTables(Connection conn, String schema, boolean objectNodes) throws SQLException {
+        ArrayNode result = MAPPER.createArrayNode();
+        String effectiveSchema = emptyToNull(schema) == null ? "PUBLIC" : schema;
```
onceMisery: `PUBLIC` 是否需要考虑大小写敏感

plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
```diff
@@ -1817,6 +1834,42 @@ static String normalizeInformationSchemaTableType(String tableType) {
...
+            "LEFT JOIN sys_catalog.sys_description d ON d.objoid = c.oid AND d.objsubid = 0 " +
+            "WHERE n.nspname = ? AND c.relkind IN ('r', 'p', 'v', 'm', 'f') " +
+            "ORDER BY c.relname";
+        try (PreparedStatement ps = conn.prepareStatement(sql)) {
+            ps.setString(1, effectiveSchema);
+            try (ResultSet rs = ps.executeQuery()) {
+                while (rs.next()) {
+                    ObjectNode item = MAPPER.createObjectNode();
+                    item.put("name", rs.getString("table_name"));
+                    if (objectNodes) {
```
onceMisery, line 1921: `rs.getString("table_name")` 存在重复调用，是否应该用变量接收

onceMisery (commented):
非常感谢这个PR，但需要讨论下review的这几个点

t8y2 (changes requested):
I’m going to request changes for now. The new Kingbase table query at `DbxJdbcPlugin.java:1853` relies on `sys_class.relkind` for every Kingbase JDBC connection. DBX removed essentially the same catalog path in `6cba08fcc` after it caused cache lookup failures in Kingbase compatibility modes.

Please make this path compatibility-aware and reuse the dedicated Kingbase agent’s cast-safe table/view/materialized-view discovery instead of applying one `relkind` query universally. The tests should also cover the compatibility path and verify that it does not use `relkind`.

t8y2 (changes requested):
I'm requesting changes because two Kingbase compatibility cases are still incorrect.

First, `kingbaseUsesInformationSchemaTables` treats every connection without `sys_catalog.sys_namespace` as an information-schema compatibility connection. The dedicated Kingbase agent separately detects `pg_catalog.pg_namespace` and uses the PostgreSQL catalog path. Collapsing that mode into `information_schema.tables` drops objects such as materialized views. Please add a PostgreSQL-catalog mode and a regression test where `sys_catalog` is unavailable but `pg_catalog` is present.

Second, the regular query selects every `sys_class` row that is not recognized as a view, index, or sequence and labels it as `TABLE`. User-defined composite types also pass those exclusions, so they appear as tables. Please use a positive table discriminator or a table-specific catalog approach that remains safe across Kingbase compatibility modes, and cover a composite type in the tests. The equivalent heuristic in the dedicated Kingbase agent should remain consistent.

t8y2 (approved):
Reviewed exact head 6afe9de. The commits after the previous Changes Requested review resolve both catalog-classification blockers: PostgreSQL-compatible mode now positively selects supported `pg_class.relkind` values (including materialized views), while regular Kingbase mode positively matches real table catalogs and excludes composite types. This aligns with DBeaver's PostgreSQL relation classification and the documented `pg_class` semantics.

Validation: targeted JDBC and Kingbase tests passed, `git diff --check` passed, and the exact-head JDBC/Agent CI checks are green.

t8y2:
Thanks for the contribution! Merged in d69f9d3, will be released in the next version.

PR #3655, opened 2026-07-16
feat(driver-store): add dedicated "Updates Available" section for pending driver updates

Abeautifulsnow:
## Summary

Move drivers with available updates into a prominent amber-highlighted section
at the top of the Agent drivers tab. Previously, updatable drivers were scattered
throughout a flat list with only a small gray banner and subtle amber version
badges — easy to miss. Now they surface immediately in a dedicated card, and
after updating they return to their original position in the "All Drivers" list
automatically.

### `DriverStoreDialog.vue`

- Added four computed properties:
  - `orderedFilteredDrivers` — splits the filtered driver list so updatable
    drivers come first, stable drivers follow
  - `filteredUpdatableCount` — count of updatable drivers in the currently
    filtered view (respects search)
  - `filteredUpdatableDrivers` / `filteredStableDrivers` — sliced views used
    directly in the template without `<template>` nesting
- Replaced the single flat `v-for` with a two-tier layout inside one
  `divide-y` container:
  1. Amber header bar ("Updates Available (N)" + "Upgrade All" button)
  2. Updatable driver rows
  3. Section divider ("All Drivers")
  4. Stable driver rows (installed + not-installed)
- Eliminated inner `<template v-if="driver.installed">` wrappers in favor of
  inline `v-if` conditions to avoid HTML parser nesting issues

### i18n (7 locales)

Added three new keys in every locale (`zh-CN`, `zh-TW`, `en`, `ja`, `es`, `it`, `pt-BR`):

| Key | Purpose |
|---|---|
| `updatesAvailableTitle` | Section heading for the updates card |
| `updatesAvailableDescription` | Help text below the heading |
| `allDriversTitle` | Divider label between updatable and stable groups |

## Behavior

| Scenario | Result |
|---|---|
| Drivers have available updates | Amber "Updates Available" section at the top |
| Driver update completes | `update_available` → `false`, row moves to "All Drivers" |
| Search filters results | Both sections filter in sync; counts reflect filtered view |
| No drivers have updates | No section header or divider; flat list as before |

## Screenshots

*Before*: flat list, small gray banner

*After*: dedicated amber section at top → divider → stable list below

## UI变化后的截图展示

- 有可用更新时
04431404-9dde-434c-9286-dfcbbeacbde0

- 全部更新完后
13243496-8ae1-4b15-91bc-1a2c540adeea


t8y2:
Thanks for the contribution! Merged in fab1fa9, will be released in the next version.

PR #3669, opened 2026-07-16
feat(mongo): improve query completion and add distinct support

jthanh8144:
## Description

Two related MongoDB editor improvements.

**1. Rewrote query completion.** Replaced the regex-heuristic suggestion engine with a bracket-aware argument scanner that tracks the enclosing `{}`/`[]`/`()` and the key each is the value of. Suggestions now match the actual cursor position instead of guessing:

- query operators vs update operators, kept to their own positions
- aggregation stages only at pipeline level (including nested `$facet`/`$lookup` pipelines)
- accumulators inside `$group` output fields, expressions in projection/expression position
- stage-option keys for `$lookup`, `$unwind`, `$merge`, etc., and quoted field references (`"$field"`)
- operator/expression coverage expanded from ~35 entries to ~270, with descriptions and insert snippets

Also stopped suggesting helpers the shell parser can't execute (`bulkWrite`, `estimatedDocumentCount`, …) and added the ones it can (`count`, `drop`, `use`, `db.version`).

**2. Added native `distinct` support.** `db.collection.distinct("field", filter)` is now executable end-to-end — driver → `mongo_ops` → web route → Tauri command → MCP bridge → TS backend adapters → shell parser → query-store dispatch. Results render as a single column named after the field. `distinct` is re-added to completion now that it actually runs. The legacy Agent pool returns a clear "not supported" error, matching how `aggregate` already behaves.

## Change type

- [x] New feature
- [ ] Bug fix
- [ ] Performance
- [x] Refactor
- [ ] Documentation
- [ ] CI / build

## Frontend impact

- [x] This PR changes frontend behavior (editor autocomplete)

> The change is editor autocomplete behavior rather than visual UI. Capturing the suggestion dropdown requires a live MongoDB connection — not yet exercised against a real server.

## Verification

- [x] `make check` passes (format, lint, typecheck, 3334 tests)
- [x] `make cargo-check-fast` passes
- [x] Relevant tests pass — added coverage in `mongoCompletion.test.ts` (context modes, operator positioning, `distinct` argument completion, snippet-placeholder guard) and `mongoShellCommand.test.ts` (`parseMongoDistinctCommand`, `mongoDistinctToQueryResult`)

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -38,7 +38,7 @@ import { sqlCompletionContextFromSemantic } from "@/lib/sql/semantic/completion"
 import { buildSqlSemanticModel } from "@/lib/sql/semantic/model";
 import { mergeSqlSemanticReferenceAnalysis, resolveSqlSemanticNavigationTarget } from "@/lib/sql/semantic/references";
 import { buildElasticsearchCompletionItemsFromContext, getElasticsearchCompletionContext, getElasticsearchCompletionResultValidFor, shouldAutoOpenElasticsearchCompletion, type ElasticsearchCompletionItem } from "@/lib/elasticsearch/elasticsearchCompletion";
-import { buildMongoCompletionItemsFromContext, getMongoCompletionContext, getMongoCompletionResultValidFor, shouldAutoOpenMongoCompletion, type MongoCompletionItem } from "@/lib/mongo/mongoCompletion";
+import { buildMongoCompletionItemsFromContext, getMongoCompletionContext, getMongoCompletionResultValidFor, mongoCompletionNeedsCollections, mongoCompletionNeedsFields, shouldAutoOpenMongoCompletion, type MongoCompletionItem } from "@/lib/mongo/mongoCompletion";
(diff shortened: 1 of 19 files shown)


t8y2 (changes requested):
有一个会让新补全器在合法查询中失效的词法扫描问题需要先修复。

1. `apps/desktop/src/lib/mongo/mongoCompletion.ts:752` 的 `findInnermostMongoCall` 仅用正则查找最后一个 `.find(`/`.aggregate(` 等片段，没有跳过字符串和注释。只要前面的字符串值包含方法形状的文本，例如 `db.users.find({ note: ".aggregate(", na`，扫描起点就会落到字符串内部，当前字段位置被误判为 `none`，字段补全完全消失。针对当前 head 的真实实现探针已稳定复现。请让调用定位与参数扫描使用同一套 quote/comment-aware 词法规则，并覆盖字符串、转义字符串和注释中包含方法名的回归测试。

Patchable by maintainer：问题局限在 completion call 定位与测试。

t8y2 (approved):
Re-reviewed the current head and targeted regression coverage. No blocking issues remain.

t8y2:
Thanks for the contribution! Merged in 640e82cdc, will be released in the next version.

PR #3688, opened 2026-07-16
feat(table-info): show column default values

### LINKED ISSUE #1240: [💡 Feature] 表属性增加默认值
### 当前痛点

当前查看表属性，看不到默认值

Image

看DDL才能看到默认值

Image

### 期望方案

期望增加一列“默认值”，展示每个字段的默认值

### 优先级（是否紧急）

🟢 P3 低优 · 有空再做

### 补充信息

SQLite数据库

LwClick:
Closes #1240

## Problem

Both table information surfaces omitted `column_default`, forcing users to inspect DDL to see field defaults.

## Changes

- Add a localized Default column to the Object Browser and DataGrid table-information panels.
- Preserve raw driver-provided expressions, including quoted, parenthesized, numeric, and function defaults.
- Show an em dash only when default metadata is absent.
- Truncate long values while exposing the complete expression through the native tooltip.
- Add a shared presentation helper and regression coverage for both surfaces.

## Reference

DBeaver exposes `JDBCTableColumn#getDefaultValue()` as a visible property and SQLite passes its raw `defaultValue` into the common column model. This change follows that behavior without parsing database-specific expressions.

## Tests

- `pnpm vitest run packages/app-tests/tableInfoDefaultValue.test.ts`
- `pnpm typecheck`
- `pnpm test` (437 files, 3595 tests)
- `pnpm build`
- `pnpm exec oxfmt --check apps/desktop/src/components/objects/ObjectBrowser.vue apps/desktop/src/components/grid/DataGrid.vue apps/desktop/src/lib/table/tableColumnDefaultPresentation.ts packages/app-tests/tableInfoDefaultValue.test.ts`
- Re-ran the targeted test, typecheck, and build after rebasing onto the latest `main`.

## Manual verification

SQLite 3.51.0 was used with a temporary table containing no default, `''`, `0`, `CURRENT_TIMESTAMP`, nested parentheses, and quoted text. `PRAGMA table_info` returned the expected raw expressions, and the temporary database was removed afterward. Authenticated UI verification was not possible because the local web environment requires an access password.

## Remaining risks

The two UI surfaces compile and share the tested presentation contract, but their final layout was not visually exercised inside an authenticated desktop session.

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -59,6 +59,7 @@ import TemporalCellEditor from "@/components/grid/TemporalCellEditor.vue";
 import EnumCellEditor from "@/components/grid/EnumCellEditor.vue";
 import type { QueryResult, ColumnInfo, DatabaseType, ForeignKeyInfo, IndexInfo, TriggerInfo, TableInfoTab } from "@/types/database";
 import { tableObjectSourceKind } from "@/lib/table/tableObjectSourceKind";
+import { tableColumnDefaultDisplayValue } from "@/lib/table/tableColumnDefaultPresentation";
 import * as api from "@/lib/backend/api";
 import { formatElapsedSeconds } from "@/lib/common/elapsedTime";
 import { dataGridCellDisplayText, dataGridCellEditorText } from "@/lib/dataGrid/dataGridCellCoercion";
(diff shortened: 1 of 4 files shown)


t8y2:
Thanks for the contribution! Merged in 63ea5f1, will be released in the next version.

PR #3689, opened 2026-07-16
fix(grid): stabilize header height during horizontal scroll

### LINKED ISSUE #2114: [Bug] 数据表格横向滚动时表头高度抖动
### 数据库类型和版本

Oracle12c

### 问题描述

当设置“在字段名下方显示注释”时，如果首屏渲染的几个字段都没有注释，而横向滚动后新显示的字段有注释，会导致表头高度在横向滚动时一直抖动，建议如果设置了“”在字段名下方显示注释“”则不管有没有注释都固定所有表头高度

### DBX 版本

v0.5.41

### 操作系统

Windows

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

_No response_

LwClick:
Closes #2114

## Problem

Column headers are horizontally virtualized, while type and comment rows were rendered only when each visible column had content. Scrolling from columns without comments to columns with comments could therefore change the header row height.

## Root cause

Header height was derived from the current virtual column window instead of the complete result and table metadata.

## Changes

- Determine type and comment row visibility from the full column set.
- Keep fixed-height placeholder rows mounted for visible columns without values.
- Omit the extra row when the setting is disabled or the complete table has no corresponding metadata.
- Preserve full type and comment tooltips while avoiding empty tooltip content.
- Keep the shared DOM header unchanged for both DOM and Canvas row renderers.

## Reference

DBeaver LightGrid computes one global `headerHeight` from the maximum header height across all top-level columns (`LightGrid#computeHeaderSizes`). This change applies the same full-column invariant to DBX virtualized headers.

## Tests

- `pnpm vitest run apps/desktop/src/lib/__tests__/dataGrid/dataGridHeaderLayout.spec.ts apps/desktop/src/components/grid/__tests__/DataGridSurfaces.spec.ts` (10 tests)
- `pnpm typecheck`
- `pnpm test` (437 files, 3597 tests)
- `pnpm build`
- Formatting check for all changed frontend files
- Re-ran the 10 targeted tests and typecheck against the final formatted commit.

## Manual verification

No reusable Oracle, PostgreSQL, or MySQL container was available. The local web client requires an access password, so authenticated horizontal-scroll testing could not be performed. Equivalent component coverage verifies full-column layout decisions, fixed placeholder height, disabled settings, no-comment tables, and tooltip behavior.

## Remaining risks

Oracle 12c and authenticated DOM/Canvas horizontal scrolling were not exercised in this environment. The fix is database-independent and does not alter the Canvas renderer or metadata loading path.

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -439,7 +439,8 @@ function headerColumnType(column: string, actualColIdx: number): string {
 }
 
 const reserveColumnTypeLine = computed(() => reserveDataGridHeaderLine(showColumnTypesInHeader.value, props.result.columns, (column, index) => headerColumnType(column, index)));
-const reserveColumnCommentLine = computed(() => reserveDataGridHeaderLine(showColumnCommentsInHeader.value, props.tableMeta?.columns ?? [], (column) => column.comment));
+// Match the rendered header columns so comments from unprojected metadata cannot add an empty row.
+const reserveColumnCommentLine = computed(() => reserveDataGridHeaderLine(showColumnCommentsInHeader.value, props.result.columns, (column) => headerColumnComment(column)));
 
 function shortTypeName(t: string): string {
   const s = t.toLowerCase();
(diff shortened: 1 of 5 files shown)


t8y2:
Thanks for the contribution! Merged in 51b4c79, will be released in the next version.

PR #3690, opened 2026-07-16
fix(results): keep output toolbar height stable

### LINKED ISSUE #3361: [Feature] 数据表切换至摘要高度不应该变化
### 数据库类型和版本

通用

### 支持信息

https://github.com/user-attachments/assets/2af5220c-e2a7-452e-be00-0f670225da55
切换【数据表】、【摘要】的时候，行高会发生变化

Image

执行计划可以一并移过去吗？ 

### 当前痛点

切换视图不是很方便

### 期望方案

修复视图

### 优先级（是否紧急）

🟢 P3 低优 · 有空再做

LwClick:
Closes #3361

## Problem

The tabular result view used the embedded DataGrid toolbar while summary, chart, and explain views used a standalone toolbar. Their height contracts differed, causing the result pane to move vertically when switching views.

## Changes

- Set both toolbar containers to the same fixed `h-7` height (28px).
- Preserve the existing compact breakpoint, filters, result switcher, and action slots.
- Keep Explain in the existing shared toolbar actions instead of moving controls in this focused fix.
- Add a regression test that compares exact class tokens on both containers.

## Reference

DBeaver ResultSetViewer keeps presentations within one shared status bar and toolbar management surface. This change follows the same stable-container behavior across DBX result presentations.

## Tests

- `pnpm vitest run packages/app-tests/queryResultToolbar.test.ts` (9 tests)
- `pnpm typecheck`
- `pnpm test` (436 files, 3594 tests)
- `pnpm build`
- Formatting check for all changed files

## Manual verification

The local web client requires an access password, so authenticated switching between table, summary, chart, and explain views was not possible. The SFC compilation and exact class-token regression verify that both mounted toolbar paths use the same 28px height.

## Remaining risks

The toolbar bounding rectangles were not measured in an authenticated desktop session. The change does not alter toolbar content, responsive breakpoints, or result-view routing.

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -7366,7 +7366,8 @@ const gridContextMenuItems = computed<ContextMenuItem[]>(() => {
     <CustomContextMenu :items="gridContextMenuItems" v-slot="{ onContextMenu }">
       <div v-if="hasData || canShowWhereSearch" class="flex-1 flex flex-col overflow-hidden" @contextmenu="onContextMenu">
         <!-- Search bar -->
-        <div ref="dataGridTopbarRef" v-if="showDataGridTopbar" class="data-grid-topbar-shell shrink-0 flex min-w-0 border-b bg-muted/20">
+        <!-- Result and standalone output views must share a fixed toolbar height to prevent layout shifts. -->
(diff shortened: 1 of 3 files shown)


t8y2:
Thanks for the contribution! Merged in c67cce5, will be released in the next version.

PR #3708, opened 2026-07-16
refactor(redis): center full value dialog

SpencerZhang:
## Summary

- open Redis member full values in a centered dialog instead of an app-height right sheet
- preserve the existing value formats, editing, save, copy, and retained-draft behavior
- keep the dialog footer within the rounded dialog boundary on zero-padded content

## Why

The right-side sheet extended to the top of the application. Rapid interactions near its close control could accidentally reach application-level controls. A centered modal provides clearer separation from the surrounding UI.

## Validation

- `pnpm exec vitest run apps/desktop/src/lib/__tests__/redis/RedisValueViewer.redisJson.spec.ts`
- `pnpm typecheck`
- `pnpm build`
- `git diff --check`


t8y2:
Thanks for the contribution! Merged in 89fca90, will be released in the next version.

PR #3755, opened 2026-07-17
feat(xugu): expose programmable object metadata

mapan0424:
## Summary

- expose XuguDB procedures, functions, packages, package bodies, triggers, sequences, and user-defined types through the Xugu Agent
- include compilation validity metadata for programmable objects
- add focused query coverage for the expanded object metadata

This is an agent-only first slice of broader XuguDB support. It deliberately avoids unrelated UI, sidebar, and workflow changes.

## Validation

- `cd agents/drivers/xugu && go test ./...`


t8y2 (changes requested):
request changes: the rebase resolves the `agents/versions.json` conflict, but the two end-to-end metadata issues remain.

1. The agent returns `valid`, while the shared Rust and TypeScript `ObjectInfo` contracts still have no validity field. The value is therefore discarded before reaching DBX consumers. Please propagate it through the shared contract and test the public `list_objects` path.

2. `TRIGGER`, `TYPE`, and `TYPE_BODY` are still not represented by the shared source contract or object-browser actions, so they fall into the procedure/function UI path. `PACKAGE_BODY` also still does not match the source query's `PACKAGE BODY` spelling.

Please either implement these object types end-to-end or restrict this first slice to the types DBX currently supports, with source-path coverage.

t8y2 (changes requested):
Two end-to-end issues still need to be addressed.

1. TYPE and TYPE_BODY source lookup currently returns a static "DDL is not exposed" comment. Xugu's ALL_TYPES metadata exposes SPEC and BODY, and the listing query already relies on BODY. Please return the actual SPEC/BODY content and keep it read-only if it cannot safely be edited.

2. Object Browser still routes TRIGGER, TYPE, and TYPE_BODY through the procedure/function menu. This exposes a misleading Drop Function action, while the Rust drop-object API cannot deserialize these object types. Please give them the same source-only actions used by the sidebar, unless drop support is implemented end-to-end.

t8y2 (commented):
Maintainer patch applied in a75880271.

The author fixed the two blocking Xugu issues, but the updated head failed the frontend format check in four touched files.

The patch applies the repository formatter to those files only; no behavior was changed.

Validation:
- focused programmable-object tests: 7 passed
- touched-file oxfmt --check
- git diff --check

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 08cd890, will be released in the next version.

PR #3769, opened 2026-07-17
优化查询编辑器，添加查找、替换、删除空行和自动列表生成功能

JeaceLiu:
## 变更说明
AI 生成
添加如标题所述的功能，方便操作使用


t8y2 (commented):
Maintainer patch applied in 3507bd1ad.

The editor actions had three boundary defects: delimited-list generation could mutate a read-only editor, selection-scoped Replace All reused the 1000-match display cap, and blank-line deletion could join adjacent SQL text.

The patch adds read-only guards, separates full replacement traversal from capped match counting while preserving one transaction, and performs blank-line deletion on CodeMirror line boundaries. It also adds focused regression coverage for read-only dispatch, 1001 replacements, and newline preservation.

Validation:
- 10 focused Vitest tests
- focused Oxlint
- `git diff --check`

t8y2:
Thanks for the contribution! Merged in 7826155, will be released in the next version.

PR #3801, opened 2026-07-17
fix(ai): preserve last-used model selection within app run

Abeautifulsnow:
## Summary

Before this fix, the AI model selection would reset to the default config in several scenarios — starting a new chat, reloading configs, or switching the default provider — discarding the user's explicit model choice within the same app session. This PR makes `activeModel` **run-scoped and sticky**: once the user selects a model, it persists across new chats and config reloads within the same run, only resetting when the app restarts or the default config is explicitly changed.

## Problem

- **New chat resets model.** `startNewChat()` in `AiAssistant.vue` explicitly reset `activeModel` to the default config, losing the user's last model pick every time they started a new conversation.
- **Config reload duplicates fallback.** `reloadAiConfigs()` implemented its own "deleted config → fall back to default" logic, duplicating what `initAiConfigs()` already does. This led to subtle order-dependency bugs where the fallback in `reloadAiConfigs` could override the intentional reset done by `initAiConfigs`.
- **Switching default config didn't update active model.** `setDefaultAiConfig()` mutated the `isDefault` flags but didn't point `activeModel` to the newly-defaulted config, leaving the UI in an inconsistent state until the next page load.

### `apps/desktop/src/stores/settingsStore.ts`

1. **`reloadAiConfigs()` simplified** — removed the redundant conditional fallback. `initAiConfigs()` (called inside) already resets `activeModel` to the default config. The only edge case `reloadAiConfigs` now handles is setting `activeModel = null` when the config list is empty.

2. **`setDefaultAiConfig()` now updates `activeModel`** — after the bhange, `activeModel` is pointed to the newly-defaulted config. Thiskeeps the UI in sync without requiring a restart or config reload.

3. **Comment clarity** — documented the intent that `activeModel` is run-scoped: both app startup and `reloadAiConfigs` intentionally discard session-level switches and return to the default.

### `apps/desktop/src/components/editor/AiAssistant.vue`

4. **`startNewChat()` no longer resets the model** — removed the 4 lines that forced `activeModel` back to the default. The store already tracks the user's last
selection; resetting it on new chat was the root cause of the "forget

### `apps/desktop/src/stores/__tests__/settingsStore.spec.ts`

5. **New test suite for `activeModel` lifecycle** — 5 test cases covering:
   - `updateActiveModel` persists the model without mutating `isDefault`
   - `setDefaultAiConfig` (success) updates `activeModel` to the new default
   - `setDefaultAiConfig` (failure) leaves all state untouched
   - `reloadAiConfigs` with empty list → `activeModel = null`
   - `reloadAiConfigs` selects the `isDefault` config, not the first in the list

## Behavior matrix

| Scenario | Before | After |
|---|---|---|
| Select model in chat, start new chat | Model resets to default ❌ | Model persists ✅ |
| Select model, reload configs | Model resets to default | Model resets to default (intentional — reload is a full sync) |
| Switch default config in settings | Active model unchanged (stale) ❌ | Active model updates to new default ✅ |
| Delete the active config | Falls back to first in list | Falls back to `isDefault` config (via `initAiConfigs`) |
| Config list becomes empty | `activeModel` stale | `activeModel = null` ✅ |
| Backend fails on `setDefaultAiConfig` | `isDefault` wrongly mutated ❌ | All state preserved (fail-atomic) ✅ |

### DIFF
--- a/apps/desktop/src/components/editor/AiAssistant.vue
+++ b/apps/desktop/src/components/editor/AiAssistant.vue
@@ -1614,10 +1614,6 @@ async function deleteConversation(id: string) {
 function startNewChat() {
   clearMessages();
   showConversationList.value = false;
-  const defaultConfig = settings.aiConfigs.find((c) => c.isDefault) || settings.aiConfigs[0];
-  if (defaultConfig) {
-    settings.updateActiveModel({ configId: defaultConfig.id, modelId: defaultConfig.model });
-  }
 }
 
 onMounted(async () => {
--- a/apps/desktop/src/stores/__tests__/settingsStore.spec.ts
+++ b/apps/desktop/src/stores/__tests__/settingsStore.spec.ts
@@ -1,5 +1,7 @@
-import { describe, expect, it } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
 import { normalizeDesktopSettings, normalizeEditorSettings } from "@/stores/settingsStore";
+import { createPinia, setActivePinia } from "pinia";
+import type { AiConfigItem } from "@/types/ai";
 
 describe("normalizeEditorSettings", () => {
   it("enables automatic table aliases by default", () => {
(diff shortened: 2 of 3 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in c417e54, will be released in the next version.

PR #3854, opened 2026-07-18
fix(tree): expand trigger and type object groups

mapan0424:
## 变更说明

- 修复 XuguDB 模式级“触发器”和“类型”对象组已展示、但无法展开的问题。
- 统一运行时与兼容树加载路径，使用共享的对象组映射，避免新增对象类型遗漏加载逻辑。
- 保留表级触发器原有的专用加载行为。

## 变更类型

- [x] Bug 修复

## 涉及前端

- [x] 本 PR 涉及前端交互修复；行为变更可通过下述测试及手动展开对象树验证。

## 验证

- [x] `corepack pnpm vitest run packages/app-tests/sidebarTreeAffordances.test.ts` 通过
- [x] 新增回归检查，覆盖运行时和兼容加载路径。

## 手动验证

连接 XuguDB 后，展开任意模式下的“触发器”或“类型”分组，均应加载并展示对应对象。


t8y2 (commented):
The implementation looks correct and I don't see a blocking issue. One non-blocking testing follow-up: the new regression test only matches implementation text, so it doesn't exercise the actual routing. Please consider adding behavioral coverage proving that schema-level trigger/type groups use `listObjects`, while table-level trigger groups continue to use `listTriggers`, ideally including empty and rejected metadata responses.

t8y2:
Thanks for the contribution! Merged in 09c3f7f44, will be released in the next version.

PR #3977, opened 2026-07-20
fix(kingbase): prevent filtering of user schemas starting with SYS

q396921921:
## Problem

In Kingbase database, when a schema is named `systems` (or any name starting with `SYS` like `syslog`, `system_backup`), it doesn't appear in the client UI even though it exists in the database and can be queried via SQL.

## Root Cause

The schema filter in `KingbaseAgent.java` uses:
```sql
AND UPPER(schema_name) NOT LIKE 'SYS%'
```

This filters out **all** schemas whose names start with `SYS` (case-insensitive due to `UPPER()`). The user's `systems` schema becomes `SYSTEMS` after uppercasing, which matches `SYS%`, so it gets incorrectly filtered out.

## Solution

Changed the filter to only match system schemas that start with `SYS_` (with underscore):
```sql
AND UPPER(schema_name) NOT LIKE 'SYS\_%' ESCAPE '\'
```

This correctly filters:
- ✅ `systems` → `SYSTEMS` → does NOT match `SYS\_%` → **not filtered** (user schema)
- ✅ `SYS_CATALOG` → matches `SYS\_%` → **filtered** (system schema)

## Changes

1. **KingbaseAgent.java**: Fixed SQL filter conditions
2. **KingbaseAgentTest.java**: Added unit test to verify `systems` schema is not filtered


t8y2 (changes requested):
The Java Agent now preserves names such as `systems`, but the native Go Agent still uses `UPPER(schema_name) NOT LIKE 'SYS%'` in `agents/drivers/kingbase-go/kingbase_metadata.go`. Users running the Go Agent will therefore continue to lose the same `systems`, `syslog`, and `system_backup` schemas described by this PR.

Please apply the escaped `SYS_` predicate to both Kingbase Agent implementations and add the equivalent Go regression test so the fix is consistent across runtimes.

t8y2 (commented):
Maintainer patch applied in 0934e1e49.

The original Java-only fix left the native Go Agent using the broad `SYS%`/`XLOG%` filters, so users of that runtime could still lose valid schemas such as `systems` and `syslog`.

The patch applies the escaped internal-prefix predicates to the Go Agent, keeps Java and Go behavior aligned, and adds a focused regression test without adding metadata queries or client-side filtering.

Validation:
- `go test ./...` in `agents/drivers/kingbase-go`
- `./gradlew :kingbase:test --no-daemon`
- `python3 scripts/validate_agents.py` from `agents/`
- `git diff --check`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 1e9e0e6dd, will be released in the next version.

PR #3986, opened 2026-07-20
Fix：优化 MySQL 创建数据库权限不足提示

zipg:
## 改动内容

- 识别 MySQL 创建数据库时的 1044 / Access denied 权限错误
- 在创建与授权结果中保留服务端原始错误，并提示 root 账号不一定具备管理员权限，需要由拥有 GRANT OPTION 的管理员授予 CREATE 权限
- 补充全部现有语言的提示文案及定向回归测试

## 验证

- `pnpm check`
- 使用真实 MySQL 临时账号验证：仅有 `GRANT USAGE` 时创建数据库返回 1044；授予 `CREATE ON *.*` 后创建成功；临时数据库和账号已清理


t8y2 (commented):
Maintainer patch applied in 1b3eb59.

The regression test used `postgresql`, which is not a valid DBX `DatabaseType`; the real PostgreSQL identifier is `postgres`. Because desktop test files are excluded from the application typecheck, the invalid test-only value was not reported by CI.

The patch changes only that test input so the non-MySQL behavior is exercised with a valid database type.

Validation:
- Focused `createDatabaseErrorHints.spec.ts`: 4/4 passed
- `pnpm typecheck`: passed
- `git diff --check`: passed

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 7d41e41, will be released in the next version.

PR #4082, opened 2026-07-21
feat(export): optional MySQL AUTO_INCREMENT strip in DDL export

soddygo:
## 变更说明

Add an opt-in export option to drop the table-level `AUTO_INCREMENT=N` clause from MySQL DDL, so the exported script can initialize a fresh database without pinning a source-specific sequence position. Defaults to off (zero behavior change); no-op for non-MySQL databases.

The strip is implemented with a winnow parser that parses the table-options tail into structured (key, value, span) records and edits the original DDL by byte span, preserving every other byte. Quote/comment-aware: an AUTO_INCREMENT inside a COMMENT or comment is never touched.

- crates/dbx-core: new mysql_ddl_normalize module; omit_auto_increment on the export request structs, threaded through format_export_table_ddl.
- apps/desktop: "Omit AUTO_INCREMENT" checkbox in the export dialog (MySQL-family only), wired through the TS binding and i18n.

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

涉及

CleanShot 2026-07-21 at 20 29 08@2x

## 验证

执行: make dev-fast , 手动导出库表DDL测试验证对比,通过;

- [x] `make check` 通过
- [x] 相关测试通过

## 关联 Issue

None

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1923,6 +1923,7 @@ dependencies = [
  "tracing-subscriber",
  "uuid",
  "webpki-roots 0.26.11",
+ "winnow 1.0.3",
  "zip 4.6.1",
 ]
 
--- a/apps/desktop/src/components/export/DatabaseExportDialog.vue
+++ b/apps/desktop/src/components/export/DatabaseExportDialog.vue
@@ -60,6 +60,10 @@ const includeStructure = ref(true);
 const includeData = ref(true);
 const includeObjects = ref(true);
 const dropTableIfExists = ref(false);
+const omitAutoIncrement = ref(false);
+// `AUTO_INCREMENT` stripping is a MySQL-only DDL transform (backend gates on
+// db_type == mysql, which also covers MariaDB / TiDB / OceanBase-MySQL-mode).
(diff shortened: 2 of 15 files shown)


t8y2 (commented):
Maintainer patch applied in 092928a8c.

Removing `AUTO_INCREMENT` could also consume the newline that terminates a preceding MySQL line comment, causing the export terminator to be commented out and potentially producing an invalid SQL script.

The patch now removes only horizontal separator whitespace and preserves CR/LF boundaries. It also adds final-format regression coverage for both `-- ` and `#` comments, LF/CRLF, spaces, and tabs.

Validation:
- Focused line-comment and whitespace regression tests passed
- MySQL DDL normalizer tests: 18 passed
- `git diff --check`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 19e5b25, will be released in the next version.

PR #4120, opened 2026-07-22
fix(clippy): resolve nonminimal_bool warnings

soddygo:
## 变更说明

`cargo clippy --all-features --all-targets --workspace` (Rust 1.97+) flags two `nonminimal_bool` expressions, which fails CI under `-D warnings`. Both are behavior-preserving simplifications:

- db/mysql.rs: `!filter.is_some_and(|f| !f.trim().is_empty())` -> `filter.is_none_or(|f| f.trim().is_empty())`
- sql.rs: collapse the statement-terminator guard's `!a && !b && ... && !(x && y)` chain into a single negated disjunction (De Morgan).

Verified: `cargo clippy --workspace --all-targets --all-features -- -D warnings` is clean; full dbx-core unit suite (2401 tests) green.

## 变更类型

- [ ] 新功能
- [ ] Bug 修复
- [x] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

None

None

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

## 关联 Issue

None

### DIFF
--- a/crates/dbx-core/src/db/mysql.rs
+++ b/crates/dbx-core/src/db/mysql.rs
@@ -1633,7 +1633,7 @@ fn should_fallback_empty_list_tables(filter: Option<&str>) -> bool {
     // MySQL proxies such as MyCat can return an empty information_schema.TABLES
     // even when SHOW TABLES exposes objects. Avoid the fallback for active
     // searches so normal MySQL "no match" queries do not rescan large schemas.
-    filter.is_none_or(|filter| filter.trim().is_empty())
+    filter.map_or(true, |filter| filter.trim().is_empty())
 }
 
 fn filter_list_tables_fallback(
--- a/crates/dbx-core/src/db/mysql.rs
+++ b/crates/dbx-core/src/db/mysql.rs
@@ -1629,6 +1629,8 @@ pub async fn list_tables_filtered(
     Ok(tables)
 }
 
+// `Option::is_none_or` requires Rust 1.82, while DBX still supports Rust 1.77.2.
+#[allow(clippy::unnecessary_map_or)]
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch applied in 10858255b.

The Clippy rewrite introduced `Option::is_none_or`, which is only available from Rust 1.82 while DBX declares Rust 1.77.2 as its minimum supported toolchain. This made the changed code fail to compile on the declared MSRV.

The patch uses the equivalent `Option::map_or` expression, preserving both the simplified condition and Rust 1.77 compatibility.

Validation:
- Rust 1.77.2 minimal reproduction: `is_none_or` fails with E0599, patched `map_or` expression compiles
- `cargo test -p dbx-core mysql_empty_list_tables_fallback_allows_sidebar_paging_and_type_filters --lib --no-default-features` — passed
- `cargo fmt --all -- --check` — passed
- `git diff --check` — passed

t8y2 (commented):
Maintainer patch updated in 0f0d8b173.

The first CI run confirmed that current Clippy recommends the Rust 1.82-only API even though the repository still supports Rust 1.77.2. The follow-up adds a narrowly scoped lint allowance with an MSRV explanation, keeping both CI and the declared minimum toolchain supported.

Additional validation:
- `cargo clippy -p dbx-core --all-targets --no-default-features -- -D warnings` — passed
- focused MySQL fallback test — passed
- `cargo fmt --all -- --check` — passed

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in b276c0c, will be released in the next version.

PR #4139, opened 2026-07-22
fix(data-grid): stop double-escaping backslashes in Postgres/SQLite literals

Abeautifulsnow:
## Summary

`format_grid_sql_literal()` in `crates/dbx-core/src/data_grid_sql.rs` unconditionally doubled every backslash (`\` → `\\`) before wrapping a cell value in single quotes for INSERT/UPDATE statements generated by the grid's add-row/edit flow. PostgreSQL (with `standard_conforming_strings = on`, the default since 9.1) and SQLite do **not** treat backslash as a string-literal escape character, so the doubling corrupted values instead of escaping them.

### PostgreSQL — jsonb INSERT fails

Typing a jsonb value containing escaped inner quotes (e.g. `{"json_raw":"{\"foo\":1,\"bar\":\"sometext\"}"}`) into a new grid row and clicking submit produced:

Statement 1 failed: ERROR: invalid input syntax for type json
DETAIL: Token "foo" is invalid.

**Cause**: The backslash-doubled SQL literal delivered `\\"` to PostgreSQL's jsonb parser, which interpreted `\\` as a literal backslash (per standard JSON rules), truncating the string early and leaving a bare `foo` token.

### SQLite — TEXT column backslashes multiply on every save

Typing `{\"foo\":1}` (one backslash before each quote) into a TEXT cell and submitting stored `{\\"foo\\":1}` (two backslashes). Two backslashes became four. Each save doubled the count.

**Cause**: SQLite stores `\\` literally — it doesn't treat backslash as an escape — so the doubled backslashes went straight into the stored value.

## Fix

Scoped to **Postgres** and **SQLite** only (others unchanged):

```rust
let escaped_text = if database_type == Some(DatabaseType::Neo4j) {
    literal_text.replace('\\', "\\\\").replace('\'', "\\'")
} else if matches!(database_type, Some(DatabaseType::Postgres) | Some(DatabaseType::Sqlite)) {
    // Postgres / SQLite: backslash has no special meaning in string literals.
    // Only the quote delimiter needs escaping.
    literal_text.replace('\'', "''")
} else {
    literal_text.replace('\\', "\\\\").replace('\'', "''")  // unchanged
};
```

MySQL and Neo4j still interpret backslash as an escape character — their behavior is preserved and verified by regression tests.

Files Changed

- crates/dbx-core/src/data_grid_sql.rs (+35 lines)

Tests

Two new unit tests, both passing:

- postgres_and_sqlite_literals_do_not_double_escape_backslashes — reproduces both reported bug scenarios and confirms fix
- mysql_and_neo4j_literals_keep_doubling_backslashes — guards against regression for dialects that need backslash escaping

Full data_grid_sql test suite: 98 passed, 0 failed.

Review

- Rating: 5 / 5 (no blocking issues, no warnings)
- Two optional suggestions (not in this PR):
  - rqlite (SQLite-based) may have the same bug
  - Postgres-derived dialects (Redshift, Kingbase, Gaussdb, etc.) may benefit from the same fix — but no reports yet, so deferred

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

Related #2055

### DIFF
--- a/crates/dbx-core/src/data_grid_sql.rs
+++ b/crates/dbx-core/src/data_grid_sql.rs
@@ -1795,13 +1795,17 @@ pub fn format_grid_sql_literal(
     } else {
         text
     };
+    if database_type == Some(DatabaseType::Postgres) && literal_text.contains('\\') {
+        // Escape strings have stable backslash semantics regardless of the
+        // session's standard_conforming_strings setting.
+        let escaped_text = literal_text.replace('\\', "\\\\").replace('\'', "''");
+        return format!("E'{escaped_text}'");
+    }
     let escaped_text = if database_type == Some(DatabaseType::Neo4j) {
         literal_text.replace('\\', "\\\\").replace('\'', "\\'")
-    } else if matches!(database_type, Some(DatabaseType::Postgres) | Some(DatabaseType::Sqlite)) {
-        // Postgres (with the default standard_conforming_strings = on) and SQLite
-        // do not treat backslash as a string-literal escape character, so doubling
-        // it here would corrupt the stored value instead of escaping it. Only the
-        // quote delimiter needs escaping for these dialects.
+    } else if is_sqlite_literal_database(database_type) {
+        // SQLite-family engines do not treat backslash as a string-literal
+        // escape character, so only the quote delimiter needs escaping.
         literal_text.replace('\'', "''")
     } else {
         literal_text.replace('\\', "\\\\").replace('\'', "''")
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks for the contribution! Merged in faca91a, will be released in the next version.

PR #4144, opened 2026-07-22
fix(mongo): ignore comments in shell JSON arguments

### LINKED ISSUE #4142: MongoDB aggregate parser rejects pipelines with comments
## Problem

MongoDB aggregate commands copied from shell/Compass can include comments inside the pipeline array. DBX currently reports:

```
MongoDB aggregate pipeline must be a JSON array (for example [{ $match: {} }]).
```

## Reproduction

Run a MongoDB query like:

```js
db.cash.aggregate([
  { $match: { portfolio_id: "b6f6ec62-8571-11f1-bbfb-000c29caf77f", date: { $gte: "20260604" } } },
  { $group: { _id: "$date", total_cash: { $sum: "$cash" } } },
  // optional reshape
  { $project: { _id: 0, date: "$_id", total_cash: 1 } }
])
```

## Cause

The Mongo shell compatibility parser normalizes shell-like JSON arguments and then validates them with `JSON.parse`. Comment tokens such as `// ...` and `/* ... */` are valid in common Mongo shell snippets but invalid JSON, so the aggregate pipeline normalization fails before execution.

## Expected

Comments outside quoted strings should be ignored when parsing Mongo shell JSON-like arguments. Comment markers inside strings, such as `https://...`, should remain intact.

## Local fix

A local fix has been prepared in commit `2fe0c28c` (`fix(mongo): ignore comments in shell JSON arguments`). It strips shell/SQL-style comments in the shared `@dbx-app/mongo-shell` argument normalizer and teaches delimiter/argument scanning to ignore comments too.

Validated with:

```
pnpm vitest run packages/app-tests/mongoShellCommand.test.ts
pnpm --filter @dbx-app/mongo-shell build
```

James-Leong:
## Summary

- strip `//`, `--`, and block comments while normalizing Mongo shell JSON-like arguments
- make Mongo aggregate delimiter matching and top-level argument splitting ignore comments outside strings
- add regression coverage for aggregate pipelines with inline comments and strings containing comment markers

Fixes #4142.

## Tests

- `pnpm vitest run packages/app-tests/mongoShellCommand.test.ts`
- `pnpm --filter @dbx-app/mongo-shell build`

### DIFF
--- a/packages/app-tests/mongoShellCommand.test.ts
+++ b/packages/app-tests/mongoShellCommand.test.ts
@@ -533,6 +533,82 @@ test("parseMongoAggregateCommand accepts an empty pipeline", () => {
   });
 });
 
+test("parseMongoAggregateCommand ignores comments inside aggregate pipelines", () => {
+  const command = parseMongoAggregateCommand(`
+    db.cash.aggregate([
+      {
+        $match: {
+          portfolio_id: "b6f6ec62-8571-11f1-bbfb-000c29caf77f",
+          date: { $gte: "20260604" }
+        }
+      },
+      { $unwind: "$cashs" },
+      {
+        $project: {
+          _id: 0,
+          date: 1,
+          trans_currency_cd: "$cashs.trans_currency_cd",
+          cash_local: { $multiply: ["$cashs.cash", "$cashs.fx_rate"] }
+        }
+      },
+      {
+        $group: {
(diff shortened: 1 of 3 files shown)


t8y2:
Thanks for the contribution! Merged in 63f375a, will be released in the next version.

PR #4153, opened 2026-07-22
fix(sql): prevent crash on comment-only SQL files

### LINKED ISSUE #4130: [Bug] SQL file preview crashes on comment-only statement with index out of bounds
## Summary

Selecting a SQL file that contains a comment-only/whitespace-only statement followed by a semicolon crashes the desktop app during the preview step, before the file is executed.

The release build uses `panic = "abort"`, so the bounds-check panic terminates the entire application.

## Minimal reproduction

Create a file named `crash.sql` with the following content:

```sql
-- DBX SQL preview crash reproducer

;
```

Then:

1. Open DBX.
2. Open **Execute SQL File**.
3. Click **Browse**.
4. Select `crash.sql`.

The application exits immediately while loading the preview. No execution is required.

## Actual result

When DBX is launched from a terminal, the following panic is printed:

```text
thread 'tokio-rt-worker' panicked at crates/dbx-core/src/sql.rs:1895:59:
index out of bounds: the len is 441 but the index is 441
```

This was reproduced with DBX 0.5.62 on macOS 15.4.1 using the x86_64 build under Rosetta.

## Expected result

The preview should load normally and comment-only/empty statements should be ignored without panicking.

## Root cause

The issue is still present on latest `main` at commit `b7f38a088e7a77eedc7c72722eacdd07d7b6a319`.

In `leading_executable_sql_with_options`, the inner whitespace loop can advance `i` to `bytes.len()`. The subsequent hash-comment check indexes `bytes[i]` without checking the bound again:

```rust
while i < bytes.len() && bytes[i].is_ascii_whitespace() {
    i += 1;
}

if options.profile.supports_hash_line_comments && bytes[i] == b'#' {
```

Source: https://github.com/t8y2/dbx/blob/b7f38a088e7a77eedc7c72722eacdd07d7b6a319/crates/dbx-core/src/sql.rs#L1882-L1922

The same unchecked pattern also exists in `first_executable_sql_token_with_options`:

https://github.com/t8y2/dbx/blob/b7f38a088e7a77eedc7c72722eacdd07d7b6a319/crates/dbx-core/src/sql.rs#L1925-L1976

The SQL-file preview path calls `mysql_like_sql_file_can_execute_without_selected_database`, so the MySQL-compatible parsing profile enables `supports_hash_line_comments` and reaches the unsafe access even when previewing a file for a PostgreSQL-compatible connection.

## Suggested regression coverage

- A MySQL-compatible parser input containing only `--` comments and trailing whitespace.
- A SQL-file preview input containing a comment-only statement followed by `;`.
- Equivalent boundary coverage for both helper functions above.

wang-zhengxin:
## Summary

- guard SQL comment skipping when it reaches the end of the input
- cover both statement splitting and executable-keyword detection paths
- add regression tests for MySQL comment-only SQL input

## Root cause

After skipping a line comment and trailing whitespace, the parser could advance its byte index to `bytes.len()`. The MySQL hash-comment check then accessed `bytes[i]` without first checking that the index was still in bounds. Release builds abort on this panic, so opening the affected SQL file exited DBX.

## Impact

Opening or importing SQL files containing comment-only statements followed by a semicolon no longer crashes DBX.

## Validation

- `cargo test -p dbx-core` (2430 passed, 0 failed, 11 ignored; integration and doc tests passed)
- `cargo clippy -p dbx-core --all-targets -- -D warnings`
- `cargo fmt --all --check`
- manually verified the minimal reproducer and the original 15 MB SQL file in the Tauri development build

Fixes #4130

### DIFF
--- a/crates/dbx-core/src/sql.rs
+++ b/crates/dbx-core/src/sql.rs
@@ -1888,6 +1888,10 @@ fn leading_executable_sql_with_options(sql: &str, options: SqlParsingOptions) ->
             i += 1;
         }
 
+        if i >= bytes.len() {
+            break;
+        }
+
         if i + 1 < bytes.len() && bytes[i] == b'-' && bytes[i + 1] == b'-' {
             i += 2;
             while i < bytes.len() && bytes[i] != b'\n' {
@@ -1929,6 +1933,10 @@ fn first_executable_sql_token_with_options(sql: &str, options: SqlParsingOptions
     while i < bytes.len() {
         while i < bytes.len() && (bytes[i].is_ascii_whitespace() || bytes[i] == b'(') {
             i += 1;
+        }
+
+        if i >= bytes.len() {
+            break;
         }
 
         if i + 1 < bytes.len() && bytes[i] == b'-' && bytes[i + 1] == b'-' {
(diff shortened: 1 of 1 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in c180c40b3, will be released in the next version.

PR #4158, opened 2026-07-22
fix(jdbc): map MySQL connect timeout to milliseconds

wbean:
## Summary
- map JDBC connect timeout seconds to milliseconds for MySQL-family drivers
- avoid overriding an explicit connectTimeout already present in the JDBC URL
- bump the optional JDBC plugin version to 0.1.25

## Why
DBX stores connection timeout values in seconds, but MySQL Connector/J expects the connectTimeout property in milliseconds. Passing the raw seconds value makes MySQL-family JDBC connections time out almost immediately, especially with custom MySQL-compatible drivers.

## Tests
- mvn test (from plugins/jdbc)
- mvn -DskipTests package (from plugins/jdbc)
- ./package.sh (from plugins/jdbc)

## Notes
I verified the fix locally with a custom MySQL Connector/J build against a MySQL-compatible distributed database where DBX previously failed with "Connect timed out" while DBeaver worked.

### DIFF
--- a/crates/dbx-core/src/agent_connection.rs
+++ b/crates/dbx-core/src/agent_connection.rs
@@ -59,6 +59,7 @@ pub fn agent_connect_params(config: &ConnectionConfig, host: &str, port: u16, da
         "ca_cert_path": config.ca_cert_path,
         "client_cert_path": config.client_cert_path,
         "client_key_path": config.client_key_path,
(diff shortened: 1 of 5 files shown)


t8y2:
Thanks for the contribution! Merged in ff46b7c, will be released in the next version.

PR #4196, opened 2026-07-22
feat(data-grid): multi-format copy extractors for cell/row/column selections

soddygo:
## 变更说明

为数据网格结果区新增一套**多格式复制 / 提取器(copy extractors)**。
用户在结果网格中选中单元格、整行或整列后,可将所选数据以多种格式复制到剪贴板,
并支持在「Configure Data Extractors…」对话框中自定义分隔符 / 引号 / NULL 文本等选项,带实时预览。

**支持的格式**
- 表格类:TSV、CSV(可带表头)、Pipe-separated、自定义 DSV、One-row、Pretty Table
- 结构化:JSON、JSON Lines
- SQL:`IN (...)` 列表、`INSERT`(可合并多行 / 逐行)、`UPDATE`、`WHERE` 子句
- 文档:Markdown、HTML、XML

**核心能力**
- 选区类型:支持行选、列选、任意单元格矩形选区;并将显示列正确映射回源列。
- 可配置项:列分隔符、行分隔符、NULL 文本、引号字符与策略(minimal/always/never)、
  是否包含列头 / 行号、跳过计算列 / 生成列 / 主键列、INSERT 合并模式、JSON 美化等。
- 预览:对话框内实时预览当前选区的提取结果(超 100 行截断提示)。
- 告警:省略列(如跳过生成列 / 主键)与 JSON 列名重复重命名都会给出明确提示。
- 持久化:默认提取器与各选项保存在编辑器设置中。
- 安全边界(后端强制):输入上限 100k 行 / 10k 列 / 200 万单元格 / 64MiB;
  输出上限 32MiB,超出时提示改用「导出到文件」。

**架构**
- Rust:`crates/dbx-core/src/data_grid_extractors/`,纯提取逻辑 + 版本化契约
  (`DATA_GRID_EXTRACTOR_CONTRACT_VERSION`),SQL 生成感知表元数据(主键驱动 UPDATE、跳过生成列)。
- 命令:Tauri `extract_data_grid_selection`(`spawn_blocking` 跑在阻塞线程池)
  与 `dbx-web` 同名路由,桌面 / Web 双端对齐。
- 前端:`lib/dataGrid/dataGridCopyExtractor.ts`(格式描述符 + 契约)、
  `composables/useDataGridExtractor.ts`(选区 → 请求构建 + 复制)、
  `components/grid/DataGridExtractorDialog.vue`(配置 + 预览),并接入 `DataGridToolbar` 的复制 split-button。
- 与既有 `data_grid_sql` 复制工具共存,未改动既有导出/保存逻辑。

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

https://github.com/user-attachments/assets/f8c47008-129f-42be-bdde-fb18b4d97915

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

## 关联 Issue

None


t8y2 (commented):
The new extractor path still needs a deeper pass before approval. In addition to the current merge conflict, the toolbar integration around `apps/desktop/src/components/grid/DataGrid.vue:5165` is not mirrored by the right-click copy menu around `:7367`. Please resolve that entry-point gap first; the INSERT/UPDATE/WHERE SQL generation also needs dialect and selection-mapping coverage beyond the current fast triage.

t8y2 (changes requested):
request changes: the current head has merge conflicts and two data-fidelity regressions that remain even after rebase.

1. **P1 — copied columns lose the user's displayed order.** `apps/desktop/src/composables/useDataGridExtractor.ts:72-80` maps displayed columns back to source indexes and then sorts those indexes. After drag-reordering columns, cell/row/column extraction outputs the original result order instead of the order visible to the user. Preserve insertion order while deduplicating and cover reordered plus hidden-column selections.

2. **P1 — SQL NULL and an ordinary string can serialize identically.** `crates/dbx-core/src/data_grid_extractors/delimited.rs:35` and `:75` send the NULL sentinel and a matching text value through the same quoting path. With the defaults, SQL NULL and the string `NULL` can produce the same output; an empty `nullText` also collides with an empty string. Keep nullness as structured state until encoding and add minimal/always quoting regressions for CSV/TSV/DSV/one-row output.

3. **P2 — the right-click copy entry is not fully integrated.** The complete extractor UI is added around `apps/desktop/src/components/grid/DataGrid.vue:5165`, but the context menu around `:7367` still exposes the old Copy/Selection subset. Users cannot reach or configure the new formats from the normal right-click workflow. Add the new Copy submenu while preserving the existing actions and default shortcuts.

The current head also conflicts with `main` in `DataGrid.vue`, `useDataGridExport.ts`, and `settingsStore.ts`; please rebase before the next review.

Validation:
- Exact-head diff reviewed: 54 files.
- `git merge-tree` reproduced the three content conflicts against current main.
- CI `frontend`, `packages`, `rust`, `rust-fmt-clippy`, `rust-test`, `autofill`, and `changes` passed, but does not cover the current merge result or the fidelity cases above.

t8y2:
这个方向我觉得是对的，不过入口还需要再调整一下。

我理解这个 PR 其实是在解决两件事：

1. 增加更多复制格式；
2. 允许用户配置 `Ctrl/Cmd + C` 默认复制出来的格式。

对于第一点，右键菜单里的「复制」应该完整接入这些新增格式。右键更适合作为临时操作入口：如果用户这次想复制成 CSV、JSON、SQL INSERT、Markdown 等格式，可以直接从右键里选。

对于第二点，我更倾向于把「默认复制格式」放到数据网格右上角的视图选项里，也就是那个 `🔧` popover。现在复制按钮旁边的位置比较紧，而且 dropdown 放在那里有点不明确，用户看到时可能会以为是在切换表格的展示格式，而不是剪贴板复制格式。

可以类似这样：

`默认复制格式：<下拉框>`

交互上建议是：

- `Ctrl/Cmd + C` 使用这里配置的默认复制格式；
- 右键复制菜单展示完整的复制格式列表；
- 当前默认格式放在右键列表第一个；
- 用户如果只是临时想用其他格式，就从右键菜单里选一次。

这样两个需求都能覆盖：默认行为放在设置类入口里，临时复制格式放在右键菜单里，语义会更清楚一些。

t8y2 (changes requested):
I'm going to request changes for now. Three behaviors still need fixing before this is safe to merge:

1. `apps/desktop/src/lib/dataGrid/gridSelection.ts:39` sorts selected source indexes numerically. After the user drags columns into a different visible order, CSV/TSV/JSON/Markdown/SQL copying reverts to the original source order. Please deduplicate while preserving the first visible occurrence, and add reordered-plus-hidden-column coverage.

2. `crates/dbx-core/src/data_grid_extractors/delimited.rs:35-118` converts SQL NULL into `null_text` before the normal string quoting path. With the default sentinel, NULL and the literal string `NULL` can produce the same output; an empty sentinel also collides with an empty string. Please preserve nullness through final encoding and define/test the sentinel behavior for the supported quoting modes.

3. `apps/desktop/src/components/grid/DataGrid.vue:7511` persists `dataGridCopyExtractor` when a format is selected from the context menu. A one-off right-click copy therefore silently changes later `Ctrl/Cmd+C` behavior. The context menu should call the selected extractor without updating the saved default; the default belongs in the separate settings entry discussed earlier.

The branch is also currently conflicting with `main`, so the conflicts need to be resolved after the behavior fixes.

soddygo:
Thanks for the detailed review — all three behaviors are fixed, and I also consolidated the right-click Copy menu onto the extractor pipeline while resolving the entry-point gap.

## Round-2 fixes

1. **Column order** — `dedupeColumnIndexes` (gridSelection.ts) preserves the first visible occurrence on the extractor path only; the shared `normalizeSelectedColumnIndexes` keeps its sort for `hasConsistentColumns`. Added reordered + hidden-column coverage.

2. **NULL / sentinel** — NULL is carried as structured state (`DsvCell`) to the field writer: the bare sentinel is never quoted, and a string equal to it is force-quoted in Minimal mode so the two stay distinguishable. CSV/one-row regressions for Minimal / Always / Never / empty-sentinel added.

3. **Context-menu default** — right-click copy no longer persists `dataGridCopyExtractor`; the ✓ marker reflects the saved default only.

## Copy menu consolidation

The toolbar and keyboard already used the extractor; the right-click menu was the last legacy holdout, showing duplicate INSERT/UPDATE/TSV actions alongside the new formats. Consolidated onto the extractor as the single copy path:

- Extractor format list flattened directly into the Copy submenu (was nested under a dead "Copy as" parent — `CustomContextMenu` only renders 2 levels).
- Legacy duplicate actions removed; `selectionSubmenu` → "Clear Selection" promoted to a top-level item; `cutSelection` (Ctrl+X) migrated to `copyWithExtractor("tsv")`.
- Net **−1677 lines** (legacy copy functions, caches, Mongo update builder, i18n keys, gridSelection formatters). Every removed format stays reachable — same Rust `build_data_grid_copy_insert/update_statements` via `extractDataGridSelection`.

Also fixed a pre-existing CI failure on main: `DataGridClipboardRegion.spec.ts` searched for a single-line `<div v-if="showTableInfo">` that was always multi-line (main CI was red for the same assertion).

CI is green (all 6 jobs pass).

## Planned follow-ups (separate PRs)

Kept these out to limit the diff — happy to do them as focused PRs:

- **Mongo INSERT → Rust**: the Mongo insert path bypasses the Rust pipeline (TS-side `buildMongoExtractorInsert`). Moving it to a Rust extractor variant would unify the pipeline and let Mongo benefit from the NULL/dialect fixes.
- **Column-classification predicates**: TS (`isAutoGeneratedColumn` / `isComputedColumn`) duplicates Rust (`is_auto_generated_column`). A Rust `extractor_capabilities` endpoint would eliminate the drift risk.
- **SQL builder error codes**: `build_data_grid_copy_update_statements` returns `Vec` and the caller infers the failure from length — should return `Result` with specific error codes.
- **`format_grid_sql_literal`**: the 110-line per-dialect formatter could be split into per-database impls.

Happy to address anything else.

soddygo:
CleanShot 2026-07-26 at 22 09 23@2x


改了一版, 比我想象中的多,还有一些代码可以拆分,或者统一下沉到rust逻辑里,但担心越改越多,后续可以单独开PR,来对部分逻辑拆分和重构,部分单文件太大了,负责的职责太多了

t8y2 (changes requested):
request changes: the latest head ee63dc00 still removes existing copy behavior and cannot be merged safely.

1. **P1 — MongoDB row updates are no longer reachable.** The previous `useDataGridExport` path generated `db.getCollection(...).updateOne(...)` statements from the original Mongo documents. The current extractor capability list explicitly disables `sql-updates` for `mongodb` (`apps/desktop/src/lib/dataGrid/dataGridCopyExtractor.ts:35-45`), while the current Copy submenu only exposes the extractor list (`apps/desktop/src/components/grid/DataGrid.vue:7526-7537`). Users lose the existing right-click “Copy Row(s) as Update” action for MongoDB. Please preserve a Mongo-specific update extractor/action rather than removing the old path.

2. **P1 — right-click SQL INSERT no longer preserves the old row-copy behavior.** The dedicated row INSERT actions were removed from `apps/desktop/src/components/grid/DataGrid.vue:7526-7537`. A cell context menu selects only the clicked cell (`DataGrid.vue:6658-6672`), and `useDataGridExtractor` then builds the request from that cell selection and its selected source columns (`apps/desktop/src/composables/useDataGridExtractor.ts:75-83`). Therefore choosing SQL INSERT after right-clicking a cell emits an INSERT for the selected cell/column instead of the previous full-row INSERT. Please retain an explicit row INSERT action, or make the extractor distinguish row-copy from cell selection while preserving the old behavior.

The targeted frontend tests pass (5 files, 30 tests), but they do not cover either removed behavior. The PR also remains conflicting with `main`.

t8y2 (changes requested):
One blocker remains in the right-click copy path.

Right-clicking an unselected cell creates a synthetic single-cell selection in `DataGrid.vue` before the extractor request is built. `useDataGridExtractor` then prioritizes `selectedCellMatrix`, so the intended context-row fallback never runs. As a result, context-menu SQL INSERT/UPDATE and Mongo INSERT can still copy only the clicked cell instead of the full row.

Please preserve the row context for context-menu copy, or make the extractor prefer `contextCell` over the synthetic selection, and add a regression covering right-click on a cell with no prior selection.

t8y2 (changes requested):
Request changes on `758e6354b9622aa5d228e163c48cf444ee47b1bb`.

The previous unselected-cell right-click issue is fixed, but three blockers remain:

1. A genuine 1×1 selection is still treated as the synthetic context-menu selection. Selecting one cell and right-clicking that same cell copies every visible column in the row instead of the selected cell. Please track whether the context menu actually created a synthetic selection and cover both genuine and synthetic 1×1 cases.

2. TSV copy/paste is misaligned when `includeRowHeader` is enabled and the data contains `NULL`. The rendered text includes the row-number column, while the internal clipboard matrix still contains the original rows. Please make the internal matrix match the effective TSV schema, or avoid storing an incompatible matrix, and add a copy-then-paste regression.

3. `selectedColumnIndexes` is not included in the input budget. The extractor allocates from this externally supplied length before rejecting invalid indexes. Please cap it against the column/max-column budget before allocation and add an oversized-index-list regression.

t8y2 (approved):
维护者补丁 `88024e3` 已基于作者最新 head `27cd479` 完成：

- 所有左键/Shift/Ctrl/Cmd 选区入口统一失效 synthetic context selection，不再复制旧右键单元格；
- TSV 行号模式的内部 clipboard 与实际 schema 等宽，保留行号、header、结构化 `null`、tab/newline，copy→paste 不再错位；
- extractor 的 separator/nullText/quote 校验与归一化统一按 Unicode code point 长度；
- `selectedColumnIndexes` allocation budget 保持在分配前校验。

验证：相关 Vitest 111 项、Rust extractor 41 项、typecheck、定向 lint/format、`git diff --check` 通过。作者 head 已包含当前 main，无需额外 merge commit。

t8y2:
Thanks for the contribution! Merged in 0b7a886, will be released in the next version.

PR #4207, opened 2026-07-22
feat(updater): support portable auto updates

wfion:
## 变更说明

已完成便携版自动更新功能，当前改动尚未提交。
主要实现：
- 应用内下载对应架构的便携 ZIP 和 .sig，使用 Tauri 公钥验证签名。
- 退出后仅替换 DBX.exe，保留 portable.dbx、data 及其他文件。
- 更新失败自动恢复旧 EXE，成功后自动重启。
- 支持 GitHub、R2、CNB 下载源。
- Release 流程会签名并上传便携 ZIP，同时保留其签名文件。
- 已更新七种语言和更新对话框。

## 变更类型

- [x] 新功能

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

## 关联 Issue

Related #4185


t8y2 (changes requested):
request changes: the portable updater currently introduces a downgrade path.

`src-tauri/src/commands/update.rs` trusts the frontend-provided `latest_version`, while the portable path in `src-tauri/src/commands/update_portable.rs` does not use the requested version at all. The ZIP signature proves that an artifact was signed by the project, but it does not prove that the signed artifact is newer than the installed app. A caller can therefore request and reinstall an older, still-valid signed portable ZIP.

Please enforce this in the backend before download/install: parse the requested and current app versions and reject `requested_version <= current_version`. The installed executable/archive version should also be checked or cryptographically bound to the signed update metadata so the version string cannot disagree with the downloaded artifact. Add regression tests for equal-version and downgrade requests.

Tauri's updater performs version comparison by default and exposes an explicit `allowDowngrades` option only when rollback is intentionally enabled: https://v2.tauri.app/reference/javascript/updater/

t8y2 (approved):
The portable update validation and frontend test coverage are fixed on c001d9eb8. Targeted updater tests and typecheck passed locally.

t8y2 (approved):
The portable update validation and frontend test coverage are fixed on c001d9eb8. Targeted updater tests and typecheck passed locally.

t8y2:
Thanks for the contribution! Merged in f7741a55f, will be released in the next version.

PR #4209, opened 2026-07-22
feat(grid): add table font shortcuts to view options

ZhongYanZhiShi:
## Summary

- Add a table font selector to View Options for both query results and table data.
- Place the font selector immediately above the table font size control.
- Reuse the persisted `tableFontFamily` setting from Appearance settings.
- Apply the selected font consistently to result grids, transposed views, cell details, bulk editing, and MongoDB JSON previews.
- Match the font selector styling in settings with the existing date and time selectors.

## Label wording

The compact View Options panel uses **“Table font”** because it appears directly above **“Table font size”**. Using paired labels makes their relationship clearer and keeps the toolbar wording concise.

The Appearance settings label remains **“Result Grid Font”** because it more precisely describes the broader scope of the setting. This change intentionally does not alter that existing settings label.

## Validation

- [x] `make check`
- [x] `make cargo-check-fast`
- [x] `pnpm build`
- [x] Manual UI verification
- [x] Regression test covering the context-specific labels

## Demo

Screenshot_20260722_205348

## Related issue

Related to #725


t8y2:
Thanks for the contribution! Merged in 5d1a6e63c, will be released in the next version.

PR #4242, opened 2026-07-23
fix(connection): keep db picker top border on hover

xddcode:
## Summary
- Database type cards used `hover:-translate-y-0.5`, which overflow clipped the top border in the connection dialog picker.
- The database search input focus ring was also clipped by a parent `overflow-hidden`.
- Replace the card lift with `hover:shadow-sm`, drop the unnecessary outer overflow clip, and leave padding so focus rings stay visible.

## Test plan
- [ ] Open New Connection and hover a database card in the icon picker
- [ ] Confirm the full border (including top edge) remains visible
- [ ] Focus the database search input and confirm the full border/focus ring is visible on all sides
- [ ] Confirm selected and unselected cards still show hover feedback

### DIFF
--- a/apps/desktop/src/components/connection/ConnectionDialog.vue
+++ b/apps/desktop/src/components/connection/ConnectionDialog.vue
@@ -4391,7 +4391,7 @@ function openExternalUrl(url: string) {
                     v-for="opt in category.options"
                     :key="opt.value"
                     type="button"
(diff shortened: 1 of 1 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in fa26a56ab, will be released in the next version.

PR #4347, opened 2026-07-23
feat(xugu): expose table child metadata

mapan0424:
## Summary

- expose Xugu table child metadata for constraints, table-level triggers, indexes, partitions, and subpartitions
- use access-scoped `ALL_*` catalogs for Xugu foreign keys and table triggers, preserving referenced schema/table/action metadata
- add the desktop tree groups and Core/Web/Tauri Agent RPC plumbing without changing non-Xugu table trees

## Validation

- `cd agents/drivers/xugu && go test ./...`
- `pnpm vitest run apps/desktop/src/stores/__tests__/connectionStore.xuguTableChildren.spec.ts`
- `pnpm typecheck`
- `cargo check -p dbx-core`
- `cargo check -p dbx-web`
- `cargo check -p dbx`
- `cargo test -p dbx-core agent_driver::tests::defines_agent_protocol_methods`
- local `SHOP_DEMO` regression with `SYSDBA`: constraints, FKs, table trigger, RANGE partitions, and LIST subpartitions

## Scope

Xugu-specific table metadata and its additive agent protocol routes. Non-Xugu table tree behavior is covered by regression tests and unchanged.


t8y2 (changes requested):
Request changes: the protocol contract is now synchronized, but two compatibility and presentation issues remain.

1. The desktop unconditionally exposes Constraints, Partitions, and Subpartitions for Xugu connections, while the runtime can still launch any previously installed Xugu Agent without a method-capability or minimum-version check. Older Agents use the same protocol generation but do not implement these methods, so expanding the new nodes returns `Unknown method`. Please add a capability/version gate or provide a backward-compatible fallback before exposing the nodes.

2. Foreign keys are still returned by both groups. `listConstraints` includes `CONS_TYPE='F'`, while `listForeignKeys` separately queries the same type and the tree displays both Constraints and Foreign Keys. This duplicates every foreign key. Please exclude foreign keys from the generic constraints query and update the test that currently expects both groups to contain them.

t8y2 (commented):
Maintainer patch applied in 5263a459.

The table-child metadata requests correctly pass the database selected in the tree, but Xugu custom native DSNs, JDBC URLs, and DBX URLs continued using the database embedded in the connection string. Expanding a table under another database could therefore query the original catalog and return missing or incorrect constraints, foreign keys, triggers, indexes, and partitions.

The patch makes an explicit selected database override the connection-string database for all three supported formats while preserving the original database when no override is provided. Native DSN replacement is quote-aware so `DB=` text inside passwords or other values is not modified.

Validation:
- `go test ./...`
- `go vet ./...`
- Xugu Agent native build
- `python3 scripts/validate_agents.py`
- `python3 -m unittest validate_agents_test.py` — 13 tests passed
- `gofmt` and `git diff --check`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 0362fef2, will be released in the next version.

PR #4380, opened 2026-07-24
fix(redis): show confirm dialog for dangerous Redis commands in SQL editor

### LINKED ISSUE #4296: [Bug] Redis SQL编辑器拦截危险命令未拦截DEL
### 来源
QQ 群反馈 · **反馈人**：NO

### 数据库类型和版本
Redis

### AI 摘要
在Redis SQL编辑器中启用了拦截危险命令功能，但执行DEL命令时未被拦截，应拦截却没有拦截。

### 原始描述
REDIS的sql编辑器开启了拦截危险命令，但是执行DEL的时候没有拦截

dienaso:
## Summary

Fix Redis `confirm`-class commands (DEL, FLUSHDB, JSON.DEL, etc.) being silently executed in the SQL editor when "Block dangerous commands" was enabled, and align the frontend command classification with the backend.

Fixes #4296

## Why

The Redis command safety classification (`Blocked` / `Confirm` / `Write` / `Allowed`) was implemented in both frontend and backend, but the SQL editor only handled `Blocked` commands (toast rejection). `Confirm` commands like `DEL` fell through and executed without any user confirmation, even though the Redis Key Browser already had this flow.

A second failure mode: the frontend defaulted unknown commands to `Allowed`, while the backend defaulted them to `Blocked`. A multi-line batch containing an unknown module command (e.g. `FCALL`) would slip through the frontend gate and reach the backend, which would then reject the whole batch — bypassing the safety classification entirely.

### `apps/desktop/src/composables/useSqlExecution.ts`

- Route Redis `Confirm` commands through the existing `DangerConfirmDialog` instead of executing silently.
- Scan the **entire** command batch for the highest safety level (`Blocked` > `Confirm`) before deciding. `Blocked` exits early; otherwise `Confirm` opens the dialog. This prevents a benign line earlier in the batch from masking a dangerous line later.
- Bind `pendingDangerKind` (`"sql" | "redis"`) to the danger dialog so `onDangerConfirm` does not need to recompute Redis state from the active connection. `suppressDangerConfirm` is reset both when the dialog opens and when it closes/cancels, to prevent state leakage from prior SQL dialogs.
- `onDangerConfirm` only disables `confirmDangerousSqlExecution` when `pendingDangerKind === "sql"`, so confirming a Redis operation cannot accidentally turn off SQL danger confirmation.
- The blocked-command toast now includes the offending command name (via `firstRedisCommandToken`) instead of a generic `"Redis"` placeholder.

### `apps/desktop/src/components/layout/AppDialogs.vue` + `App.vue`

- `AppDialogs` accepts an `activeDatabaseType` prop. The "suppress future prompts" toggle is hidden for Redis sessions, since that toggle mutates `confirmDangerousSqlExecution` which is a SQL-only setting.
- `App.vue` passes `activeConnection?.db_type` through.

### `apps/desktop/src/lib/redis/redisCommandSafety.ts`

- Add an explicit `ALLOWED_COMMANDS` read-command allowlist matching the backend classification.
- Unknown commands now default to `Blocked` (fail-closed) instead of `Allowed`, mirroring `crates/dbx-core/src/db/redis_driver.rs::classify_command`.
- Add missing module/Redis 7+ commands to `CONFIRM_COMMANDS`: `JSON.DEL`, `JSON.FORGET`, `JSON.CLEAR`, `JSON.ARRPOP`, `JSON.ARRTRIM`, `BLMOVE`, `BLMPOP`, `BLPOP`, `BRPOP`, `BRPOPLPUSH`, `LMOVE`, `LMPOP`, `RPOPLPUSH`, `BZMPOP`, `BZPOPMAX`, `BZPOPMIN`.
- Move `FLUSHDB` from implicit-blocked (via the old unknown → allowed fallback no longer applying) to `Confirm` so the user gets a confirmation prompt before wiping the current DB.
- Header comment points at the Rust mirror — both sides must be kept in sync.

### `crates/dbx-core/src/db/redis_driver.rs`

- Add `FLUSHDB` to the `Confirm` arm to match the frontend classification.
- Header comment on `classify_command` points at the TS mirror.
- Update regression test: previously `flushdb` was asserted as `Blocked`; now `Confirm`. Add assertions for `JSON.DEL`, `JSON.FORGET`, `JSON.CLEAR`.
- Add `classify_command_batch_blocks_on_any_unknown_member` regression covering both `["DEL victim", "FCALL wipe 0"]` and `["FCALL wipe 0", "DEL victim"]` orderings (see Follow-up below).

### Follow-up: batch-order regression + typecheck cleanup (commit 59a06023)

Addresses review feedback "include both dangerous-command batch orders". Adds one Rust test and one Vitest test that exercise the per-command classifier across both DEL/FCALL orderings; the batch scan must short-circuit to `Blocked` whenever any line is unknown. Also removes a dead `firstRedisCommandToken` import and a redundant `highestSafety !== "blocked"` check in the Redis batch scan (both flagged by `pnpm typecheck`).

- `apps/desktop/src/lib/ai/__tests__/aiRedisConsoleRouting.spec.ts`: new `it("blocks a batch whenever any line is an unknown command", …)` covers both orderings via `classifyRedisCommandSafety`.

## Semantic decisions worth calling out

- **`FLUSHDB` is now `Confirm`, not `Blocked`.** It only wipes the *current* logical DB (vs `FLUSHALL` which still wipes everything and remains `Blocked`). The user gets one click to abort; this is intentional and reversible. `FLUSHALL` is unchanged.
- **Unknown commands default to `Blocked` on both sides.** Anything not on the allowlist fails closed. This rejects new Redis-version commands and module commands until DBX explicitly classifies them. Tradeoff: stricter safety, slightly more friction for users on cutting-edge Redis builds. Follow-up: an in-app way to override or report unknown commands.
- **`FLUSHDB` test was deleted from `Blocked` and re-asserted as `Confirm`.** Same test name, different expected value — intentionally surfaces the semantic change.

## Test coverage

- **`apps/desktop/src/lib/redis/__tests__/redisCommandSafety.spec.ts`** (new): unit tests for each safety branch plus the `FCALL`/`UNKNOWN_CMD` fail-closed guarantee.
- **`apps/desktop/src/composables/__tests__/useSqlExecution.spec.ts`** (extended): new `describe("Redis safety", …)` block covering single-line confirm, single-line blocked, multi-line blocked-before-confirm, the `block-dangerous` off bypass, and the `pendingDangerKind` discriminator (Redis confirm must not disable `confirmDangerousSqlExecution`, SQL confirm must). The pre-existing "dangerous script pending" test continues to cover the SQL path.
- **`apps/desktop/src/lib/ai/__tests__/aiRedisConsoleRouting.spec.ts`** (extended, commit 59a06023): batch-order regression — both `["DEL", "FCALL"]` and `["FCALL", "DEL"]` must produce at least one `blocked` line.
- **Backend**: existing tests in `redis_driver.rs::tests` updated; new assertions for `JSON.DEL` / `JSON.FORGET` / `JSON.CLEAR`. New `classify_command_batch_blocks_on_any_unknown_member` (commit 59a06023) covers both batch orderings end-to-end through `classify_command`.

## Testing done

- `pnpm lint` — 0 errors
- `pnpm typecheck` — no new errors (commit 59a06023 also clears the two pre-existing `useSqlExecution.ts` errors left by 31726ab5)
- `pnpm test -- useSqlExecution` — all new and existing cases pass
- `pnpm test -- redisCommandSafety` — all classifier branches pass
- `pnpm test -- aiRedisConsoleRouting` — 4 cases pass including the new batch-order test
- `cargo test -p dbx-core --no-default-features redis_driver` — updated assertions and the new batch-order test pass
- Manual: in DBX with "Block dangerous commands" enabled, `DEL key:1` shows the confirmation dialog and "Suppress future prompts" is hidden; `KEYS *` is blocked with a toast naming the offending command.

## Frontend/backend drift

`redisCommandSafety.ts` and `redis_driver.rs::classify_command` are now mirrored by header comments in both files. A follow-up issue should generate one side from the other (build script / `include_str!`) to eliminate manual drift.


t8y2 (dismissed):
Request changes: the Redis confirmation path currently weakens the existing command safety gate in multi-command input.

1. `continueExecute()` returns as soon as it encounters the first `confirm` command, so it never checks whether a later command is `blocked`. For example, `DEL victim\nFLUSHALL` opens the confirmation dialog for `DEL`; after confirmation, execution can delete the key before the backend rejects `FLUSHALL`. Please scan the complete batch first and apply the highest safety level (`blocked > confirm > write > allowed`), with regression tests for both command orders.

   Evidence: [Redis DEL](https://redis.io/docs/latest/commands/del/) states that it deletes keys, while [Redis FLUSHALL](https://redis.io/docs/latest/commands/flushall/) removes keys from all databases.

2. The pending confirmation does not record which safety flow opened it. `onDangerConfirm()` recomputes Redis state from the current connection and the complete SQL text, while the shared suppress flag can retain state from a previously cancelled SQL dialog. Please bind the danger kind/database type to the pending operation and reset the suppress state when opening or cancelling the dialog.

3. `FLUSHDB` is classified as `confirm` in the frontend, but the Rust Redis safety classifier falls through to `blocked`. The advertised flow therefore confirms the operation and then rejects it. Please use one consistent classification or add an explicit, narrowly scoped approval path.

   Evidence: [Redis FLUSHDB](https://redis.io/docs/latest/commands/flushdb/) documents that it deletes all keys in the selected database.

t8y2 (dismissed):
Request changes: the latest commit fixes the pending-confirmation reset and improves the FLUSH handling, but the confirmation scan can still be bypassed and the exact head has a failing Rust test.

1. `apps/desktop/src/lib/redis/redisCommandSafety.ts:117` treats unknown commands as allowed, while `crates/dbx-core/src/db/redis_driver.rs:1694` rejects unknown commands and commands such as `FCALL`. For a batch such as `DEL victim` followed by `FCALL wipe 0`, the confirmation path can execute the destructive first command before the backend rejects the second command. The frontend scan must fail closed using the same classification source as execution.

   Evidence: [Redis FCALL](https://redis.io/docs/latest/commands/fcall/):
   > “Invokes a function.”

2. The frontend confirmation list also omits destructive module commands that the backend already classifies as confirmation-required, including `JSON.DEL`, `JSON.FORGET`, and `JSON.CLEAR`. `JSON.DEL doc $` currently executes without a prompt.

   Evidence: [Redis JSON.DEL](https://redis.io/docs/latest/commands/json.del/):
   > “Deletes the JSON value at path in key.”

3. `crates/dbx-core/src/db/redis_driver.rs:3927` still expects `FLUSHDB` to be blocked, while the implementation now classifies it as confirmation-required, so `classifies_safe_confirmed_and_blocked_commands` fails. Please update the regression coverage and include both dangerous-command batch orders.

dienaso:
Addressed both review requests in commit `59a06023`:

1. **Frontend fail-closed on unknown commands** — `redisCommandSafety.ts:268-277` now returns `"blocked"` for any command not in the explicit lists. Mirrors `redis_driver.rs:1694`. The `DEL victim\nFCALL wipe 0` path is closed: the batch scan in `useSqlExecution.ts` short-circuits to `Blocked` on the `FCALL` line and never reaches `DEL`.

2. **`JSON.DEL` / `JSON.FORGET` / `JSON.CLEAR`** — added to `CONFIRM_COMMANDS` (`redisCommandSafety.ts:153-155`) and to the backend `Confirm` arm (`redis_driver.rs:1681`). `JSON.DEL doc $` now prompts.

3. **`FLUSHDB` regression test** — `classifies_safe_confirmed_and_blocked_commands` now asserts `Confirm` (was `Blocked`); matches the new implementation.

4. **Batch-order regression coverage** — added:
   - Rust: `classify_command_batch_blocks_on_any_unknown_member` runs `classify_command` over both `["DEL victim", "FCALL wipe 0"]` and `["FCALL wipe 0", "DEL victim"]`, asserting the batch converges to `Blocked` whenever any line is unknown.
   - Frontend: `aiRedisConsoleRouting.spec.ts > blocks a batch whenever any line is an unknown command` mirrors the assertion via `classifyRedisCommandSafety`.

5. **Drive-by typecheck cleanup** (same commit): removed unused `firstRedisCommandToken` import and a redundant `highestSafety !== "blocked"` check in `useSqlExecution.ts`. Both were flagged by `pnpm typecheck` after 31726ab5.

Verification on the new commit:
- `cargo test --no-default-features -p dbx-core --lib -- classify_command_batch classifies_safe_confirmed` → 2 passed
- `vitest run apps/desktop/src/lib/ai/__tests__/aiRedisConsoleRouting.spec.ts` → 4 passed
- `pnpm typecheck` → exit 0 (was failing on 2 pre-existing errors before this commit)
- `cargo fmt --check`, `cargo clippy -D warnings`, `pnpm lint`, `oxfmt --check` → all clean

Ready for re-review.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 355ca7d3a, will be released in the next version.

PR #4396, opened 2026-07-24
feat(sidebar): add persistent table filters

### LINKED ISSUE #4385: [Feature] add persistent table filters for large schemas
**Background**
When a database/schema contains hundreds or thousands of tables, the sidebar Tables tree becomes noisy and hard to navigate. Most users only care about a subset of business tables, but currently they need to browse or search repeatedly.

**Proposal**
Add a persistent table filter configuration on the sidebar Tables node so users can focus on the tables they care about.

**Suggested behavior:**

Add a Filters... action to the Tables context menu.
- Support include and exclude pattern lists.
- Persist filters per connection/database/schema/table-group scope.
- Apply filters when loading/paginating table metadata.
- Show an indicator near the table count when a filter is active, e.g. filtered / 存在过滤.

**User Value**
This helps users working with very large schemas quickly narrow the sidebar tree to relevant tables, reducing visual noise and repeated manual searching.

**Notes**
The behavior is similar to DBeaver's table filter capability, but should fit DBX's sidebar tree and metadata loading model.

wbean:
## Summary

- Add persistent Include/Exclude table name filters to the sidebar `Tables` context menu.
- Apply SQL LIKE-style table filters to metadata loading, with MySQL pushdown and local filtering fallback for non-pushdown paths.
- Show an active filter indicator next to the `Tables` count and add localized strings.

Closes #4385

## Test Plan

- `pnpm typecheck`
- `cargo check -p dbx-core -p dbx-web -p dbx-mcp`
- `cargo test -p dbx-core table_name_filter`
- Manual Tauri dev app check for `Tables -> Filters...`

filter-button

filter-dialog


t8y2 (dismissed):
Request changes: the rebase resolves the merge conflict, but the three correctness blockers remain unchanged.

1. `crates/dbx-core/src/schema.rs:58` still implements `%` by recursively trying every remaining position without memoization. Adversarial failing LIKE patterns grow exponentially and can stall metadata loading once the fallback runs across a large schema. Please replace this with a bounded DP/automaton implementation and add an adversarial regression test.

2. `apps/desktop/src/components/sidebar/ConnectionTree.vue:1146` starts an independent forced refresh after every save, while `apps/desktop/src/stores/connectionStore.ts:3587` only checks that the node/search context is still valid. If save A is slower than save B, A can finish last and overwrite the tree with results for the old filter. Please add a per-scope generation/revision check and test the out-of-order completion case.

3. `apps/desktop/src/stores/connectionStore.ts:1882` still removes the connection without removing its entries from `sidebarTableNameFilters` or persisting the cleaned map. Re-importing a connection with the same ID can unexpectedly restore the deleted connection's filters. Please cover both single and batch deletion.

wbean:
Thanks for the review. I addressed all three blockers in the latest push:

1. Replaced the recursive SQL LIKE matcher with a bounded dynamic-programming implementation (`O(pattern_len * value_len)`) and added an adversarial failing-pattern regression test.
2. Added per-scope table filter revisions. Filter-triggered refreshes now carry the expected revision and stale out-of-order completions are ignored before applying tree children. Added a regression test where save/refresh A finishes after save/refresh B.
3. Connection removal now removes and persists cleanup for all matching `sidebarTableNameFilters` entries. Added coverage for both `removeConnection` and batch `removeConnections`.

Validation run locally:

- `pnpm test --run apps/desktop/src/stores/__tests__/connectionStore.metadataLoading.spec.ts packages/app-tests/connectionStorePinnedCleanup.test.ts packages/app-tests/sidebarRuntimeDecomposition.test.ts`
- `pnpm typecheck`
- `cargo check -p dbx-core -p dbx-web -p dbx-mcp`
- `cargo test -p dbx-core table_name_filter`
- `git diff --check`

t8y2 (dismissed):
The three previous blockers are addressed, but the current PR merge result does not compile.

`crates/dbx-core/src/schema.rs:2385` still has the Cloudberry branch calling `filter_table_infos(tables, filter, limit, offset, object_types)` with the old five-argument shape, while this PR changed the helper to also require `table_name_filter`. This is why `packages`, `rust-fmt-clippy`, and `rust-test` are failing with Rust `E0061`.

Please update the branch against current `main` and pass `table_name_filter` through the Cloudberry path as well, then rerun the Rust checks.

t8y2 (dismissed):
I'm going to request changes because the current head fails the required Rust lint job. `crates/dbx-core/src/schema.rs:91` performs an element-by-element slice copy, which triggers `clippy::manual_memcpy` in `rust-fmt-clippy`. Please replace it with the repository-supported slice-copy idiom, keep the behavior unchanged, and rerun the lint job.

t8y2 (commented):
Maintainer patch applied in 5f83f77d6.

The table-name filter matcher copied the previous DP row into the next row one element at a time. Clippy correctly reported this as `manual_memcpy`; the warning caused the required `rust-fmt-clippy` job and the Rust aggregate check to fail.

The patch replaces only that unconditional slice copy with `copy_from_slice`, preserving the existing matching behavior while removing the lint failure.

Validation:
- `cargo fmt --check`
- `cargo test -p dbx-core --locked --no-default-features --lib table_name_filter` (4 passed)
- `cargo clippy -p dbx-core --locked --lib --no-default-features -- -D warnings`
- all-targets Clippy check for `manual_memcpy` passed; the repository's unrelated pre-existing `questdb.rs` `items_after_test_module` lint was isolated
- Sidebar and metadata Vitest suites (95 passed)
- `git diff --check`

t8y2 (approved):

t8y2 (approved):

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 98d1ab3ab, will be released in the next version.

PR #4402, opened 2026-07-24
feat(agents): concurrent all-agent-driver upgrade

Abeautifulsnow:
## Summary

Replace the sequential **"Upgrade All"** driver update loop with concurrent execution via `futures::future::join_all`. All updatable agent drivers now download and install in parallel instead of one at a time.

## Motivation

The existing `upgrade_all_agent_drivers_from` uses a plain `for` loop with `.await` — each driver waits for the previous one to fully complete. With a handful of drivers this adds up to noticeable wall-clock time; the downloads could run concurrently.

### Backend (`crates/dbx-core`)

**`agent_manager.rs`** — two new locks on `AgentManager`:

| Lock | Purpose |
|------|---------|
| `state_lock: Mutex<()>` | Serialises `load_state → modify → save_state`, preventing lost updates when multiple driver installs write `state.json` concurrently |
| `jre_install_locks: Mutex<HashMap<String, Arc<Mutex<()>>>>` | Per-JRE-key install deduplication via double-checked locking — multiple drivers sharing the same JRE (e.g. JRE 21) download and extract it exactly once |

**`agent_service.rs`** — concurrent execution + safety:

- `upgrade_all_agent_drivers_from`: `for` loop → `futures::future::join_all` with borrow-safe references (no `'static`/`Arc` wrapping needed)
- `ensure_jre_from_registry`: DCL pattern — fast-path skip if installed, per-JRE lock, re-check, then download
- Stash persistence split into two phases under `state_lock`: stash path persisted **before** extraction (crash recovery), JRE version persisted **after** extraction succeeds
- `stop_daemons()` → `stop_daemons_using_jre()` — stops only daemons using the specific JRE key, avoiding mid-install kills of unrelated drivers
- Per-JRE temporary archive names (`jre-{key}-download.tar.gz`) replace the shared `jre-download.tar.gz`

### Frontend (`apps/desktop`)

**`driverInstallProgressUi.ts`** — per-driver progress model:

- `DriverInstallProgressTargetState.progress` (single ref) → `progressMap` (per-driver record)
- New `updatePerDriverProgress()` helper routes events by `db_type`
- Batch counter uses a **monotonic completed count** (`upgradingCompletedCount`) instead of a fixed list index — no more misleading `[3/5] → [2/5]` regression on out-of-order events

**`DriverStoreDialog.vue`** — adapted to the new model:

- Global `agentProgress` ref replaced with `reactive(agentProgressByDbType)` map
- Each driver row reads progress from `getAgentProgressPercent(dbType)` + `getAgentProgressTitle(dbType, ...)`
- JRE reinstall progress uses its own isolated ref
- "Upgrade All" button shows `completedCount / totalCount`

**Tests** (`driverInstallProgressUi.spec.ts`) — 2 new test cases:
- Concurrent done event clears only the target driver; others remain visible
- `all-done` event clears the entire progress map

## Performance

| Scenario | Before | After |
|----------|--------|-------|
| N drivers | sum of all download + install times | ≈ max single download + install time |
| Shared JRE (e.g. 3 Java drivers all on JRE 21) | downloads & extracts 3 times | downloads & extracts once |

## Risk & Safety

- State mutations (`state.json`) are now always under `state_lock` — no lost updates
- JRE deduplication uses proven DCL pattern with immediate version persistence
- Stash crash recovery: stash is persisted before extraction, so a crash midway still leaves the orphan directory in `pending_jre_cleanup` for startup cleanup
- `replace_old_jre_dir` no longer writes state directly — all state writes flow through the same `state_lock`-guarded path

## 变更类型

- [ ] 新功能
- [ ] Bug 修复
- [x] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [x] 相关测试通过


t8y2 (dismissed):
Request changes: the concurrent batch path has useful failure isolation, but the shared installation state and duplicate-operation boundaries are not yet safe.

1. `crates/dbx-core/src/agent_service.rs:574` and the local-JAR fallback around `:309` still perform independent `load_state` / `save_state` sequences without the new shared state lock. Running JRE reinstall or single-driver fallback alongside the batch upgrade can write an older snapshot over successful driver/JRE updates. Please centralize all installation-state mutations behind the same lock and audit reinstall, uninstall, and fallback paths.

2. `crates/dbx-core/src/agent_service.rs:929` uses a fixed `<agent>.download` temporary path, but the single and batch commands have no per-driver or operation-level mutual exclusion. Duplicate requests can concurrently truncate/write/rename the same file. Progress is also emitted globally from `src-tauri/src/commands/agents.rs:210` without an operation ID, so another upgrade's `done` or `all-done` event can corrupt the current dialog's progress. Please add backend deduplication/locking and operation-scoped progress events, with concurrent-request regression coverage.

t8y2 (dismissed):
Request changes: the latest commits resolve the shared-state overwrite and duplicate online-download races, but offline import still sits outside the same operation-safety contract.

1. `crates/dbx-core/src/agent_service.rs:1428` runs `import_offline_zip()` without the new installation/JRE/driver locks while replacing JRE and driver directories. The import button at `apps/desktop/src/components/config/DriverStoreDialog.vue:1273` is disabled only by `importingZip`, so an offline import can still overlap Upgrade All or a JRE operation. Please put offline import under the same operation-lock model, or centrally block all conflicting entry points, and add a concurrent import/upgrade regression test.

2. `crates/dbx-core/src/agent_service.rs:622` puts the display label into `AgentProgressEvent.db_type`, while `apps/desktop/src/lib/connection/driverInstallProgressUi.ts:55` keys progress by the real database type. A label such as `MySQL` will not match the `mysql` row, so offline-import progress can disappear. Please carry the real `db_type` separately from the display label and cover this path in a focused test.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 95db01e35, will be released in the next version.

PR #4405, opened 2026-07-24
feat(postgres,opengauss,kingbase)/Add ddl  triggers in postgres, opengauss and kingbase

amwps290:
## 变更说明

+ postgresql, opengauss 表 ddl 新增触发器
+ kingbase 表 ddl 新增触发器，注释，索引

## 变更类型

- [x] 新功能

## 涉及前端

image
image

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

## 关联 Issue

Related #4123 #4000


t8y2 (commented):
The PostgreSQL direction looks reasonable, but this needs real-version validation before approval. `crates/dbx-core/src/schema.rs:7092` now joins trigger metadata with the rest of table metadata; a trigger query failure caused by permissions, version differences, or compatibility mode can therefore fail the whole DDL request. Please verify PostgreSQL, openGauss, and Kingbase success and failure paths, including both catalog variants used by `agents/drivers/kingbase-go/kingbase_metadata.go:677` and `:681`, or isolate optional trigger metadata failures.

t8y2 (dismissed):
request changes: the Kingbase catalog-mode handling is incomplete in the new trigger/table DDL path.

1. **P1 — sys catalog mode still calls the PostgreSQL index function.** `agents/drivers/kingbase-go/kingbase_metadata.go:739` unconditionally uses `pg_get_indexdef`. In `sys_catalog/sys` mode, every table DDL request still executes that query even when the table has no secondary indexes, so older Kingbase catalog variants can fail before returning any DDL. Select `sys_get_indexdef` or the explicitly qualified `pg_catalog.pg_get_indexdef` according to the detected mode.

2. **P1 — trigger definition has the same catalog mismatch.** `agents/drivers/kingbase-go/kingbase_metadata.go:775` hard-codes `pg_get_triggerdef` while the surrounding view logic already switches between `sys_get_viewdef` and `pg_get_viewdef`. A sys-catalog connection can still fail after the index path is corrected. Apply the same mode-specific function selection for trigger definitions.

3. **P2 — the fake driver validates the incorrect SQL.** `agents/drivers/kingbase-go/main_test.go:133` directly matches these `pg_get_*` queries and returns fixed rows, so CI cannot detect the missing catalog split. Add tests for both catalog modes and real V8/V9 coverage before relying on the green Agent job.

Validation:
- Exact-head diff and DBeaver PostgreSQL trigger/table-DDL path reviewed.
- PostgreSQL 16 quoted schema/table/function/trigger probe passed.
- CI `agents`, `rust`, `rust-fmt-clippy`, `rust-test`, `changes`, and `notify` passed.
- Kingbase version integration was not completed and is not claimed as passing.

t8y2 (commented):
Maintainer patch applied in 9fb616362.

The Kingbase `sys_catalog` path selected `sys_*` catalog relations but still called PostgreSQL-style deparser functions. On compatibility modes that expose only the `sys_catalog` functions, table DDL export could fail while reconstructing secondary indexes or triggers.

The patch now selects catalog-qualified `sys_catalog.sys_get_indexdef` / `sys_catalog.sys_get_triggerdef` or `pg_catalog.pg_get_indexdef` / `pg_catalog.pg_get_triggerdef` according to the detected metadata mode. The regression test executes both DDL metadata queries in both modes.

Validation:
- `go test ./...`
- `go vet ./...`
- `git diff --check`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 736ba75c9, will be released in the next version.

PR #4414, opened 2026-07-24
修复：优化 Oracle 首次别名字段补全

zipg:
## 改动说明

- Oracle 未显式选择 Schema 时，按真实对象、当前用户私有同义词、PUBLIC 同义词的优先级解析字段来源
- 编辑器首次请求直接使用支持同义词的字段查询，避免先执行一次无结果的 Agent 元数据调用
- 收窄主键与字段注释查询范围，改善 Oracle 11g 数据字典查询性能
- 保持显式 Schema、非编辑器元数据请求以及其他数据库的现有行为

## 验证

- Oracle 11g、18c：私有同义词别名补全均返回 9 个字段，字段类型、注释和主键标记正确
- Oracle 11g：验证真实表、PUBLIC 同义词 `DUAL` 和不存在对象边界
- Oracle 18c：验证真实表、PUBLIC 同义词 `DUAL` 和不存在对象边界
- 最新 main 实测首次请求：Oracle 11g 约 1.24 秒，Oracle 18c 约 1.32 秒；原反馈约 3-4 秒
- `cargo test -p dbx-core --no-default-features oracle_columns --lib`
- `pnpm check`

这是 #4075 中 Oracle 别名字段补全的后续优化。

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -2690,7 +2690,7 @@ function buildLocalSqlCompletionResult(completionContext: ReturnType<typeof getS
       continue;
     }
     const target = completionMetadataTarget(refTable);
-    const localColumns = target && !usesOracleSessionCompletionColumns(target.schema) ? connectionStore.lookupLocalCompletionColumns(props.connectionId, target.database, refTable.name, target.schema, target.catalog) : [];
+    const localColumns = target && !usesOracleSessionCompletionColumns(target.schema) ? connectionStore.lookupLocalCompletionColumns(props.connectionId, target.database, refTable.name, target.schema, target.catalog, refTable) : [];
     if (localColumns.length > 0) {
       columnsByTable.set(cacheKey, localColumns);
     }
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -414,25 +414,27 @@ function usesOracleSessionCompletionColumns(schema?: string | null): boolean {
   });
 }
 
-function completionColumnRequestContext() {
+function completionColumnRequestContext(reference?: Pick<SqlCompletionReferencedTable, "nameQuoted" | "schemaQuoted">) {
   return {
     clientSessionId: props.clientSessionId,
     version: props.completionContextVersion,
+    tableQuoted: reference?.nameQuoted,
+    schemaQuoted: reference?.schemaQuoted,
   };
 }
 
-async function listCompletionColumnsForEditor(connectionId: string, database: string, table: string, schema?: string, catalog = props.catalog) {
+async function listCompletionColumnsForEditor(connectionId: string, database: string, table: string, schema?: string, catalog = props.catalog, reference?: Pick<SqlCompletionReferencedTable, "nameQuoted" | "schemaQuoted">) {
   const requestedVersion = props.completionContextVersion;
   const sessionScoped = usesOracleSessionCompletionColumns(schema);
-  const columns = await connectionStore.listCompletionColumns(connectionId, database, table, schema, completionColumnRequestContext(), catalog);
+  const columns = await connectionStore.listCompletionColumns(connectionId, database, table, schema, completionColumnRequestContext(reference), catalog);
   if (sessionScoped && requestedVersion !== props.completionContextVersion) throw new Error("Stale Oracle completion context");
   return columns;
 }
 
-async function refreshCompletionColumnsForEditor(connectionId: string, database: string, table: string, schema?: string, catalog = props.catalog) {
+async function refreshCompletionColumnsForEditor(connectionId: string, database: string, table: string, schema?: string, catalog = props.catalog, reference?: Pick<SqlCompletionReferencedTable, "nameQuoted" | "schemaQuoted">) {
   const requestedVersion = props.completionContextVersion;
   const sessionScoped = usesOracleSessionCompletionColumns(schema);
-  const columns = await connectionStore.refreshCompletionColumns(connectionId, database, table, schema, completionColumnRequestContext(), catalog);
+  const columns = await connectionStore.refreshCompletionColumns(connectionId, database, table, schema, completionColumnRequestContext(reference), catalog);
   if (sessionScoped && requestedVersion !== props.completionContextVersion) throw new Error("Stale Oracle completion context");
   return columns;
 }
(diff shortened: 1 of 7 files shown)


t8y2 (commented):
The Oracle alias completion change needs real synonym coverage before approval. `crates/dbx-core/src/schema.rs:1462` resolves only one synonym layer; a synonym that points to another synonym can still produce no columns. Please add or run coverage for direct, two-level, public, private, and quoted-case synonyms on the supported Oracle versions.

zipg:
Addressed the synonym coverage feedback on top of the latest `main`.

Changes:
- Replaced the one-hop `ALL_SYNONYMS` lookup with Oracle 10g-compatible `CONNECT BY NOCYCLE` traversal.
- Resolve only chain nodes that end at a table, view, or materialized view.
- Preserve Oracle precedence: direct object, private root synonym, public root synonym, then shortest chain.
- Exclude database-link synonyms at the root and every recursive step.
- Preserve the exact synonym/object name for quoted mixed-case identifiers.
- Added generated-SQL coverage for direct objects, private/public precedence, two-level chains, cycle protection, quoted-case names, and DB-link exclusion.

Local verification:
- `cargo test -p dbx-core --no-default-features oracle_columns --lib` (7 passed)
- `cargo clippy -p dbx-core --no-default-features --lib -- -D warnings`

I do not have live Oracle 10g/legacy instances on this machine, so I have not claimed a live version matrix run. The hierarchy syntax was kept to `CONNECT BY NOCYCLE` / `CONNECT_BY_ROOT`, which is available in Oracle 10g, instead of using recursive subquery factoring.

t8y2 (changes requested):
request changes: I re-reviewed the new head 9aabb05ff. The recursive synonym lookup is now corrected with `CONNECT BY NOCYCLE`, but quoted/unquoted Oracle name handling remains a blocker.

**P1 — quotedness is still lost before schema lookup.** `crates/dbx-core/src/schema.rs:1544` still compares the incoming name as an exact string, while the frontend strips identifier quotes and passes only the text. The lookup therefore cannot distinguish an ordinary unquoted `orders_alias` (which Oracle resolves using uppercase folding) from a quoted `"orders_alias"` (which preserves lowercase). This can return no columns or the wrong object for first-time alias completion. Preserve quotedness through the call chain, or normalize unquoted names according to Oracle identifier rules before querying the catalog. See [Oracle Database Object Names](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Database-Object-Names-and-Qualifiers.html) and [CREATE SYNONYM](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/CREATE-SYNONYM.html).

The previous multi-level synonym blocker is addressed in this head. Please add a regression covering ordinary lowercase, quoted lowercase, mixed-case, and a synonym chain before re-requesting review.

Validation:
- Current head `9aabb05ff` is mergeable and conflict-free.
- Recursive synonym change was statically verified.
- Rust CI was still running during this review; no completed exact-head local Oracle validation is claimed.

zipg:
Addressed the quoted/unquoted Oracle identifier blocker in `fa2b5e40`.

The completion path now preserves `nameQuoted` and `schemaQuoted` from both the semantic parser and the legacy reference parser, carries those flags through `QueryEditor`, and normalizes identifiers immediately before the metadata request in `connectionStore`:

- unquoted `orders_alias` -> `ORDERS_ALIAS`
- quoted `"orders_alias"` -> `orders_alias`
- quoted `"Orders_Alias"` -> `Orders_Alias`

This is explicit quotedness propagation rather than a mixed-case heuristic. Existing callers that do not provide quotedness keep their previous behavior.

Regression coverage now includes:
- semantic alias references for ordinary lowercase, quoted lowercase, and quoted mixed-case Oracle names
- the exact table names sent to `getColumns` for all three cases
- the existing direct/private/public/two-level synonym-chain and cycle-protection Rust coverage

Validation:
- `vitest run apps/desktop/src/lib/__tests__/sql/semantic/completion.spec.ts apps/desktop/src/stores/__tests__/connectionStore.completion.spec.ts` (50 passed)
- `pnpm typecheck`
- targeted `oxfmt --check` and `oxlint`
- `cargo test -p dbx-core --no-default-features oracle_columns --lib` (7 passed)

t8y2 (changes requested):
request changes: the new head `fa2b5e40` fixes quotedness propagation into the metadata request, but the shared column index still collapses distinct Oracle identifiers.

**P1 — explicit-schema quoted and unquoted objects share the same cache key.** `apps/desktop/src/stores/connectionStore.ts:4538` lowercases both the schema and table in `completionColumnsKey`. As a result, unquoted `APP.orders_alias` (resolved as `APP.ORDERS_ALIAS`) and quoted `APP."orders_alias"` both index under the same key. Loading the second object overwrites the first, and `lookupLocalCompletionColumns` can silently return the wrong columns for alias completion.

Oracle documents the required distinction in [Database Object Names and Qualifiers](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Database-Object-Names-and-Qualifiers.html):

> “Nonquoted identifiers are not case sensitive. Oracle interprets them as uppercase. Quoted identifiers are case sensitive.”

Please make the completion index and lookup preserve the normalized exact Oracle identifier (or include quotedness in the key), then add a regression that loads both objects with an explicit schema in the same store lifecycle.

Focused verification on this exact head:
- Added a temporary regression that loads unquoted `orders_alias` as `UPPER_ID`, then quoted `"orders_alias"` as `LOWER_ID` under schema `APP`.
- The unquoted lookup returned `LOWER_ID`; targeted Vitest result: 1 failed, 18 passed.

zipg:
Addressed the shared Oracle completion-column index collision in `70139b44`.

The column index now uses the exact normalized Oracle schema/table identifiers instead of lowercasing them. Local lookups also receive quotedness from the parsed table reference, so unquoted lowercase names are folded to uppercase before index lookup while quoted names retain their exact case. Non-Oracle cache-key behavior is unchanged.

Added a same-store-lifecycle regression that loads both `APP.orders_alias` (normalized to `APP.ORDERS_ALIAS`, column `UPPER_ID`) and `APP."orders_alias"` (column `LOWER_ID`) and verifies both exact and source-form lookups remain isolated.

Validation:
- targeted Vitest: 2 files, 51 tests passed
- `pnpm typecheck`
- targeted `oxfmt --check`
- targeted `oxlint`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in c1270ba, will be released in the next version.

PR #3855, opened 2026-07-18
fix(xugu): avoid privileged metadata DDL lookups

mapan0424:
## Summary

- Read Xugu view, trigger, procedure, function, and package source through access-scoped `ALL_*` catalogs instead of privileged `SYS_*` catalogs.
- Reconstruct Xugu table DDL from columns, primary keys, and indexes rather than calling `DBMS_METADATA.GET_DDL`, which can block indefinitely.

## Validation

- `go test ./...` in `agents/drivers/xugu`
- Manual local XuguDB regression: trigger source returned successfully; `QRTZ_BLOB_TRIGGERS` table DDL returned immediately.

## Scope

Xugu Agent only; no frontend changes or bundled Agent binary.

PR #4422, opened 2026-07-24
fix(xugu): make reconstructed table DDL executable and case-preserving

mapan0424:
## Summary

This PR completes the Xugu catalog-reconstructed table DDL path introduced after #3855. The generated script is intended to be executable while preserving the catalog's original quoted identifier spelling.

- Load foreign keys from the dedicated foreign-key catalog query and emit them after `CREATE TABLE` as `ALTER TABLE ... ADD CONSTRAINT`. This preserves ordinary and self-referencing foreign keys.
- Do not emit a duplicate `CREATE UNIQUE INDEX` when an equivalent primary-key or unique-constraint backing index is already represented by the table definition.
- Preserve unsupported default expressions verbatim. Only normalize exact, proven-equivalent function forms outside quoted literals.
- Replace the quoted-identifier regexp with a lexer that handles doubled double quotes, commas, and parentheses.
- Resolve the stored schema/table spelling before reconstruction: exact catalog lookup is always attempted first; a case-insensitive fallback is allowed only when it identifies exactly one candidate.
- Use strict case-sensitive comparison when deciding whether a unique index is already represented by a constraint.

## Why

`DBMS_METADATA.GET_DDL` can block on XuguDB even when the catalog is readable. Reconstructing from `ALL_*` catalog views avoids that failure mode, but it must not silently lose foreign keys, alter valid default-expression semantics, or change double-quoted identifier casing.

## Review follow-up

This update addresses both rounds of review:

1. The DDL path now uses the foreign-key catalog query rather than the non-FK constraint query, so foreign-key rendering is reachable from real metadata.
2. Default handling is token-aware and literal-safe; quoted `'UUID()'`, `CASE` expressions, zero-date values, and unsupported values are preserved rather than rewritten heuristically.
3. The identifier lexer correctly handles escaped quotes such as `"a""b"`, plus commas and parentheses in quoted identifiers.
4. The catalog resolver gives exact schema/table spelling priority. A folded lookup cannot silently pick one candidate when it is ambiguous.
5. Constraint/index column-set deduplication is strict-case, not `EqualFold`.

## Validation

- `cd agents/drivers/xugu && GONOSUMDB=gitee.com/XuguDB/go-xugu-driver go test ./...`
- `cd agents/drivers/xugu && go vet ./...`
- `gofmt` clean and `git diff --check` clean.
- Local XuguDB `SHOP_DEMO` replay checks for ordinary tables, self-referencing foreign keys, unique constraints/indexes, and quoted mixed-case table/column identifiers.
- Regression tests cover catalog-to-DDL foreign-key wiring, literal-safe defaults, escaped identifiers, exact-name priority, ambiguous folded lookup rejection, strict-case unique-index handling, and exact catalog predicates across the DDL metadata queries.

## Compatibility note

The XuguDB test environment rejects creation of two schema objects whose names differ only by case. The regression suite therefore uses synthetic catalog candidates to exercise the defensive ambiguity and strict-case logic, while the integration replay checks run against actual Xugu metadata.

## Scope

Xugu Agent only (`agents/drivers/xugu`). No frontend, protocol, or bundled-binary changes.

### DIFF
--- a/agents/drivers/xugu/main.go
+++ b/agents/drivers/xugu/main.go
@@ -100,6 +100,24 @@ WHERE UPPER(s.SCHEMA_NAME) = UPPER(?)
   AND UPPER(t.TABLE_NAME) = UPPER(?)
   AND c.CONS_TYPE <> 'F'
 ORDER BY c.CONS_NAME`
+
+// Keep foreign keys separate from the generic constraint query. The desktop
+// presents them in their own group, while table DDL must replay them only after
+// CREATE TABLE so that self-referencing tables can be restored safely.
+const xuguTableForeignKeysSQL = `
+SELECT c.CONS_NAME, c.CONS_TYPE, c.DEFINE,
+       rs.SCHEMA_NAME, rt.TABLE_NAME,
+       c.MATCH_TYPE, c.UPDATE_ACTION, c.DELETE_ACTION,
+       c.DEFERRABLE, c.INITDEFERRED, c.ENABLE, c.VALID
+FROM ALL_CONSTRAINTS c
+JOIN ALL_TABLES t ON t.DB_ID = c.DB_ID AND t.TABLE_ID = c.TABLE_ID
+JOIN ALL_SCHEMAS s ON s.DB_ID = t.DB_ID AND s.SCHEMA_ID = t.SCHEMA_ID
+LEFT JOIN ALL_TABLES rt ON rt.DB_ID = c.DB_ID AND rt.TABLE_ID = c.REF_TABLE_ID
+LEFT JOIN ALL_SCHEMAS rs ON rs.DB_ID = rt.DB_ID AND rs.SCHEMA_ID = rt.SCHEMA_ID
+WHERE UPPER(s.SCHEMA_NAME) = UPPER(?)
+  AND UPPER(t.TABLE_NAME) = UPPER(?)
+  AND c.CONS_TYPE = 'F'
+ORDER BY c.CONS_NAME`
 const xuguTablePartitionsSQL = `
 SELECT p.PARTI_NO, p.PARTI_NAME, p.PARTI_VAL, p.ONLINE,
        t.PARTI_TYPE, t.PARTI_KEY, t.AUTO_PARTI_TYPE, t.AUTO_PARTI_SPAN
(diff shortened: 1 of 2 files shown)


t8y2 (changes requested):
Request changes: the generated DDL still loses foreign keys and can silently change valid metadata expressions.

1. `xuguTableConstraintsSQL` explicitly filters out `CONS_TYPE = 'F'`, so `tableConstraints()` can never return foreign keys. The later `foreignKeys` branch and `renderXuguForeignKeyAlterDDL()` are unreachable in the real catalog path; the new tests only construct an `F` constraint directly. Reconstructed tables therefore silently lose ordinary and self-referencing foreign keys. Please connect the DDL path to the foreign-key catalog query and add an end-to-end catalog-to-DDL regression test.

2. Default normalization uses substring matching for `uuid()` and lowercases the whole expression. It turns a string literal such as `'UUID()'` into `'sys_guid()'` and changes the case of unrelated `CASE` expressions. Replacing zero-date defaults with `1970-01-01` also invents a different value rather than preserving metadata. Please only transform exact, proven-equivalent function tokens outside literals and preserve unsupported defaults verbatim or report them explicitly.

3. `quotedIdentifierRegexp` does not recognize doubled quotes inside a quoted identifier. For example, `"a""b"` is split into two columns, which produces incorrect foreign keys and can break unique-index deduplication. Please use a lexer that understands doubled-quote escaping and add PK/UNIQUE/FK/index tests containing quotes, commas, and parentheses.

t8y2 (commented):
The Xugu DDL changes need a deeper compatibility pass before approval. `agents/drivers/xugu/main.go:2055` uses `UPPER(...) = UPPER(?)` without exact-match priority, and `:3217` uses `EqualFold` when deduplicating indexes. That can select the wrong object or merge distinct quoted-case objects. Please validate ordinary and quoted mixed-case schemas/tables plus indexes with names that differ only by case, and add regression coverage for the supported Xugu versions.

mapan0424:
@t8y2 I addressed the remaining catalog-casing review in `01bef22a`.

- Catalog resolution now tries the exact stored `SCHEMA_NAME` + `TABLE_NAME` pair first. The `UPPER(...)` lookup is a compatibility fallback only; it is accepted only for a single candidate and otherwise returns an ambiguity error.
- All downstream table-DDL catalog queries use the resolved, exact identifier spelling.
- Unique-index/constraint deduplication is now strict-case rather than `EqualFold`.
- Added regression coverage for exact-name priority, ambiguous folded lookup rejection, escaped mixed-case identifiers, strict-case column-set handling, and exact catalog predicates across the DDL metadata queries.

I also re-ran the agent test suite, vet, formatting/diff checks, and local `SHOP_DEMO` DDL replay checks for ordinary and quoted mixed-case identifiers.

The available XuguDB environment rejects creating two schema objects whose names differ only by case. The regression suite therefore supplies representative catalog candidates for the ambiguity/strict-case protection rather than claiming a server fixture that Xugu rejects. Could you please re-review the updated commits?

t8y2 (changes requested):
request changes: the current head still has two correctness risks in the reconstructed Xugu DDL path.

1. **P1 — index expressions are emitted as quoted identifiers.** `agents/drivers/xugu/main.go:3495` treats keys such as `"CODE" DESC` or `LOWER("CODE")` as one identifier, so `quoteIdentifier` can generate invalid SQL such as `"\"CODE\" DESC"`. A replay can fail or lose the index ordering/expression. Parse the expression while preserving `ASC`/`DESC`, and add coverage for ordered and expression index keys.

2. **P2 — explicit UNIQUE constraints can be dropped.** `agents/drivers/xugu/main.go:2932-2941` skips any single-column UNIQUE constraint matching an identity column, but that does not distinguish an identity backing constraint from a user-declared, named UNIQUE constraint. Please use the catalog's backing/generated metadata before suppressing a constraint and add a regression case for an explicit UNIQUE constraint on an identity column.

Validation on the exact head:
- `go vet ./...`
- `gofmt`
- `git diff --check`
- CI `agents` and `changes` passed

t8y2 (commented):
Maintainer patch applied in 55f49c793.

Root cause: the DDL path used the identity sequence's `IS_SYS` flag as the only signal for suppressing a single-column UNIQUE constraint. A user-declared UNIQUE constraint on the same IDENTITY column could therefore be omitted from reconstructed DDL.

The patch reads the constraint-level `IS_SYS` flag from `ALL_CONSTRAINTS` and suppresses the UNIQUE clause only when both the identity sequence and the constraint are system-generated. It also covers the catalog-to-DDL path and keeps the existing foreign-key metadata readers aligned with the expanded result set.

Validation:
- `go test ./... -count=1`
- `go vet ./...`
- `git diff --check`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in b5435e779, will be released in the next version.

PR #4454, opened 2026-07-24
perf: 优化导出大量数据到sql的ui体验

### LINKED ISSUE #4305: [Feature] 结果集全部数据希望有进度条显示进度
### 数据库类型和版本

通用

### 支持信息

DBX 版本: v0.5.65
运行环境: 桌面端
操作系统: Windows 10.0.26100.8737
系统架构: x86_64

### 当前痛点

功能：导出当前结果集全部数据 SQL INSERT 
现象：当导出数据为4万多条的时候，界面没有及时响应而是超过20多秒的时候才弹窗提示保存。
当数据到10多万的时候甚至需要1分多才显示弹窗。体验非常不好。一度误以为是功能bug。

Image

### 期望方案

查询结果后导出当前结果集全部数据 SQL INSERT 功能是希望立即有弹窗提示保存进度，给人非常好的交互体验。

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须做

### 补充信息

_No response_

DengQingNian:
## 变更说明

close #4305

1. SQL 格式导出
- 流式导出新增 sql 格式，按 batch（100 行）构建 INSERT 语句写入 .sql 文件
- row_limit 不走 XLSX 硬上限，直接使用用户设置
- 前端 SQL 导出链路增加 query-result 上下文的后端流路径

2. 查询结果导出统一化
- 移除 queryResultExportRequest 外部回调，useDataGridExport 内联构造请求
- 所有格式（CSV/XLSX/TXT/SQL）统一走 startQueryResultExport 和 background task popover
- 进度展示从旧弹窗迁移到统一的 useExportTracker

3. Web 兼容
- Web 后端 SSE route 只实现了 csv/xlsx 的临时文件流程
- sql / txt 两个格式在 web 下回退到本地 resultToExport + Blob 下载，与改动前行为一致


t8y2 (changes requested):
request changes: the new query-result export path regresses existing execution and failure semantics.

1. **Blocker — the request drops the existing export execution context.** `apps/desktop/src/composables/useDataGridExport.ts:1386-1429` bypasses `queryStore.buildQueryResultExportRequest`, uses the grid SQL directly, hard-codes `pageSize: 1000`, and omits `rowLimit`, `totalRows`, `timeoutSecs`, `useAgentCursor`, `setupSql`, `clientSessionId`, `executionId`, and keyset settings. This can change ordering/SQL, ignore export limits, and break agent-cursor, temporary setup SQL, cancellation, and timeout behavior. Reuse the existing builder or extract a shared builder and preserve all fields.

2. **Blocker — plain SELECT SQL export has no target table.** The query-result request at `apps/desktop/src/composables/useDataGridExport.ts:1427-1428` only sets `exportTableName` from `tableMeta`, which is absent for an ordinary query result. The backend then reaches `crates/dbx-core/src/database_export.rs:900-903` without a table name and fails with `tableName is required when qualifiedTableName is not provided`. Define a valid query-result SQL target-table contract or retain a safe fallback, and cover a plain SELECT without `tableMeta`.

3. **High — startup rejection leaves a permanent Running task.** `apps/desktop/src/composables/useDataGridExport.ts:1431-1443` adds the tracker task before `startQueryResultExport`; if startup rejects before terminal progress, the catch only shows a toast and the task remains Running. The cancel handler is then unregistered, so the background popover can remain stuck and later cancellation can use the wrong path. Mark the task Error/remove it on startup failure and add a rejected-start/cancel regression.

Validation:
- Exact-head diff and call-chain review completed.
- CI `frontend`, `rust`, `rust-fmt-clippy`, `rust-test`, and `changes` passed; skipped jobs do not cover these paths.
- Local focused Vitest and Rust export tests were started but not completed, so they are not claimed as passing.

PR #4455, opened 2026-07-24
fix(xugu): preserve programmable object DDL during execution

mapan0424:
## Summary
- classify Xugu as an Oracle-style procedural SQL dialect so selected procedures, functions, triggers, and package bodies remain a single executable statement
- retain the required terminal `END;` in the Xugu Agent while preserving ordinary SQL trimming behavior
- prevent declaration semicolons and nested package-body routine endings from prematurely ending a statement

## Validation
- `pnpm exec vitest run apps/desktop/src/lib/__tests__/sql/sqlStatementRanges.spec.ts`
- `cargo test -p dbx-core xugu_split_keeps_create_package_body_together`
- manual XuguDB validation on `SHOP_DEMO.SYSDBA`: created and removed a procedure, function, trigger, package, and package body through the Agent

## Scope
This is independent of table DDL reconstruction and does not change #4422.

### DIFF
--- a/agents/drivers/xugu/main.go
+++ b/agents/drivers/xugu/main.go
@@ -3363,7 +3363,61 @@ func errorResponse(id json.RawMessage, err error) response {
 }
 
 func trimStatementSQL(sqlText string) string {
-	return strings.TrimRight(strings.TrimSpace(sqlText), "; \t\r\n")
+	trimmed := strings.TrimSpace(sqlText)
+	if isXuguProgrammableObjectDDL(trimmed) {
+		// Xugu's compiler requires the terminator after END. The desktop
+		// statement splitter already removes only client-side delimiters, while
+		// retaining this one for Oracle-style procedural objects.
+		return trimmed
+	}
+	return strings.TrimRight(trimmed, "; \t\r\n")
+}
+
+func isXuguProgrammableObjectDDL(sqlText string) bool {
+	fields := strings.Fields(strings.ToUpper(stripLeadingSQLComments(sqlText)))
+	if len(fields) < 2 || fields[0] != "CREATE" {
+		return false
+	}
+
+	index := 1
+	if len(fields) > index+1 && fields[index] == "OR" && fields[index+1] == "REPLACE" {
+		index += 2
+	}
+	if len(fields) > index && (fields[index] == "EDITIONABLE" || fields[index] == "NONEDITIONABLE") {
+		index++
+	}
+	if len(fields) <= index {
+		return false
+	}
+
+	switch fields[index] {
+	case "PROCEDURE", "FUNCTION", "TRIGGER", "PACKAGE":
+		return true
+	default:
+		return false
+	}
+}
+
+func stripLeadingSQLComments(sqlText string) string {
+	remaining := strings.TrimLeft(sqlText, " \t\r\n")
+	for {
+		switch {
+		case strings.HasPrefix(remaining, "--"):
+			lineEnd := strings.IndexByte(remaining, '\n')
+			if lineEnd < 0 {
+				return ""
+			}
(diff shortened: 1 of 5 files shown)


t8y2 (changes requested):
request changes: the Xugu programmable-DDL splitter is still incomplete across the frontend, Rust, and Agent layers.

1. **P1 — TYPE BODY loses the terminal semicolon in the Agent.** `agents/drivers/xugu/main.go:3365-3395` only recognizes PROCEDURE/FUNCTION/TRIGGER/PACKAGE when preserving programmable DDL terminators. For `CREATE OR REPLACE TYPE BODY ... END;`, the Agent sends `... END` to `ExecContext`, while the frontend splitter already treats the statement as complete. Add TYPE/TYPE BODY handling consistently and cover execution.

2. **P1 — FORCE splits valid package DDL.** `apps/desktop/src/lib/sql/sqlStatementRanges.ts:1617-1623`, `crates/dbx-core/src/sql.rs:2124-2127`, and `agents/drivers/xugu/main.go:3382-3387` do not accept the valid `FORCE`/`NOFORCE` modifier. A probe with `CREATE OR REPLACE FORCE PACKAGE BODY ...` produced partial fragments instead of one executable statement. Use one dialect-specific optional-prefix rule across all three layers and add FORCE fixtures. Xugu documents FORCE as a valid package creation modifier: [package reference](https://docs.xugudb.com/content/reference/object/package).

3. **P2 — package specifications without a slash can consume the next SQL statement.** `apps/desktop/src/lib/sql/sqlStatementRanges.ts:1666-1710` and `crates/dbx-core/src/sql.rs:2086-2121` require `BEGIN` to complete the programmable block, but a package specification has declarations and `END` without `BEGIN`. `CREATE OR REPLACE PACKAGE ... END ...; SELECT 1;` was returned as one range. Add an `END [name];` completion path for package specifications and test both slash and no-slash forms.

Validation on the exact head:
- `sqlStatementRanges.spec.ts`: 128 passed
- focused Xugu split test and FORCE probe
- `git diff --check`
- CI `frontend`, `rust`, `rust-fmt-clippy`, `rust-test`, and `agents` passed

t8y2 (changes requested):
request changes: the current head still has a P1 splitter regression around `CASE` expressions inside Oracle/Xugu-style programmable objects.

1. **P1 — desktop splitter can consume the following statement.** In `apps/desktop/src/lib/sql/sqlStatementRanges.ts:1716`, every `CASE` token is pushed as a PL/SQL block and only popped by `END CASE`. A normal SQL expression such as `RETURN CASE WHEN 1=1 THEN 1 ELSE 0 END;` ends with plain `END;`, so the statement range remains open through the real function `END;` and can include the following `SELECT 1` as part of the object.

2. **P1 — backend splitter can cut the function body too early.** In `crates/dbx-core/src/sql.rs:2120`, the Rust splitter does not track `CASE` expression nesting and treats an `END;` as a block terminator unless followed by `IF`, `LOOP`, or `CASE`. The same function can be split before the outer object `END;`, causing SQL file execution to send an incomplete function body.

Repro shape for both paths:

```sql
CREATE OR REPLACE FUNCTION f RETURN NUMBER AS
BEGIN
  RETURN CASE WHEN 1=1 THEN 1 ELSE 0 END;
END;
SELECT 1;
```

Please distinguish `CASE ... END` expressions from PL/SQL `CASE ... END CASE` statement blocks in both the desktop and Rust splitters, and add regressions for Xugu/Oracle programmable DDL containing `RETURN CASE ... END;` followed by another statement.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 3836142f7, will be released in the next version.

PR #4480, opened 2026-07-25
docs(i18n): add 12 missing zh-CN keys for grid, tree, processList

dienaso:
Closes the gap between \`en.ts\` (5220 keys) and \`zh-CN.ts\` (5283 keys) where 12 keys present in the English source were missing from the Chinese locale file.

## What changed

Added 12 missing keys to \`apps/desktop/src/i18n/locales/zh-CN.ts\`, grouped by source component:

**\`grid\`** (generic SQL result grid):
- \`cachedResultUnavailable\`: "缓存结果缺失或不兼容。"
- \`reexecuteQuery\`: "重新运行查询"

**\`tree\`** (MongoDB sidebar tree):
- \`gridfs\`: "GridFS"
- \`buckets\`: "Buckets"

**\`processList\`** (MySQL/Postgres session manager):
- \`colPid\`: "PID"
- \`colClient\`: "客户端"
- \`colApp\`: "应用"
- \`colWait\`: "等待"
- \`colQuery\`: "查询"
- \`previewTitle\`: "语句"
- \`copy\`: "复制"
- \`copied\`: "已复制"

All keys already exist in \`en.ts\`; \`zh-CN.ts\` was simply missing them. Terms that are technical identifiers (\`PID\`, \`GridFS\`, \`Buckets\`) are kept in English to match the project convention.

## Why

When DBX runs under Chinese locale, the affected components fall back to the English string (via \`withEnglishFallback\`), which is correct but inconsistent — the locale file should be complete. Without these keys:
- The grid "Run query again" button shows "Run query again" instead of "重新运行查询"
- The process list column headers render in English instead of Chinese
- MongoDB GridFS / Buckets tree items have no Chinese translation

## Testing done

- \`git diff --numstat\`: \`12 0\` (12 insertions, 0 deletions, single file)
- \`pnpm typecheck\` exit 0
- \`pnpm lint\` exit 0 (only pre-existing warnings in unrelated files)
- Verified all 12 keys appear at the correct indentation in their target sections (\`grid\`, \`tree\`, \`processList\`)

## Scope notes

I deliberately did not touch the 75 keys that exist in \`zh-CN.ts\` but are absent from \`en.ts\` (e.g. \`editor.unsavedChangesMessage\`, \`connectionGroup.deleteGroupConfirmMessage\`). These look like legacy Chinese-only strings that predate the English source sync; reversing them is a separate cleanup that should be coordinated with the maintainer to avoid accidentally removing Chinese-only fallbacks.


t8y2 (commented):
Maintainer patch applied in 6b5f7a878.

The cached-result fallback requested `editor.cachedResultUnavailable` and `editor.reexecuteQuery`, while the locale messages are defined under `grid`. This made the new translations unreachable and could render untranslated key paths.

The patch switches both call sites to the existing `grid` namespace and adds a focused regression test covering the source wiring and every supported locale.

Validation:
- `pnpm exec vitest run apps/desktop/src/i18n/__tests__/cachedResultMessages.spec.ts` — 8 tests passed
- `pnpm typecheck`
- `pnpm exec oxfmt --check` on the changed files
- `pnpm exec oxlint --vue-plugin` on the changed files
- `git diff --check`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 43f73ca4c, will be released in the next version.

PR #4491, opened 2026-07-26
feat: add static musl browser builds for linux x64/arm64

HandSonic:
## Summary

Adds fully static musl-linked builds of the browser (`dbx-web`) variant for **linux x64 and arm64** to the release workflow. The resulting tarballs have **no glibc dependency** and run on any Linux distribution — verified down to Ubuntu 14.04 and on KylinOS aarch64.

### Source (minimal, default behavior unchanged)
- Make `font-kit` optional behind a new `system-fonts` feature in `dbx-core` (font-kit needs fontconfig/freetype C libraries, which can't link in a static musl build). The feature is in the default set for `dbx-core`, `dbx-web`, and `src-tauri`, so all existing builds keep the exact same behavior — only `--no-default-features` builds exclude it, and `list_system_fonts()` gracefully returns an empty list there.

### CI
- New `static-browser` job in `release.yml` (matrix: x64 on `ubuntu-22.04`, arm64 on `ubuntu-22.04-arm`):
  - builds `dbx-web` with `cargo-zigbuild` for `{x86_64,aarch64}-unknown-linux-musl` with `+crt-static`
  - arm64 links with `max-page-size=64K` so the binary works on 4K/16K/64K page-size kernels
  - asserts the ELF has no interpreter and no `DT_NEEDED` entries
  - runs the packaged tarball inside the oldest pullable Ubuntu container (14.04/16.04) and checks the HTTP endpoints
  - uploads `DBX_<version>_<arch>-browser-static.tar.gz` to the release; `publish` waits for it
- `scripts/package-web-static.sh` — packages the static binary + built frontend + launcher
- `scripts/verify-web-static.sh` — in-container verification (static ELF checks + HTTP smoke test)

## Validation

- CI run with both matrix jobs green (including ubuntu:14.04 in-container verification): https://github.com/HandSonic/dbx/actions/runs/30194788232
- The arm64 static build was additionally deployed and manually verified on KylinOS V10 aarch64 (glibc-independent, 4K page kernel): server starts, UI and `/api/auth/check` respond normally.

## Notes

- The static variant is built with `--features duckdb-bundled,mq-admin`: `sqlite-sqlcipher` (vendored OpenSSL) and `system-fonts` are excluded from the static build; SQLCipher-encrypted databases report a clear error message there (existing `#[cfg]` fallback).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -424,9 +424,17 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           VERSION="${GITHUB_REF_NAME#v}"
+          PACKAGE_VERSION="$(cargo metadata --no-deps --format-version 1 | jq -r '.packages[] | select(.name == "dbx-web") | .version')"
+          if [ "$PACKAGE_VERSION" != "$VERSION" ]; then
+            echo "Release tag version $VERSION does not match dbx-web package version $PACKAGE_VERSION" >&2
+            exit 1
+          fi
+          SOURCE="dist-web-static/dbx-linux-${{ matrix.arch }}-browser-static.tar.gz"
           ASSET="DBX_${VERSION}_${{ matrix.arch }}-browser-static.tar.gz"
-          cp "dist-web-static/dbx-linux-${{ matrix.arch }}-browser-static.tar.gz" "$ASSET"
-          gh release upload "${GITHUB_REF_NAME}" "$ASSET" \
+          CHECKSUM="${ASSET}.sha256"
+          cp "$SOURCE" "$ASSET"
+          sha256sum "$ASSET" | tee "$CHECKSUM"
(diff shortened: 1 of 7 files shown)


t8y2 (commented):
Maintainer patch applied in e433fcc0.

The static browser workflow generated a checksum during packaging but did not publish it with the release asset, and it did not verify that the release tag matched the `dbx-web` package version. The patch now publishes a tag-named `.sha256` asset and fails before upload when the tag/package versions differ.

Validation:
- `cargo check -p dbx-web --no-default-features --features 'duckdb-bundled,mq-admin'`
- `bash -n scripts/package-web-static.sh scripts/verify-web-static.sh`
- Release workflow YAML parse
- `git diff --check`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 2bfbe67, will be released in the next version.

PR #4559, opened 2026-07-27
fix(xugu): return result sets for SHOW statements

mapan0424:
## Summary
- Classify Xugu `SHOW` commands as row-producing SQL and route them through `QueryContext`.
- Keep `SELECT` and `WITH` on the same query path, but match complete leading SQL keywords rather than raw prefixes.
- Add regression coverage for standard and paginated execution of `SHOW DB_INFO`.

## Problem
XuguDB returns a result set for statements such as `SHOW DB_INFO`. The Xugu Agent previously recognized only `SELECT` and `WITH` as queries, so it sent `SHOW` through `ExecContext`. DBX therefore reported success but had no columns or rows to render.

## Implementation
The change is intentionally limited to `agents/drivers/xugu`:
- `isQuerySQL` now recognizes the `SHOW` keyword after whitespace or leading SQL comments.
- The behavioral regression driver returns a `DB_INFO` row only from `QueryContext` and fails if `ExecContext` is used.
- The test exercises both `execute_query` and the paginated `execute_query_page` path, asserts all five `SHOW DB_INFO` columns and the returned row, and verifies that no execution-path call occurred.

No frontend or Rust changes are required: the desktop already renders the `columns` and `rows` returned by the Agent; the defect was solely the native Xugu Agent's dispatch choice.

## Validation
- [x] `GONOSUMDB=gitee.com/XuguDB/go-xugu-driver go test ./...` in `agents/drivers/xugu`
- [x] `GONOSUMDB=gitee.com/XuguDB/go-xugu-driver CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -trimpath -ldflags="-s -w" -o /tmp/dbx-agent-xugu-linux-x64 .`
- [x] `go vet .`
- [x] `git diff --check`
- [x] Manual Agent RPC validation against a local XuguDB instance with DBX's pinned `go-xugu-driver v1.0.12`:
  - `SHOW DB_INFO` returned `DB_NAME`, `DB_ID`, `DB_OWNER`, `DB_CHARSET`, and `DB_TIMEZ` plus one row.
  - `SHOW CHARSETS` returned a paginated result set.
  - `SHOW CURRENT_SCHEMA` returned its single row.

## Compatibility evidence
A separate 73-statement comparison of xgconsole and `go-xugu-driver v1.0.13` found 66 commands succeeding in both clients. This confirms that ordinary Xugu `SHOW` commands provide real driver result sets. MySQL-style `SHOW DATABASES`, `SHOW TABLES`, and `SHOW VARIABLES` remain unsupported by the Xugu driver and are intentionally not emulated by this PR.

## Scope
- [x] Bug fix
- [ ] Frontend change
- [ ] Documentation change
- [ ] CI / build change

### DIFF
--- a/agents/drivers/xugu/main.go
+++ b/agents/drivers/xugu/main.go
@@ -16,6 +16,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode"
+	"unicode/utf8"
 
 	_ "gitee.com/XuguDB/go-xugu-driver"
 )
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch `790f9100` fixes the SQL keyword boundary regression while preserving the intended Xugu `SHOW` result-set path. `SELECT(1)`, adjacent hints/comments, and `SHOW/*...*/` now use `QueryContext`, while identifier continuations such as `SELECTIVE` and `SHOWCASE` remain excluded.

Validation passed: focused and full Xugu Go tests, `go vet`, `gofmt`, Rust protocol tests, `cargo fmt`, and `git diff --check`.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 7e024b8, will be released in the next version.

PR #4570, opened 2026-07-27
fix(ai): harden write-SQL confirmation flow with precise SQL binding

Abeautifulsnow:
### Problem

When the AI agent proposed a write/DDL operation (e.g. "Should I create this table?"), the
user clicking ✅ or typing "yes" granted the agent **blanket write permission for the
entire run** — the agent could then execute _any_ write/DDL statement, not just the one the
user confirmed. Five gaps contributed:

1. **dbx-mcp didn't compile** — `AgentSqlPermissions` grew a `confirmed_write_sql` field, but
   `mcp_permissions()` in the MCP server wasn't updated (E0063).
2. **No SQL code block → silent blanket permission** — when `extractFirstSqlCodeBlock()`
   returned `undefined`, the frontend still set `allowWriteSqlForNextRun = true` and the
   backend's `sql_matches_confirmed_write` returned `true` for `None`.
3. **CLI providers (Codex / Claude Code) entirely bypassed SQL binding** —
   `CliAgentRunOptions` and `dbx_mcp_scope_env` never carried `confirmed_write_sql`.
4. **Normalization broke string-literal semantics** — `to_lowercase()` on the entire SQL
   meant `VALUES ('Alice')` matched `VALUES ('alice')`, silently changing data values.
5. **WebBackend dropped permissions** — `execute_agent_tool` named the parameter
   `_permissions` and forwarded to `/api/query/execute` without the binding check.

### Solution

Each layer now participates in the binding:

| Layer | Change |
|-------|--------|
| **Frontend** | `allowWriteSqlForNextRun` only set when `extractFirstSqlCodeBlock` returns a non-`undefined` value. `looksLikeWriteSqlProposal` gains a `lastLineMentionsWriteSql` safety gate (proposals about reads with write keywords elsewhere are now rejected). |
| **Tauri commands / Web routes** | `allow_writes` / `allow_dangerous` gated on `write_sql_confirmed`: requires non-empty `confirmed_write_sql`. |
| **agent_tools** | `normalize_sql_for_confirmation` rewritten as a per-character parser that preserves single-quoted string literals verbatim while normalizing keywords/whitespace in non-literal segments. `truncate_sql_for_error` fixed to use `.chars().take()` instead of byte-index slicing. |
| **CLI agents** | `CliAgentRunOptions` + `agent_loop.rs` + `dbx_mcp_scope_env()` + MCP server all now carry `confirmed_write_sql` end-to-end. |
| **WebBackend** | `execute_agent_tool` replicates the `confirmed_write_sql` binding check before forwarding to `/api/query/execute`. |

### Tests

- **Rust** (`agent_tools::tests`): 13 new tests — 7 for `normalize_sql_for_confirmation`
  (keyword-case, reformatting, string-literal case/whitespace/escaped-quote preservation,
  empty literal), 6 for `sql_matches_confirmed_write` (match, mismatch, same-table-different-
  statement, case-in-data, whitespace-in-data).
- **Frontend** (`aiWriteSqlConfirmation.spec.ts`): 11 new tests — 5 for
  `extractFirstSqlCodeBlock`, 6 for the new `looksLikeWriteSqlProposal` safety rules
  (rejects proposals where last line asks about reads despite write keywords elsewhere).
- **All passing**: 53 frontend tests + 31 Rust tests. `cargo check -p dbx-mcp` and
  `cargo check -p dbx-web` both clean.

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,4 +64,6 @@ DBX_*_x64-portable.zip
 .agents/skills
 .pnpm-store/*
 .dbx-web
+# Local devtest connections (contains credentials; do not commit)
+dbx-devtest-connections.json
 .npmrc
--- a/apps/desktop/src/components/editor/AiAssistant.vue
+++ b/apps/desktop/src/components/editor/AiAssistant.vue
@@ -79,7 +79,7 @@ import { parseExplainResult, parseOracleExplainText, type ParsedExplainPlan } fr
 import { copyToClipboard } from "@/lib/common/clipboard";
 import { AI_TABLE_MENTION_CANDIDATE_LIMIT, AI_TABLE_MENTION_SCHEMA_LIMIT, filterAiTableMentionCandidates, formatAiTableMention, parseAiTableMentions, type AiTableMention } from "@/lib/ai/aiTableMentions";
 import { isAiPromptImeCompositionEvent, shouldSubmitAiPromptOnKeydown } from "@/lib/ai/aiPromptKeyboard";
-import { looksLikeActionProposal, containsChinese } from "@/lib/ai/aiProposalDetect";
+import { looksLikeActionProposal, containsChinese, looksLikeWriteSqlProposal, shouldGrantWriteSqlOnShortAffirmative } from "@/lib/ai/aiProposalDetect";
 import { visibleToActualIndex } from "@/lib/ai/aiMessageEdit";
 import { shouldShowReasoningCharCount, reasoningCharCountClass } from "@/lib/ai/aiReasoningPresentation";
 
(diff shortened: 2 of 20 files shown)


t8y2 (changes requested):
Request changes: the exact-SQL binding is improved, but three confirmation boundaries are still unsafe.

1. CLI providers can still execute unconfirmed writes when the user has an explicitly configured writable MCP policy. `dbx_mcp_scope_env()` sends `DBX_MCP_ALLOW_WRITES=0` for an unconfirmed run, but `effective_mcp_policy_with_legacy_allow_writes()` intentionally ignores that value when `McpGlobalPolicyState.configured` is true. `mcp_permissions()` then restores `allow_writes=true`; with no `DBX_MCP_CONFIRMED_WRITE_SQL`, `sql_matches_confirmed_write(..., None)` returns true. The existing exact-head tests reproduce both links. Run-scoped CLI restrictions must override the persistent MCP policy, and an AI-scoped write permission must fail closed without a non-empty confirmed SQL.

2. Both the button and typed-affirmative paths bind `extractFirstSqlCodeBlock()`, regardless of which block the proposal asks the user to confirm. A message can show an old “do not execute” DELETE first, a recommended DELETE second, and ask “Should I execute the recommended SQL?”; the UI grants the first DELETE, and the backend accepts that first statement as the confirmed write. The exact-head probe reproduces this. Bind the confirmation to the specific proposed block, or reject ambiguous multi-block proposals.

3. The grant is bound only to SQL text, not to the connection/database snapshot that was visible when the user confirmed it. `sendProposalReply()` sets mutable permission state before `send()` awaits prompt/context loading, and the later request is built from mutable `props.tab` / `props.connection`. Changing the active database during that window can execute the same confirmed SQL against a different target. Carry the confirmed connection/database through the request and verify it at the backend boundary; also clear the pending grant on every early-return/failure path.

Please add regression tests for configured writable MCP policy + no confirmation, ambiguous multi-block proposals, and database/connection changes between confirmation and execution.

t8y2 (approved):
Reviewed current head `48938f57a8f9ccc2f4b764b4606576dd82141e`.

The three previously requested confirmation boundaries remain addressed: unconfirmed CLI runs force MCP read-only, ambiguous multi-block proposals fail closed, and the confirmed connection/database snapshot is verified before execution. The latest commit only adjusts a test comment and does not change the implementation path.

Targeted frontend confirmation tests passed previously; current CI is still pending for the new head. Approve the code review; clear the required checks before merging.

t8y2 (approved):
Reviewed current head `5f8df3947`. The contributor removed the unused test import that failed clippy. Local `cargo fmt --all -- --check`, targeted clippy with warnings denied, and all five confirmation-target tests pass.

t8y2:
Thanks for the contribution! Merged in cb9d0f7, will be released in the next version.

PR #4574, opened 2026-07-27
fix(i18n): route hardcoded Chinese UI strings through i18n

heyrmi:
Several UI strings were hardcoded in Chinese and never went through vue-i18n, so they stayed Chinese for every locale. The locale files themselves were already in full key parity — the gap was untranslated literals in components, composables and backend error messages.

- connection dialog: driver install progress/failure dialog, connection failure dialog, optional-field suffixes, Turso host/token hints
- data generation: labels, placeholders and option lists across the 40 generator param panels, plus the generate dialog itself
- layer preview, editor settings, data grid insert error
- history AI analysis prompt (English metadata with Chinese instructions)
- backend: JRE/driver install, DuckDB drain, export limit, Kafka, fs_open and web auth error messages

Adds the corresponding keys to all seven locale files. Chinese/Japanese sample-data payloads, native language names in language pickers, Chinese-vendor product names and error-text matching predicates are left untouched.

## 变更类型

- [ ] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过


t8y2 (changes requested):
request changes: the component-side i18n work looks good, but the backend part is not localized yet.

1. The changed backend errors are only being replaced from hardcoded Chinese with hardcoded English. For example, `crates/dbx-core/src/query_result_export.rs:37`, `crates/dbx-core/src/query.rs:49`, `crates/dbx-core/src/agent_service.rs:64`, and `crates/dbx-web/src/auth.rs:82` still return English text directly. Those strings are then rendered without translation, e.g. `apps/desktop/src/components/export/ExportProgressDialog.vue:109`, `apps/desktop/src/components/config/DriverStoreDialog.vue:607`, and `apps/desktop/src/components/auth/LoginPage.vue:43`. `apps/desktop/src/i18n/backend-errors.ts` has no matching entries for them either.

   This regresses Chinese users from Chinese to English and still leaves Spanish, Italian, Japanese, and Portuguese users with English errors. Please return stable error codes plus parameters and translate them at the UI boundary. A smaller acceptable fix would be to add mappings for every changed standard error and make sure every consumer goes through the same translation path. Please add an interpolation test for `zh-CN` and at least one other non-English locale.

2. Three new Simplified/Traditional Chinese labels have the wrong meaning: `dataGenerate.rawPattern` is translated as “raw data mode”, `dataGenerate.uuidNoHyphens` as “unformatted”, and `dataGenerate.nameFormat` as “format type” (`zh-CN.ts:3547/3564/3581`, `zh-TW.ts:3024/3041/3058`). They should describe raw regex mode, UUID without hyphens, and name format respectively.

heyrmi:
@t8y2 Both valid, thanks. Took the smaller fix — mappings plus one translation path — rather than error codes, to keep the diff reviewable. Happy to do codes + params as a follow-up if you want that end state.

**1. Backend errors** (`dfd29ff`)

- `backend-errors.ts`: patterns for every message this PR changed — export limits, DuckDB drain, JRE dir remove (Windows + POSIX), JRE in use, offline package registry, driver update blocked, Kafka unload, missing file, login rate limit.
- `paramNames` now accepts `string | string[]`. Messages with two placeholders (`{path}` + `{error}`) need more than `match[1]`; a bare string still means one capture.
- Routed the consumers that bypassed `translateBackendError`: `ExportProgressDialog`, `ExportProgressPopover`, `DriverStoreDialog` (12 raw `{ error: e }` interpolations), `LoginPage`, `ProducerConsumerPanel`, reveal-file toast, `ConnectionDialog` install error.
- `queryStore.toErrorResult` translates once — single funnel for every query execution failure, so the DuckDB message and anything mapped later is covered.
- Side finding: `LoginPage` rendered the raw response body, so the 429 case showed the literal `{"error":...}` JSON. Now unwrapped, then translated.
- `translateBackendError` took `ComposerTranslation`, which rejects `i18n.global.t`. Replaced with an exported structural type, same approach as `connectionErrorHints.ts`.

**2. Labels** — fixed in zh-CN and zh-TW: `原始正则模式`, `不含连字符`, `姓名格式`. I had carried the old literals over verbatim; promoting them to shared keys made the imprecision systematic, so that was my mistake.

**Tests** — `apps/desktop/src/i18n/__tests__/backendErrors.spec.ts`, 158 cases: each message across all seven locales, asserting the resolved key, parameter interpolation, no leaked `{placeholder}`, and that non-English output is not the raw English. Mutation-checked — breaking one regex fails 6. I also pinned the wording to the Rust sources: text matching degrades silently to raw English if a message is reworded, this fails loudly instead.

**Not verified:** no Rust toolchain on this machine, so `cargo check` / `cargo test` and `make cargo-check-fast` have not run — the two test assertions updated in `2bf5617` are uncompiled. `make check` is green apart from a pre-existing `objectBrowserRows` digit-grouping failure that also fails on clean `main` here (locale-dependent).

t8y2 (changes requested):
这轮已经修正 `rawPattern`、`uuidNoHyphens`、`nameFormat` 的简繁中文错义，但后端英文错误的 i18n 消费链仍未闭环：

1. 查询结果 CSV/TXT/XLSX 导出在终态 `Error` 时会用原始 `errorMessage` reject，外层 catch 又直接把 `e.message` 放进 toast。XLSX 行数限制等标准错误会先显示翻译后的进度信息，随后再弹出原始英文。
2. `queryStore.toErrorResult` 不是所有查询错误的统一入口。侧边栏 JSON/SQL 导出直接调用 `api.executeQuery` 并原样展示异常，DuckDB draining 等本 PR 改成英文的错误仍会绕过翻译。
3. `file does not exist` 仍有多个直接展示入口，包括导出进度框、SQL 文件面板和定时备份设置；后端改成英文后这些路径会向用户泄漏原始文本。

请让这些入口统一经过 `translateBackendError`（或等价的稳定错误码映射），并补上述导出、侧边栏查询和文件定位路径的回归测试。

t8y2 (commented):
Maintainer merge patch `c1cef41b` resolves the current-main import conflict and closes the remaining backend-error i18n gaps.

Root cause: several export, sidebar-query, file-reveal, and scheduled-backup UI boundaries still rendered raw backend error strings even though the shared translator already contained stable mappings.

Patch:
- normalizes string, Error, and `{ message }` inputs in `translateBackendError`;
- applies it at DataGrid export toasts, sidebar direct-query/export catches, and file reveal failures;
- translates scheduled backup completion toasts and persistent inline history errors;
- preserves raw progress state where components perform reactive translation.

Checks run:
- focused Vitest: 190 tests passed, plus the inline scheduled-backup regression
- pnpm typecheck / vue-tsc
- oxfmt --check
- oxlint with Vue plugin
- git diff --check and conflict-marker scan

t8y2 (approved):
Reviewed the maintainer merge patch at c1cef41b; backend errors now pass through the shared translator at all identified UI boundaries.

t8y2:
Thanks for the contribution! Merged in 2f44d360, will be released in the next version.

PR #4651, opened 2026-07-28
fix(tree): label Xugu table child groups

mapan0424:
## Summary

Fix the sidebar presentation for Xugu table child metadata groups introduced by the table-child metadata feature.

## Problem

The `group-constraints`, `group-table-partitions`, and `group-table-subpartitions` nodes were already created and routed to their metadata loaders, but `TreeItem.vue` did not classify them as group nodes. The sidebar therefore rendered raw i18n keys such as `tree.constraints` and used the default database icon.

## Changes

- Register the three Xugu child node types as tree groups so their existing localized labels are resolved.
- Render the constraint group with the key icon and partition groups with the folder icon.
- Move the group-type classification into a small shared helper and add regression coverage for the new groups and their leaf counterparts.

## Scope

- Frontend presentation only.
- No Xugu agent, metadata SQL, protocol, or DDL changes.
- The affected node types are emitted only for Xugu table child metadata, so other database trees retain their existing behavior.

## Validation

- Added `treeNodeGroup.spec.ts` regression coverage.
- `git diff --check` passed.
- Local `pnpm` test/typecheck/build commands were not run because this isolated worktree has no installed frontend dependencies; CI will run them in the repository dependency environment.

## Change Type

- [x] Bug fix

## Frontend

- [x] This PR changes frontend tree presentation.
- [ ] Screenshot/recording attached. Not available from this isolated worktree because the frontend dependencies are not installed.

### DIFF
--- a/apps/desktop/src/components/sidebar/TreeItem.vue
+++ b/apps/desktop/src/components/sidebar/TreeItem.vue
@@ -64,6 +64,7 @@ import { focusSidebarRenameInput } from "@/lib/sidebar/sidebarRenameFocus";
 import { useDragSort } from "@/composables/useDragSort";
 import { sidebarTreeRuntimeKey } from "@/lib/sidebar/sidebarTreeRuntime";
 import { treeNodePinKey } from "@/lib/app/pinnedItems";
+import { isTreeGroupNodeType } from "@/lib/sidebar/treeNodeGroup";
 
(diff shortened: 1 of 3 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 3e26b77, will be released in the next version.

PR #4668, opened 2026-07-28
fix(xugu): expose user-managed sequence DDL

mapan0424:
## Summary
- Hide XuguDB internal sequences created for IDENTITY columns from the schema tree.
- Reconstruct read-only, executable CREATE SEQUENCE DDL for user-managed sequences.

## Details
- Restrict discovery and source lookup to ALL_SEQUENCES.IS_SYS = FALSE.
- Preserve catalog schema and sequence spelling, and quote identifiers.
- Render INCREMENT BY, START WITH, MINVALUE/NOMINVALUE, MAXVALUE/NOMAXVALUE, CACHE/NOCACHE, CYCLE/NO CYCLE, and COMMENT.
- Avoid DBMS_METADATA, which can be unavailable or permission-restricted.

## Scope
- Xugu Go agent only: agents/drivers/xugu.
- No frontend, core, PostgreSQL, or other-driver changes.

## Validation
- [x] cd agents/drivers/xugu && go test .
- [x] cd agents/drivers/xugu && go build .
- [x] Live XuguDB (SHOP_DEMO) replay: create a temporary sequence, fetch generated DDL, drop and replay it, verify start, bounds, increment, cache, cycle, and comment, then remove the temporary object.

### DIFF
--- a/agents/drivers/xugu/main.go
+++ b/agents/drivers/xugu/main.go
@@ -40,6 +40,11 @@ const xuguCatalogTableNameSelectSQL = `
 SELECT s.SCHEMA_NAME, t.TABLE_NAME
 FROM ALL_TABLES t
 JOIN ALL_SCHEMAS s ON s.DB_ID = t.DB_ID AND s.SCHEMA_ID = t.SCHEMA_ID`
+const xuguCatalogSequenceNameSelectSQL = `
+SELECT s.SCHEMA_NAME, q.SEQ_NAME
+FROM ALL_SEQUENCES q
+JOIN ALL_SCHEMAS s ON s.DB_ID = q.DB_ID AND s.SCHEMA_ID = q.SCHEMA_ID
+WHERE q.IS_SYS = FALSE`
 const xuguPrimaryKeyColumnsSQL = `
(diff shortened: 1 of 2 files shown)


t8y2 (changes requested):
Request changes: the system-sequence filter and escaping look good, but two correctness issues remain.

1. The renderer emits `NO CYCLE`, while Xugu documents the sequence option as `CYCLE | NOCYCLE`. Please render `NOCYCLE` and add replay-oriented coverage.

   Xugu sequence documentation:
   https://docs.xugudb.com/content/reference/object/sequence

   > “`[CYCLE | NOCYCLE]`”

2. The source lookup case-folds schema/name with `UPPER()` and accepts the first row. Quoted identifiers that differ only by case can therefore export the wrong sequence. Please resolve the exact catalog spelling first and reject ambiguous case-insensitive matches, consistent with the existing table DDL resolver. Add coverage for quoted case-sensitive sequence names.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 99cfa1b, will be released in the next version.

PR #4682, opened 2026-07-28
fix: parse GaussDB boolean columns encoded as 0x74/0x66 (#4599)

gggaiitx:
### 问题根因
GaussDB 在二进制协议中使用 ASCII 字节 t(0x74)/f(0x66) 编码布尔值，而非标准 PostgreSQL 的 0x00/0x01。tokio_postgres 的 FromSql<bool> 只接受 [0]/[1]，导致布尔列解码失败显示为 NULL。

### 修复方案
修改 [crates/dbx-core/src/db/postgres.rs](file:///D:/DBX/dbx-source/crates/dbx-core/src/db/postgres.rs)：

新增 pg_bool_value_to_json() — PgColType::Bool 的新入口，先尝试标准 bool 解码，失败后读取原始字节通过 decode_bool_bytes 解码 GaussDB 编码
新增 decode_bool_bytes() — 识别标准编码(0x00/0x01)和 GaussDB 编码(0x74/'t' => true, 0x66/'f' => false)
增强 pg_row_try_bool() — 元数据查询同样增加原始字节回退，保持一致性
### 测试验证
新增 decode_bool_bytes_handles_standard_and_gaussdb_encodings 单元测试覆盖所有编码场景
全部 114 个 postgres 测试通过（10 个需要 live 数据库的已忽略）

### DIFF
--- a/crates/dbx-core/src/db/postgres.rs
+++ b/crates/dbx-core/src/db/postgres.rs
@@ -523,7 +523,7 @@ pub(crate) fn pg_value_to_json_classified(row: &Row, idx: usize, col_type: PgCol
             }
             serde_json::Value::Null
         }
-        PgColType::Bool => row.try_get::<_, bool>(idx).map(serde_json::Value::Bool).unwrap_or(serde_json::Value::Null),
+        PgColType::Bool => pg_bool_value_to_json(row, idx),
         PgColType::Interval => row
             .try_get::<_, PgInterval>(idx)
             .map(|interval| serde_json::Value::String(format_pg_interval(interval)))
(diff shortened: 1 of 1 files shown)


t8y2 (commented):
Maintainer patch `4667007` fixes the boolean decoding order.

Root cause: the pinned GaussDB PostgreSQL driver accepts any non-zero byte as `true`, so ASCII `f` (`0x66`) was decoded as `true` before the raw-byte fallback could run.

Fix: route both boolean conversion paths through the same helper and prefer recognized raw PostgreSQL/GaussDB encodings before the driver's standard boolean result. Added a regression test proving raw `f` overrides the driver's `true` result.

Checks run:
- `cargo fmt --all --check`
- `cargo test -p dbx-core --no-default-features raw_gaussdb_boolean_takes_precedence_over_standard_decoder`
- `cargo test -p dbx-core --no-default-features decode_bool_bytes_handles_standard_and_gaussdb_encodings`
- `git diff --check`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 5f2fd52, will be released in the next version.

PR #4685, opened 2026-07-28
feat(sql-file): preview multiple SQL files via side file list

Abeautifulsnow:
## Summary

When executing multiple SQL files, the preview area now uses a **side file list + single-file preview** layout instead of rendering every file's full preview at once.

## Problem

- The dialog originally previewed only `previews[0]`; an earlier change rendered *all* selected files' previews, which became heavy and cluttered as the number of files
grew.
- A tab-strip approach truncated long file names (e.g. `05_e...`), so files sharing a prefix were indistinguishable and the strip could not scale to dozens of files.

## Changes

- Replaced the all-files render / tab strip with a two-pane layout:
- **Left**: a scrollable file list showing each file name with a count header; clicking a file selects it for preview.
- **Right**: only the **active** file's preview is mounted (line numbers, SQL syntax highlight, line/byte summary), keeping the DOM light regardless of how many files
are selected.
- The list height is capped to match the preview pane (`max-h-[min(46vh,420px)]`) so the two panes stay equal-height and the list scrolls internally when there are many
files.
- Added a hover **tooltip** (side=right, wraps long names) that shows the full file name on truncated entries.
- Single-file selection hides the list and keeps the original full-width preview experience.

## Verification

- `pnpm typecheck` (vue-tsc): passes.
- `pnpm lint` (oxlint): 0 errors.
- Manual UI check (recommended): select many `.sql` files → the list scrolls and stays equal-height with the preview, hovering a truncated name shows the full name, and
a single file shows no list.

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

356fd1c6-5d06-44c4-a37f-acca694dd734

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

### DIFF
--- a/apps/desktop/src/components/sql-file/SqlFileExecutionDialog.vue
+++ b/apps/desktop/src/components/sql-file/SqlFileExecutionDialog.vue
@@ -54,6 +54,41 @@ watch(previews, (list) => {
 const activePreview = computed<SqlFilePreview | null>(() => {
   return previews.value.find((item) => item.filePath === activePreviewPath.value) ?? previews.value[0] ?? null;
 });
+
+// When multiple files share the same fileName (e.g. migration/create.sql and
+// seed/create.sql), disambiguate by prepending parent directory segments until
+// each file gets a unique label.  Single-name files stay as just fileName.
+const displayFileNames = computed(() => {
+  const items = previews.value;
+  const byFileName = new Map<string, SqlFilePreview[]>();
+  for (const item of items) {
+    const list = byFileName.get(item.fileName) ?? [];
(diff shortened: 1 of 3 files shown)


t8y2 (approved):
Looks good. The preview-only file switching preserves the complete execution list and keeps web temporary paths out of the UI.

t8y2:
Thanks for the contribution! Merged in 8872192, will be released in the next version.

PR #4721, opened 2026-07-29
revert(data-grid): remove configurable default sorting

t8y2:
Reverts #2442 because the configurable default sorting behavior needs to be withdrawn. This restores the previous default table-opening and column-menu behavior.


t8y2 (commented):
Revert patch verified on exact head `29b4c6a1`.

The revert removes exactly the 24 files introduced by #2442 and restores the previous default table-opening and column-menu behavior. No later commits existed on `main` when the revert was generated.

GitHub required checks are all green, including the frontend check. The isolated local `pnpm check` had 20 failures in two localStorage-dependent suites, while format, lint, and typecheck passed; the targeted baseline comparison passed 33/33 tests, and the GitHub frontend check is the merge gate.

t8y2:
Revert merged in ed619284. This restores the pre-#2442 default sorting behavior.

emrecpp:
I’d really like to see this feature brought back. Having tables automatically sorted in descending order by default on the database side is very practical, especially when working with records where the most recent entries are usually the most relevant.

It makes browsing data much faster and reduces the need to manually apply sorting every time a table is opened or refreshed.

Please consider reintroducing this option, ideally as a configurable setting so users can choose their preferred default sorting behavior.

PR #4744, opened 2026-07-29
fix(ai): honour Retry-After headers, preserve HTTP status in stream errors, and add retry regression tests

Abeautifulsnow:
## Summary

Streaming LLM API requests can fail transiently (429, 504, connection drops, empty responses). This branch adds automatic retry with `Retry-After` header support to both `with_retry` (completion) and `with_stream_retry` (streaming) paths, along with 20+ unit and integration tests covering the full `HTTP response → classify → retry` pipeline.

## Motivation

- **No retry tests existed.** `with_stream_retry` is called by 8 streaming paths but had zero test coverage — the most critical behaviour (no duplicate output after content emitted, maxRetries=0, cancellation during back-off) was unverified.
- **`Retry-After` was ignored.** The back-off was purely client-side (500ms × 2ⁿ). If a provider returned `Retry-After: 5`, DBX would exhaust its default 2 retries before the server-allowed window.
- **`stream_error` lost HTTP status codes.** On non-JSON or empty-body error responses (e.g. a proxy returning a bare `429`), `response.json()` consumed the body, the parse error discarded the status code, and `classify_error` mapped the result to `"unknown"` — not retryable. A real 429 was silently dropped.

### 1. Retry-After support (`ai.rs:2090–2125`)

New helpers:
- `retry_after_secs(headers: &HeaderMap) -> Option<u64>` — extract from HTTP headers.
- `parse_retry_after(error: &str) -> Option<Duration>` — parse `[retry-after:N]` prefix from error strings.
- `maybe_tag_retry_after(headers, error) -> String` — encode Retry-After into the error string before it reaches the retry loop.

`categorized_http_error` clones headers **before** consuming `response.text()`, then tags the output.

`stream_error()` — new helper replacing 8 duplicate `res.json() → extract_error()` patterns across streaming call sites (Claude, OpenAI, Gemini, Custom). Preserves Retry-After the same way.

Both `with_retry` and `with_stream_retry` now apply `base_ms = max(client_backoff, server_delay)` so the retry delay never violates the server's `Retry-After` directive.

### 2. Preserve HTTP status in stream errors (`ai.rs:1816–1834`)

`stream_error` now uses `response.text()` + `serde_json::from_str()` instead of `response.json()`. This guarantees the diagnostic always includes `"HTTP {status}: {detail}"`, so `classify_error` can match on status codes even for empty or non-JSON bodies.

| Scenario | Before | After |
|---|---|---|
| Empty body 429 | `"failed to parse error response: …"` → `"unknown"` → **no retry** | `"[rateLimit] … (HTTP 429: empty response body)"` → **retried** |
| HTML 429 from proxy | same | `"[rateLimit] … (HTTP 429: <html>…)"` → **retried** |
| JSON 401 | works, but no Retry-After | works, + Retry-After if present |

### 3. Tests (`ai.rs:5395–5814`)

**20 tests total** (14 new + 6 existing):

`with_retry` unit tests:
- First-attempt success, transient recovery, exhaustion, `maxRetries=0`, non-retryable (401), Retry-After timing (≥1s enforced)

`with_stream_retry` unit tests:
- First-attempt success, emitted=true blocks retry, emitted=false allows retry, cancellation during back-off

Helper unit tests:
- `parse_retry_after`, `retry_after_secs`, `maybe_tag_retry_after`, `is_retryable_error` category coverage, `classify_error` diagnostic coverage (HTTP 429/504/503/502/401/403 with empty + non-JSON bodies)

Integration tests (real `reqwest::Response` via local test servers):
- `spawn_error_server` / `spawn_error_server_with_body` — serve arbitrary HTTP status, headers, and body
- `stream_error_empty_body_429_is_retryable` — real 429 → `stream_error` → `classify` + `is_retryable`
- `stream_error_empty_body_429_triggers_retry` — real 429 → `stream_error` → `with_stream_retry` → successful retry
- `stream_error_preserves_retry_after_from_headers` — `Retry-After: 5` header encoded as `[retry-after:5]` prefix
- `stream_error_retry_after_extends_backoff` — `Retry-After: 1` forces ≥800ms delay (verified with `Instant::elapsed`)
- `stream_error_non_json_429_body_is_retryable` — HTML body containing `rate` keyword → still classifies as `rateLimit`
- `stream_error_401_is_not_retryable` — negative case: auth errors are never retried

### 4. Formatting

`cargo fmt --check` passes on `dbx-core`.

## Files changed

| File | +/− |
|---|---|
| `crates/dbx-core/src/ai.rs` | +1461 / −726 |

## Verification

`cargo fmt --check --package dbx-core` passes.
`cargo test` could not run locally due to a pre-existing OpenSSL build dependency (missing Perl module in this environment); CI is the source of truth for test results.

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

e91ed6e9-1978-41f1-b735-813121fdff79

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

Related #4620


t8y2 (commented):
Maintainer patch e40ed38 closes the confirmed blockers on the current PR head.

Root causes and fixes:
- HTTP-date `Retry-After` values were rounded down, so a still-future deadline under one second could trigger an early retry. The parser now uses the HTTP-date parser with deterministic current-time injection and rounds positive delays up.
- The new global `max_retries` value could be lost when a stale app-settings snapshot was written afterward. Its dedicated writer now updates atomically and legacy settings saves preserve the latest owned value.
- Two newly added regression tests did not compile or reach the HTTP server; the moved provider is cloned and the connection test now supplies the required API key.

Checks run:
- `cargo fmt --all --check`
- `cargo test -p dbx-core --no-default-features retry_after -- --nocapture`
- `cargo test -p dbx-core --no-default-features max_retries -- --nocapture`
- `cargo test -p dbx-core --no-default-features ai::tests::retry_ -- --nocapture`
- `cargo test -p dbx-core --no-default-features ai::tests::stream_retry_ -- --nocapture`
- `cargo test -p dbx-core --no-default-features ai::tests::stream_error_ -- --nocapture`
- `git diff --check`

t8y2 (commented):
CI follow-up patch ab3d805 fixes the remaining current-head test and clippy blockers:

- Enables Tauri's test support for dev/test builds and updates the mock-app call to the current API.
- Removes two unused test imports.
- Rewrites the Web entry-point assertion to satisfy `clippy::single_match`.

Checks run:
- `cargo clippy -p dbx -p dbx-web --locked --all-targets --no-default-features -- -D warnings`
- `cargo test -p dbx --no-default-features tauri_entry_respects_global_max_retries_zero -- --nocapture`
- `cargo test -p dbx-web --no-default-features web_test_connection_respects_global_max_retries_zero -- --nocapture`
- `cargo fmt --all --check`
- `git diff --check`

t8y2:
Thanks for the contribution! Merged in 52d5c99, will be released in the next version.

PR #4786, opened 2026-07-29
fix(starrocks): preserve external catalog query context

jischeng:
## Summary

- model an external StarRocks/Doris catalog independently from its database throughout query-tab creation, persistence, selection, and duplication
- initialize MySQL-family sessions with `SET catalog` before `USE database`, including multi-statement, transaction, count, explain, import, and export paths
- keep completion metadata, generated object SQL, and session cleanup scoped to the selected catalog

## Problem

Opening `paimon_catalog.bi` through a StarRocks connection previously retained only `bi`. DBX then initialized a fresh session with `USE bi`, which failed because `bi` exists under the external catalog rather than the internal catalog. Completion metadata was scoped incorrectly for the same reason.

## Verification

- `pnpm typecheck`
- `pnpm exec vitest run` — 639 files / 5545 tests passed
- `cargo check --workspace --no-default-features`
- `cargo test -p dbx-core catalog` — 51 tests passed
- local packaged DBX.app smoke test against `paimon_catalog.bi`


t8y2 (commented):
Maintainer patch pushed in `37e01a109`.

Root cause: Doris and StarRocks do not share one catalog-switching statement. Applying a generic `SET catalog = ...` across MySQL-compatible databases could fail to establish or restore the requested query context.

Patch: added database-specific catalog switching and quoting (`SWITCH` for Doris and `SET CATALOG` for StarRocks), then applied the same setup consistently during pool creation, connection reset, single and multi-statement execution, batch transactions, and manual transactions.

Validation:

- targeted Rust regression tests — 6 tests passed
- `cargo check -p dbx-core --no-default-features` — passed
- shared agent `:common:test` — passed
- `cargo fmt --all -- --check` and `git diff --check` — passed

t8y2 (commented):
Conflict-resolution follow-up pushed in `1a9209a03`.

The latest `main` changed query-tab context restoration in `App.vue`. The resolution keeps that schema behavior and preserves this PR's catalog fallback across the active tab, object browser, and table metadata.

Validation after merging `main`:

- catalog-focused Rust suite — 52 tests passed
- `cargo check -p dbx-core --no-default-features` — passed
- `pnpm typecheck` — passed
- `cargo fmt --all -- --check` and `git diff --check` — passed

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 2152f8c, will be released in the next version.

PR #4833, opened 2026-07-29
feat(xugu): expose private synonyms in schema tree

mapan0424:
## Summary

- Add a **Synonyms** group to Xugu schema trees and open its children in the object-source viewer.
- Read only private synonyms from `ALL_SYNONYMS`; public synonyms intentionally remain outside schema groups because they have no owning schema and would otherwise be duplicated under every schema.
- Reconstruct executable, quoted `CREATE SYNONYM ... FOR ...;` DDL from `ALL_SYNONYMS` and `ALL_SCHEMAS`.
- Preserve catalog spelling through exact-match lookup first, then a unique case-insensitive fallback. Ambiguous fallback results return an error rather than selecting a different quoted object.

## Scope

- [x] New feature
- [ ] Bug fix
- [ ] Performance
- [ ] Refactor
- [ ] Documentation
- [ ] CI / build

This is Xugu-only: no other driver receives a synonym tree group or executes Xugu synonym catalog queries.

The historical Xugu implementation branches contain no synonym metadata/source path. The private-versus-public split follows the DBeaver Xugu extension: private synonyms belong to a schema, while public synonyms are datasource-level objects.

## Frontend

- [x] The schema tree gains an Xugu-only object group and routing coverage.
- Automated tree/group/source-routing tests are included.

## Validation

- [x] `pnpm check`
- [x] `go test -count=1 .` in `agents/drivers/xugu`
- [x] Xugu Agent build
- [x] Regression coverage for object listing, private/public filtering, source reconstruction, quoted names, and exact-match/case-insensitive fallback behavior.
- [x] Validated against `SHOP_DEMO` / `SYSDBA`: listed existing private synonyms; created, read, replayed, and removed a quoted cross-schema synonym; confirmed a temporary public synonym is excluded from schema-level results.

### DIFF
--- a/agents/drivers/xugu/main.go
+++ b/agents/drivers/xugu/main.go
@@ -45,6 +45,12 @@ SELECT s.SCHEMA_NAME, q.SEQ_NAME
 FROM ALL_SEQUENCES q
 JOIN ALL_SCHEMAS s ON s.DB_ID = q.DB_ID AND s.SCHEMA_ID = q.SCHEMA_ID
 WHERE q.IS_SYS = FALSE`
+const xuguCatalogSynonymSelectSQL = `
+SELECT s.SCHEMA_NAME, y.SYNO_NAME, t.SCHEMA_NAME AS TARGET_SCHEMA, y.TARG_NAME
+FROM ALL_SYNONYMS y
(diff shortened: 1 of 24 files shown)


t8y2 (approved):
Looks good to me.

t8y2:
Thanks for the contribution! Merged in 1774126, will be released in the next version.

PR #4931, opened 2026-07-30
fix(sidebar): apply Doris table name filters

wbean:
## Summary
- align table-name filter lookup with the sidebar node schema
- restore Doris internal include filters such as ads_pgc_%
- add a Doris internal table-group regression test

## Test
- pnpm vitest run apps/desktop/src/stores/__tests__/connectionStore.metadataLoading.spec.ts --reporter=dot --silent


t8y2:
Thanks for the contribution! Merged in 380c4bb21, will be released in the next version.

PR #4935, opened 2026-07-30
fix(xugu): accept synonym object source requests

mapan0424:
## Summary

- add `SYNONYM` to the shared object-source contract so the desktop command accepts the Xugu synonym nodes already exposed in the sidebar
- keep conservative fallback handling for drivers that do not expose synonyms
- mark reconstructed Xugu synonym DDL as read-only source

## Why

The Xugu sidebar and agent already enumerate private synonyms, but `get_object_source` was rejected by the Rust/Tauri `ObjectSourceKind` deserializer before the agent received the request.

## Regression coverage

- `CARGO_NET_OFFLINE=true cargo test -p dbx-core --lib`
- direct serde coverage for the `SYNONYM` wire value
- programmable metadata parser and SQL routing coverage

### DIFF
--- a/agents/drivers/xugu/main.go
+++ b/agents/drivers/xugu/main.go
@@ -2285,6 +2285,7 @@ func (s *server) getSynonymSource(schema, name string) (map[string]any, error) {
 		"object_type": "SYNONYM",
 		"schema":      synonym.Schema,
 		"source":      builder.String(),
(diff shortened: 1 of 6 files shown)


t8y2:
Thanks for the contribution! Merged in 44f49253d, will be released in the next version.

PR #4951, opened 2026-07-30
fix(schema): preserve varchar and serial aliases in PostgreSQL type display and export

xue-yuxin:
## 改动说明
- 编辑/新增 PostgreSQL 字段时保持 varchar 而非 character varying
- DDL 视图保持 serial 而非展开为 bigint NOT NULL DEFAULT nextval(...)
- 导出表结构时同样应用友好类型名
- 迁移路径仍使用完整 nextval(...)，保证序列依赖不被破坏

## 验证方式
- 在 PostgreSQL 连接中编辑/新增字段，确认类型显示与 DDL 视图均为 varchar / serial
- 执行导出，确认导出结果使用友好类型名
- cargo check 与 cargo test 通过


t8y2 (changes requested):
The current head `75f5a20e` improves PostgreSQL type display, but the `serial` reconstruction changes schema semantics and breaks the existing export flow:

1. `crates/dbx-core/src/database_export.rs:1716` emits explicit `CREATE SEQUENCE` statements before table DDL, while `get_table_ddl_for_export_core` now folds the same integer `NOT NULL DEFAULT nextval(...)` column back into `serial`. Restoring that export attempts to create the sequence twice. Please either preserve the raw `nextval` form in full exports or redesign sequence/table emission so the dump restores without duplicate objects and preserves sequence attributes.

2. `crates/dbx-core/src/schema.rs:7481` treats any non-null integer column with a `nextval(...)` default as `serial`. Custom or shared sequences are therefore renamed/recreated with different ownership and dependency semantics. Detect only a sequence proven to be owned by that exact column, using PostgreSQL dependency metadata rather than the expression shape alone.

Please add round-trip tests covering a normal serial column, an explicitly created owned sequence, a shared/custom sequence, and full schema export ordering.

t8y2 (changes requested):
request changes: the merge-main update does not close the PostgreSQL sequence-semantic blockers.

1. Full exports now fold integer `NOT NULL DEFAULT nextval(...)` columns back into `serial`, while the exporter still emits sequences separately. A normal serial column can therefore attempt to create the same sequence twice during restore.

2. Both the Rust and TypeScript analyzers identify serial solely from an integer type plus a `nextval(...)` default. This also matches custom and shared sequences. The editor then removes the original default, so saving or exporting can recreate a different sequence and change ownership and dependency semantics.

Please only reconstruct `serial` when PostgreSQL dependency metadata proves that the sequence is owned by that exact table column. Full exports should preserve the raw `nextval(...)` form unless sequence enumeration, ownership restoration, and output ordering are redesigned together.

Please add round-trip coverage for normal serial ownership, an explicitly created owned sequence, a shared/custom sequence, and full-export ordering without duplicate sequence creation.

PR #4996, opened 2026-07-31
feat(sql): add ClickHouse function completion

encyc:
## Summary

- add a static ClickHouse registry covering regular, aggregate, window, and table functions with canonical casing and aliases
- generate valid aggregate-combinator completions without runtime metadata queries
- integrate ClickHouse-only function and table-function suggestions into SQL completion
- support overload-aware and parametric signature help, including multi-overload tooltip rendering
- preserve the legacy signature-help shape for existing non-ClickHouse callers

## Why

ClickHouse SQL editing did not offer useful function completion or signatures, while other database types already had dedicated inventories. This adds broad checked-in coverage without coupling completion to a live ClickHouse connection.

## Validation

- `pnpm exec vitest run ...` — 9 relevant files, 363 tests passed
- `pnpm typecheck`
- scoped `oxlint --vue-plugin`
- full-suite audit retains only the existing 20 `localStorage` environment failures in the two pre-existing grid test files


t8y2 (commented):
Maintainer patch added in `61285701d`.

Root cause: the checked-in aggregate snapshot also contains window-function names, so the registry generated invalid aggregate combinators for them; several window functions also used generic placeholders or missed their documented aliases.

Patch: model exact zero-argument/ranked, `ntile`, `lag`, and `lead` signatures, register camelCase aliases, and exclude every window-function spelling from aggregate-combinator generation.

Validation: 3 focused Vitest files, 67 tests passed; `pnpm typecheck`, focused formatting checks, and `git diff --check` passed.

t8y2:
Thanks for the contribution! Merged in 9bc1c1b0f, will be released in the next version.

PR #5028, opened 2026-07-31
feat(grid): navigate to referenced record from foreign key cells

### LINKED ISSUE #329: [Feature] 加入外键导航 跳转功能
### 提交前确认

- [x] 我已搜索现有 issues，确认没有重复建议。
- [x] 这个建议与 DBX 的数据库管理、查询、浏览、导出或桌面体验有关。

### 建议类型

其他

### 相关数据库

不限数据库

### 当前痛点

外键导航也是一个非常实用的功能，在进行数据查看的时候不用打开多个数据表进行查询对比，直接点击对应的外键数据即可跳转查询

### 期望方案

点击外键参数跳转查询对应的数据，可以自由跳转以及回退

Image

Image

### 期望工作流

_No response_

### 你试过的替代方案

_No response_

### 参考截图 / 产品 / 文档

_No response_

### 需求紧急程度

只是想法，不急

difagume:
## What

Adds foreign key navigation from the data grid, as requested in #329.

When a table is opened, its foreign keys are loaded in the background. Hovering a cell in an FK column with a non-null value shows a navigation button next to the existing "Info" button (tooltip: "Go to {referenced table}"). Clicking it opens the referenced table in a new tab, filtered by the cell value (`ref_column = <value>`), so the original tab stays intact and going back is natural. The same command is available in the cell context menu for discoverability.

Closes #329

## How

- New pure helper `dataGridForeignKeyNavigation.ts`: column → FK map (case-insensitive, composite FKs navigate per column), null-safety check, and navigation target builder (`ref_schema` falls back to the current schema, like the ER diagram does).
- `DataGrid.vue`: eager FK fetch gated by `tableMetadataCapabilities.foreignKeys` (reuses the existing `fetchForeignKeys` dedupe/reset), hover button in the floating action overlay, and cell context menu entry. Values are quoted per dialect through the existing backend filter builder (`buildColumnValueFilterCondition`), the same path used by column filters.
- `canvasDataGridRenderer.ts`: action overlay width generalized to reserve room for the third button.
- i18n: new `grid.foreignKeyNavigate` key in all 8 locales.

Only enabled when the engine supports foreign keys and the result has table identity; no Rust changes.

## Testing

- New unit spec `dataGridForeignKeyNavigation.spec.ts` (10 tests: map building, composite/duplicate FKs, navigability of null/0/""/false, schema fallback, `whereInput` propagation) — all passing via `pnpm vitest run`.
- All existing dataGrid specs pass (33 files, 199 tests).
- `pnpm typecheck` clean.
- Verified manually against a PostgreSQL database (screenshot below).

## Screenshot

_FK hover button on an `id_sede` cell, tooltip "Go to sede" (Spanish locale):_
pr


t8y2 (commented):
Maintainer patch added in `0f9e65ded`.

Root cause: foreign-key navigation treated each metadata row as an independent relationship, used displayed aliases as a fallback for query results, and allowed stale metadata responses to populate a reused data tab.

Patch: group every column pair in the same FK association and build one `AND` condition, resolve query-result cells only through physical `sourceColumns` bindings, and guard metadata results by table identity plus request generation. Regression coverage includes composite keys, aliases, missing bindings, nulls, generated conditions, and stale requests.

Validation: focused Vitest suite, 19 tests passed; `pnpm typecheck`, focused formatting checks, and `git diff --check` passed.

t8y2 (commented):
当前 exact HEAD 为 `023a5f8a2`。

首轮 CI 的 frontend 失败来自既有结构测试仍只匹配单参数 `canvasDataGridActionReservedWidth`；本 PR 为 FK 导航图标增加了第二个布尔参数，实现与类型检查均已通过。跟进提交 `023a5f8a2` 已同步该断言。

重新验证：FK 导航与 canvas renderer 两个聚焦 Vitest 文件共 27 项通过；`pnpm typecheck`、格式检查及 `git diff --check` 通过。后续合并以当前 HEAD 的新 CI 为准。

t8y2:
Thanks for the contribution! Merged in ac6b53365, will be released in the next version.

PR #5047, opened 2026-07-31
fix(ai): preserve Gemini thought_signature and model parts in tool calls

xEdenx:
### Summary
Fixes an issue where multi-turn Tool Calls using Gemini Thinking models (such as `gemini-3.5-lite` / `gemini-2.0-flash-thinking`) fail with:
`Agent stream stopped before completion: Function call is missing a thought_signature in functionCall parts.`

### Root Cause
When Gemini Thinking models respond with tool calls, they output thought parts containing `thought_signature` along with or prior to `functionCall` parts. In `build_gemini_contents`, previous Assistant tool call messages were reconstructed into Gemini `parts` by only serializing the single `functionCall` payload, discarding any associated `thought_signature` or model thought parts. When sending the multi-turn history back to Gemini, the missing `thought_signature` validation in `functionCall` parts causes Google's API to abort the stream.

### Solution
- Updated `emit_gemini_tool_call_part` to bundle candidate `model_parts` into `provider_payload` if thought signatures or multiple parts are present.
- Updated `build_gemini_contents` to restore full `model_parts` (including `thought_signature` parts) when reconstructing historical `model` messages.

### DIFF
--- a/crates/dbx-core/src/ai.rs
+++ b/crates/dbx-core/src/ai.rs
@@ -3547,22 +3547,32 @@ fn build_gemini_contents(messages: &[AiMessage]) -> Vec<serde_json::Value> {
             if m.role == "assistant" && !m.tool_calls.is_empty() {
                 let mut parts: Vec<serde_json::Value> = Vec::new();
                 let mut has_text_part = false;
-                for tc in &m.tool_calls {
-                    if let Some(payload) = &tc.provider_payload {
-                        if let Some(model_parts) = payload.get("model_parts").and_then(|p| p.as_array()) {
-                            for p in model_parts {
-                                if p.get("text").is_some() {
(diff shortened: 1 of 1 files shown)


t8y2 (commented):
Maintainer patch pushed: `75ddc13a` → `189dc0de`.

The patch fixes the compile errors, accumulates Gemini model parts across SSE events, delays tool-call publication until the stream completes, and replays the complete model parts exactly once for parallel calls. This preserves thought signatures and original part ordering without duplicating function calls or retaining partial tool history after an interrupted stream.

Validation: 15 focused Gemini tests, `cargo fmt --check`, and `git diff --check` passed.

t8y2:
Thanks for the contribution! Merged in 4a2f8f1, will be released in the next version.

PR #5048, opened 2026-07-31
fix(grid): grow default column widths for large page values

xddcode:
## Summary
- Short headers like `id` kept a narrow default width from early/small samples, so late-page integers rendered as leading ellipses (`...17001`).
- Sample both the head and tail of the current row window, grow cached widths only when larger values appear, and add a small measurement slack for tabular digits.

fb5c422f-2fc4-4b2b-82d7-b53b70ed847c

2f941847-b74b-48a3-9da3-6e29aeab6d61

## Test plan
- [x] `npx vitest run apps/desktop/src/composables/__tests__/useDataGridColumnResize.spec.ts packages/app-tests/dataGridColumnWidth.test.ts`
- [ ] Open a large table, jump to a late page with 7-digit `id` values, confirm the first data column shows full numbers by default
- [ ] Navigate from page 1 (small ids) to a late page and confirm the column grows instead of staying ellipsized
- [ ] Manually widen a column, confirm grow-only does not shrink it on later pages

### DIFF
--- a/apps/desktop/src/composables/__tests__/useDataGridColumnResize.spec.ts
+++ b/apps/desktop/src/composables/__tests__/useDataGridColumnResize.spec.ts
@@ -1,19 +1,29 @@
 // @vitest-environment happy-dom
 
-import { computed, nextTick, ref } from "vue";
+import { computed, isRef, nextTick, ref } from "vue";
 import { beforeEach, describe, expect, it } from "vitest";
-import { DATA_GRID_COL_AUTO_FIT_MAX_WIDTH, DATA_GRID_COL_MIN_WIDTH } from "@/lib/dataGrid/dataGridColumnWidth";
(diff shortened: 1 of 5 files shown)


t8y2 (commented):
维护者补丁已推送：`5ec599e7` → `e7fa680a`。

自动列宽增长现在直接使用实际百分位宽度作为变化签名，避免最大字符串长度相同但内容分布变化时不更新；同时持久化用户手动调整/显式 auto-fit 的列索引，翻页和重新挂载不会覆盖用户宽度。

验证：聚焦 Vitest 37/37、`pnpm typecheck`、`oxfmt --check`、`oxlint`、`git diff --check` 均通过。

t8y2:
Thanks for the contribution! Merged in 584848f73, will be released in the next version.

PR #5053, opened 2026-07-31
fix(xugu-agent): degrade SYSTEM control for normal user load

mapan0424:
## Summary
- Xugu `open_session` / `connect` previously required a SYSTEM control connection and `SYS_SESSIONS` access to track sessions for cancel/kill.
- Ordinary business users (user exists only in the app database, or no `SYS_SESSIONS` privilege) failed with Agent RPC `-1` / `E18012` even though the business DB credentials were valid (`test_connection` OK).
- **P0:** treat SYSTEM control as optional. If control open/ping/`SYS_SESSIONS`/session-identity fails, still open the configured business database session and degrade cancel/kill to no-op. When control works (e.g. SYSDBA), behavior is unchanged.

## Files
- `agents/drivers/xugu/main.go`
- `agents/drivers/xugu/main_test.go`

## Test plan
- [x] `go test ./...` in `agents/drivers/xugu`
- [x] Live local OrbStack: `DBX_LOCAL_TEST` @ `SHOP_DEMO` → `open_session` OK, `list_databases` → `SHOP_DEMO`, `list_schemas` OK
- [x] Live: `SYSDBA` @ `SHOP_DEMO` → `open_session` OK, multi-db list still works
- [x] Live: `test_connection` for normal user still OK
- [ ] Desktop: connect as normal user to business DB → load tree succeeds (cancel may be unavailable)

### DIFF
--- a/agents/drivers/xugu/main.go
+++ b/agents/drivers/xugu/main.go
@@ -620,7 +620,7 @@ func (r *runtimeServer) dispatch(method string, params map[string]json.RawMessag
 			return map[string]bool{"ok": true}, false, nil
 		}
 		r.connectMu.Lock()
-		err = session.server.connectWithControl(session.server.params, session.server.cancelDB, false)
+		_, err = session.server.connectWithControl(session.server.params, session.server.cancelDB, false)
 		r.connectMu.Unlock()
 		return map[string]bool{"ok": true}, false, err
 	case "cancel_session":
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch applied on `2dd7bcf18` (previous head `e48d070a9`, rebased onto `c1be3da13`).

The conflict came from unrelated generated contributor data; the current `main` version was retained while preserving the Xugu control-session fallback changes.

The patch also releases and clears a detached shared control reference when business reconnection succeeds without control capability, preventing later sessions from repeatedly reusing a stale control connection. A regression test covers control failure, successful business reconnection, old-control cleanup, and establishing a replacement control connection.

Checks run:
- `gofmt`
- `git diff --check`
- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- old/new `git range-diff`

No dependencies were added.

t8y2 (commented):
The patched branch was rebased again from `2dd7bcf18` to `b8b1251fc` on the new `main` head `0e414e638`, which includes the repaired RabbitMQ CI environment.

Both Xugu commits are patch-equivalent according to `git range-diff`. The final PR diff still contains only `agents/drivers/xugu/main.go` and `agents/drivers/xugu/main_test.go`; the unrelated generated contributor data remains excluded.

Checks rerun after the rebase:
- `git diff --check`
- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- old/new `git range-diff`

No dependencies were added.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in aef61df30, will be released in the next version.

PR #5076, opened 2026-07-31
feat(export): split XLSX rows across sheets when exceeding the per-sheet limit

### LINKED ISSUE #4671: [Feature] XLSX导出最多支持 1,048,575 行数据，建议数据超过1,048,575条后，多余数据写到新建的sheet页
### 数据库类型和版本

通用

### 支持信息

DBX 版本: v0.5.65
运行环境: 桌面端
操作系统: Windows 10.0.19044.7417
系统架构: x86_64

### 当前痛点

用xlsx格式导出大量数据时会报 “XLSX 最多支持 1,048,575 行数据，请改用 CSV 导出完整结果。”； 但是csv不利于数据的展示，如果数据中包含0开头的数字，用excel打开csv文件，数字前的0会被去掉导致数据错乱。

Image

Image

### 期望方案

建议数据超过1,048,575条后，多余数据写到新建的sheet页

### 优先级（是否紧急）

🟢 P3 低优 · 有空再做

### 补充信息

_No response_

Abeautifulsnow:
## Summary

Fixes #4671. Excel hard-limits every worksheet to **1,048,576 rows** (header + 1,048,575 data rows). Previously, exporting a result set larger than that failed with an error telling the user to fall back to CSV. With this change, oversized exports are **automatically split across multiple sheets** (`Result`, `Result (2)`, `Result (3)`, …), so large result sets export successfully while staying within Excel's constraints.

This applies to **both export paths**:

- **Streaming path** (`StreamingXlsxWriter`, used by query result exports) — the writer rotates to a new data sheet as soon as `max_data_rows_per_sheet` is reached, so it stays **constant-memory** and never buffers the full result set.
- **In-memory path** (`build_xlsx_workbook` / `build_xlsx_workbook_multi`, used by non-streaming Tauri exports) — sheets are split via borrow-only `WorksheetSegment` descriptors, so **no cell data is deep-copied** even for million-row exports.

### Rust — `crates/dbx-core/src/xlsx_export.rs` (+798)
- `XLSX_MAX_DATA_ROWS = 1_048_575` — the effective per-sheet data-row ceiling.
- `StreamingXlsxWriter.write_row()` now triggers `finish_current_sheet()` + `start_next_data_sheet()` when the current sheet is full; per-sheet row counters track rotation.
- `SheetNameAllocator` deduplicates names across the whole workbook and keeps the trailing **SQL sheet last**, naming overflow chunks `Result`, `Result (2)`, `Result (3)`, … with 31-char truncation + illegal-character sanitization.
- `split_sheets_for_max_rows()` splits overflowing sheets for the in-memory path; `WorksheetSegment<'a>` borrows columns/types/comments/rows instead of cloning them (P1 memory fix), with a defensive `max(1)` guard against `slice::chunks(0)`.
- ZIP entry ordering note: metadata files are written after sheet data in `finish()`, which is valid because ZIP readers locate entries via the central directory.

### Rust — `crates/dbx-core/src/query_result_export.rs` (-120)
- Removed the XLSX-specific hard-limit error and implicit `1_048_575` query-layer cap.
- `effective_row_limit()` now honors only the **user's explicit `row_limit`**, so the "export as CSV" limitation message is gone and no query layer artificially truncates exports.

### Frontend — dead-code cleanup
- Removed the obsolete `xlsxRowLimit` backend-error mapping from `apps/desktop/src/i18n/backend-errors.ts` and all 8 locale files (`en`, `zh-CN`, `zh-TW`, `ja`, `ko`, `pt-BR`, `it`, `es`).
- Retargeted `backendErrors.spec.ts` and `packages/app-tests/useDataGridExport.test.ts` to the surviving streaming-unsupported error path.

### Rust — `crates/dbx-core/src/ai_codex_cli.rs` (+2/-1, separate commit)
- Moved `codex_process_env_with_system_proxy` into the unconditional test import list; it was mis-gated under `#[cfg(not(windows))]` even though Windows tests reference it, breaking `cargo test --lib` compilation on Windows.

## Behavior notes

- The user's explicit `row_limit` is **still respected** (truncation happens at the query layer, not per-sheet).
- Sheets that don't overflow are passed through unchanged — only overflowing sheets are split.
- The SQL/comment trailing sheet always appears **after** all data sheets.

## Tests

- `xlsx_export::tests` — **27 passed** (incl. new: splitting rows over the limit, multi-sheet only-overflowing split, under-limit single sheet, header-only sheet, identically-named trailing sheets)
- `query_result_export::tests` — **32 passed** (no query-layer XLSX cap; user row-limit still respected; total rows above sheet limit no longer errors)
- `table_export::tests` — **23 passed**
- `cargo check --no-default-features` and `cargo fmt --check` — clean
- Frontend: `backendErrors.spec.ts` (159) and `useDataGridExport.test.ts` (30) — pass

## Known limitations

- Not yet validated against a real million-row database export (would require a large live dataset).
- If two sheets share the same base name *and* both overflow, the dedup suffix can nest (e.g. `Result (2) (2)`) — still a valid, unique Excel name, but cosmetically odd. Unreachable on the streaming path (data-sheet names are guaranteed unique).

## 变更类型

- [ ] 新功能
- [ ] Bug 修复
- [x] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

Related #4671


t8y2 (commented):
Maintainer patch applied on top of the reviewed exact head.

- Adds equivalent multi-sheet splitting to complete local Web XLSX exports.
- Streams Rust worksheet XML directly into ZIP entries to reduce peak memory.
- Adds Web-path large-result coverage and a 25,000-row bounded-write Rust regression.

Validation: 42 Vitest tests, 28 targeted Rust tests, `pnpm typecheck`, `cargo check -p dbx-core --no-default-features`, formatter checks, and `git diff --check`.

t8y2:
Thanks for the contribution! Merged in c43e845, will be released in the next version.

PR #5101, opened 2026-07-31
fix: keep connection tooltip open while copying

James-Leong:
## What changed

Increase the sidebar connection detail tooltip's `close-delay` from 0ms to 30ms.

## Why

The tooltip is rendered through Teleport and sits 8px away from the tree row. With an immediate close delay, pointer movement across that gap closes the tooltip before the pointer can enter it. A short delay lets `LightTooltip`'s existing tooltip `mouseenter` handler cancel the pending close timer, so users can select and copy connection details.

## Validation

- `pnpm exec vitest run apps/desktop/src/lib/__tests__/sidebar/sidebarLabelTooltip.spec.ts apps/desktop/src/components/ui/__tests__/LightTooltip.spec.ts`
- `git diff --check`
- pre-commit `oxfmt` hook passed


t8y2:
Thanks for the contribution! Merged in b5ba730, will be released in the next version.

PR #5127, opened 2026-08-01
Feature/data transfer tree selection

GIWTO:
## 变更说明
本 PR 主要实现了以下两个核心功能及相关修复：
1. **跨数据库非表对象传输（后端实现）**：
   - 适配并支持了 MySQL、SQL Server、达梦（Dameng）之间的非表对象传输。
   - 采用“只支持可转换对象”策略，跨库时仅限传输 **视图 (View)** 和 **序列 (Sequence)**，其余不支持的类型在前端置灰。
   - 实现了 DDL 的跨库方言转换（剥离 DEFINER、WITH SCHEMABINDING、AS 类型等，转换引号风格及归一化序列关键字），并对视图的基表引用自动加上目标 schema 限定符。
2. **前端对象选择树优化（前端实现）**：
   - 重构了 `ObjectSelectionTree` 组件：将搜索框与操作按钮固定在顶部，分类列表独立滚动，彻底消除了嵌套双滚动条和滚动遮挡问题。
   - 引入 Vue 的 `computed` 过滤优化，解决大表/大对象场景下的渲染性能瓶颈。
   - 优化了搜索状态下的“全选”与“组全选”行为（仅勾选/取消过滤出的可见项）。
## 变更类型
- [x] 新功能
- [x] 性能优化
- [x] 代码重构
- [ ] Bug 修复
- [ ] 文档更新
- [ ] CI / 构建
## 涉及前端
image
## 验证
- [x] `make check` 通过
- [x] 相关测试通过
  - 后端：`dbx-core` 传输单元测试全部通过（共 157 个用例，包含中文标识符转换等新增 TDD 用例）。
  - 前端：`ObjectSelectionTree` 相关 vitest 测试全部通过（共 15 个用例，包含源码特征值分析测试）。


t8y2 (changes requested):
Request changes for three compatibility problems:

1. The new selection tree does not model `none / partial / all` state. Clearing a filtered group removes hidden selections, while disabled object types can remain selected and are still serialized into the request. Please scope bulk actions to visible enabled objects, filter disabled types at the request boundary, and add real component interaction tests.

2. `selectedObjects` starts empty and the Tauri/Web paths now invoke schema-object transfer only when the list is non-empty. The previous PostgreSQL-to-PostgreSQL flow always transferred views, functions, triggers, policies, ownership, and grants. Following the old table-selection flow now silently drops that behavior. Please define and test a backwards-compatible default instead of treating an empty new selection as “transfer none”.

3. The cross-database capability matrix enables VIEW/SEQUENCE directions that the executor rejects or rewrites using regexes that do not distinguish identifiers from strings/comments. Please narrow the matrix to validated directions and use dialect-aware conversion before exposing those combinations. Also restore the ten unrelated `docs/superpowers` files removed by this branch.

t8y2 (changes requested):
request changes: the selection filtering issue is mostly fixed, but three blockers remain on this head.

1. `src-tauri/src/commands/transfer.rs:250` and `crates/dbx-web/src/routes/transfer.rs:343` still call `transfer_schema_objects` only when `objects` is non-empty. This makes the new PostgreSQL empty-selection compatibility logic unreachable from both real entry points, so the legacy table-only flow can still omit views, functions, triggers, policies, ownership, and grants. Please route both entry points through the shared Core decision and add entry-point regression tests.

2. `apps/desktop/src/lib/database/transferObjectKinds.ts:71` still enables cross-family VIEW transfer between MySQL, SQL Server, and Oracle families, while `crates/dbx-core/src/transfer.rs:384` only rewrites DDL wrappers, quoting, and schema qualifiers. It does not translate the view query body, so source-specific constructs such as `IFNULL`, `TOP`, or `GETDATE` can be executed unchanged on an incompatible target. Please disable unverified cross-family VIEW directions or introduce dialect-aware conversion with live target coverage.

3. `apps/desktop/src/components/transfer/ObjectSelectionTree.vue:198` renders the same checked icon for partial and complete group selection. Please add an indeterminate state and a component interaction test covering none / partial / all.

t8y2 (commented):
Maintainer patch pushed in `cf5679875`.

Root cause: the shared schema-object decision only considered database families and object selection, not `TransferContent`. As a result, PostgreSQL-family `DataOnly` transfers with an empty object selection could enter the legacy “transfer all schema objects” fallback.

Fix: include the content mode in the Core decision, make `DataOnly` an unconditional schema-object no-op before database-type lookup, and keep the legacy PostgreSQL-family empty-selection behavior only for structure-bearing modes. Web and Tauri continue through the same Core path.

Validation: Core content/database matrix test passed; Web data-only no-op test passed; `cargo fmt --all --check`, `cargo check --no-default-features`, and `git diff --check` passed.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 473ef910b, will be released in the next version.

PR #5136, opened 2026-08-01
fix(xugu): degrade metadata operations for normal users

mapan0424:
## Summary

Normal XuguDB users can now browse and inspect accessible objects even when SYSTEM control or ALL_* metadata views are unavailable.

## Root cause

The existing connection fallback only degraded the SYSTEM control session. Metadata handlers still propagated E18012 permission errors from ALL_* catalog views as Agent RPC errors. Several child-object and DDL paths also failed before they could use the direct table connection.

## Changes

- Expand Xugu metadata-access error detection to sequences, synonyms, procedures, packages, types, partitions, and subpartitions.
- Fall back from ALL_TABLES/ALL_VIEWS to USER_TABLES/USER_VIEWS for the current user schema.
- Fall back to SELECT * ... WHERE 1 = 0 for column discovery when catalog table resolution is unavailable.
- Treat inaccessible indexes, constraints, foreign keys, triggers, partitions, and subpartitions as empty metadata instead of RPC failures.
- Generate a best-effort column-only table DDL when catalog metadata is unavailable.
- Return an explicit read-only source explanation when programmable-object source metadata cannot be reconstructed.
- Preserve normal privileged metadata behavior and keep all changes scoped to the Xugu Go agent.

## Compatibility

The fallback is only activated for Xugu permission or catalog-access errors. Syntax errors, network errors, ambiguous identifiers, and genuinely missing objects continue to return errors. Cross-schema objects that are not exposed by Xugu ALL_* views remain subject to the database privilege model, but directly accessible tables can still expose columns and best-effort DDL.

## Verification

- \`go test -count=1 ./...\`
- \`go test -count=1 -race ./...\`
- \`go vet ./...\`
- \`go build\`
- \`git diff --check\`
- Real XuguDB 12.10.13 test on SHOP_DEMO with a non-DBA user:
  - open_session, list_databases, list_schemas, list_tables, list_objects
  - columns, indexes, constraints, foreign keys, triggers, partitions, subpartitions
  - table DDL, procedure source, and sequence source
  - cross-schema SELECT-granted table column discovery and DDL fallback
- Temporary test user and objects were removed after replay.

### DIFF
--- a/agents/drivers/xugu/main.go
+++ b/agents/drivers/xugu/main.go
@@ -1516,23 +1516,18 @@ func configuredDatabaseName(params connectParams) string {
 
 func isXuguMetadataAccessError(err error) bool {
 	message := strings.ToUpper(err.Error())
-	return strings.Contains(message, "E18012") ||
-		strings.Contains(message, "权限不够") ||
-		strings.Contains(message, "ALL_DATABASES") ||
-		strings.Contains(message, "SYS_DATABASES") ||
-		strings.Contains(message, "ALL_SCHEMAS") ||
-		strings.Contains(message, "SYS_SCHEMAS") ||
-		strings.Contains(message, "ALL_TABLES") ||
-		strings.Contains(message, "SYS_TABLES") ||
-		strings.Contains(message, "ALL_VIEWS") ||
-		strings.Contains(message, "SYS_VIEWS") ||
-		strings.Contains(message, "ALL_COLUMNS") ||
-		strings.Contains(message, "ALL_CONSTRAINTS") ||
-		strings.Contains(message, "ALL_INDEXES") ||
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Patched the Xugu metadata fallback so only concrete permission or missing-catalog conditions degrade gracefully. Errors from `USER_TABLES`/object fallback queries and direct DDL column probes now propagate instead of being converted into empty metadata or placeholder success.

Validation passed: focused permission/error-propagation tests, full Xugu driver `go test -count=1 ./...`, `gofmt`, and `git diff --check`.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 4ee1f2b94, will be released in the next version.

PR #5149, opened 2026-08-02
fix(ai): keep saved models selectable and persist manual models

Abeautifulsnow:
## Summary

The AI model list in the assistant dialog consistently failed for providers whose model-discovery endpoint (`/models`) returns empty or errors, and manually entered ("manual") models were never persisted — they were only set as the active selection in memory, so they vanished on restart.

This PR makes the model list resilient to discovery failure by always merging saved models with discovered ones, and persists manual models through the existing config storage pipeline so they survive restarts.

## Root Cause

1. **Empty model list on discovery failure** — `getModelsForConfig` returned only `modelCatalogs.get(configId)?.models ?? []`, i.e. models discovered from the provider's `/models` endpoint. When discovery failed or returned an empty list, the dialog showed a "Failed to load models" state with no usable entries, even if the user had already saved models for that config.

2. **Manual models never saved** — `applyManualModel` only called `handleModelSelect` (which sets the active model in memory). Nothing wrote the model back to `config.models`, so it was not stored in `ai_configs.models` and disappeared after restart.

### `apps/desktop/src/lib/ai/aiConfigList.ts`

- **`aiModelOptions(config, discovered)`** — new pure helper that combines the user's saved models (`config.model` + `config.models`) with the provider-discovered models, deduplicated by trimmed model id:
  - `displayName` / `supportedEffortLevels` prefer the saved entry (user labels win)
  - `effortCapability` prefers the discovered entry
  - So saved models remain selectable even when discovery fails or is empty.
- **`addConfiguredAiModel(models, modelId)`** — new pure helper that appends a trimmed model id to `config.models`, skipping empty ids and duplicates.

### `apps/desktop/src/components/editor/AiAssistant.vue`

- `getModelsForConfig` now resolves the config from `settings.aiConfigs` and returns `aovered)` instead of the raw catalog.
- `applyManualModel` is now async and:
  1. persists the manual model via `settings.updateAiConfigItem(configId, { models: add — verified to persist through the Rust backend into the `ai_configs.models` column,
  2. skips the redundant write when the entered id equals the config's default `model`,
  3. only then selects the model and clears the form,
  4. surfaces save failures through `toast(translateBackendError(...))`, consistent wit

### `apps/desktop/src/lib/ai/__tests__/aiConfigList.spec.ts`

Added 4 tests covering:

- saved models remain selectable when discovery returns nothing
- manual model persisted only once (trim + dedupe)
- merge precedence between saved metadata and discovered capability
- empty configured model / blank manual id edge cases

## Testing

- `vitest run` — **701 test files / 6,203 tests, all passing**
- `typecheck` (vue-tsc) — passing
- `oxlint` + `oxfmt` — clean

## Follow-up (not blocking)

- `applyManualModel` integration logic in `AiAssistant.vue` has no component-level testinto the unit-tested helpers. A component test harness for the assistant dialog could beadded later.

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

image

image

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

## 关联 Issue

Related #4859

### DIFF
--- a/apps/desktop/src/components/editor/AiAssistant.vue
+++ b/apps/desktop/src/components/editor/AiAssistant.vue
@@ -57,6 +57,7 @@ import { useToast } from "@/composables/useToast";
 import { useNavigationTargets } from "@/composables/useNavigationTargets";
 import { buildAiContext, resolveAiDatabaseTarget, resolveAiNamespaceSelection, resolveDefaultAiSchema, runAgentStream, isVectorDbType, isValidActionForMode, defaultActionForMode, type AiAction, type AiAssistantMode, type AiSqlFileContext, type CustomPromptContext } from "@/lib/ai/ai";
 import { isAiConfigModelCandidate } from "@/lib/ai/aiConfigCandidates";
+import { addConfiguredAiModel, aiModelOptions } from "@/lib/ai/aiConfigList";
 import { orderAiConfigsForDisplay } from "@/lib/ai/aiConfigOrdering";
 import { effortSelectionEquals, runtimeEffortFromPreference } from "@/lib/ai/aiEffortPreference";
 import { useAiModelCatalog } from "@/composables/useAiModelCatalog";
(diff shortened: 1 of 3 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in b053adf17, will be released in the next version.

PR #5160, opened 2026-08-02
feat(explain): Plan Canvas — unified graphical execution plan view, with actual-plan support

### LINKED ISSUE #5161: Feature: graphical execution plan view (Plan Canvas) — one visual language across engines
## Problem

DBX already normalizes execution plans from six engines (mysql, postgres, dameng, questdb, oracle, sqlserver) into a single `ExplainPlanNode` tree, but the Explain results panel renders it only as an indented list, a flat summary table, and raw JSON/XML/text. There is no graphical plan view, no visual cost weighting, and no way to spot the expensive operator or a misestimate at a glance — the things DBA-oriented tools (SSMS, dbForge, pgAdmin) offer, each with a different engine-specific visual language.

## Proposal

A **Plan Canvas** view rendering the normalized tree as an interactive diagram with one visual language for every engine:

- **Category, not operator, drives icon + color** — ~11 semantic categories (table scan, index scan, row lookup, join, sort, aggregate, materialize, parallelism, modify, result, other). `Seq Scan`, `ALL`, `Clustered Index Scan` and `TABLE ACCESS FULL` render identically; cards keep the engine's native operator names.
- **Cost is heat** — green/amber/red badge by share of total plan cost.
- **Data volume is ink** — edge thickness scales with log(rows).
- **Progressive disclosure** — card → click-to-inspect side panel with all engine-native details → raw plan untouched.
- **Est vs actual** — amber `×N.N` badge when measured rows diverge ≥ 2× from the estimate (Dameng autotrace, PostgreSQL `EXPLAIN ANALYZE`).

Plan Canvas — PostgreSQL EXPLAIN ANALYZE

Engines without numeric costs (QuestDB) degrade to neutral badges; unknown operators fall back to *Other* — the view never blocks an engine. No new dependencies (custom SVG layout, ~160 lines of logic; Tailwind tokens and Lucide icons already in the stack).

## Implementation

PR #5160 implements this (Plan Canvas + PostgreSQL EXPLAIN ANALYZE support), with 52 new frontend tests and 3 new `dbx-core` unit tests.

/claim

possebon:
Closes #5161

## What

A new **Plan Canvas** view for execution plans: one visual language regardless of the connected engine. It renders the already-normalized `ExplainPlanNode` tree (mysql, postgres, dameng, questdb, oracle, sqlserver) as an interactive diagram and becomes the default view when a structured plan exists. The existing Tree / Summary / Raw / Table views are unchanged.

**Visual grammar (engine-agnostic by construction):**
- **Category, not operator, drives icon + color** — ~11 semantic categories (table scan, index scan, row lookup, join, sort, aggregate, materialize, parallelism, modify, result, other). `Seq Scan`, `ALL`, `Clustered Index Scan` and `TABLE ACCESS FULL` all look the same; cards keep the engine's native operator name.
- **Cost is heat** — green < 5%, amber 5–20%, red > 20% of total plan cost (self-cost derived from cumulative subtree costs).
- **Data volume is ink** — edge thickness scales with log(rows), row counts labeled on edges.
- **Progressive disclosure** — card (4 facts) → side inspector (all engine-native details) → raw JSON/XML/text untouched.
- **Est vs actual** — nodes whose measured rows are ≥ 2× off the estimate get an amber `×N.N` badge.

## Actual-plan support (commits 2 and 3)

The Dameng autotrace toggle is generalized into an actual-plan toggle:

| Engine | Estimated | Actual |
|---|---|---|
| PostgreSQL | `EXPLAIN (FORMAT JSON)` | `EXPLAIN (ANALYZE, FORMAT JSON)` |
| SQL Server | `SET SHOWPLAN_XML` | `SET STATISTICS XML` |
| Dameng | `EXPLAIN` | autotrace *(pre-existing)* |

Both new paths keep the existing safety gate — explain sources are restricted to `SELECT/WITH/TABLE/VALUES`, so a statement that actually executes can never be DML — and engines that don't support the flag ignore it (pinned by unit tests).

**Row-semantics normalization (the subtle part).** PostgreSQL reports `Actual Rows` per loop, matching `Plan Rows`. SQL Server reports `ActualRows` **cumulatively across all executions** while `EstimateRows` stays **per execution** — so the inner side of a nested loop would flag a bogus misestimate of exactly the execution count. The parser therefore emits `Actual Rows Per Execution` for SQL Server and the canvas prefers it when present. Crucially, normalization is gated on executions *per thread*, not total executions: a parallel operator reports one execution per worker, and dividing there would under-report every parallel node by its degree of parallelism. Runtime counters are aggregated across threads (rows and executions summed, elapsed = slowest thread, CPU summed).

**SQL Server actual plan — 37 000 rows over 3 700 executions normalizes to exactly the 10-row estimate, so no false badge, while the genuine ×2.3 misestimates upstream still fire:**
SQL Server actual plan

**PostgreSQL EXPLAIN ANALYZE — actual rows drive ×2.3 misestimate badges on Sort and Hash Join:**
PostgreSQL EXPLAIN ANALYZE

## Screenshots

Same query, same visual language, different engines:

**PostgreSQL — light**
PostgreSQL light

**PostgreSQL — dark**
PostgreSQL dark

**SQL Server — dark, estimated plan**
SQL Server dark

## How

- `apps/desktop/src/lib/diagram/planCanvas.ts` — pure, fully unit-tested logic: ordered categorization rules (handles overlaps like *Merge Join* vs *Merge*, *SORT AGGREGATE* vs *Sort*, *Clustered Index Scan* vs *Index Scan*), engine-native number parsing (`0.29..1830.12`, `1,234`, `96M`), cumulative-cost self-share computation, tidy left-to-right tree layout.
- `apps/desktop/src/components/explain/ExplainPlanDiagram.vue` — SVG elbow edges, node cards, heat bars/badges, zoom in/fit/out, drag-to-pan, click-to-inspect panel, legend. Light + dark themes via existing tokens.
- `ExplainPlanViewer.vue` — Canvas added as first tab and default; table-only fallback behavior (MySQL 5.x path) preserved.
- `crates/dbx-core` — optional, serde-defaulted `analyze` field on `ExplainSqlOptions`, backward-compatible across the Tauri and HTTP transports. `is_sqlserver_showplan_xml_set` widened to `is_sqlserver_plan_capture_set` so the read-only guard and the simple-query-batch rule both cover `SET STATISTICS XML`.
- i18n — keys added to all 8 locales, parity verified.
- **No new dependencies.**

## Tests

- `pnpm typecheck` — clean.
- Frontend suite: **675 files / 5,887 tests passing** (64 new across the three commits), no regressions.
- `cargo test -p dbx-core` — **3,744 passed**; the touched modules verified independently (`query_execution_sql` 43, `sqlserver` 220).
- CI on the fork (same workflows as here) passes `frontend`, `rust-test`, and `rust-fmt-clippy`.

## Notes for reviewers

- Engines without numeric costs (QuestDB) degrade to neutral badges; unknown operators fall back to the *Other* category — the diagram never blocks an engine.
- The categorization table lives in one place (`CATEGORY_RULES`) and is ordered to resolve cross-engine naming overlaps; supporting a new engine is a one-line rule.
- One pre-existing sharp edge left untouched as out of scope: `sqlServerExplainResult` locates the plan by scanning for the first cell containing `<ShowPlanXML`. Under `SHOWPLAN_XML` no user rows come back, but under `STATISTICS XML` the real result set precedes the plan — so explaining a query that itself selects plan XML (e.g. from `sys.dm_exec_query_plan`) could match the user's own row first. Happy to tighten it here if you'd prefer.

https://claude.ai/code/session_01TKcC15gEuQBHMidPCFFYwn


possebon:
Feature issue with the full rationale: #5161 (claimed). This PR closes #5161.

possebon:
Note on the `db/postgres` label (auto-applied): only the second commit (EXPLAIN ANALYZE) is PostgreSQL-specific. The Plan Canvas itself is engine-agnostic — it renders the normalized plan tree for **all six engines with explain support today: MySQL, PostgreSQL, SQL Server, Oracle, Dameng and QuestDB**, keeping each engine's native operator names. Engines without numeric costs (QuestDB) degrade to neutral badges, and unknown operators fall back to an *Other* category, so no engine is blocked.

t8y2 (commented):
Maintainer patch applied in `12318ec1b`.

Root causes addressed:
- PostgreSQL `EXPLAIN ANALYZE` ran on a reusable session without a guaranteed rollback, so side effects or failed transaction state could leak into later queries.
- The isolated PostgreSQL execution mode was not fail-closed for multi-statement requests.
- SQL Server actual-plan validation missed dialect-specific `SELECT INTO` table creation.
- PostgreSQL `Merge Append` was classified as a DML operator because of the generic `Merge` prefix rule.

Patch summary:
- Run PostgreSQL analyzed plans in a unique client session inside `BEGIN READ ONLY`, apply the selected schema transaction-locally, and always `ROLLBACK` after success, failure, cancellation, or timeout.
- Discard the exact isolated pool without retry when rollback cleanup fails, and close the frontend session before completing the analyzed-plan request.
- Require PostgreSQL read-only transaction mode to use an isolated session and exactly one statement.
- Reject SQL Server analyzed plans whose source is a dialect-specific write such as `SELECT INTO`.
- Keep bare `Merge` as modify while leaving `Merge Append` in the non-DML category.

Checks run:
- Vitest: 3 files, 42 tests passed.
- Targeted dbx-core tests for PostgreSQL transaction mode, rollback paths, pool discard, and SQL Server analyze safety passed.
- `pnpm typecheck`
- Targeted `oxlint` and `oxfmt --check`
- `cargo check --no-default-features`
- `git diff --check`

The PostgreSQL rollback design was also cross-checked against DBeaver's PostgreSQL plan analyzer, which disables auto-commit for plan execution and rolls back in cleanup. No live PostgreSQL integration test was run.

t8y2 (commented):
Maintainer conflict resolution applied on `ea1dc2b9b`.

- Previous patched head: `12318ec1b48ed88bcafed4cac8960fad2d1e2dfd`
- Rebased onto `main` at `cc7b2b17c57af480555631fb7f659f479b7e5f19`
- New head: `ea1dc2b9bf5e517728f474d51a708cf4b81e3b21`

Conflict cause:
- `main` added PostgreSQL text-protocol selection and structured agent recovery in the same `query.rs` / `postgres.rs` call path after the PR's original base.

Resolution:
- Preserved `main`'s Redshift text-protocol behavior, localized PostgreSQL decoding tests, and no-replay agent recovery policy.
- Preserved the PR's PostgreSQL isolated read-only transaction, guaranteed rollback, pool discard on cleanup failure, SQL Server `SELECT INTO` safety gate, and plan-node classification fix.
- The final base advance had no overlapping PR files; `git range-diff` reports all four PR commits unchanged after that rebase.
- No dependencies were added or changed.

Validation:
- Vitest: 3 files, 42 tests passed.
- Conflict-owner dbx-core tests passed for PostgreSQL rollback, JSON-array decoding, Redshift text protocol, non-agent reconnect behavior, and SQL Server analyze safety.
- `pnpm typecheck`, targeted `oxlint`, and `oxfmt --check` passed.
- `cargo fmt --all -- --check`, `cargo check -p dbx-core --no-default-features`, and workspace `cargo check --no-default-features` passed.
- Final-base focused tests and `cargo check -p dbx-core --no-default-features` passed again.
- `git diff --check` and complete range-diff review passed.

No live PostgreSQL integration test was run.

possebon:
Heads-up on the red `rust-fmt-clippy`: it does not originate from this branch.

```
error: this `map_or` can be simplified
  --> crates/dbx-core/src/table_structure_sql/triggers.rs:58:32
   = note: `-D clippy::unnecessary-map-or` implied by `-D warnings`
help: use `is_none_or` instead
```

That line comes from `cc7b2b17c` ("fix(postgresql): allow table edits with unchanged triggers") on `main` — the commit this PR was rebased onto — and `triggers.rs` does not appear in this PR's diff. It reproduces on `main` alone, so every PR rebased onto the current tip will fail the same check.

Suggested one-liner (clippy's own suggestion):

```rust
trigger.marked_for_drop || trigger.original.as_ref().is_none_or(|original| has_trigger_change(trigger, original))
```

I have deliberately not touched `triggers.rs` here, since it is unrelated to this PR's scope — happy to include the fix if you would rather have it land in this branch. Everything else on the current head is green: `rust-test`, `frontend`, `changes`, `label`, `autofill`.

Thanks for the hardening patch, by the way — the reusable-session rollback gap and the `SELECT INTO` hole in the explain safety gate were both real and both mine. The `Merge Append` classification too: I had ordered joins ahead of the DML rule to protect `Merge Join`, but the `merge\b` prefix still caught `Merge Append`; `^merge$` is the right narrowing.

t8y2:
Thanks for the contribution! Merged in ed1e36912, will be released in the next version.

PR #5165, opened 2026-08-02
fix: 保留 SQL Server 批处理结果集

### LINKED ISSUE #4371: [Compatibility] 批处理脚本在工具中执行成功但未返回结果集
### 来源
QQ 群反馈 · **反馈人**：自由自在

### 数据库类型和版本
Microsoft SQL Server

### AI 摘要
同一个批处理脚本在MSSQL中原语句能执行成功并返回完整结果集，但在某工具中执行时，初始报错，调整后虽执行成功，却只返回执行结果信息，未返回数据列表。

### 原始描述
同一个批处理脚本执行报错且调整后执行成功也只返回执行结果未有返回结果数据列表，但是在MSSQL中原语句都能执行成功且返回结果集

zipg:
## 问题

SQL Server 批处理以 `SET`、`DECLARE` 等准备语句开头、后续再执行 `SELECT` 时，单查询入口没有复用已有的批处理结果集判定，而是走 `Client::execute`。该路径只汇总 affected rows，不读取 TDS 结果集，因此批处理执行成功后只显示执行信息。

## 修复

- 单查询入口统一使用 `sqlserver_batch_can_use_execute` 判断是否可走纯执行路径。
- 可能返回结果集的批处理复用现有 `execute_simple_batch_with_max_rows`，返回首个结果集。
- 纯 DML 继续使用原路径统计 affected rows；分页上限、截断状态和 SQL Server 消息处理保持现有行为。
- 增加分流单测和真实 SQL Server 集成测试。

## 验证

- 最新 `main` + 真实 SQL Server：`SET NOCOUNT OFF; DECLARE ...; INSERT ...; SELECT ...` 经 `/api/query/execute` 返回空列、空行和 `affected_rows: 4`，稳定复现。
- 修复分支同一真实库、同一 API 请求：返回 `batch_value` 列以及 `1`、`2` 两行。
- 真实库集成测试同时验证 `maxRows=1` 时只返回一行且 `truncated=true`。
- SQL Server 模块单测：90 passed，4 ignored。
- `cargo fmt --check --all`、`cargo clippy -p dbx-core --lib --tests -- -D warnings`、`git diff --check` 通过。

Fixes #4371

### DIFF
--- a/crates/dbx-core/src/db/sqlserver.rs
+++ b/crates/dbx-core/src/db/sqlserver.rs
@@ -2497,8 +2497,7 @@ pub async fn execute_query_with_max_rows(
         strip_dbx_sqlserver_row_number_column(&mut result, sql);
         Ok(result)
     } else if !sqlserver_batch_can_use_execute(sql) {
-        let mut results = execute_simple_batch_with_max_rows(client, sql, max_rows).await?;
-        Ok(results.remove(0))
+        execute_simple_batch_first_result_with_max_rows(client, sql, max_rows).await
     } else {
         let (result, messages) = capture_sqlserver_messages(sqlserver_driver_result(client.execute(sql, &[]))).await;
         let result = result?;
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch `6698f04f4` has been pushed to this PR.

Root causes:
- `INSERT ... SELECT` was classified as result-set-producing solely because it contained a top-level `SELECT`, so the normal SQL Server execute path no longer returned the real affected-row count.
- the single-result API collected every later result set before discarding them, causing unnecessary memory growth for large SQL Server batches.

Patch:
- statement-aware routing keeps DML-leading `INSERT/UPDATE/DELETE/MERGE` on the affected-row path while still routing genuinely mixed batches through result-set handling.
- the single-result path now retains only the first result set while fully draining the remaining TDS stream, preserving connection usability without accumulating discarded rows.
- regression coverage includes `INSERT ... SELECT` affected rows and a follow-up query after draining a large later result.

DBeaver comparison: its JDBC execution loop reads each update count separately and repeatedly advances through `nextResults()`, so later results are drained even when only one result is presented.

Checks run:
- SQL Server module tests: 90 passed / 4 ignored
- 3 live integration tests compiled; they were not executed because SQL Server test environment variables are not configured locally
- `cargo check` passed
- Clippy with `-D warnings` passed
- `cargo fmt --check` passed
- `git diff --check` passed

t8y2 (approved):
Approved after maintainer patch validation. The exact-head Rust tests passed; the Clippy failure is the pre-existing manual_async_fn failure already reproduced on main at 43b24c43d.

t8y2:
Thanks for the contribution! Merged in 02bad5d75, will be released in the next version.

PR #5222, opened 2026-08-03
feat(redis): add decompressed value view for compressed values

### LINKED ISSUE #4248: [Feature] redis新增展示压缩数据的视图
### 数据库类型和版本

通用

### 支持信息

DBX 版本: v0.5.64
运行环境: 桌面端
操作系统: macOS 26.5.2
系统架构: aarch64

### 当前痛点

如果redis中存储了压缩后的数据，不能直观查看内容

Image

### 期望方案

可以参考 another redis desktop的方式

Image

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

Abeautifulsnow:
## Summary

Add a read-only **decompressed value view** for Redis values, so users can read
gzip/zlib/deflate-compressed business content (e.g. gzip caches) directly in the
value viewer and member detail sheet instead of inspecting raw bytes.

Closes #4248

## What changed

- **New `redisCompression` module** (`apps/desktop/src/lib/redis/redisCompression.ts`)
  - `decompressRedisValue()` — gzip magic (`1f 8b`) detection → zlib (`DecompressionStream("deflate")`) → raw deflate (`deflate-raw`) fallback, using the browser-native Compression Streams API (no new dependencies)
  - Returns the algorithm that actually succeeded so the UI can label it
  - **50 MiB streaming output cap** (`REDIS_DECOMPRESS_MAX_OUTPUT_BYTES`) via a counting TransformStream — the primary zip-bomb defense; over-limit cancels immediately and keeps no partial content
  - Corrupt data maps to a clean `corrupt` / `limit` / `unsupported` result instead of throwing

- **`RedisValueFormat` extended with `"decompressed"`** — Decompressed appears as the last tab in the existing format picker (STRING value + member detail sheet)

- **Gzip badge** — a `1f 8b` header check shows a one-click "Gzip" badge on the STRING toolbar and the member-detail title bar. The badge never auto-decompresses and is never hidden for large inputs.

- **Failure behavior** — a failed or non-compressed value keeps the original content visible with a non-blocking notice ("Unable to decompress with the selected format" / 50 MiB-limit message). It never replaces the content with an error string and never fakes success.

- **JSON chaining** — decompressed output that is valid JSON renders via the existing `JsonTree` viewer.

- **Copy semantics** — while the Decompressed view is active, the Copy buttons copy the decompressed text (raw value copy remains available on other views).

- **i18n** — 6 new keys across all 8 locales.

## Why this approach

- **Backend untouched** — the frontend already has lossless raw bytes (`RedisBlob.raw_base64`), so no Redis read-protocol changes are needed.
- **Zero new dependencies** — `DecompressionStream` supports gzip/zlib/deflate-raw on WebView2 (Windows/Linux) and WKWebView (macOS 13.3+). Brotli/zstd are not viable cross-platform in the browser; zstd/LZ4/Snappy are deferred for a follow-up evaluation.
- **Read-only** — the decompressed view never writes back, so the original compressed bytes and write semantics are unchanged.

## Testing

- 12 new unit tests for `redisCompression.ts` (gzip/zlib/raw-deflate success, corrupt input, output-cap enforcement, unsupported runtime, empty input)
- Existing redis suites green (132 tests)
- `oxlint` clean; `vue-tsc` clean (no new type errors)
- Note: tests run a spec-faithful `DecompressionStream` mock because Node 22's native adapter crashes the process on corrupt input even when the reader error is caught — the production browser engines follow the spec contract.

## Out of scope (follow-ups)

- fflate fallback for macOS < 13.3 (no `DecompressionStream`) — currently surfaces a non-blocking "not supported" notice
- zstd / LZ4 / Snappy decompression
- Editable/write-back of decompressed values

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

c8ac297f-0429-4ee7-9ea6-fd6009efe6ae

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

Related #4248


t8y2 (changes requested):
There are two correctness/resource-safety issues to address before merging:

1. `apps/desktop/src/lib/redis/redisCompression.ts:95-106` automatically retries every non-gzip value as raw DEFLATE after zlib fails. Raw DEFLATE has no framing or checksum, so arbitrary Redis binary data can accidentally be accepted and displayed, searched, or copied as incorrect decompressed content. Please remove raw DEFLATE from automatic detection, or require an explicit user-selected algorithm, and add an ambiguous-binary regression test.

2. `apps/desktop/src/lib/redis/redisCompression.ts:34-61` applies the 50 MiB counter downstream of `DecompressionStream` while writing the complete compressed value as one chunk. That limits delivered output, but does not prove that the decompressor cannot allocate substantially more output internally first. Please use a decompression path with an enforceable output/memory limit and add a platform-representative high-compression-ratio test.

Abeautifulsnow:
> There are two correctness/resource-safety issues to address before merging:
> 
> 1. `apps/desktop/src/lib/redis/redisCompression.ts:95-106` automatically retries every non-gzip value as raw DEFLATE after zlib fails. Raw DEFLATE has no framing or checksum, so arbitrary Redis binary data can accidentally be accepted and displayed, searched, or copied as incorrect decompressed content. Please remove raw DEFLATE from automatic detection, or require an explicit user-selected algorithm, and add an ambiguous-binary regression test.
> 2. `apps/desktop/src/lib/redis/redisCompression.ts:34-61` applies the 50 MiB counter downstream of `DecompressionStream` while writing the complete compressed value as one chunk. That limits delivered output, but does not prove that the decompressor cannot allocate substantially more output internally first. Please use a decompression path with an enforceable output/memory limit and add a platform-representative high-compression-ratio test.

All comments have been addressed.

t8y2 (commented):
Maintainer patch added on top of `fe2976f8` as `e11a58b1`.

Root cause: regenerating `pnpm-lock.yaml` replaced 12 workspace-local CLI/MCP links with registry packages and refreshed unrelated snapshots, which triggered the Nix minimum-release-age failure.

Patch: restored the lockfile to the PR base and retained only the intended `pako@3.0.1` importer, package, and snapshot entries. The dependency ordering in `package.json` is restored as well.

Validation:
- `pnpm install --lockfile-only --frozen-lockfile --ignore-scripts`
- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm exec vitest run apps/desktop/src/lib/redis/__tests__/redisCompression.spec.ts` (15 passed)
- `git diff --check`

t8y2 (commented):
Follow-up maintainer update on the current head `e42ac9c6`.

The refreshed Nix job proved that the lockfile drift was fixed, then reported the expected fixed-output change caused by the intended `pako` dependency: `sha256-h4KlJml4J7nIyPQRcpSJ1b3buwt81B3gIaL5Ekj38xo=`. The branch is now merged with the latest `main` to avoid a same-line `flake.nix` conflict, and that exact hash is committed.

Revalidation on the updated base:
- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm exec vitest run apps/desktop/src/lib/redis/__tests__/redisCompression.spec.ts` (15 passed)
- PR diff retains only the Redis feature, `pako` lock entries, and the Nix hash update
- `git diff --check`

t8y2 (approved):
Maintainer patches and the refreshed Nix closure are validated on this exact head; all current checks pass.

t8y2:
Thanks for the contribution! Merged in ec44e09e, will be released in the next version.

PR #5230, opened 2026-08-03
fix(xugu): recover metadata access after ordinary-user EOF

mapan0424:
## Summary
- recover Xugu metadata browsing for ordinary users when denied catalog queries surface as EOF
- keep the low-privilege USER_TABLES/USER_VIEWS fallback for table and view visibility
- query object families independently so one restricted ALL_* catalog does not hide accessible metadata
- add EOF and metadata-root regression coverage

## Validation
- `go test -count=1 ./...`
- `go test -count=1 -race ./...`
- `go build -o /tmp/dbx-xugu-agent-test .`
- Live APP_TEST validation against `SHOP_DEMO` on Xugu: databases, schemas, tables, objects, columns, indexes, constraints, foreign keys, triggers, partitions, and subpartitions returned without EOF/RPC errors.

This change is limited to `agents/drivers/xugu` and does not modify other database drivers or database data.


t8y2 (commented):
Maintainer patch added on top of `3614f318` as `d8019dea`.

Root causes: the EOF helper replayed ordinary user queries, reconnects lost the live database selected with `USE`, and the per-source object fallback ignored requested object families and omitted synonyms. The unconstrained fallback could also return low-privilege tables early and hide other accessible object families.

Patch: limited replay to metadata paths, restored the selected database before retrying metadata, normalized and respected object-type constraints, included `SYNONYM`, combined low-privilege tables with independently accessible object families, and preserved global pagination semantics.

Validation:
- `go test -count=1 ./...`
- `go test -race -count=1 ./...`
- `go build -o /tmp/dbx-xugu-pr5230 ./...`
- `git diff --check`

t8y2:
Thanks for the contribution! Merged in 06838c42, will be released in the next version.

PR #5237, opened 2026-08-03
feat: add MySQL-compatible types for GaussDB M-mode and migrate JDBC …

AndrewDi:
…plugin to Gradle

- Add MySQL-compatible types (MEDIUMINT, FLOAT, DOUBLE, TINYTEXT, MEDIUMTEXT, LONGTEXT, TINYBLOB, MEDIUMBLOB, LONGBLOB, DATETIME, YEAR, ENUM, SET, GEOMETRY) to gaussdb.yaml for M-mode support
- Make DECIMAL and NUMERIC both primary types with mutual aliases to support both A-mode and M-mode
- Replace Maven (pom.xml) with Gradle build for JDBC plugin
- Use Gradle 9.5.0 (matching agents/ convention) with Shadow plugin for fat JAR packaging


AndrewDi:
@t8y2 PTAL

t8y2 (commented):
Maintainer patch added on top of `6b44abd9` as `ebd9d5fd`.

Root causes: the Gradle shadow JAR name did not match the packaging scripts, Gradle selected SLF4J 2.0.18 instead of the Maven-era 1.7.36 API, the JDBC CI job still requested Maven caching after removing `pom.xml`, and the packaging version parser was not portable to macOS `sed`. The generic GaussDB dialect also advertised `ENUM` and `SET` outside a mode-aware type boundary.

Patch: restored the expected unversioned shadow JAR, strictly constrained SLF4J API 1.7.36, moved CI to Java 21 with Gradle caching, made version parsing POSIX-portable, and removed the unsafe generic `ENUM`/`SET` entries.

Validation:
- `./gradlew test shadowJar`
- dependency insight resolves `slf4j-api` to 1.7.36
- `./package.sh`
- verified `build/libs/dbx-jdbc-plugin-all.jar` and `dist/dbx-jdbc-plugin-0.1.26.zip`
- `git diff --check`

t8y2 (commented):
Follow-up maintainer patch on the current head `21559357`.

The approved CI run reached the JDBC version guard and proved another Maven-to-Gradle migration gap: both the CI guard and release automation still read or copied `plugins/jdbc/pom.xml`.

Patch:
- read and update the plugin version in `build.gradle`
- keep `manifest.json` synchronized as the explicit version-change signal
- update release commit, Java 21/Gradle cache, version read, and asset packaging paths
- add focused Node tests for Gradle parsing, mismatch detection, automatic patch bumps, and explicit bumps

Validation:
- `node --test .github/scripts/jdbc-plugin-version.test.mjs` (4 passed)
- `node .github/scripts/check-jdbc-plugin-version.mjs HEAD~0 HEAD`
- `./plugins/jdbc/gradlew -p plugins/jdbc test shadowJar`
- `./plugins/jdbc/package.sh`
- release workflow YAML parse and `git diff --check`

t8y2 (commented):
The current head `ee9a3452` also migrates the existing `packages/app-tests/jdbcPluginRelease.test.ts` coverage from the removed Maven inputs to Gradle inputs. The focused file now passes all 9 tests locally, and no `pom.xml` assumptions remain in that coverage.

t8y2:
Thanks for the contribution! Merged in 1dd005db, will be released in the next version.

PR #5247, opened 2026-08-03
Feat/gaussdb multi host

AndrewDi:
GaussDB支持添加多主机地址
image


AndrewDi:
@zipg PTAL

t8y2 (dismissed):
request changes: the multi-host representation still has three correctness/security issues.

1. `crates/dbx-core/src/models/connection.rs:983` decides whether ports are embedded solely by checking for a comma. A valid single-node host-list value such as `db.example.com:5432` has no comma, so serialization appends the scalar port again and produces an invalid endpoint. Please parse host entries structurally instead of using comma presence as the representation test, and cover one-node plus multi-node forms.

2. `crates/dbx-core/src/connection.rs:4639` probes only the first node. This makes connection testing fail whenever the first node is down even though a later node is reachable, defeating the failover semantics that this PR adds. Please probe the nodes using the same failover order/timeout policy as the real connection path and add a first-node-down regression test.

3. `apps/desktop/src/lib/connection/connectionPresentation.ts:34` redacts each hostname but returns every real per-node port unchanged. The single-host redacted endpoint deliberately hides the port, so the multi-host path now leaks information that the redacted UI is expected to conceal. Please replace each embedded port with the redacted marker and cover IPv4, hostname, and bracketed IPv6 lists.

AndrewDi:
@t8y2 already fix issue

t8y2 (commented):
Maintainer patch `f8603c37b` fixes single-host serialization while preserving multi-host support.

Root cause: the UI always embedded the port in `host`, while core only treated comma-separated values as multi-host. A single `db.example.com:5432` was therefore parsed as an IPv6-like host and received a second port.

Fix: single-node configs keep host and port separate; legacy single `host:port` values are normalized; multi-host values retain embedded ports; bracketed IPv6 is handled consistently by form parsing, URL creation, endpoint probing, and presentation.

Checks: frontend tests (14), `pnpm typecheck`, focused Rust GaussDB URL/probe tests (4), `cargo fmt --check`, and `git diff --check` passed.

t8y2 (commented):
Maintainer follow-up: fixed the CI formatting failure only. The previously reviewed behavior and targeted validation are unchanged, and formatter checks now pass on this exact head.

t8y2:
Thanks for the contribution! Merged in 336fbe094, will be released in the next version.

PR #5259, opened 2026-08-03
fix(grid): strip thousand separators in numeric grid cells

### LINKED ISSUE #5210: [Bug] 网格数据编辑无法识别千分位
### 数据库类型和版本

SQL Server

### 问题描述

查询窗口输入select * from tablename
表中列类型是decimal(18,2)
粘贴一个带千分位的数值文本，比如10,000.00
提交保存报错
Token error: '从数据类型 nvarchar 转换为 numeric 时出错。' on server xxx executing  on line 1 (code: 8114, state: 5, class: 16)
Statement 1 failed; previous 0 statement(s) may have been committed.
SQL text omitted from user-facing error; enable debug SQL diagnostics for a redacted statement.

如果字段类型是数值型的，可以自动去除千分位的逗号，从Excel复制粘贴过来的容易带上千分位

### 支持信息

DBX 版本: v0.5.74
运行环境: 桌面端
操作系统: Windows 10.0.26200.8875
系统架构: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

dienaso:
Closes #5210

## Problem

Pasting a value with thousands separators (e.g. `10,000.00`) from Excel into a numeric grid cell fails to save — the value is sent to the server as the literal `10,000.00` and SQL Server rejects it:

> Token error: '从数据类型 nvarchar 转换为 numeric 时出错'

Root cause: `Number("10,000.00")` returns `NaN` in `coerceDataGridCellValue`, so the string is passed through untouched instead of being coerced to a number.

## Change

`dataGridCellCoercion.ts` — when the target column is numeric and the value is a valid thousand-grouped number, strip the separators before normal coercion. Non-numeric columns and invalid groupings (`1,23`) are untouched.

## Test

Added 4 cases to `dataGridCellCoercion.spec.ts` (separator stripping, negative values, non-numeric column preservation, invalid grouping). Watched the stripping test fail before the fix, then pass after.

- `pnpm vitest run apps/desktop/src/lib/__tests__/dataGrid/` → 236 passed (only `geometryProjection.spec.ts` fails, pre-existing missing `proj4` dependency)
- `pnpm oxlint` clean
- `pnpm typecheck` clean except pre-existing missing `proj4` module

### DIFF
--- a/apps/desktop/src/lib/__tests__/dataGrid/dataGridCellCoercion.spec.ts
+++ b/apps/desktop/src/lib/__tests__/dataGrid/dataGridCellCoercion.spec.ts
@@ -65,4 +65,55 @@ describe("coerceDataGridCellValue", () => {
     expect(coerceDataGridCellValue(options)).toBeNull();
     expect(coerceDataGridCellValue({ ...options, preserveEmptyString: true })).toBe("");
   });
+
+  it("strips thousands separators before numeric coercion", () => {
+    expect(
+      coerceDataGridCellValue({
+        value: "1,234.50",
+        oldValue: 1234.5,
+        databaseType: "sqlserver",
+        columnInfo: { data_type: "float" },
+      }),
+    ).toBe(1234.5);
+
+    expect(
+      coerceDataGridCellValue({
+        value: "10,000",
+        oldValue: 10000,
+        databaseType: "mysql",
+        columnInfo: { data_type: "int" },
+      }),
+    ).toBe(10000);
+
+    expect(
+      coerceDataGridCellValue({
+        value: "-10,000.00",
+        oldValue: "-10000.00",
+        databaseType: "sqlserver",
+        columnInfo: { data_type: "decimal(18,2)" },
+      }),
+    ).toBe("-10000.00");
+  });
+
+  it("does not strip commas when the column is not numeric", () => {
+    expect(
+      coerceDataGridCellValue({
+        value: "10,000.00",
+        oldValue: "10,000.00",
+        databaseType: "sqlserver",
+        columnInfo: { data_type: "varchar(255)" },
+      }),
+    ).toBe("10,000.00");
+  });
+
+  it("leaves invalid thousand-grouping values untouched", () => {
+    expect(
+      coerceDataGridCellValue({
+        value: "1,23",
+        oldValue: 123,
+        databaseType: "sqlserver",
+        columnInfo: { data_type: "decimal(18,2)" },
+      }),
+    ).toBe("1,23");
+  });
 });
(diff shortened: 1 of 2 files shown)


t8y2 (changes requested):
request changes: there are two correctness blockers in the coercion path.

1. `apps/desktop/src/lib/dataGrid/dataGridCellCoercion.ts:16` treats every `1,000`-shaped value in a numeric column as a grouped number. In decimal-comma locales, `10,000` can mean decimal `10.000`, so pasting it now silently stores `10000`. The grid has no locale contract that makes this transformation safe. Please make separator handling explicit/locale-aware, or avoid silently removing commas from ambiguous input.

2. The code rewrites `value` before the existing precision-preservation checks run. A value such as `9,007,199,254,740,993` becomes a JavaScript `number` and can lose precision, while `shouldPreserveNumericText(...)` no longer sees the original formatted literal. Please preserve exact integer/decimal text before numeric conversion and add regression coverage for grouped values beyond `Number.MAX_SAFE_INTEGER` and exact decimal columns.

Also, please reuse the existing numeric type classification helpers instead of maintaining a second, divergent regex in `isNumericColumn`; otherwise type coverage and precision behavior will drift again.

dienaso:
Thanks for the review — all three points addressed:

**1. Locale ambiguity** — A lone `1,000` (single comma group, no decimal point) is now left untouched: in comma-decimal locales it reads as 1.000, so silently stripping it would corrupt data. Commas are only stripped when the grouping is unambiguous — a decimal point is present (`1,234.56`) or there are multiple comma groups (`1,234,567`). Added `leaves ambiguous single-group values untouched`.

**2. Precision** — The normalized (comma-stripped) text is now threaded through `shouldPreserveNumericText`, so the precision checks see the numeric literal instead of the original formatted string. Regression coverage added: `9,007,199,254,740,993` (beyond Number.MAX_SAFE_INTEGER) in a bigint column is kept as exact text `"9007199254740993"`; a grouped decimal `10,000.00` in a decimal(18,2) column is kept as `"10000.00"` text instead of being rounded by JS Number.

**3. Reuse** — Removed the hand-rolled `isNumericColumn` regex and switched to the shared `isNumericColumnType` (`dataGridColumnType.ts`, already used by xlsxExport etc.), so type coverage stays consistent with the rest of the grid.

Verification: `pnpm vitest run apps/desktop/src/lib/__tests__/dataGrid/` → 247 passed (the one failing file is the pre-existing missing `proj4` dependency); oxlint clean; typecheck has no errors in the changed files (remaining ones are all missing dependencies in this environment).

t8y2 (commented):
Maintainer patch pushed as `f74cfe666`.

Root cause: grouped-number normalization only accepted a plain mantissa, while the existing canonical numeric parser already supports an exponent. Values such as `1,234.50e2` therefore kept the comma and followed the original conversion-error path.

Patch: split the optional exponent from the mantissa, normalize only an unambiguous grouped mantissa, and preserve the ambiguity guard for values such as `1,000e3`.

Checks: 15 focused Vitest tests, `pnpm typecheck`, changed-file `oxfmt --check`, `oxlint`, and `git diff --check` passed.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 343ef6fa8, will be released in the next version.

PR #5293, opened 2026-08-04
feat(agent): add get_current_time tool to AI assistant

Abeautifulsnow:
## 变更说明

Adds a DB-independent `get_current_time` tool to the AI assistant so the model can resolve relative time expressions ("last 7 days", "yesterday", "this month") into concrete dates when building SQL. Previously the agent had no time awareness at all — neither a tool nor a current-date injection in the system prompt — so it could only guess or hardcode wrong absolute dates.

The tool returns a rich payload so the model can pick whichever representation fits:

```json
{
  "utc": "2026-08-04T12:34:56Z",
  "local": "2026-08-04T20:34:56+08:00",
  "utc_offset_minutes": 480,
  "readable": "2026-08-04 20:34:56 (UTC+08:00)"
}
```

### Changes

**Backend — `crates/dbx-core/src/agent_tools.rs`**
- New `get_current_time` tool definition (`read_only = true`, `parallel_ok = true`, no parameters).
- New `execute_get_current_time()` helper returning the payload above via `chrono::Utc::now()` / `chrono::Local::now()`.
- Dispatch arm wired into `execute_tool` before the unknown-tool catc
- Registered in **all four** tool-set branches (`read_only_tools` + `all_tools` × SQL + vector), so it's available in every mode (Agent + Ask) and every DB type. It is automatically reachable on Desktop, Web, and MCP through the shared dispatch.

**Frontend — `apps/desktop/src/lib/ai/ai.ts`**
- `get_current_time` added to the Agent-mode tool list in the SQL prompt (`buildModePromptLines`) and the vector prompt (`buildVectorModePromptLines`), in both zh-CN and
en.

### Design decisions

- **Tool-only, no system-prompt injection** — injected timestamps go l returns a fresh value on each call.
- **Rich payload (UTC + local + offset)** — covers the web/MCP case where the server timezone may differ from the user's; UTC is always correct and the offset disambiguates.
- **No separate "compute time range" tool** — range math is delegated to SQL date functions (`NOW() - INTERVAL 7 DAY`) or done by the model after reading `now`.

### Tests

- `cargo test -p dbx-core --lib agent_tools`: **42 passed / 0 failed**.
- 7 new tests: presence in all 4 tool sets (Postgres + Qdrant), `read_only` / `parallel_ok` flags, valid RFC3339 output within tolerance, UTC-offset consistency, and end-to-end `execute_tool` dispatch.
- 2 existing vector tests switched from exact-`vec` to `contains` assfuture tool additions.

### Notes

- `cargo clippy` / full build cannot run locally on this Windows machine due to a pre-existing `openssl-sys` build issue (unrelated to this change). CI is the source of truth.
- A spec note ("AI Agent Tool Registration") was added under `.trellis/spec/` (local-only, gitignored): DB-independent tools must be registered in all 4 branches and listed in the frontend prompt strings (zh + en).

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

### DIFF
--- a/apps/desktop/src/lib/ai/ai.ts
+++ b/apps/desktop/src/lib/ai/ai.ts
@@ -351,24 +351,27 @@ function buildVectorSystemPrompt(context: AiContext, mode: AiAssistantMode, cust
 }
 
 function buildVectorModePromptLines(context: AiContext, mode: AiAssistantMode, isZh: boolean): string[] {
+  const currentTimeGuidance = currentTimeToolGuidance();
   if (mode === "agent") {
-    return [isZh ? "你处于 Agent 模式。你有以下工具可用：list_collections、browse_collection、get_current_time。" : "You are in Agent mode. You have the following tools available: list_collections, browse_collection, get_current_time."];
+    return [isZh ? "你处于 Agent 模式。你有以下工具可用：list_collections、browse_collection、get_current_time。" : "You are in Agent mode. You have the following tools available: list_collections, browse_collection, get_current_time.", currentTimeGuidance];
   }
   return [
     isZh
       ? `你处于 Ask 模式。你只能使用 list_collections 确认集合清单；不要浏览集合数据。${dbLabel(context.databaseType)} 的查询格式为 REST API（METHOD /path + JSON body），具体格式因数据库类型而异。只生成查询请求文本和说明，不要暗示已经执行。`
       : `You are in Ask mode. You may only use list_collections to inspect collection names; do not browse collection data. ${dbLabel(context.databaseType)} uses a REST API query format (METHOD /path + JSON body) that varies by database type. Generate query strings and explanations only; do not imply execution.`,
-    "When a request needs a relative time expression such as yesterday, last 7 days, this month, or today resolved into a concrete filter, call get_current_time first; do not guess the current date or timezone.",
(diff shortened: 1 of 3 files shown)


t8y2 (commented):
Maintainer patch `1647132d8` fixes the timezone boundary.

Root cause: `chrono::Local::now()` resolved “local” time from the backend process, which can differ from the desktop client and database session timezone.

Fix: the client timezone name and current UTC offset are now passed explicitly to `get_current_time`; the backend converts from UTC using that offset and safely falls back to UTC when context is absent.

Checks: AI prompt tests (15) passed, focused Rust current-time tests (3) passed, `pnpm typecheck`, `cargo fmt --check`, and `git diff --check` passed.

t8y2 (commented):
Maintainer follow-up: fixed the CI formatting failure only. The previously reviewed behavior and targeted validation are unchanged, and formatter checks now pass on this exact head.

t8y2:
Thanks for the contribution! Merged in c0a6d6f2f, will be released in the next version.

PR #5336, opened 2026-08-04
feat(influxdb3): add InfluxDB 3.x engine with Flight SQL and HTTP tra…

Clarence404:
## 变更说明

Add InfluxDB 3.x (Core) as a first-class engine, sharing the InfluxDB connection card with 1.x/2.x via a version picker.

- New native driver `crates/dbx-core/src/db/influxdb3_driver.rs`, two transports selectable via `external_config.transport`:
  - `flightsql` (default): SQL over Arrow Flight SQL (`arrow-flight` + `tonic`, no `protoc` build dep needed — `arrow-flight` ships pre-generated protobuf).
  - `http`: SQL over `POST /api/v3/query_sql` (JSON response), useful when Flight SQL is unavailable or firewalled.
- `list_databases` always goes through the admin HTTP API (`GET /api/v3/configure/database?format=json`) — Core's SQL dialect does not accept `SHOW DATABASES`.
- `list_tables` filters `table_schema == "iox"` so `system` / `information_schema` internals don't clutter the sidebar.
- Auth: `Authorization: Bearer <token>` from `config.password`. Empty password skips auth (works with `--without-auth` Core deployments).
- Backend wiring: new `DatabaseType::InfluxDb3`, `PoolKind::InfluxDb3`, `KeepaliveTarget::InfluxDb3`; SQL dialect / DDL / admin / query-execution branches added parallel to `InfluxDb`.
- Driver manifest entry with `runtimeMode: "native"`, default port 8181.
- Frontend: InfluxDB card gains a 3rd version option (1.x / 2.x / 3.x); selecting 3.x switches port default to 8181 and shows database + token fields. Save-side translates `db_type` from `influxdb` to `influxdb3`; load-side reverses so the same card round-trips.
- Table sub-tree hides 索引/外键/触发器 for `influxdb3` (mirrors existing `influxdb`).

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

- [x] 本 PR 涉及前端改动，已附截图（连接对话框版本选择 → 3.x；侧栏显示 mydb 下的 cpu/mem/sensor）

图像 (8)
图像 (9)
图像 (10)
图像 (11)

## 验证

- `cargo check -p dbx-core --lib`（含 default features）通过
- `cargo check -p dbx --lib`（Tauri 层）通过
- `cargo test -p dbx-core --lib db::influxdb3_driver` 5/5 通过
- `pnpm typecheck` 通过
- 端到端：
  - 启动 InfluxDB 3.10.3 Core（`--without-auth`）
  - dbx 连接 → 侧栏正确显示 `mydb` + 三张 `iox` 表（`cpu`/`mem`/`sensor`）
  - `SELECT * FROM cpu ORDER BY time DESC LIMIT 100` 正常返回

- [x] `cargo check -p dbx-core --lib` 通过
- [x] `pnpm typecheck` 通过
- [x] 相关测试通过

## 关联 Issue

…nsports

## 变更类型

- [ ] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过


t8y2 (changes requested):
Request changes. The current implementation still has three blocking correctness/resource issues:

1. Database creation is exposed even though InfluxDB 3.x does not support that operation through this path, so the UI/API advertises behavior that cannot succeed.
2. `max_rows` is applied only after the complete Flight result has been downloaded and materialized. Large queries therefore still consume unbounded network and memory before truncation.
3. The connection test does not complete a real Flight SQL handshake/query, so configurations can pass testing and fail on their first actual query.

Please remove or explicitly reject unsupported creation, enforce the row limit while streaming batches, and make connection testing validate the same Flight SQL path used by execution. Add focused coverage before merging; this PR also needs another dependency/architecture pass after those changes.

Clarence404:
> Request changes. The current implementation still has three blocking correctness/resource issues:
> 
> 1. Database creation is exposed even though InfluxDB 3.x does not support that operation through this path, so the UI/API advertises behavior that cannot succeed.
> 2. `max_rows` is applied only after the complete Flight result has been downloaded and materialized. Large queries therefore still consume unbounded network and memory before truncation.
> 3. The connection test does not complete a real Flight SQL handshake/query, so configurations can pass testing and fail on their first actual query.
> 
> Please remove or explicitly reject unsupported creation, enforce the row limit while streaming batches, and make connection testing validate the same Flight SQL path used by execution. Add focused coverage before merging; this PR also needs another dependency/architecture pass after those changes.

Thanks — will address all three.

One question first, tied to your dep/architecture note:
I'd like to drop Flight SQL entirely and keep only the
HTTP transport (POST /api/v3/query_sql). Rationale:

- Drops 6 heavy deps: tonic, arrow-flight, arrow-array,
  arrow-schema, arrow-cast, prost
- 2's streaming truncation becomes moot — HTTP responses
  are atomic, and dbx metadata queries already carry LIMIT
- 3 collapses to a single-path test
- Flight SQL can land as an opt-in follow-up once a
  concrete analytical use case surfaces

Would that work, or should I keep both transports?

Clarence404:
> Thanks — will address all three.
> 
> One question first, tied to your dep/architecture note: I'd like to drop Flight SQL entirely and keep only the HTTP transport (POST /api/v3/query_sql). Rationale:
> 
> * Drops 6 heavy deps: tonic, arrow-flight, arrow-array,
>   arrow-schema, arrow-cast, prost
> * 2's streaming truncation becomes moot — HTTP responses
>   are atomic, and dbx metadata queries already carry LIMIT
> * 3 collapses to a single-path test
> * Flight SQL can land as an opt-in follow-up once a
>   concrete analytical use case surfaces
> 
> Would that work, or should I keep both transports?

Thanks — will address all three.

One question first, tied to your dep/architecture note:
I'd like to drop Flight SQL entirely and keep only the
HTTP transport (POST /api/v3/query_sql). Rationale:

Drops 6 heavy deps: tonic, arrow-flight, arrow-array,
arrow-schema, arrow-cast, prost
2's streaming truncation becomes moot — HTTP responses
are atomic, and dbx metadata queries already carry LIMIT
3 collapses to a single-path test
Flight SQL can land as an opt-in follow-up once a
concrete analytical use case surfaces
Would that work, or should I keep both transports?

Clarence404:
Superseded by https://github.com/t8y2/dbx/pull/7744. That branch drops Flight SQL per
the dep/architecture concern and addresses the three blocking issues.

PR #5338, opened 2026-08-04
fix(xugu): return EXPLAIN statements as query results

mapan0424:
## Summary

Xugu `EXPLAIN` and `EXPLAIN VERBOSE` statements were incorrectly routed through `ExecContext`, so the query editor reported successful execution with no result rows. This change routes Xugu `EXPLAIN` statements through the existing result-set path.

## Changes

- Recognize `EXPLAIN` as a query statement in the Xugu agent.
- Preserve support for `EXPLAIN VERBOSE` and other statements beginning with the `EXPLAIN` keyword.
- Add regression coverage for direct and paged query execution.
- Add keyword-boundary coverage so identifiers such as `EXPLAINATION` are not misclassified.

## Validation

- Manually validated the native Xugu agent against `SHOP_DEMO` with the `APP_TEST` user:
  - `EXPLAIN SELECT 1` returned a `PLAN_PATH` result set.
  - `EXPLAIN VERBOSE SELECT 1` returned a `PLAN_PATH` result set including verbose plan text.

The change is isolated to the Xugu agent; no other database agents or frontend code are modified.


t8y2:
Thanks for the contribution! Merged in 76501d6ae, will be released in the next version.

PR #5361, opened 2026-08-04
fix: prevent IF...BEGIN...END blocks from being split in SQL Server scripts

EstoyAburridowGH:
## Problem

SQL Server scripts containing `IF...BEGIN...END` blocks are incorrectly split into individual statements by the SQL splitter. This causes `Incorrect syntax near 'NULL'` errors when executing scripts that use conditional blocks.

**Example that fails:**
```sql
IF OBJECT_ID ('CATEGORIA') IS NOT NULL
BEGIN
DROP TABLE CATEGORIA
END
CREATE TABLE CATEGORIA (...)
```

The splitter treats `BEGIN`, `DROP`, `END` etc. as separate statements because they appear at the start of a line after a newline.

## Solution

Added detection of T-SQL `IF` blocks in both:

- **Rust backend** (`crates/dbx-core/src/sql.rs`): `starts_with_tsql_if_block()` function detects `IF` keyword at start of statement text and guards in `split_statement_range_at_blank_lines`
- **TypeScript frontend** (`apps/desktop/src/lib/sql/sqlStatementRanges.ts`): `startsWithSqlServerTsqlBlock()` function with regex detection, guard in `splitStatementRangeAtSoftStarts`

The guard returns the entire statement as-is (no splitting) when the statement starts with an `IF` block.

## Tests

- Rust: `sqlserver_keeps_tsql_if_begin_end_block_together`
- TypeScript: 3 tests covering `splitSqlStatementRanges`, `executableStatementRanges`, and `statementRangeAtCursor`

All 147 TS tests and Rust unit tests pass.

### DIFF
--- a/apps/desktop/src/lib/__tests__/sql/sqlStatementRanges.spec.ts
+++ b/apps/desktop/src/lib/__tests__/sql/sqlStatementRanges.spec.ts
@@ -335,6 +335,44 @@ describe("splitSqlStatementRanges", () => {
     expect(rangeSqlTexts(splitSqlStatementRanges(sql, "sqlserver"))).toEqual(["SELECT 'GO'\n-- GO\nSELECT 2", "SELECT 3"]);
   });
 
+  it("keeps SQL Server IF...BEGIN...END block together as a single statement", () => {
+    const sql = `IF OBJECT_ID ('CATEGORIA') IS NOT NULL
+BEGIN
+DROP TABLE CATEGORIA
+END
+CREATE TABLE CATEGORIA
+(
+  COD_CATE CHAR(3) NOT NULL PRIMARY KEY,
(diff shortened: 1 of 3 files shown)


t8y2 (changes requested):
Request changes. The current head still has three SQL-splitting correctness issues:

1. Semicolon-delimited `IF ... BEGIN ... END` blocks can still be split into separate statements, so valid SQL Server scripts may execute with changed control flow.
2. `BEGIN` inside string literals or comments can enter block mode and consume later statements.
3. The Rust-side special case is not scoped to the SQL Server dialect, so scripts for other databases can be parsed with SQL Server semantics.

Please make the scanner token-aware for strings/comments, scope this behavior to SQL Server, and add regression coverage for semicolon-delimited blocks plus literal/comment occurrences of `BEGIN` and `END`.

PR #5474, opened 2026-08-06
feat(oracle): query editor manual transaction (Tx / Commit / Rollback)

Yy-702:
## Summary

- Enable Oracle query-editor **manual transactions**: toolbar **Tx / Commit / Rollback**, default **manual** mode for new Oracle query tabs (`autoCommit=false`).
- Add oracle-go sticky TX RPCs (`begin_manual_transaction` / `commit` / `rollback`) and wire Rust `TxnConnection::Agent` through existing Tauri manual-txn APIs.
- Protect open manual TX from shared multi-session agent runtime kills (skip keepalive on `manual-txn` sessions; do not replace/kill shared runtime on session close timeout when sticky TX may be active).
- Fix idle-timeout banner so statement failures no longer look like “5 minutes of inactivity”.

Tracks #5473. Related: #3495, #5177, #5414 (Oracle), #3813 (Oracle).

## Test plan

- [x] `cd agents/drivers/oracle-go && go test .`
- [x] `cargo check -p dbx-core`
- [x] Protocol unit tests (`agent_protocol_matches_contract_file`, `defines_agent_protocol_methods`)
- [x] Local desktop: Oracle query tab shows Tx (default on); UPDATE then SELECT in same TX; Commit/Rollback; reconnect after agent install
- [ ] CI full suite
- [ ] After install, native agent must include new methods (rebuild/copy `agents/drivers/oracle-go` → `~/.dbx/agents/drivers/oracle/agent` or ship updated agent)

### Agent install note (dev)

```bash
(cd agents/drivers/oracle-go && go build -o agent .)
cp agent ~/.dbx/agents/drivers/oracle/agent
chmod +x ~/.dbx/agents/drivers/oracle/agent
```

### DIFF
--- a/agents/common/src/main/java/com/dbx/agent/AgentProtocol.java
+++ b/agents/common/src/main/java/com/dbx/agent/AgentProtocol.java
@@ -42,6 +42,9 @@ public final class AgentProtocol {
     public static final String METHOD_GET_EXPLAIN_INFO = "get_explain_info";
     public static final String METHOD_EXECUTE_BATCH = "execute_batch";
     public static final String METHOD_EXECUTE_TRANSACTION = "execute_transaction";
+    public static final String METHOD_BEGIN_MANUAL_TRANSACTION = "begin_manual_transaction";
+    public static final String METHOD_COMMIT_MANUAL_TRANSACTION = "commit_manual_transaction";
+    public static final String METHOD_ROLLBACK_MANUAL_TRANSACTION = "rollback_manual_transaction";
     public static final String METHOD_DISCONNECT = "disconnect";
     public static final String METHOD_SHUTDOWN = "shutdown";
 
@@ -188,6 +191,9 @@ public final class AgentProtocol {
         METHOD_GET_EXPLAIN_INFO,
         METHOD_EXECUTE_BATCH,
         METHOD_EXECUTE_TRANSACTION,
+        METHOD_BEGIN_MANUAL_TRANSACTION,
+        METHOD_COMMIT_MANUAL_TRANSACTION,
+        METHOD_ROLLBACK_MANUAL_TRANSACTION,
         METHOD_DISCONNECT,
         METHOD_SHUTDOWN
     ));
--- a/agents/common/src/main/resources/agent-protocol-v1.json
+++ b/agents/common/src/main/resources/agent-protocol-v1.json
@@ -36,6 +36,9 @@
     "get_explain_info",
     "execute_batch",
     "execute_transaction",
+    "begin_manual_transaction",
+    "commit_manual_transaction",
+    "rollback_manual_transaction",
     "disconnect",
     "shutdown"
   ],
--- a/agents/common/src/main/resources/agent-protocol-v2.json
+++ b/agents/common/src/main/resources/agent-protocol-v2.json
@@ -40,6 +40,9 @@
     "get_explain_info",
     "execute_batch",
     "execute_transaction",
+    "begin_manual_transaction",
+    "commit_manual_transaction",
+    "rollback_manual_transaction",
     "disconnect",
     "shutdown"
   ],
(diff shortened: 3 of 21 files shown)


t8y2 (commented):
Maintainer patch here — fixes a correctness issue in the manual-transaction path that the original tests didn't catch (the mock driver doesn't emulate Oracle's server-side implicit COMMIT on DDL).

**Root cause**: `ALTER SESSION SET CURRENT_SCHEMA = <schema>` is DDL. Per Oracle's transactions docs, "the database issues an implicit COMMIT statement before and after every DDL statement; if the current transaction contains DML statements, then Oracle Database first commits the transaction." `executeQuery` / `executeQueryPage` / `startTableRead` each called `s.setSchema(opts.Schema)` on every statement, and `setSchema` routes through `manualTx.ExecContext` when a manual TX is open — so each statement after the first silently committed the preceding DML, breaking manual-transaction atomicity. Example: `UPDATE` then `DELETE` in the same TX got the `UPDATE` auto-committed when `DELETE`'s preceding `ALTER SESSION` fired; `Rollback` then only undid the `DELETE`.

**Fix**: `beginManualTransaction(schema)` already pins the schema on the reserved connection at BEGIN time, so the per-statement `setSchema` is redundant during an open manual TX. Guard the three call sites with `&& !s.hasManualTransaction()`:
- `executeQueryPage` (main.go:2886)
- `startTableRead` (main.go:2962)
- `executeQuery` (main.go:3095)

**Regression test**: added `TestManualTransactionSkipsPerStatementSetSchema` — opens a manual TX with a schema, runs `UPDATE` + `DELETE` with `Schema` set, and asserts exactly **1** `ALTER SESSION SET CURRENT_SCHEMA` in the driver's exec log (the one from `beginManualTransaction`). Without the guard the test fails with "expected 1, got 3" (one per statement + the begin), proving it catches the regression. All existing manual-tx tests still pass (`TestManualTransactionBeginCommitRollback`, `TestManualTransactionDisconnectRollsBack`, `TestManualTransactionBlocksOneShotTransaction`, `TestManualTransactionRPCDispatch`, etc.). `go build` + `go test` clean on `agents/drivers/oracle-go`.

New head: `e9aef57b9` (fast-forwarded from `41989accc`). 2 files, +46/-3. Once CI is green this is good to merge.

t8y2 (changes requested):
Follow-up on the maintainer patch (`e9aef57b9`, setSchema guard): while verifying locally I found three `connection::tests` failures that the original review missed (the reviewer only read the code and didn't run the full `rust-test`). These are from the PR's own `connection.rs` changes, not the setSchema guard.

Failing on `e9aef57b9` (confirmed locally with `cargo test -p dbx-core --no-default-features --lib`):
- `connection::tests::inserting_replacement_returns_error_when_previous_close_replaces_shared_runtime` — `assertion failed: result.is_err()` (connection.rs:6983)
- `connection::tests::metadata_close_replace_runtime_detaches_shared_runtime_siblings` — `assertion failed: !state.connections.read().await.contains_key(workload_pool_key)` (connection.rs:6936)
- `connection::tests::reconnect_waits_for_agent_runtime_replacement_before_publishing_new_pool` — `assertion failed: runtime.is_failed()` (connection.rs:7084)

Root cause: the PR's `connection.rs` changes intentionally protect shared Agent runtimes when a `:session:manual-txn-` session is active — `replace_runtime_after_close_failure` / `PoolRoutingControl` now skip killing sibling pools and leave the shared runtime running so manual transactions on sibling sessions aren't aborted. That's the right behavior for the feature. But the three existing tests still assert the **old** expectations (runtime gets killed/replaced/err returned), so they now fail.

Requested change: update these three tests' expectations to match the new "protect shared runtime when a manual-txn session is active" semantics — i.e. assert the runtime stays alive / siblings are preserved / the call succeeds (not `is_err()`) when a `:session:manual-txn-` pool is present. The setSchema guard patch (`e9aef57b9`) is correct and can stay; only the connection tests need their assertions updated to the new semantics. Once the tests are updated and `cargo test -p dbx-core --no-default-features --lib` passes, this is good to merge.

PR #5487, opened 2026-08-06
fix(sql): keep Ctrl+Enter on the current statement when a middle statement's line ends with trailing whitespace

Abeautifulsnow:
## Summary

When a multi-line script contains a **middle** statement (not the last) whose line ends
with trailing spaces, pressing Ctrl+Enter with the cursor on those spaces executed the
wrong target:

- **No `;` before the middle statement** → execution fell back to the first statement.
- **`;` before the middle statement** → the middle statement was concatenated with the
  following statement and MySQL reported a syntax error.

## Root cause

Statements without a `;` terminator merge into one range and are re-split into "soft"
statement ranges. Gap attribution only checked whether the cursor's line contained the
range's **start** line (`isCursorOnStatementLine`). For **multi-line soft statements**
(`WITH … SELECT`, `EXPLAIN … SELECT`, `UNION` operands, `INSERT … SELECT`) the start is
on an earlier line, so the check failed and the cursor fell back to the whole merged
script — reproducing both reported symptoms.

## Changes

- `apps/desktop/src/lib/sql/sqlStatementRanges.ts`
  - New `isCursorOnRangeEndLine`: attributes the trailing-whitespace gap by the range's
    **content end line** (`to`) instead of its start line.
  - Applied in `rangeForCursorInSoftRanges` and the `statementRangeAtCursor` gap fallback.
- `apps/desktop/src/lib/sql/executableStatementRangeCache.ts`
  - Same fallback added to `executableStatementRangeAtCursor` (the current-statement
    frame cache) so the highlight matches what actually executes.
- Tests (`.spec.ts` ×2)
  - WITH / UNION / EXPLAIN multi-line tails, both no-`;` and `;`-before variants.
  - The exact reported screenshot shape (`… AS uu   ` with `WHERE id = 2` following),
    asserting the **submitted SQL** (`executionCandidateForMode(…, "current")`), not only
    the highlight range.
  - Last-line trailing-whitespace case that the fix targets (pre-patch it fell back to
    the merged script).

## Verification

- `npx vitest run apps/desktop/src/lib/__tests__/sql/ apps/desktop/src/lib/__tests__/editor/`
  → **58 files / 896 tests passed**.
- `oxlint` on changed files: clean.
- Brute-force scan of every cursor position across 13 script shapes confirms
  `statementRangeAtCursor` and `executableStatementRangeAtCursor` resolve identically.

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

## 关联 Issue

Related #5193

### DIFF
--- a/apps/desktop/src/lib/__tests__/sql/executableStatementRangeCache.spec.ts
+++ b/apps/desktop/src/lib/__tests__/sql/executableStatementRangeCache.spec.ts
@@ -124,6 +124,19 @@ describe("executableStatementRangeCacheForDoc", () => {
     expect(executableStatementRangeAtCursor(cache, contentEnd + 1)?.sql).toBe("SELECT 2");
   });
 
+  it("keeps a trailing-whitespace cursor on a non-final line of a multi-line statement in the frame cache", () => {
+    const sql = "SELECT * FROM `profiles`;\nSELECT * FROM `users` AS uu   \nWHERE id = 2\nSELECT * FROM `orders`;";
+    const doc = Text.of(sql.split("\n"));
+    const cache = executableStatementRangeCacheForDoc(null, doc, "mysql");
+    const markerEnd = sql.indexOf("uu") + 2;
+    const lineEnd = sql.indexOf("\n", markerEnd);
+    const expected = "SELECT * FROM `users` AS uu   \nWHERE id = 2";
+
+    for (let pos = markerEnd; pos < lineEnd; pos += 1) {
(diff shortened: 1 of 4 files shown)


t8y2:
Thanks for the contribution! Merged in b437c470e144, will be released in the next version.

PR #5531, opened 2026-08-06
fix(xugu): scope metadata queries to current database

mapan0424:
## Summary

- Scope Xugu metadata lookups to `CURRENT_DB_ID` after the selected database is applied.
- Keep database enumeration global so users can still see all databases exposed by their account.
- Cover schemas, tables, columns, indexes, constraints, foreign keys, partitions, subpartitions, sequences, synonyms, programmable objects, object source, and table comments.

## Compatibility

This change is isolated to `agents/drivers/xugu` and does not modify other database drivers or frontend routing. Schema listing remains database-wide for the selected database; schema-level filtering is applied when listing objects and resolving table metadata.

## Testing

- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- Built the Xugu agent and verified database switching and metadata access against SHOP_DEMO and TIBMS_NEW using SYSDBA.
- Verified APP_TEST ordinary-user access to schemas, tables, programmable objects, columns, constraints, and indexes without RPC/EOF failures.
- Verified SHOP_DEMO table child metadata and executable table DDL retrieval after switching databases.


t8y2:
Thanks for the contribution! Merged in 00bbfad05549, will be released in the next version.

Issue #5262, opened 2026-08-03
[Bug] macOS: app becomes permanently unlaunchable when the running instance loses its WindowServer connection (single-instance guard has no liveness check)


possebon:
> Reported from a macOS install of DBX 0.5.74. Not database-specific — this is a desktop app lifecycle bug.
> 本问题与具体数据库无关，属于桌面端应用生命周期问题。数据库类型：通用。

## Support info / 支持信息

```
DBX 版本：v0.5.74
运行环境：桌面端 (Desktop)
操作系统：macOS 26.5.2 (Build 25F84)
系统架构：aarch64 (Apple Silicon, Mac16,8)
```

(Filled in manually — the app could not be opened to use "设置 → 关于我们 → 复制支持信息".)

---

## Summary

On macOS, DBX can reach a state where **the app is permanently unlaunchable with no error, no window, and no crash report**. Clicking the app in Finder/Dock does nothing at all.

The cause is a running DBX process that has **lost its WindowServer (CoreGraphics) connection**. It stays alive and idle, but can no longer draw a window or a tray icon. Because `tauri-plugin-single-instance` is enabled for release builds, every subsequent launch detects that instance, hands off, and exits `0` — so the user gets no feedback whatsoever.

The single-instance guard checks *"does a process hold the lock"*, not *"can that process actually show a window"*. One process in this state makes the app unusable until it is killed from a terminal.

## Evidence

The stuck process was alive for 5h22m and completely invisible to LaunchServices:

```console
$ lsappinfo info 60873
[ NULL ]  [ NULL ]
    bundleID=[ NULL ]
    bundle path=[ NULL ]
    executable path=[ NULL ]
    pid =  !cgsConnection !signalled type=[ NULL ] flavor=[ NULL ] Version=[ NULL ] Arch=!!none
```

`!cgsConnection` = no CoreGraphics connection. The process was also absent from `lsappinfo list` entirely, and `System Events` could not see it (`-1728 Can't get process`).

The process itself was **healthy and idle, not hung** — `sample` showed a normal Cocoa event loop at 0% CPU:

```
2602 Thread_20159001   DispatchQueue_1: com.apple.main-thread  (serial)
  2602 -[NSApplication run] + 368
    2602 -[NSApplication(NSEventRouting) nextEventMatchingMask:untilDate:inMode:dequeue:] + 72
      ...
        2602 __CFRunLoopServiceMachPort + 160
          2602 mach_msg2_trap + 8
```

```console
[...]

PR #5552, opened 2026-08-06
fix(macos): surface failed main-window reveals instead of discarding them

possebon:
Refs #5262

## Problem

`show_main_window` discarded all three of its calls:

```rust
fn show_main_window<R: tauri::Runtime>(app: &tauri::AppHandle<R>) {
    if let Some(window) = app.get_webview_window("main") {
        let _ = window.show();
        let _ = window.unminimize();
        let _ = window.set_focus();
    }
}
```

So a failed reveal produced **no output at all**, on any platform.

That silence is what makes #5262 so hard to diagnose. On macOS an instance can lose its WindowServer (CoreGraphics) connection and keep running, idle: it can no longer present a window *or* render a tray icon, but it still holds the single-instance lock. Every later launch is handed to that instance and exits `0` in ~115 ms. From the user's side the app simply does not start — no error, no window, no crash report.

In the case I hit, the stuck process had been alive 5h22m and was invisible to LaunchServices:

```console
$ lsappinfo info 60873
    pid =  !cgsConnection !signalled  Arch=!!none      # all fields NULL
```

...while `sample` showed a perfectly normal, idle Cocoa event loop at 0% CPU (`-[NSApplication run]` → `mach_msg2_trap`) — so the process was *healthy, just blind*. With `quit_on_close: false` the window had been hidden rather than quit, and all three ways back (window, tray icon, Dock reopen) fail together.

## What this PR does

Narrow scope: it makes the failure **observable**. No behaviour change on the success path.

- `show_main_window` now returns whether the window actually ended up visible, and logs each failing call (`show`, `unminimize`, `set_focus`, `is_visible`) rather than swallowing it.
- The single-instance handoff logs `main_window_probe_state` when it cannot reveal a window. That helper already existed and computed exactly the right thing — it just was not used on this path.
- The macOS `RunEvent::Reopen` path gets the same diagnostic, since it fails identically.

`eprintln!` is deliberate rather than `log::`. `append_startup_probe` routes to `startup_recovery::record`, which is the no-op implementation everywhere except Windows, and `log` is set to `LevelFilter::Off` unless the user has enabled debug logging — so neither leaves a trace in exactly the situation that needs one.

## What this PR deliberately does *not* do

It does not attempt automatic recovery, and I want to be upfront about why rather than ship something that looks like a fix.

**I could not reproduce the lost WindowServer connection on demand.** Without a reproduction I cannot verify that any liveness probe actually detects the broken state — an `NSWindow` in a process that has lost its CGS connection may well still report `isVisible == true`, in which case a probe would silently never fire. And a probe that gets it *wrong in the other direction* would tear down a healthy instance and lose the user's unsaved editor state, which is considerably worse than the bug it is trying to fix.

So the recovery step needs a liveness signal that is known-good, and this PR is the prerequisite: with these logs, the next occurrence produces the evidence needed to build that check on something real.

If you would like to go further in this PR, the options I can see are:

1. On failed reveal, have the stale instance exit so the *next* launch succeeds (one dead click instead of a permanent lockout).
2. Same, but re-spawn a detached replacement first — mirroring what `startup_recovery` already does on Windows.
3. If the tray icon fails to install, fall back to `quit_on_close: true` so a hidden unreachable process cannot happen in the first place.

Happy to implement whichever you prefer, or to drop the diagnostics into a different shape if you would rather they went through `log::`.

## Testing

```
cargo check   --no-default-features   # clean
cargo test    --no-default-features --lib   # 196 passed
cargo clippy  --no-default-features -- -D warnings   # clean
cargo fmt -- --check   # clean
```

No new tests: the changed function needs a live Tauri window to exercise, and the failure mode it reports is the one I could not reproduce. The pure helpers around it are already covered by the existing suite.

Manual verification of the underlying bug and the workaround: `SIGTERM` on the stuck process exited it in ~2s, and the relaunched instance registered correctly (`lsappinfo` → `LSDisplayName="DBX"`, RSS 115 MB vs the stuck process's 1.2 MB) and came to the front.

---

### 中文说明

`show_main_window` 用 `let _ =` 丢弃了 `show()`、`unminimize()`、`set_focus()` 的全部返回值，因此窗口显示失败时**没有任何输出**。

这正是 #5262 难以定位的原因：在 macOS 上，实例可能丢失与 WindowServer (CoreGraphics) 的连接后继续空闲运行 —— 既无法显示窗口，也无法渲染托盘图标，但仍持有 single-instance 锁。之后每次启动都会把参数转交给它并在约 115ms 内以 `0` 退出，用户看到的就是"点了没反应"，且没有报错、没有窗口、没有 crash report。

本 PR 只做一件事：**让失败可见**，成功路径行为不变。

- `show_main_window` 返回窗口最终是否可见，并记录每个失败的调用。
- single-instance 回调在无法显示窗口时输出 `main_window_probe_state`（该函数已存在，只是没用在这条路径上）。
- macOS `RunEvent::Reopen` 路径同样加上该诊断。

使用 `eprintln!` 而非 `log::` 是有意的：`append_startup_probe` 在非 Windows 平台是 no-op，而 `log` 在未开启 debug logging 时为 `LevelFilter::Off`，两者恰好在最需要日志的场景下都不会留下记录。

**本 PR 不包含自动恢复。** 我无法稳定复现 CGS 连接丢失的状态，因此无法验证任何存活检测是否真的能识别该状态（丢失 CGS 连接的 `NSWindow` 可能仍然返回 `isVisible == true`）；而误判则会导致健康实例被强制重启、丢失用户未保存的编辑内容，后果比原 bug 更严重。上面列出了三种可选的后续方案，您倾向哪种我都可以实现。

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -476,11 +476,34 @@ fn apply_linux_webkit_rendering_workarounds() {
     }
 }
 
-fn show_main_window<R: tauri::Runtime>(app: &tauri::AppHandle<R>) {
-    if let Some(window) = app.get_webview_window("main") {
-        let _ = window.show();
-        let _ = window.unminimize();
-        let _ = window.set_focus();
+/// Brings the main window to the foreground, reporting whether it ended up visible.
+///
+/// The individual calls used to be discarded with `let _ =`, which made a failed
+/// reveal completely silent. That matters on macOS: an instance that has lost its
+/// WindowServer connection stays alive and idle but can no longer present a window
+/// or a tray icon, and the single-instance guard keeps handing later launches to it,
+/// so the app looks like it simply does not start. Logging here is what makes that
+/// state diagnosable at all.
+fn show_main_window<R: tauri::Runtime>(app: &tauri::AppHandle<R>) -> bool {
+    let Some(window) = app.get_webview_window("main") else {
+        eprintln!("[WINDOW] show_main_window: no \"main\" webview window to reveal");
+        return false;
+    };
+    if let Err(err) = window.show() {
+        eprintln!("[WINDOW] show_main_window: show() failed: {err}");
+    }
+    if let Err(err) = window.unminimize() {
+        eprintln!("[WINDOW] show_main_window: unminimize() failed: {err}");
+    }
+    if let Err(err) = window.set_focus() {
+        eprintln!("[WINDOW] show_main_window: set_focus() failed: {err}");
+    }
+    match window.is_visible() {
+        Ok(visible) => visible,
+        Err(err) => {
+            eprintln!("[WINDOW] show_main_window: is_visible() failed: {err}");
+            false
+        }
     }
 }
 
@@ -736,7 +759,9 @@ fn setup_desktop_tray<R: tauri::Runtime, M: Manager<R>>(
     })
     .on_tray_icon_event(|tray, event| match event {
         TrayIconEvent::Click { button: MouseButton::Left, button_state: MouseButtonState::Up, .. }
-        | TrayIconEvent::DoubleClick { button: MouseButton::Left, .. } => show_main_window(tray.app_handle()),
+        | TrayIconEvent::DoubleClick { button: MouseButton::Left, .. } => {
+            show_main_window(tray.app_handle());
+        }
         _ => {}
     })
     .build(manager)?;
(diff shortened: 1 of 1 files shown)


possebon:
@t8y2 CI is green on this one — `rust-test` (16m44s), `rust-fmt-clippy` (9m7s), `rust`, `changes`, `label` and `notify` all pass. The other matrix jobs are skipped by the `changes` filter, which is correct: the diff touches only `src-tauri/src/lib.rs`.

The change itself is small and behaviour-preserving on the success path — `show_main_window` stops discarding `show()` / `unminimize()` / `set_focus()` with `let _ =`, returns whether the window actually became visible, and logs failures. The single-instance handoff and the macOS reopen path then report `main_window_probe_state` when no window could be revealed.

There is one decision I would rather you make than guess at, and it is the reason I kept the scope this narrow. I could not reproduce the lost WindowServer connection on demand, so I did not add automatic recovery: an `NSWindow` in a process that has lost its CGS connection may still report `isVisible == true`, which would make any liveness probe silently never fire, and a probe wrong in the other direction would tear down a healthy instance and lose the user's unsaved editor state. That trade seemed yours to make, not mine.

If you want recovery in this PR, the three options I see are listed in the description — stale instance exits so the next launch succeeds, exits after re-spawning a detached replacement (mirroring what `startup_recovery` already does on Windows), or falling back to `quit_on_close: true` when the tray icon fails to install. Happy to implement any of them, reshape the diagnostics to go through `log::` instead of `eprintln!`, or close this if you would rather solve it differently.

No urgency from my side — I know the issue volume here is high. Context is in #5262.

t8y2:
Thanks for the contribution! Merged in 4efe579d9005, will be released in the next version.

PR #5555, opened 2026-08-06
feat(sidebar): show invalid object status with a visual marker

mapan0424:
## Why

DBX already receives a `valid=false` state for invalid database objects and shows an `· INVALID` suffix in the tree. The suffix alone is easy to miss in a large schema, especially when several object groups are expanded. This change adds a small, consistent visual marker so invalid objects can be identified at a glance.

## What changed

- Overlay the existing object-type icon with a red `CircleX` marker when `TreeNode.valid === false`.
- Keep the original table, view, procedure, function, trigger, package, type, and other object icons unchanged.
- Use the same visible label in the tooltip, including the existing `· INVALID` suffix.
- Add regression coverage for invalid, valid, and unknown validity states.

## Compatibility and scope

This is intentionally a frontend-only change. It does not change database drivers, metadata SQL, permissions, caching, object loading, or DDL generation. It therefore does not alter behavior for any database. Any driver that already reports `valid=false` benefits from the marker; objects with `valid=true` or an unknown validity state render exactly as before.

## Validation

- `pnpm vitest run apps/desktop/src/components/sidebar/__tests__/TreeItem.loadMoreActivation.spec.ts` — 4 tests passed.
- `pnpm vitest run apps/desktop/src/components/sidebar/__tests__` — 6 files / 19 tests passed.
- `pnpm typecheck` — passed.
- `pnpm lint` — passed with existing repository warnings only.
- `pnpm exec oxfmt --check ...` — passed.

The local production build is currently blocked by the checkout dependency resolution for `@noble/ciphers/aes.js` imported by `apps/desktop/src/lib/imports/navicatImport.ts`; this error is outside the files changed by this PR.


mapan0424:
The failed CI run appears to be infrastructure-related rather than a code failure. The `changes` job failed while resolving/download­ing GitHub Actions, ending with `Service Unavailable`; the remaining checks were skipped during setup. I attempted to rerun the workflow, but GitHub requires repository administrator permissions for this action. Could a maintainer please rerun the failed workflow when convenient?

t8y2:
Thanks for the contribution! Merged in c32f92a90232, will be released in the next version.

PR #5559, opened 2026-08-06
feat(docs): database documentation viewer and DBML export

possebon:
## 变更说明

Adds **database documentation** to DBX: a dbdocs.io-style wiki for a live schema, editable in place, plus **DBML export** for dbdiagram.io and CI drift checks.

The organising idea is that `SchemaSnapshot` is the product and **every output is a serializer over it**. The collector reads a schema once; the viewer, the DBML writer and (later) a standalone HTML export are all just renderings of that one structure. Nothing re-queries the database to render.

```
collect ──▶ SchemaSnapshot ──┬──▶ Vue viewer      (in-app, editable)
   ▲                          ├──▶ DBML           (dbx dbml)
   │                          └──▶ standalone HTML (not in this PR)
annotations (notes + groups, a JSON file you can commit)
```

**What a user gets**

- Right-click any table in the Object Browser → **Documentation**. A browsable wiki of the schema: tables, columns, indexes, enums, and relationships in **both** directions (References / Referenced by).
- Write Markdown notes on the project, any table and any column. Group tables and give each group a colour. Everything autosaves.
- Notes live in a plain JSON file. Point a connection's **Notes file** at a path in your repository and schema documentation becomes reviewable in pull requests, next to the migration that changed the schema.
- `dbx dbml <connection> --notes docs/dbx-docs.json --out docs/schema.dbml` — commit the result and let CI fail on unreviewed schema drift.

**Three design decisions worth flagging to a reviewer**

1. **The viewer makes zero backend calls.** Everything under `apps/desktop/src/docs/**` is pure: components emit edit intents upward and the dialog outside that directory does all I/O. A static contract test enforces it (no `@/lib/backend`, no `invoke(`, no stores, no `fetch(`, and — for the same reason — no `vue-i18n`, since `useI18n()` throws with no Vue app around it). This is what lets the identical components later be bundled into a self-contained HTML file.

2. **Groups store a hue (0–359), not a colour.** Lightness and chroma come from the theme, so one stored number stays legible on both light and dark grounds. The group editor previews both grounds side by side because that is the property being chosen. Tokens are declared in `hsl()` first and re-declared inside `@supports (color: oklch(1 0 0))`, since DBX still supports WebViews without `oklch`.

3. **Annotations are applied in Rust, never re-implemented in TypeScript.** `collect` returns the *raw* snapshot; `applyDocsAnnotations(raw, file)` is a pure Rust call with no database access. Doing it optimistically in TS would mean a second implementation of the rule where a local note shadows a database `COMMENT ON` while preserving it underneath — in a second language.

**Not built on purpose:** a second ER diagram. DBX already ships `SchemaDiagramDialog.vue`; the docs viewer links to it.

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [x] 文档更新
- [ ] CI / 构建

## 涉及前端

Screenshots are against a live PostgreSQL database, not mocks.

**Documentation index, grouped by table group.** Notes render inline; each group's colour comes from its stored hue. The group editor below shows the same hue on a light and a dark ground at once.

Documentation index grouped by table group

**Table page.** Columns with types, constraints and enum values; indexes; relationships in both directions with inferred cardinality. `LOCAL` marks a note written in DBX — the database `COMMENT ON` it shadows is preserved, never overwritten.

Table page for public.orders

**Search** spans tables, columns, groups and enums.

Search across tables, columns and enums

**Enum page**, with the reverse lookup of every column using it.

Enum page showing values and using columns

**Dark mode.** The swatches shift with the theme while the stored hue is unchanged, and the two-ground preview still shows both.

The same documentation in dark mode

**DBML export** of the same schema:

```dbml
Enum order_status { pending  paid  packed  shipped  delivered  refunded  cancelled }

TableGroup Ordering [color: #007cbb] { order_items  orders  payments  shipments }

Table orders {
  id bigint [pk, default: `nextval('orders_id_seq'::regclass)`]
  status order_status [not null, default: `'pending'::order_status`, note: '''Only moves forward, except `refunded`.''']
  ...
}

Ref inventory_product_id_fkey: inventory.product_id - products.id   // 1–1: the PK *is* the FK
Ref orders_customer_id_fkey:   orders.customer_id     > customers.id // many–1
```

Cardinality is inferred rather than assumed: `inventory` gets `-` because its primary key *is* the foreign key, while the rest get `>`.

Constructs DBML cannot express are reported rather than dropped silently:

```console
$ dbx dbml shop --notes docs/dbx-docs.json
warning: idx_products_active on public.products is documented but omitted from the DBML:
         partial index filter has no DBML equivalent.
```

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

```console
$ cargo test -p dbx-core --lib docs        # 111 passed
$ pnpm vitest run apps/desktop/src/docs apps/desktop/src/i18n   # 304 passed
$ pnpm typecheck                            # clean
```

Full-stack manual verification against a live PostgreSQL schema (9 tables, 2 enums, 10 foreign keys, a partial index): opened the viewer from the Object Browser, wrote project/table/column notes, created and recoloured a group, confirmed the notes file was written to the configured path, reopened the dialog and confirmed everything survived, then exported DBML from the CLI over the same notes file.

One caveat, stated plainly: on the first `make check` run, `queryStore.test.ts > result cache eviction keeps recently accessed inactive tabs` failed on a wall-clock timeout. It passes in isolation and on a re-run of the full suite; it is a timing-sensitive test under load and is not touched by this change.

**On the size of this PR.** It is large (89 commits, ~117 files) because it was built in parts that each stand alone: snapshot + DBML + CLI verb, then the annotation store, then the viewer, then mounting and editing it in the app. Every part is independently reviewable and the commit history keeps that order. Happy to split it into separate PRs along those seams if that is easier to review — just say which shape you prefer.

**Known boundaries** (also stated in the docs page): relational engines only; tables, columns, indexes, relationships and enums are documented, but not triggers, procedures, functions, sequences or grants; DBML export is one-way — DBX never imports it; engines with no foreign-key metadata produce documentation with no edges and say so in a warning rather than rendering an empty page.

## 关联 Issue

None — this was not tracked by an existing issue. Happy to open one first if you would rather discuss the design before reviewing code.

---


t8y2:
Thanks for the contribution! Merged in f30d59b9, will be released in the next version.

PR #5100, opened 2026-07-31
feat(grid): add boolean cell checkbox toggle

difagume:
## Description

Boolean-typed columns (`boolean` / `bool` / `bit` / `bit(1)`) in the data grid now render as a clickable checkbox instead of raw text.

- Click the checkbox in the canvas grid to cycle the value: `true -> false -> null` (nullable columns) -> back to `true`.
- The same toggle is available in the DOM grid cells.
- Keyboard entry and double-click text editing are suppressed for boolean columns in favor of the toggle interaction.
- A new `dataGridBooleanColumn` helper centralizes type detection, value normalization, and tri-state cycling logic.

## Change Type

- [x] New feature
- [ ] Bug fix
- [ ] Performance optimization
- [ ] Code refactoring
- [ ] Documentation update
- [ ] CI / Build

## Frontend Change

- [x] This PR changes the frontend; a screenshot / recording is attached below.

## Demo
[check-boolean.webm](https://github.com/user-attachments/assets/dab38a81-e599-4bb6-82a3-f650b6ed5af0)

## Files Changed

| File | Change |
|------|--------|
| `apps/desktop/src/lib/dataGrid/dataGridBooleanColumn.ts` | New helper: boolean column detection, value normalization, and tri-state cycle logic |
| `apps/desktop/src/lib/dataGrid/canvasDataGridRenderer.ts` | Render themed checkbox for boolean cells, centered null text, strike-through for deleted rows |
| `apps/desktop/src/components/grid/DataGrid.vue` | Wire canvas and DOM click handlers to toggle boolean values; skip text editing for boolean columns |
| `apps/desktop/src/composables/useDataGridEditor.ts` | Add `cycleBooleanCellValue` and allow explicit boolean values through the edit pipeline |
| `packages/app-tests/dataGridBooleanColumn.test.ts` | Unit tests for type detection, normalization, and both nullable / non-nullable cycle orders |

## Verification

- [ ] `make check` passed
- [ ] `make cargo-check-fast` passed
- [x] Relevant tests passed

Verification details:

- `pnpm exec vitest run packages/app-tests/dataGridBooleanColumn.test.ts packages/app-tests/canvasDataGridRenderer.test.ts` - 13 tests passed
- `pnpm typecheck` - passed

## Related Issue

No related issue.

PR #5572, opened 2026-08-07
feat(grid): add checkbox/dropdown display mode for boolean cells

difagume:
﻿## What changed

Following [PR #5100](https://github.com/t8y2/dbx/pull/5100), boolean editing settled on the enum dropdown. This PR adds the checkbox back as a **user-selectable display mode** instead of replacing one approach with the other:

- New setting `dataGridBooleanDisplayMode` (`checkbox` / `dropdown`), **default `dropdown`** — current behavior unchanged.
- Toggle available in the "View options" popover of both the SQL results and table data grids.
- Checkbox mode renders interactive checkboxes (canvas + DOM); a single click cycles `true -> false -> null` for nullable columns.
- In dropdown mode, boolean `null` cells align left like `true`/`false` values.
- Labels localized across 9 locales.

## Demo
[56.851Z.webm](https://github.com/user-attachments/assets/3525e3db-a1eb-463b-8871-386061195902)

## How it was tested

- `pnpm exec vitest run apps/desktop/src/stores/__tests__/settingsStore.spec.ts packages/app-tests/dataGridBooleanColumn.test.ts packages/app-tests/canvasDataGridRenderer.test.ts` — 84 tests passed
- `pnpm fmt` / `pnpm lint` — clean
- Manually verified: toggle in both grids, cycle `true -> false -> null`, default stays dropdown

## Related Issue

Related #5100 — follow-up on the checkbox vs. dropdown discussion, making the boolean display mode user-selectable.


t8y2 (commented):
Maintainer patch + rebase before squash.

Old head: `077ea798` — New head: `baaf8ed8b` (rebased onto current main `9fab90b6`, then patch commit added on top).

Rebase: branch was 3+ commits behind main (base `518f60fc`); three-way merge was clean (no conflicts), rebased for a tidy history.

Patch (DOM null-boolean edit gap from review): in checkbox mode the DOM/RecycleScroller path rendered an interactive checkbox only for non-null boolean cells (`item.data[...] !== null`), so null boolean cells fell through to the static `v-else` text — neither clickable to cycle nor double-clickable to edit (short-circuited by `onDomCellDblClick`). Canvas already handled this via `booleanNullTextHitFromCanvasEvent` (click the NULL text to cycle). The patch adds a dedicated `v-else-if` for null boolean cells in checkbox mode that renders the NULL text with a `@click.stop="cycleBooleanGridCell"` handler, aligning DOM behavior with canvas. Double-click short-circuit is kept unchanged (matches canvas `onCanvasCellDblClick`).

Verified: `oxfmt --check` clean; `vue-tsc --noEmit` clean; `vitest run dataGridBooleanColumn.test.ts + settingsStore.spec.ts + canvasDataGridRenderer.test.ts` 85 passed; new regression test `DOM checkbox mode renders a clickable placeholder for null boolean cells so they can be cycled like canvas` fails on the pre-patch source and passes with the patch.

t8y2:
Thanks for the contribution! Merged in 1940111, will be released in the next version.

PR #5582, opened 2026-08-07
fix(xugu): preserve invalid view status

mapan0424:
## Summary

- Preserve XuguDB view validity metadata when listing schema objects.
- Show invalid views with the same `INVALID` state used by other programmable objects.
- Add a regression test to prevent the catalog validity flag from being discarded.

## Root cause

The Xugu view listing query selected `NULL AS VALID`, even though `ALL_VIEWS.VALID` exposes the actual object state. As a result, invalid views could be listed but their invalid status was lost before reaching the object tree.

## Validation

- Built the Xugu agent and verified against `SHOP_DEMO`:
  - `APP_TEST.V_DBX_INVALID_DEMO` returns `valid: false`
  - `APP_TEST.V_CUSTOMER_ORDERS` returns `valid: true`

This change is isolated to the Xugu agent and does not alter metadata behavior for other database drivers.


t8y2:
Thanks for the contribution! Merged in b8144da, will be released in the next version.

PR #5662, opened 2026-08-07
feat(docs): export database documentation as a standalone HTML file

possebon:
## 变更说明

Follow-up to #5559, which added database documentation and DBML export. This makes the documentation **leave the application**: `dbx docs <connection> --out schema.html` produces **one self-contained HTML file** that opens with no server, no network and no DBX installed. The same file is produced by an **Export HTML** button in the docs dialog.

It is the same viewer #5559 already ships — the purity rule that kept `apps/desktop/src/docs/**` free of backend calls, stores and `vue-i18n` existed precisely so those components could be bundled and run somewhere DBX isn't.

```
collect ──▶ SchemaSnapshot ──┬──▶ Vue viewer       (in-app, editable)   #5559
   ▲                          ├──▶ DBML            (dbx dbml)           #5559
   │                          └──▶ standalone HTML (dbx docs)           this PR
annotations (notes + groups, a JSON file you can commit)
```

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [x] 文档更新
- [x] CI / 构建

## 涉及前端

Everything below is **one 250 KB HTML file** opened directly. No server, no network, no DBX running. Generated from a live PostgreSQL database.

**The exported index.** Theme and language controls are for the *reader*, who is usually not the exporter — you send this file to a colleague.

The exported documentation, rendered from a single HTML file

**A minimal read-only ER diagram**, cards tinted by table group, edges clipped to card borders. Clicking a table navigates to its page, so the diagram doubles as a table of contents. This is deliberately *not* a second `SchemaDiagramDialog` — that one reaches into stores and the backend and cannot be inlined. In-app, the docs viewer still links to the full diagram; only the export carries this one.

The exported ER diagram

**Switched to 日本語.** `カラム`, `インデックス`, `リレーションシップ`, `参照`, `参照元`, and the `ローカル` badge.

The same export switched to Japanese

### One fix to the in-app viewer, worth calling out separately

While wiring the language switcher I found the viewer's chrome was **mostly hardcoded English**. The `docs` namespace shipped in #5559 with 40 keys across all 8 locales and a passing parity test, but most components never called it — `docs.groupBySchema: "Schemas"` existed specifically for `DocsSidebar`, which rendered the word as a literal. `docs.noGroup`, `docs.noSchema` and `docs.localNote` were translated in every locale and called from nowhere.

The parity test compares locale files to *each other*, so a key no component calls is invisible to it. That is fixed here, and a new guard derives its forbidden literals **from the English namespace**, so a key added later is covered without anyone remembering to register it.

So a non-English user gets a correctly translated viewer in the app too, not only in exports.

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

```console
$ cargo test -p dbx-core --lib docs                          # 118 passed
$ pnpm vitest run apps/desktop/src/{docs,docs-export,i18n}   # 381 passed
$ pnpm typecheck                                             # clean
```

The artefact is verified three ways rather than only inspected:

- **It is executed.** A `happy-dom` test builds a fixture by calling the real `to_standalone_html`, runs the actual bundle, and asserts rendered column data reaches the DOM. Verified to catch regressions by breaking a component, rebuilding, and watching it fail.
- **It references nothing external.** Every `url()` in the emitted CSS must resolve to a `data:` URI, and the generated shell must contain no `src=`/`href=`.
- **The committed bundle cannot drift from its sources.** A sha256 manifest derived from Rollup's module graph *and* the CSS `@import` graph.

That last guard has already earned its keep twice in this branch's own history: rebasing onto current `main` pulled a newer `erDiagram.ts`, which `SchemaDiagram.vue` imports — the test failed, named that file, and the bundle was rebuilt. `erDiagram.ts` lives *outside* `src/docs/`, so a directory glob would have stayed green over a stale artefact. That is why the manifest is derived rather than listed.

## Three things a reviewer will want answered up front

**1. Why is a build artefact committed?** `crates/dbx-core/assets/docs-export.{js,css}` are generated. Rust cannot run Vite, and `dbx docs` must work from a plain `cargo install` on a machine with no Node — a `build.rs` shelling out to pnpm would break exactly that. Mitigations: the manifest guard above; a non-blocking CI step that rebuilds and diffs; and the build is **byte-for-byte reproducible** — four independent rebuilds produced an empty `git diff`. Both files are marked `-diff linguist-generated` so they don't drown review.

**2. `dbx docs` does not read a connection's stored `docs_notes_path`.** It requires an explicit `--notes`, exactly as `dbx dbml` does. So the same connection currently yields an annotation-full export from the GUI and an annotation-free one from the CLI. This is a **deliberate decision, not an oversight**: changing one verb alone would make two sibling verbs disagree about the same flag. If you'd prefer the fallback, it should change in both verbs together — happy to do that here or as a follow-up.

**3. This adds CI cost.** The `frontend` job gains a Rust toolchain and one crate build, because the smoke test builds its fixture by calling the real exporter. It reuses the existing sccache + `rust-cache` setup from the other Rust jobs, with a distinct `shared-key` (this job builds one crate with `--no-default-features`, a fingerprint the other keys don't match). If you'd rather not pay that on frontend PRs, the alternative is to assemble the test's HTML shell in JS instead of calling Rust — cheaper, but it stops proving that the bytes Rust actually emits will execute.

## Design notes

- **The payload is base64, not escaped JSON.** A note discussing HTML is plausible in a schema document, and inlined as text a literal `</script>` would terminate the script element early and inject the rest as markup. base64's alphabet contains no `<`, so the property holds by construction rather than by remembering to escape on every path.
- **The bundle is an `iife` in a classic `<script>`, not a module.** This is the single property separating "works from `file://`" from "works only over http", and it is invisible to any test that serves over localhost.
- **Hash routing lives in the export shell, not in `DocsApp`.** DBX has no router; a viewer that wrote to `location.hash` would hijack the host application's address bar. `DocsApp` gained an *optional* `v-model:route` — absent, it behaves exactly as it does today.
- **Design tokens moved to `apps/desktop/src/styles/tokens.css`**, imported by both `globals.css` and the export's stylesheet. The export needs its own Tailwind entry, because `globals.css` sources the whole application and `@source` is additive with no way to un-source; without the extraction the tokens would have to be duplicated.

**Known boundaries:** relational engines only; tables, columns, indexes, relationships and enums are documented, but not triggers, procedures, functions, sequences or grants; the export is read-only.

## 关联 Issue

Follows #5559. Not tracked by an issue — happy to open one if you'd rather discuss the design first.

---


t8y2 (commented):
Maintainer patch — rebase onto latest main to clear the pre-existing rustfmt failure.

- Old head: `eeb599df5640c236d33781bc4c98ffd1e0993376`
- New head: `878b50c68`
- Old base: `29fe57143640a025e7a3624876e997bfbc833425`
- New base: `9057be0e781d40dadffe85709b54649d07f69e15` (latest main)

## 根因
CI `rust-fmt-clippy` 失败的唯一 diff 在 `crates/dbx-core/src/db/sqlserver.rs:3220`，该文件本 PR 未触碰。这是 main 分支预存的 rustfmt 折行偏差（CI rustfmt 1.97.1 与该文件最后格式化时的版本不一致）。#5659 合入 main 时已修复该偏差（`style(dbx-core): reformat sqlserver.rs imports to satisfy cargo fmt`）。

## Patch 内容
纯 rebase：将本 PR 24 个 commit replay 到最新 main 之上，带入 #5659 的 fmt 修复作为 base，无需修改本 PR 任何代码。

- `git range-diff` 确认 24 个 commit 全部 `=` 对应，无 commit 丢失、无语义变化。
- 本地 `cargo fmt --check` 全量 EXIT=0（rebase 后无任何 fmt 偏差）。
- `git diff --check` clean，无冲突标记/空白错误。

## 行为保留
本 PR 的代码（`crates/dbx-core/src/docs/export.rs`、`apps/desktop/` 等）与逻辑完全未变，仅 base 前移。rebase 无冲突（`merge-tree` 预检 0 conflict markers）。

## 已跑检查
- `cargo fmt --check`：EXIT=0
- `git range-diff old-base...old-head new-base...new-head`：24 commit 全部等价对应
- `git diff --check`：clean

## CI
已记录当前状态，标记 ci-pending，等待 rebase 后 CI 重新触发，rust-fmt-clippy 应转 SUCCESS 后再进入 merge。

t8y2:
Thanks for the contribution! Merged in 63498da, will be released in the next version.

PR #5663, opened 2026-08-07
feat(xugu): expose routine parameter metadata

mapan0424:
## Change summary

This PR adds source-backed parameter metadata for XuguDB stored procedures and functions.

- Parse parameter names, declaration types, `IN`, `OUT`, and `IN OUT` modes, ordinals, and default expressions from the routine definition.
- Show procedure and function parameter details above the source preview.
- Show the declared `RETURN` type for functions.
- Show the actual XuguDB default expression in the procedure execution dialog.
- Generate executable XuguDB procedure calls that supply `NULL` for `OUT`, preserve input values for `IN OUT`, omit trailing defaults, and switch to named notation when a middle default is omitted.

## Why source-backed metadata

The tested XuguDB server does not expose `ALL_ARGUMENTS`, so the Oracle-style catalog query is not portable across supported XuguDB installations. The existing Xugu Agent already returns the authoritative routine definition through `get_object_source`.

This follows the same high-level approach used by the mature XuguDB DBeaver extension: obtain the stored routine definition and parse only its declaration header. The PL/SQL body is never interpreted. Malformed or unterminated definitions fail closed and return no inferred metadata.

The parser handles:

- `CREATE` and `CREATE OR REPLACE`
- qualified and quoted routine names
- omitted `IN`, explicit `IN`, `OUT`, `IN OUT`, and `INOUT`
- nested declaration parentheses such as `NUMERIC(12, 3)`
- `DEFAULT`, `:=`, and `=` expressions
- commas and comment markers inside string literals
- line and block comments between declaration tokens
- escaped quoted parameter names
- parameterless procedures and functions
- function return declarations

## Compatibility and scope

The new behavior is explicitly gated to `databaseType === "xugu"`.

- Other databases keep their existing metadata queries.
- Other databases keep their existing procedure execution SQL generation.
- Other databases keep the previous parameter counter and DEFAULT column layout.
- The parameter details panel is only populated for XuguDB procedures and functions.
- No Go Agent, RPC protocol, Rust core, driver version, or release version is changed.
- Routine definitions are loaded on demand when the dialog opens, so schema tree loading does not fetch or parse CLOB definitions.

## Frontend

- [x] This PR changes the XuguDB procedure/function source and execution dialogs.
- [ ] Screenshot or recording is not attached yet; automated component coverage is included below.

## Validation

Automated validation:

- [x] `pnpm check`
- [x] Formatter
- [x] Lint
- [x] Vue/TypeScript typecheck
- [x] Full frontend suite: 824 test files, 7,377 tests passed
- [x] Focused XuguDB metadata, SQL generation, and component tests: 21 tests passed

Focused regression coverage includes:

- source loading without querying `ALL_ARGUMENTS`
- ordinary, quoted, and escaped parameter names
- nested precision and scale declarations
- string defaults containing commas and comment markers
- malformed and unterminated definitions
- function return types
- `OUT` placeholders and `IN OUT` inputs
- positional omission of trailing defaults
- named notation after omission of a middle default
- unchanged existing MySQL and SQL Server routine execution coverage

Real XuguDB verification was also performed with an ordinary, non-administrator user in a local business database:

- procedure and function definitions loaded successfully
- a positional procedure call with `IN`, `IN OUT`, and `OUT` arguments executed successfully
- a named procedure call omitting a middle default executed successfully
- function calls with all arguments and with an omitted trailing default returned the expected values
- temporary generic test routines were removed after verification

## Known limitation

The current XuguDB Go driver does not surface procedure `OUT` values as a result set. This PR makes the metadata visible and generates a valid call, but it intentionally does not claim to solve driver-level output retrieval.

## Change type

- [x] New feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Documentation
- [ ] CI / build

## Related issue

None.

### DIFF
--- a/apps/desktop/src/components/objects/ObjectSourceDialog.vue
+++ b/apps/desktop/src/components/objects/ObjectSourceDialog.vue
@@ -9,9 +9,11 @@ import { copyToClipboard } from "@/lib/common/clipboard";
 import { formatSqlForDisplay, type SqlFormatDialect } from "@/lib/sql/sqlFormatter";
 import { buildEditableObjectSource, buildExecutableObjectSourceStatements, executeObjectSourceSave, formatObjectSourceSaveError } from "@/lib/table/objectSourceEditor";
 import { loadObjectSourceWithRoutineFallback } from "@/lib/table/objectSourceLoad";
+import { xuguRoutineMetadataFromDefinition, type XuguRoutineMetadata } from "@/lib/table/routineParameters";
 import { executeWithProductionSqlGuard } from "@/lib/database/productionExecutionGuard";
 import * as api from "@/lib/backend/api";
 import QueryEditor from "@/components/editor/QueryEditor.vue";
+import RoutineMetadataPanel from "@/components/objects/RoutineMetadataPanel.vue";
 import { Button } from "@/components/ui/button";
 import { Dialog, DialogContent, DialogFooter, DialogHeader, DialogTitle } from "@/components/ui/dialog";
 import type { DatabaseType, ObjectSourceKind } from "@/types/database";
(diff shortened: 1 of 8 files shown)


t8y2:
Thanks for the contribution! Merged in 159e813, will be released in the next version.

PR #5664, opened 2026-08-07
fix(docs): move Documentation to the database context menu

possebon:
## 变更说明

Moves the **Documentation** entry from the Object Browser's table and view menus to the sidebar's database and schema context menu, next to **View Diagram**.

PR #5559 put the entry on table and view rows. That placement misrepresents what the action does, and it made the feature hard to find — users look for it on the database node, where the neighbouring schema-wide tools already live.

The action was never table-scoped:

```ts
// ObjectBrowser.vue, before this PR
function openDocs(row: ObjectBrowserRow) {
  connectionStore.docsSource = {
    connectionId: props.connection.id,
    database: props.database,
    schema: row.schema || selectedSchema.value,   // the row's table is discarded
  };
}
```

`DatabaseDocsDialog` accepts `prefillConnectionId`, `prefillDatabase` and `prefillSchema` — there is no table prop, so the viewer cannot deep-link to a table even in principle. Right-clicking `orders` and picking **Documentation** produced documentation for the whole schema.

**What changes for a user**

- Right-click a **database** node → **Documentation** documents every schema in that database. This scope was reachable from the CLI but had no UI entry point at all before this PR.
- Right-click a **schema** node → **Documentation** documents that schema.
- The entry no longer appears on table and view rows. **View Diagram** is untouched in both menus.

**How it works**

The new sidebar action mirrors `openDiagram` in `useSidebarTreeToolRuntime.ts`, minus the `tableName` the docs dialog has no prop for:

```ts
function openDocs() {
  const node = activeNode.value;
  if (!node.connectionId || !node.database) return;
  connectionStore.docsSource = {
    connectionId: node.connectionId,
    database: node.database,
    // A database node has no schema, and an absent schema is what tells the
    // collector to document every schema in the database.
    schema: node.schema,
  };
}
```

Forwarding `node.schema` verbatim is the whole mechanism. A database node has none, so `DatabaseDocsDialog.vue` sends an empty schema list and the collector documents everything — the behaviour its own comment already described as "everything the collector finds". A schema node has one, so the collector is narrowed. **No backend, dialog, or viewer code changes**; this is a menu-placement fix on top of the plumbing #5559 already shipped.

Registration is a single `items.push` inside `buildDatabaseSidebarMenu`, whose guard is `node.type === "database" || node.type === "schema"` — so one line serves both node types.

**One thing worth your call.** The entry reuses the existing `canOpenDiagram` gate, exactly as #5559 did, so the set of drivers showing it is unchanged. That coupling is not principled: documentation and diagrams are separate features, and a driver supporting one but not the other would be misrepresented. The principled fix is a `docs` key in `DATABASE_PRODUCT_CAPABILITY_KEYS`, but that list is backed by `database-drivers.manifest.json` and shared with the Rust side, so it needs a per-support-level default and per-driver overrides across every driver. That felt like your decision rather than something to bury inside a placement fix — happy to do it in a follow-up if you want it.

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

**Database node** — **Documentation** directly under **View Diagram**:

Documentation on a database node, below View Diagram

**Schema node** — the same single `items.push` serves both node types, because `buildDatabaseSidebarMenu` is guarded by `node.type === "database" || node.type === "schema"`:

Documentation on a schema node, below View Diagram

**What opens from a database node** — the whole database, not one schema:

The documentation viewer opened at database scope

## 验证

```console
$ pnpm vitest run apps/desktop/src/composables apps/desktop/src/components/objects apps/desktop/src/components/sidebar
  445 passed
$ pnpm test
  825 files, 7374 passed
$ pnpm typecheck
  clean
```

Two specs are added:

- `useSidebarTreeToolRuntime.docs.spec.ts` covers the behaviour that matters — a database node yields `schema: undefined` (the whole-database case), a schema node propagates its schema, and a node without a database is a no-op.
- `docsMenuPlacement.spec.ts` guards the placement itself: the entry is registered in `buildDatabaseSidebarMenu` and absent from both `buildObjectSidebarMenu` and `ObjectBrowser.vue`. It slices the source by builder function rather than matching a formatted line, so `oxfmt` cannot break it.

Manual verification against a live PostgreSQL instance (screenshots above): **Documentation** appears under **View Diagram** on both a database node and a schema node; invoking it from the database node loads the viewer with the database's tables and schema grouping, confirming the `schema: undefined` path reaches the collector intact; the entry is gone from table and view menus in the Object Browser, while **View Diagram** remains in both.

One limit worth stating rather than glossing: the database I verified against has a single schema, so the screenshots show database scope working end to end but do not by themselves demonstrate *multi-schema* breadth. That behaviour is the collector's pre-existing contract for an empty schema list, unchanged by this PR.

## 关联 Issue

None. This follows up #5559 based on a user reporting they could not find the feature in v0.5.77.

---

### DIFF
--- a/apps/desktop/src/components/objects/ObjectBrowser.vue
+++ b/apps/desktop/src/components/objects/ObjectBrowser.vue
@@ -6,7 +6,6 @@ import {
   ArrowDown,
   ArrowRightLeft,
   ArrowUp,
-  BookOpen,
   Braces,
   CheckSquare,
   Clipboard,
@@ -1463,16 +1462,6 @@ function openDiagram(row: ObjectBrowserRow) {
   };
 }
 
-function openDocs(row: ObjectBrowserRow) {
-  // The docs viewer documents the whole schema rather than one object, so the
-  // row only supplies which schema to collect.
-  connectionStore.docsSource = {
-    connectionId: props.connection.id,
-    database: props.database,
-    schema: row.schema || selectedSchema.value,
-  };
-}
-
 function openTableImport(row: ObjectBrowserRow) {
   if (row.type !== "TABLE") return;
   connectionStore.tableImportSource = {
@@ -2663,12 +2652,7 @@ function getTableMenuItems(item: ObjectBrowserRow): ContextMenuItem[] {
     ...(canOpenStructureEditor.value ? [{ label: t("contextMenu.editStructure"), action: () => openStructureEditor(item), icon: PencilRuler }] : []),
     ...(canRename(item) ? [{ label: t("contextMenu.renameObject"), action: () => requestRename(item), icon: Pencil }] : []),
     { label: t("contextMenu.newQuery"), action: () => openNewQuery(item), icon: TerminalSquare },
-    ...(canOpenDiagram.value
-      ? [
-          { label: t("diagram.open"), action: () => openDiagram(item), icon: Network },
-          { label: t("docs.title"), action: () => openDocs(item), icon: BookOpen },
-        ]
-      : []),
+    ...(canOpenDiagram.value ? [{ label: t("diagram.open"), action: () => openDiagram(item), icon: Network }] : []),
     ...(canOpenTableImport.value ? [{ label: t("contextMenu.importData"), action: () => openTableImport(item), icon: Download }] : []),
     { label: t("dataCompare.title"), action: () => openDataCompare(item), icon: ArrowRightLeft },
     { label: "", separator: true },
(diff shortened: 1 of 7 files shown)


t8y2:
Thanks for the contribution! Merged in bdf87ac, will be released in the next version.

PR #5704, opened 2026-08-09
fix(etcd): make gRPC inbound message limit configurable

tyloryang:
## Problem

DBX's etcd key browser can fail to load keys when a prefix contains many keys or large values. jetcd returns the matching key/value data in a single gRPC range response; once that response exceeds gRPC's default 4 MiB inbound limit, key loading stops with an error similar to:

```text
RESOURCE_EXHAUSTED: gRPC message exceeds maximum size 4194304: 6451419
```

The etcd cluster itself remains healthy and the same data can still be queried with `etcdctl`, which makes this look like a DBX key-loading problem rather than a server problem.

## Solution

- configure jetcd's `maxInboundMessageSize` for etcd connections
- default the receive limit to 32 MiB so the key browser can load larger prefix responses
- clamp configured values to a safe 1-256 MiB range
- persist the byte value in the existing `url_params` field as `grpc_max_inbound_message_size`
- expose the setting in the etcd advanced connection UI with translations
- add unit coverage for defaults, explicit values, bounds, and URL parameters

## Connection setting

The new field controls the maximum size of a single etcd response received while loading keys:

etcd gRPC receive limit setting

## Testing

- `gradle :etcd:test :etcd:shadowJar`
- `pnpm typecheck`
- live etcd 3.7.0 validation with seven values of about 900 KiB each:
  - a forced 4 MiB limit reproduced the key-loading failure on a 6,451,419-byte response
  - the new default 32 MiB limit loaded the same keys successfully

### DIFF
--- a/agents/drivers/etcd/src/main/java/com/dbx/agent/etcd/EtcdAgent.java
+++ b/agents/drivers/etcd/src/main/java/com/dbx/agent/etcd/EtcdAgent.java
@@ -87,6 +87,10 @@ public final class EtcdAgent {
     private static final Gson GSON = new Gson();
     private static final int DEFAULT_LIMIT = 100;
     private static final int RPC_TIMEOUT_SECONDS = 30;
+    static final int DEFAULT_GRPC_MAX_INBOUND_MESSAGE_SIZE = 32 * 1024 * 1024;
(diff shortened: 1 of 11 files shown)


t8y2:
Thanks for the contribution! Merged in f1c998c9c, will be released in the next version.

PR #5774, opened 2026-08-10
docs: fix Homebrew tap name for dbx-cli

solarhell:
The docs tell users to run `brew tap t8y2/dbx`, but that tap repository (`t8y2/homebrew-dbx`) does not exist, so the command fails with:

```
fatal: repository 'https://github.com/t8y2/homebrew-dbx/' not found
```

The actual tap lives at [t8y2/homebrew-tap](https://github.com/t8y2/homebrew-tap), so the correct command is `brew tap t8y2/tap`. Verified locally on macOS that `brew tap t8y2/tap && brew install dbx-cli` resolves and installs the formula.

Updated all five occurrences: README.md, README.zh-CN.md, packages/cli/README.md, and both CLI docs pages (en/cn).

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ DBX also provides a dedicated CLI package for terminal, script, and Codex workfl
 ```bash
 npm install -g @dbx-app/cli
 # or via Homebrew
-brew tap t8y2/dbx && brew install dbx-cli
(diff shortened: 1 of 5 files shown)


t8y2:
Thanks for the contribution! Merged in a9736f0, will be released in the next version.

PR #5776, opened 2026-08-10
feat(xugu): expose object type members in schema tree

mapan0424:
## Summary

This adds lazy schema-tree members for Xugu object types. Expanding an object `TYPE` now shows its attributes and declared procedures/functions, including parameter signatures and function return types.

## Design

- Uses the same catalog-first model as the DBeaver Xugu extension: `ALL_TYPE_ATTRS`, `ALL_TYPE_METHODS`, and `ALL_METHOD_PARAMS`.
- Resolves the owning type from `ALL_TYPES`, scoped to `CURRENT_DB_ID`. Exact schema/type matching is attempted first; the case-folded fallback preserves exact-match priority for quoted mixed-case identifiers.
- Some supported Xugu deployments expose `ALL_TYPES` but not the auxiliary type-member views. Only in that compatibility case, the implementation parses the top-level `OBJECT (...)` declaration from `ALL_TYPES.SPEC`. It never infers members from a type body.
- Object collection types remain leaf nodes and intentionally expose no object members.
- Type attributes and type methods are new read-only sidebar node kinds, so they are not treated as table columns or executable top-level routines.

## Isolation / compatibility

- The new RPC handler exists only in the Xugu agent.
- The desktop tree marks a type as expandable only when it was produced for an Xugu connection.
- Non-Xugu type nodes do not receive the marker, do not issue member metadata requests, and retain their existing source-opening behavior.

## Tests

- [x] `go test .` in `agents/drivers/xugu`
- [x] Live Xugu integration test with an administrator account and a normal business account. The test creates uniquely prefixed temporary object/collection types, verifies member metadata, then drops them.
- [x] Xugu unit coverage for RPC dispatch, catalog queries, current-database filtering, case-folded exact-match priority, object vs. collection types, filtering/limits, quoted identifiers, overloads, comments, and `IN` / `OUT` / `IN OUT` signatures.
- [x] `vitest run` for the affected sidebar suites (14 tests).
- [x] `pnpm lint` completed successfully; it reports only pre-existing repository warnings.

The repository-wide Vue typecheck is currently blocked by an unrelated missing `@dbx-app/mongo-shell` workspace package, so the focused sidebar tests are included to cover this UI path.

## UI

- [x] This PR changes sidebar behavior. Focused tests verify the Xugu-only expand affordance and that type bodies/non-Xugu types remain non-expandable.

## Related

No issue reference.


mapan0424:
Rebased onto the current `main` and addressed the previous frontend CI failures.

- Formatted the affected Xugu sidebar test with the repository formatter.
- Guarded the type-member metadata requests with required connection and database context before constructing the strict completion request.
- Resolved the sidebar click-test conflict by preserving the current Consul coverage and retaining the Xugu expandable-type assertion.

Local validation after the rebase:

- Focused Vitest sidebar suites: 15 passing tests.
- Xugu agent `go test .` and `go vet .`: passed.
- Formatter check and lint: passed (lint reports only existing repository warnings).

The updated branch is mergeable and CI is rerunning.

mapan0424:
Rebased this PR onto the current main branch after the package-member work merged.

During the rebase I found that the package and type member implementations had independently introduced the same completion-assistant request/response types and handler name. This update unifies that boundary instead of keeping competing definitions:

- `parent_type: "type"` routes Xugu object-type member requests to the type reader.
- `parent_type: "package"` routes package requests to the existing package reader.
- Existing Xugu package requests without `parent_type` still use the package path for compatibility with already-running desktop clients.
- The frontend sends the discriminator only for Xugu, so the existing Oracle package request contract is unchanged.

Validation:

- `go test .` and `go vet .` in `agents/drivers/xugu`
- opt-in live Xugu integration test covering object type attributes, methods, collection types, and cleanup
- focused frontend metadata/sidebar suite: 65 tests passed
- `pnpm lint` passes with existing repository warnings only
- full `pnpm typecheck` now has no type-member errors; it remains blocked locally by an unrelated missing `@dbx-app/mongo-shell` workspace dependency

The branch was force-pushed because rebasing rewrote the previous PR commits.

t8y2:
Thanks for the contribution! Merged in 1b37e97, will be released in the next version.

PR #5780, opened 2026-08-10
feat(xugu): expose detailed trigger metadata

mapan0424:
## Summary

Expose documented XuguDB trigger metadata in both table-level and schema-level trigger trees.

### Backend

- Extends `list_triggers` with trigger level, condition, language, enabled/valid state, comment, and creation time from `ALL_TRIGGERS`.
- Adds a trigger-only schema-object query for the schema trigger group, scoped to the current database and requested schema.
- Keeps source/DDL loading on the existing on-demand path; expanding a tree does not fetch trigger definitions.
- Preserves low-privilege degradation: a denied trigger catalog yields an empty trigger group instead of failing unrelated metadata browsing.

### Desktop

- Shows row/statement level plus disabled/invalid state in Xugu table-trigger labels.
- Preserves the metadata on schema trigger nodes and exposes it through a Xugu-only hover tooltip.
- Adds English and Simplified Chinese labels for the new metadata.
- Guards the presentation with `effectiveDatabaseType === "xugu"`; other database trigger labels and metadata presentation are unchanged.

## Validation

- `go test ./...`
- `go vet ./...`
- Targeted Vitest: 22 tests across table metadata and tree construction.
- `pnpm typecheck`
- `pnpm lint` (repository warnings only; no errors)
- Opt-in live regression test against XuguDB under both a non-admin account and an administrator account. The test creates row- and statement-level triggers, checks table-level and schema-level discovery, validates metadata fields and disabled state, and cleans up temporary objects.

## Scope

No query execution, DDL/source retrieval, or non-Xugu driver behavior is changed.


t8y2:
Thanks for the contribution! Merged in d0535f9, will be released in the next version.

PR #5896, opened 2026-08-11
fix(web): redirect bare base path to trailing slash for subpath deployments

cheshireez:
## Problem

When `DBX_PUBLIC_BASE_PATH` is set (e.g. `/dbx` or `/xxxx/rsu`), visiting the bare path (without the trailing slash) does not redirect to the trailing-slash form. Browsers then resolve relative assets against the site root, which leaves a blank homepage, and reverse proxies that only forward the `/prefix/` form return 404 for the bare path. Related: #2099, #5518.

PR #5538 fixed `/prefix/` returning index.html but did not add the bare-path redirect, so this scenario still fails.

## Fix

After nesting the app under the public base path, register an exact route at `{public_base_path}` that responds with a 308 Permanent Redirect to `{public_base_path}/`. The browser ends up at `/dbx/` (or `/xxxx/rsu/`) where index.html is served, and relative asset URLs keep working.

No path is hardcoded: the redirect target is always derived from the configured `DBX_PUBLIC_BASE_PATH`, so any single- or multi-segment prefix works. `/dbx` and `/xxxx/rsu` below are only illustrative examples.

## Verification

- Added regression test covering representative examples `/dbx` (single segment) and `/xxxx/rsu` (multi segment).
- `cargo test -p dbx-web`: 78 passed.
- Manual black-box run with `DBX_PUBLIC_BASE_PATH=/dbx`:
  - `/dbx` -> 308 `location: /dbx/`
  - `/dbx/`, `/dbx/index.html`, `/dbx/api/auth/check` -> 200
  - `curl -L /dbx` -> 200

## Notes for reverse proxies

The proxy must forward the bare prefix too (e.g. `location /dbx { proxy_pass http://127.0.0.1:4224; ... }` without a trailing slash), otherwise the request never reaches the app.

Refs #2099, #5518

### DIFF
--- a/apps/desktop/vite.config.ts
+++ b/apps/desktop/vite.config.ts
@@ -2,6 +2,7 @@ import { defineConfig } from "vite";
 import vue from "@vitejs/plugin-vue";
 import tailwindcss from "@tailwindcss/vite";
 import path from "path";
+import { publicBasePathRedirectPlugin } from "./vitePublicBasePathRedirect";
 
 const host = process.env.TAURI_DEV_HOST;
 const isTauri = !!host || !!process.env.TAURI_ENV_ARCH;
@@ -63,7 +64,7 @@ const backendUrl = process.env.DBX_BACKEND_URL || "http://localhost:4224";
 export default defineConfig(async () => ({
   root: import.meta.dirname,
   base: viteBase,
-  plugins: [vue(), tailwindcss()],
+  plugins: [publicBasePathRedirectPlugin(publicBasePath), vue(), tailwindcss()],
   resolve: {
     alias: {
       "@": path.resolve(import.meta.dirname, "./src"),
--- a/apps/desktop/vitePublicBasePathRedirect.ts
+++ b/apps/desktop/vitePublicBasePathRedirect.ts
@@ -1,4 +1,4 @@
-import type { Plugin } from "vite";
+import type { Connect, Plugin } from "vite";
 
 export function publicBasePathRedirectLocation(requestUrl: string | undefined, publicBasePath: string): string | null {
   if (!requestUrl || !publicBasePath || publicBasePath === "/") return null;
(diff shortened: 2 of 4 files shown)


t8y2 (commented):
Maintainer patch updated through `ef00709f`.

The original redirect fix remains unchanged: `dbx-web` preserves the bare-base query string, and Vite redirects only the exact bare public base path while passing root, trailing-slash, static asset, and API requests through.

CI follow-up root cause:
- The first regression test started a full Vite server and a proxy backend, then waited for both server lifecycles to close.
- That passed alone locally, but exceeded Vitest's default 5-second timeout when the complete frontend suite ran concurrently in GitHub Actions.

Follow-up fix:
- Extract the exact Connect middleware used by the Vite plugin and exercise it directly.
- Keep the same redirect/query/pass-through assertions without network startup, proxy sockets, environment mutation, or asynchronous cleanup.
- No timeout increase was added.

Verified:
- The focused Vitest file passed 10 consecutive runs (`20/20` tests); each run completed in about `114-137ms`, with test execution taking `3-4ms`.
- `pnpm exec oxlint apps/desktop/vitePublicBasePathRedirect.ts packages/app-tests/vitePublicBasePathRedirect.test.ts`.
- `pnpm typecheck`.
- `pnpm exec oxfmt --check apps/desktop/vite.config.ts apps/desktop/vitePublicBasePathRedirect.ts packages/app-tests/vitePublicBasePathRedirect.test.ts`.
- `git diff --check`.

No dependency, manifest, or lockfile changes.

t8y2 (commented):
Maintainer follow-up on `ef00709f`:

The subpath redirect fix now preserves the original query in both `dbx-web` and Vite development mode. The CI-only follow-up replaces the full Vite/proxy server smoke test with a direct test of the production Connect middleware, avoiding the previous 5-second full-suite timeout without increasing the timeout.

Verified with 10 repeated test runs (20/20 assertions each), `pnpm typecheck`, `oxlint`, `oxfmt --check`, and `git diff --check`.

t8y2:
Thanks for the contribution! Merged in 9808660, will be released in the next version.

PR #5906, opened 2026-08-11
feat(xugu): expose public synonyms in an isolated scope

mapan0424:
## Summary

Expose XuguDB database-global (public) synonyms without conflating them with a real schema. The agent now advertises a reserved protocol scope, while private synonyms remain schema-scoped and the desktop tree renders the reserved scope as `Public synonyms`.

## Protocol and metadata behavior

- Public synonyms are detected with a current-database probe and are added to `list_schemas` only when present. This works whether or not a real `GUEST` schema exists.
- Public object listing and source lookup are routed only through `__DBX_XUGU_PUBLIC_SYNONYMS__`; `GUEST` is treated as an ordinary private schema.
- Public/private aliases with the same name are disambiguated by their scope identity. Exact catalog spelling is retained, with the existing case-insensitive fallback confined to the selected scope.
- Public source reconstruction emits `CREATE PUBLIC SYNONYM ... FOR ...;`; private source reconstruction remains `CREATE SYNONYM ... FOR ...;`.
- Public-scope probe failures are non-fatal, so restricted or older catalogs still expose ordinary schemas.

## Desktop behavior

- The raw scope key remains on tree nodes for routing and API calls.
- The sidebar displays `Public synonyms` instead of the implementation key.
- Schema-node lookup compares the raw schema identity, preserving deep links and refresh behavior.
- The Xugu schema-tree cache version is bumped so stale `GUEST`-based entries are not reused.

## Test matrix

- No real `GUEST` + no public synonym.
- No real `GUEST` + public synonym.
- Real `GUEST` + public synonym.
- Private-only and public-only listing paths.
- Real `GUEST` private lookup remains private.
- Public/private same-name selection is deterministic.
- `list_schemas -> list_objects -> get_object_source` is covered by an opt-in live XuguDB test.

## Validation

- `go test ./...`
- `go test -race ./...`
- `go build ./...`
- Live XuguDB test passed with a temporary public/private synonym fixture and cleanup.
- Targeted desktop Vitest passed (2 tests).
- Oxfmt and `git diff --check` passed.

The changes are isolated to Xugu agent metadata plus the desktop tree display/routing needed for the explicit scope; no other database driver behavior is changed and no dependencies were added.

### DIFF
--- a/agents/drivers/xugu/main.go
+++ b/agents/drivers/xugu/main.go
@@ -28,6 +28,11 @@ const defaultMaxRows = 1000
 const defaultXuguPort = 5138
 const legacyAgentSessionID = "__legacy__"
 const maxAgentSessions = 256
+
+// xuguPublicSynonymScope is a protocol-only namespace for database-global
+// synonyms. It is deliberately not a real schema name (and must never be
+// interpreted as one by metadata queries).
+const xuguPublicSynonymScope = "__DBX_XUGU_PUBLIC_SYNONYMS__"
 const xuguListDatabasesSQL = `
 SELECT DB_NAME
 FROM ALL_DATABASES
@@ -37,6 +42,11 @@ SELECT SCHEMA_NAME
 FROM ALL_SCHEMAS
 WHERE DB_ID = CURRENT_DB_ID
 ORDER BY SCHEMA_NAME`
+const xuguPublicSynonymScopeProbeSQL = `
+SELECT 1
+FROM ALL_SYNONYMS
+WHERE DB_ID = CURRENT_DB_ID
+  AND IS_PUBLIC = TRUE`
 const xuguCatalogTableNameSelectSQL = `
 SELECT s.SCHEMA_NAME, t.TABLE_NAME
 FROM ALL_TABLES t
@@ -48,7 +58,7 @@ JOIN ALL_SCHEMAS s ON s.DB_ID = q.DB_ID AND s.SCHEMA_ID = q.SCHEMA_ID
 WHERE s.DB_ID = CURRENT_DB_ID
   AND q.IS_SYS = FALSE`
 const xuguCatalogSynonymSelectSQL = `
-SELECT CASE WHEN y.IS_PUBLIC = TRUE THEN 'GUEST' ELSE s.SCHEMA_NAME END AS SCHEMA_NAME,
+SELECT CASE WHEN y.IS_PUBLIC = TRUE THEN '__DBX_XUGU_PUBLIC_SYNONYMS__' ELSE s.SCHEMA_NAME END AS SCHEMA_NAME,
        y.SYNO_NAME, t.SCHEMA_NAME AS TARGET_SCHEMA, y.TARG_NAME, y.IS_PUBLIC
 FROM ALL_SYNONYMS y
 LEFT JOIN ALL_SCHEMAS s ON s.DB_ID = y.DB_ID AND s.SCHEMA_ID = y.SCHEMA_ID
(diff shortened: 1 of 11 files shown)


t8y2 (changes requested):
当前实现还不能安全合入，主要阻塞在 public synonym 的 namespace 身份设计上：

`GUEST` 同时被当作真实 schema 名和 synthetic public scope。`list_schemas` 并不会在 catalog 中没有 `GUEST` 时创建这个 synthetic 节点，因此 public synonym 可能完全不可发现；反过来，如果数据库确实有名为 `GUEST` 的 schema，`list_objects` / source lookup 又会因为 `schema == GUEST` 无条件切到 `IS_PUBLIC = TRUE`，导致该真实 schema 的 private synonym 不可达。public/private 同名时，现有 `schema/name/object_type` 标识也无法消歧。

这里需要先明确协议和 UI 中 synthetic scope 的独立身份，或者定义两类对象在同一 scope 内的共存与同名消歧规则；这属于产品/协议语义选择，不适合维护者直接做局部代修。

请补充完整协议矩阵测试：
- catalog 中有 / 无真实 `GUEST` schema；
- private-only / public-only；
- public/private 同名；
- 两类 synonym 的 `list_schemas -> list_objects -> get_object_source` 路径。

本 review 绑定 exact head `943bd39e4`。当前改动未新增依赖，`git diff --check` 已通过，exact-head agents CI 已成功。

mapan0424:
Thanks for the detailed review. I updated the PR to make the public-synonym namespace an explicit protocol/UI scope instead of overloading the real `GUEST` schema.

What changed:
- Added the reserved scope `__DBX_XUGU_PUBLIC_SYNONYMS__` and switched public-synonym catalog rows, object listing, and source lookup to use it. A real `GUEST` schema now follows the normal private-synonym path.
- Added a low-risk public-synonym probe to `list_schemas`. The scope is added only when a public synonym exists; probe errors are ignored so restricted/older catalogs still expose ordinary schemas.
- Kept the raw scope value on tree nodes for routing, but render it as `Public synonyms` in the desktop tree. The schema lookup now compares the raw schema identity, so deep links and refreshes continue to work with the friendly label. Xugu schema-cache version was bumped to avoid stale `GUEST` entries.
- Kept exact/case-insensitive synonym resolution scoped by the namespace, including deterministic disambiguation when private and public aliases share a name.

Coverage added:
- Go protocol matrix: no real GUEST/public absent, public-only without GUEST, public plus real GUEST, real GUEST private routing, public/private same-name selection, and DDL source paths.
- Opt-in live Xugu test covers schema discovery, private/public object listing, and both source DDL forms against a real database.
- Desktop unit test covers reserved-scope identity/display behavior.

Validation performed:
- `go test ./...` passed.
- `go test -race ./...` passed.
- `go build ./...` passed.
- Live test passed with XuguDB (`TestLivePublicSynonymScope`).
- Targeted desktop Vitest passed (2 tests).
- Oxfmt check and `git diff --check` passed.

The commit is `b233459f4` on the PR branch. No new dependencies were added.

t8y2 (commented):
Maintainer patch pushed from `d59b341f` to `f6d1261ce6` against merge base `5737bb49` (current `main` was re-fetched before the push).

The patch closes the remaining safety and performance gaps without changing the contribution's intended behavior:

- Replaced the legal schema-like sentinel with a NUL-prefixed protocol identity, so a real schema named `__DBX_XUGU_PUBLIC_SYNONYMS__` can coexist with the synthetic public-synonym scope.
- Combined ordinary schema discovery and public-scope discovery into one normal-path metadata query. If the combined query is unsupported or restricted, the agent falls back to the ordinary `ALL_SCHEMAS` query; no extra request is added to the normal object-tree path.
- Kept the synthetic scope out of SQL catalog values and reconstructs it from `IS_PUBLIC` after scanning, avoiding sentinel-dependent SQL behavior.
- Restricted the synthetic tree node to copy/refresh actions and added guards against setting it as the default schema or dropping it.
- Restored the unrelated Hive coverage-row change.

Validation completed on the patched exact head:

- `go test ./...`
- `go test -race ./...`
- `go build -o /workspace/cogvlm/tty/tmp/dbx-pr5906-xugu-agent .`
- Targeted Vitest: 3 files, 7 tests passed
- `pnpm typecheck`
- scoped `oxlint --vue-plugin`
- scoped `oxfmt --check`
- `git diff --check`

No dependency changes, polling, watchers, background I/O, or normal-path extra database requests were added.

t8y2 (commented):
CI follow-up on `7613240e00`: the failed `:common:test` showed that the contributor's deletion of the retired Java `hive` coverage entry was required. I had incorrectly restored that line while removing what looked like unrelated scope.

The coverage test discovers metadata drivers from the current source tree; only `hive-go` remains, so the matrix must not contain `hive`. I restored the contributor's deletion and reproduced the test's exact discovered/matrix set comparison locally with no missing or extra keys. The Xugu patch and its validation are unchanged.

t8y2 (approved):
Exact-head maintainer patch and CI verified. Approving for squash merge.

t8y2:
Thanks for the contribution! Merged in ea9a27ce11, will be released in the next version.

PR #5932, opened 2026-08-11
fix(kingbase): fall back from LIMIT pagination when a TOP clause bounds the query

Abeautifulsnow:
Fixes **t8y2/dbx#5910** (P0): running normal SQL on KingbaseES in **SQL Server
compatibility mode** (`compat=sqlserver`) failed with
`kb: multiple TOP/LIMIT clauses not allowed`, while the official Kingbase
management tool executed the same statement successfully.

**DBX 0.5.81** desktop · **KingbaseES V009R001C010** · driver **0.1.46**

## Root cause

Commit `378674253` (`fix(kingbase): optimize large query pagination`) made
Kingbase prefer server-side `LIMIT/OFFSET` pagination over the Agent cursor for
the first page. `add_standard_limit()` only detected an existing top-level
`LIMIT` and did not recognize a top-level `TOP` clause. In SQL Server
compatibility mode `TOP` is a real clause, so a statement such as
`SELECT TOP 100 * FROM t ORDER BY id` was rewritten to
`SELECT TOP 100 * FROM t ... LIMIT {pageSize} OFFSET {offset}` — a mix the
server rejects at parse time. Official tools do not append `LIMIT`, which is why
they succeed.

### 1. Grid/query path — Agent-cursor fallback for top-level `TOP`

`build_paginated_query_sql` (`LimitOffset` arm) returns `unsupported` when the
database is Kingbase and the statement has a top-level `TOP`
(`has_top_level_top`, based on the existing `top_level_sql_tokens` scanner). The
pagination plan then falls back to the existing bounded Agent result-session
path. The guard is **Kingbase-only** (a column named `top` on other dialects is
unaffected) and top-level only (a `TOP` inside a subquery still permits outer
`LIMIT/OFFSET`).

### 2. Export path — single-execution fallback (no Agent cursor)

Query-result export runs with `use_agent_cursor=false`. When offset pagination
is impossible for a Kingbase TOP statement, the export now executes the query
**once** with the whole result streamed to file, instead of failing with
`Streaming export is unsupported`. A `single_execution` flag on the pagination
plan (`#[serde(skip)]`, never serialized to the frontend) signals the export
loop to size the page to the enforceable row bound and stop after the first
response.

### 3. Keyset must never drop the TOP bound (P0)

`safe_keyset_candidate()` now rejects `select.top.is_some()`. Previously a
simple `SELECT TOP 100 * FROM users` qualified for keyset export, whose SQL
reconstruction dropped the user's `TOP 100` — combined with an unbounded page
this could export the whole table.

### 4. Single-execution uses an enforceable row bound, never `i32::MAX` (P1)

`top_level_top_row_count()` extracts a **concrete** bound only for a bare
literal (`TOP n`) or a parenthesized form that is exactly one integer literal
followed by `)` (`TOP (100)`, `TOP ( 100 )`). It refuses:

- `TOP n PERCENT` and `TOP n WITH TIES` (server may return more than `n` rows)
- parenthesized expressions — `TOP (100 + 1)` returns 101 rows, so treating the
  leading `100` as the bound would silently truncate the export
- `TOP` inside a subquery, and absent `TOP`

`single_execution_row_bound = min(concrete TOP count, configured row_limit)`.
Without any enforceable bound (e.g. unbounded `TOP PERCENT`/`WITH TIES` and no
row limit) the export is **rejected honestly** at the streaming guard rather
than buffering an unbounded result in memory.

## Files changed

- `crates/dbx-core/src/query_result_sql.rs` — `has_top_level_top`,
  `top_level_top_row_count`, `single_execution` plan field, Kingbase TOP guard
  in `build_paginated_query_sql` and the plan fallback branches.
- `crates/dbx-core/src/query_result_export.rs` — keyset rejection of TOP,
  `single_execution_row_bound` / single-execution guard and loop handling.

## Tests

- `query_result_sql` (130): TOP variants → Agent-session fallback with
  `use_agent_cursor=true`; single-execution plan with `use_agent_cursor=false`;
  subquery-TOP still server-paginated; non-Kingbase `top` column unaffected;
  `top_level_top_row_count` bound matrix (`TOP n`/`TOP(n)`/`TOP (n)`/`TOP ( 100 )`
  → bound; PERCENT/WITH TIES/`TOP (100 + 1)`/`TOP (100 * 2)`/subquery → `None`).
- `query_result_export` (38): non-keyset TOP join export falls back to
  single execution with bound 100; single-table `SELECT TOP 100 * FROM users`
  never keysets and is bounded; PERCENT/WITH TIES require a row limit;
  `TOP (100 + 1)` export requires a row limit or cursor; no-TOP Kingbase keeps
  offset pagination.
- `cargo clippy -p dbx-core --lib -- -D warnings` clean.
- Full `cargo test -p dbx-core --lib` (this env): 4278 passed, 4 failed — the 4
  are pre-existing Windows-only failures unrelated to this change
  (`agent_manager` path separator, `docs_export_bundle_is_current`, two legacy
  XLS parsers). CI (`--features "$RUST_FAST_FEATURES"`) is the source of truth.

## Notes for reviewers

- Root cause was confirmed by code inspection and is consistent with the issue's
  error text; a real Agent SQL log would upgrade it to direct evidence. The
  regressions added here cover the decision points so the same class of bug
  cannot reappear.
- Scope is intentionally Kingbase-only; no change to `LimitOffset` pagination
  for any other database.

### DIFF
--- a/crates/dbx-core/src/query_result_export.rs
+++ b/crates/dbx-core/src/query_result_export.rs
@@ -461,6 +461,22 @@ fn supports_streaming_offset_pagination(request: &QueryResultExportRequest, page
         && !first_sql.trim().eq_ignore_ascii_case(second_sql.trim())
 }
 
+/// True when a non-agent export can still stream this query by executing it
+/// exactly once and writing the whole result. Used for statements that cannot
+/// be rewritten with LIMIT/OFFSET (Kingbase SQL Server compatibility mode
+/// top-level TOP), where the TOP clause itself bounds the row count.
+fn supports_single_execution_export(request: &QueryResultExportRequest) -> bool {
+    let plan = build_query_pagination_execution_plan(QueryPaginationExecutionPlanOptions {
+        sql: request.sql.clone(),
+        query_base_sql: request.query_base_sql.clone(),
+        database_type: Some(request.database_type),
+        pagination: QueryPagination { limit: request.page_size.max(1), offset: 0, session_id: None },
+        use_agent_cursor: false,
+        first_page_uses_actual_sql: true,
+    });
+    plan.single_execution
+}
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 struct SafeKeysetCandidate {
     schema: Option<String>,
@@ -650,7 +666,15 @@ async fn export_query_result_core_inner(
     let mut offset: usize = 0;
     let mut wrote_text_header = false;
     let mut keyset_plan = build_keyset_plan(state, request).await;
-    if keyset_plan.is_none() && !request.use_agent_cursor && !supports_streaming_offset_pagination(request, page_size) {
+    // A statement that can only be exported with a single execution (e.g.
+    // Kingbase SQL Server compat top-level TOP) sizes its page to the whole
+    // remaining row budget and stops after the first response.
+    let single_execution_capable = supports_single_execution_export(request);
+    if keyset_plan.is_none()
+        && !request.use_agent_cursor
+        && !supports_streaming_offset_pagination(request, page_size)
+        && !single_execution_capable
+    {
         return Err(STREAMING_PAGINATION_UNSUPPORTED_ERROR.to_string());
     }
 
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch pushed in 070ba66ca.

Root cause: the non-cursor fallback could request the entire literal `TOP` bound in one response, recreating the large-result memory spike, and the `WITH TIES` check treated unrelated later `WITH` tokens such as table hints as TOP modifiers.

Patch: single-execution export is now limited to one normal export page; larger bounds require a real result session instead of being silently truncated or buffered. TOP modifier parsing now rejects only adjacent `PERCENT` or `WITH TIES` clauses.

Checks: `cargo fmt --check`; `cargo test --no-default-features kingbase_` (41 focused unit tests plus the matching integration test passed); `git diff --check`.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in b814b3d9be, will be released in the next version.

PR #5937, opened 2026-08-11
feat(xugu): organize package members and preserve body access

mapan0424:
## Summary

This change improves the Xugu package tree without changing package handling for other database providers. A Xugu package specification and package body represent one logical package, but exposing both as separate top-level rows makes the tree look duplicated and makes it unclear which node should be managed.

The package body remains fully accessible as a source variant from the package actions, while package members are organized into focused procedure and function folders.

### User-facing rationale

- Shows one package entry instead of repeating the same package name for the specification and body.
- Makes larger packages easier to browse by separating procedures from functions.
- Preserves overloads by keeping the complete member signature in each leaf node.
- Keeps the package body easy to inspect through an explicit `View Source (Package Body)` action.
- Avoids empty folders when a package contains only procedures or only functions.

### Implementation rationale

- The specification and body are coalesced only for Xugu metadata responses; the body availability and validity are retained on the package node.
- Package member grouping is enabled only when the active database type is Xugu. Existing Oracle and other provider behavior remains flat and unchanged.
- Package member source routing continues to use the owning package name and signature, so overload-specific source loading is preserved.
- The package body action creates a source target of `PACKAGE_BODY` without introducing a duplicate tree object.

## Files changed

- `apps/desktop/src/lib/table/tableTree.ts`: coalesces Xugu package specification/body metadata and preserves body state.
- `apps/desktop/src/lib/sidebar/packageMembers.ts`: builds Xugu procedure/function member folders while retaining the existing non-Xugu layout.
- `apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue`: adds the package-body source action.
- `apps/desktop/src/stores/connectionStore.ts`: passes the effective database type into package member construction.
- Related type definitions and regression tests.

## Validation

- 40 targeted Vitest cases passed, covering package/body coalescing, body-only metadata, invalid state propagation, overloads, source routing, Xugu grouping, and non-Xugu compatibility.
- Targeted Oxlint passed.
- `git diff --check` passed.
- Full Vue typecheck still reports the repository baseline errors for the missing `@dbx-app/mongo-shell` package and existing implicit-any diagnostics; no errors are reported in the changed files.

## Scope

No database schema, SQL, driver, or non-Xugu metadata behavior is changed. This PR is limited to the Xugu package tree presentation and source navigation.

### DIFF
--- a/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
+++ b/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
@@ -1728,15 +1728,17 @@ async function refreshDropTableChildObjectPreviewSql() {
   dropTableChildObjectPreviewSql.value = options ? await buildDropTableChildObjectSql(options).catch(() => "") : "";
 }
 
-function openObjectSourceDialog(initialEditing: boolean) {
+function openObjectSourceDialog(initialEditing: boolean, viewPackageBody = false) {
   const node = activeNode.value;
   if (!node.connectionId || !node.database) return;
+  if (viewPackageBody && (node.type !== "package" || node.xuguPackageBodyAvailable !== true || currentDatabaseType() !== "xugu")) return;
+  const sourceNode: TreeNode = viewPackageBody ? { ...node, type: "package-body" } : node;
   // TYPE/TYPE_BODY only have a source implementation on Xugu; PostgreSQL-family
   // connections list user-defined types without a CREATE TYPE getter this cycle.
-  if ((node.type === "type" || node.type === "type-body") && !supportsTypeObjectSource(currentDatabaseType())) return;
+  if ((sourceNode.type === "type" || sourceNode.type === "type-body") && !supportsTypeObjectSource(currentDatabaseType())) return;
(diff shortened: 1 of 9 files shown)


t8y2 (commented):
Maintainer patch pushed in `bf336bf4` (from `387cb0e0`).

Root cause: paginated tree merging treated an existing package node as a duplicate and discarded the later page’s package-body metadata. When the package spec appeared on page one and its body on page two, the node never gained body availability or validity state.

Fix: keep the existing node identity, merge `xuguPackageBodyAvailable`, `xuguPackageBodyValid`, and aggregate validity from later pages, and add a page-size-1 regression covering spec/body split across pages.

Validation:
- `pnpm exec vitest run apps/desktop/src/stores/__tests__/connectionStore.metadataLoading.spec.ts --config vitest.config.ts --no-file-parallelism` — 49/49 passed
- `pnpm typecheck` — passed
- `git diff --check` — passed

No dependencies, requests, polling, or background I/O were added.

t8y2:
Thanks for the contribution! Merged in 275b4d58, will be released in the next version.

PR #5993, opened 2026-08-12
fix(sidebar): don't force-expand located tables before their groups load

### LINKED ISSUE #5850: [Bug] 侧边栏定位表展开不统一
### 数据库类型和版本

通用

### 问题描述

定位某个未展开的表时箭头变成了向下，但是未展开具体内容，定位展开过的表时会正常展开，个人建议定位到表就行了，不用展开展示字段索引这些信息
Image

### 支持信息

DBX 版本: v0.5.81
运行环境: 桌面端
操作系统: Windows 10.0.26220.7872
系统架构: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## What

Fixes #5850

Locating a table in the sidebar (via the SQL editor cursor / active tab context) always forced every ancestor node in its tree path — including the table itself — into the expanded state, regardless of whether that node's children were actually loaded yet.

For a table that had never been expanded before, this flipped the arrow to "expanded" while the Columns/Indexes/etc. groups underneath were never fetched, so the tree showed an expanded arrow with nothing inside. Previously-expanded tables looked fine because their children were already cached — hence the "不统一" (inconsistent) behavior reported in the issue.

## Fix

`locateActiveTabInSidebar` now only auto-expands a node in the located path when `connectionStore.canUseLoadedTreeNodeToggle(node)` is true — the same check the normal click-to-expand handler (`SidebarTreeRuntimeHost.vue`'s `toggle()`) uses to decide whether cached children can be shown without a reload. Database/schema ancestors are unaffected since `ensureTreeLoadedForTarget` already populates their children before this loop runs; only a not-yet-expanded table/collection leaf is left collapsed, matching the reporter's own suggestion that locate should just reveal/select the table rather than force its structure open.

## Testing

- `pnpm exec vitest run apps/desktop/src/lib/__tests__/sidebar/treeLoadedChildrenMarker.spec.ts` — pass (covers the reused `canUseLoadedTreeNodeToggle` semantics)
- `pnpm exec vitest run apps/desktop/src/stores/__tests__/connectionStore.treeMemory.spec.ts apps/desktop/src/stores/__tests__/connectionStore.metadataLoading.spec.ts` — pass
- `pnpm exec vue-tsc --noEmit --project apps/desktop/tsconfig.json` — pass

I wasn't able to exercise this against a live DB connection in my current environment (no GUI/Tauri runtime available here), so I'd appreciate a sanity check from anyone who can reproduce with a real connection: locate a table via cursor position in the SQL editor that has never been expanded in the sidebar before, and confirm it now stays collapsed (arrow not flipped) instead of showing an empty expanded state.

### DIFF
--- a/apps/desktop/src/components/sidebar/ConnectionTree.vue
+++ b/apps/desktop/src/components/sidebar/ConnectionTree.vue
@@ -1065,7 +1065,11 @@ async function locateActiveTabInSidebar() {
   if (!nodePath) return;
 
   for (const ancestor of nodePath) {
-    if (!ancestor.isExpanded) {
+    // Only flip the arrow when this node's own children are already loaded
+    // (e.g. by ensureTreeLoadedForTarget above). Forcing isExpanded on a
+    // table/collection whose column/index groups were never fetched shows an
+    // "expanded" arrow with no content underneath (issue #5850).
+    if (!ancestor.isExpanded && store.canUseLoadedTreeNodeToggle(ancestor)) {
       ancestor.isExpanded = true;
     }
   }


t8y2:
Thanks for the contribution! Merged in 463c3cd7f0, will be released in the next version.

PR #6000, opened 2026-08-12
fix(gaussdb-m): quote reserved keyword aliases and use table_privileg…

AndrewDi:
- Quote column aliases (user, host, plugin) in postgresListRolesSql and kingbaseListRolesSql with double quotes to avoid reserved keyword errors.
- Add gaussdbMListRolesSql with backtick-quoted aliases for GaussDB M mode.
- Add gaussdbMShowGrantsSql that uses information_schema.table_privileges instead of role_table_grants (which does not exist in M mode), uses concat() instead of || (treated as OR in MySQL mode), and avoids CAST(... AS text) since text is a reserved keyword in M mode.
- Add gaussdbMUserAdminProvider and update provider resolution to detect GaussDB M mode via gaussdbConnectionMode().

### DIFF
--- a/apps/desktop/src/lib/__tests__/database/databaseUserAdmin.spec.ts
+++ b/apps/desktop/src/lib/__tests__/database/databaseUserAdmin.spec.ts
@@ -1,9 +1,34 @@
 import { describe, expect, it } from "vitest";
-import { dorisGrantPrivilegesSql, dorisGrantsResult, dorisPrivilegeTargetSql, dorisUsersResult, getDatabaseUserAdminProvider, kingbaseShowGrantsSql, mysqlPrivilegeSelectionFromGrants } from "@/lib/database/databaseUserAdmin";
-import type { QueryResult } from "@/types/database";
+import {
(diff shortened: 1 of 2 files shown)


AndrewDi:
@zipg PTAL

t8y2 (commented):
Maintainer test patch pushed after rebasing onto the latest main.

Previous head: 84ddf952f01e1327ab8e5f2002f5049768885d25
New head: c648800ce4b0d5aeda47d075bb85719421670cc4

Added focused coverage for GaussDB M-mode provider resolution, quoted result aliases, table_privileges and concat-based grant SQL, and unchanged native GaussDB/PostgreSQL/Kingbase provider behavior. No additional database requests were introduced.

Validation: 13 focused tests, full frontend typecheck, scoped lint, changed-file formatting, and git diff --check. The SQL was structurally validated; a matching live GaussDB M-mode instance/vendor JDBC environment was not available.

t8y2:
Thanks for the contribution! Merged in 6429334549, will be released in the next version.

PR #6010, opened 2026-08-12
fix(editor): keep full-width characters from crashing SQL formatting

### LINKED ISSUE #5988: [Bug] sql格式化失败
### 数据库类型和版本

mysql8.0

### 问题描述

sql格式化失败
Image

### 支持信息

DBX 版本: v0.5.82
运行环境: 桌面端
操作系统: Windows 10.0.26200.8973
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Summary

- `sql-formatter`'s tokenizer throws a bare `Parse error: Unexpected "..."` when it hits characters it can't tokenize at all — e.g. full-width `≠`/`（）`, which MySQL accepts in identifiers/expressions but `sql-formatter` doesn't recognize.
- `isSqlFormatterParseError` (used by `formatSqlForEditing`, the function the editor's "Format SQL" command calls) only recognized the *grammar-level* nearley error shape (`Parse error at token: ...` with `.offset`/`.token`), not this *lexer-level* one, so the tokenizer error propagated and surfaced as a "SQL 格式化失败" toast instead of leaving the SQL unchanged like other unparseable-editor-SQL cases already do.
- Widened the check to also recognize the `Parse error: Unexpected ...` shape, so it falls back to the original SQL instead of throwing — consistent with how the existing "incomplete mid-edit SQL" case is already handled.

## Test plan

- [x] Added a regression test in `sqlFormatter.spec.ts` using the exact repro SQL from the issue (full-width `≠` and `（）`), asserting `formatSqlText` still throws (raw formatter behavior unchanged) while `formatSqlForEditing` now resolves to the original SQL unchanged.
- [x] `pnpm test` (`apps/desktop/src/lib/__tests__/sql/sqlFormatter.spec.ts`) — 19/19 passing.
- [x] `vue-tsc --noEmit` — clean.
- [x] Manually verified in the desktop app: pasted the repro SQL into the query editor and clicked "格式化" — before the fix it threw the error toast, after the fix it no-ops silently (no crash) and normal SQL still formats correctly.

Fixes #5988

### DIFF
--- a/apps/desktop/src/lib/__tests__/sql/sqlFormatter.spec.ts
+++ b/apps/desktop/src/lib/__tests__/sql/sqlFormatter.spec.ts
@@ -57,6 +57,13 @@ describe("sqlFormatter", () => {
     await expect(formatSqlForEditing(sql, "mysql")).resolves.toBe(sql);
   });
 
+  it("keeps editor SQL unchanged when it contains full-width characters the tokenizer can't parse", async () => {
+    const sql = "update t set a=concat(t.入池时间（审核通过时间）,' 00:00:00') where t.入池时间（审核通过时间） ≠ '';";
+
+    await expect(formatSqlText(sql, "mysql")).rejects.toThrow("Parse error: Unexpected");
+    await expect(formatSqlForEditing(sql, "mysql")).resolves.toBe(sql);
+  });
+
   it("keeps non-parse editor formatting failures visible", async () => {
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in 2a26ea5429, will be released in the next version.

PR #6017, opened 2026-08-12
fix(sidebar): reuse existing tab when reopening a stored procedure source

### LINKED ISSUE #6003: [Bug] 侧边栏多次点击同一个存储过程重复打开标签页
### 数据库类型和版本

Mysql8

### 问题描述

期望相同的存储过程不重复打开新的标签页

### 支持信息

DBX 版本: v0.5.82
运行环境: 桌面端
操作系统: Windows 10.0.26200.8894
系统架构: x86_64

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

_No response_

q396921921:
## Summary

Repeatedly clicking the same stored procedure in the sidebar opened a new "Source - x" tab every time instead of reusing the already-open one — tables/views and the Ctrl/Cmd+P quick-open path already deduped correctly.

The cause was `openObjectSourceDialog` (`SidebarTreeRuntimeHost.vue`) passing `{ forceNew: true }` to `createTab`, bypassing `createTab`'s built-in lookup of an existing tab by connection/database/schema/catalog identity. After dropping that flag, a reused tab needs to actively refresh its content (`updateSql` + `markTabClean`), so that if the stored procedure was `ALTER`ed externally, the reused tab doesn't keep showing stale source.

Also updated the corresponding assertion in `packages/app-tests/sidebarContextMenuHost.test.ts` — that test was a regex match against the literal source of the `openObjectSourceDialog` function body, and it had `forceNew: true` plus "does not call `updateSql`" baked in as expected behavior. Checking the history, that assertion was only mechanically kept in sync through variable renames across past refactors (`result.source` → `editableSource`, etc.), not an intentional design decision — so it had effectively locked the bug in as expected behavior. Fixed that assertion along with the bug.

## Change Type

- [x] Bug fix

## Frontend Changes

- [x] This PR touches the frontend (interaction behavior); verification steps are described below.

## Validation

- [x] `make check` (`pnpm run check`: format / lint / typecheck / test) — all pass
- [x] `make cargo-check-fast` — pass
- [x] Relevant tests pass (and the corresponding assertion in `sidebarContextMenuHost.test.ts` was updated)

Manual repro and verification:
1. Built a fresh desktop app from current `main` (before the fix), connected to MySQL, expanded a stored procedure in the sidebar, and clicked the same stored procedure 4 times — reproduced the bug: each click opened a new "Source - x" tab.
2. Switched to this branch, rebuilt, repeated the same 4 clicks — only 1 tab remained.
3. Also verified content refresh on reuse: after `ALTER`ing / recreating the stored procedure in the database and clicking again, the reused tab shows the latest source instead of staying on stale content (this is why `updateSql`/`markTabClean` were kept rather than simply removing `forceNew: true`).

## Related Issue

Fixes #6003

### DIFF
--- a/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
+++ b/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
@@ -1813,15 +1813,24 @@ function openObjectSourceDialog(initialEditing: boolean, viewPackageBody = false
           databaseType,
           signature: sourceNode.signature,
         });
-        const tabId = queryStore.createTab(connectionId, database, `Source - ${node.label}`, "query", schema, editableSource, node.catalog, { forceNew: true });
         const sourceIsEditable = raw.editable !== false && !["SEQUENCE", "TRIGGER", "TYPE", "TYPE_BODY"].includes(resolvedType);
         if (sourceIsEditable) {
-          queryStore.setObjectSource(tabId, {
+          queryStore.openObjectSourceTab({
+            connectionId,
+            database,
+            title: `Source - ${node.label}`,
             schema,
-            name: objectName,
-            objectType: resolvedType,
-            signature: node.signature,
+            catalog: node.catalog,
+            sql: editableSource,
+            objectSource: {
+              schema,
+              name: objectName,
+              objectType: resolvedType,
+              signature: node.signature,
+            },
           });
+        } else {
+          queryStore.createTab(connectionId, database, `Source - ${node.label}`, "query", schema, editableSource, node.catalog, { forceNew: true });
         }
       })
       .catch((e: any) => {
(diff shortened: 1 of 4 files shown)


t8y2 (commented):
Maintainer patch pushed after rebasing onto the latest main.

Previous head: 276c486fdd3b63510e6a2c04ebb02ef41f43e667
New head: 0811ae2df6fbd0ea1eee5933103a74379f818325

Object-source tabs now use canonical object identity including object type and routine signature. Reopening a matching dirty tab activates it without replacing unsaved SQL or marking it clean; clean tabs may still refresh from the server.

Validation: 2 focused test files / 172 tests, desktop typecheck, scoped lint and formatting, and git diff --check. Coverage includes dirty preservation, package spec/body separation, and overloaded routine signatures.

t8y2:
Thanks for the contribution! Merged in a9194ba3b5, will be released in the next version.

PR #6018, opened 2026-08-12
fix(gaussdb): use SHOW CREATE VIEW for view DDL in M-mode

AndrewDi:
GaussDB M-mode uses an ExternalDriver (JDBC) connection, but the JDBC plugin's getObjectSource only handles Oracle, throwing 'Object source is not supported by this JDBC driver' for other databases.

Add a special case in get_object_source_once: when the object type is View and the database uses the GaussDB M-mode JDBC driver, execute SHOW CREATE VIEW directly through the JDBC connection instead of calling getObjectSource.

### DIFF
--- a/crates/dbx-core/src/schema.rs
+++ b/crates/dbx-core/src/schema.rs
@@ -2652,9 +2652,9 @@ mod tests {
         dameng_object_statistics_rows_only_sql, dameng_object_statistics_user_segments_sql, deduplicate_column_infos,
         ephemeral_agent_metadata_session_id, external_driver_uses_mysql_ddl, filter_mongodb_agent_collections,
         filter_mysql_system_databases_for_config, filter_object_infos, filter_table_infos, filter_visible_schema_names,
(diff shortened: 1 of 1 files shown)


AndrewDi:
@t8y2 PTAL

t8y2 (commented):
Maintainer patch pushed after rebasing onto the latest main.

Previous head: 1ec5261ba7b97457003ab19c97945c471690315a
New head: f5bd992055a37be1479b19045ca3e6209d2c13a8

The GaussDB M-mode view fallback now reuses the canonical MySQL object-source SQL and result-parsing helpers, including database qualification, rather than duplicating an unqualified SHOW CREATE VIEW path.

Validation: 4 focused helper/gating tests, dbx-core check with no default features, cargo formatting, and git diff --check. The fallback still performs exactly one explicit metadata query.

t8y2:
Thanks for the contribution! Merged in 58f05d91c3, will be released in the next version.

PR #6021, opened 2026-08-12
fix(editor): center search match instead of scrolling to viewport edge

### LINKED ISSUE #5520: [Bug] 编辑器界面搜索选中滚动异常
### 数据库类型和版本

Oracle

### 问题描述

1. 复现步骤：ctrl+F 搜索的时候，当 sql 很长的时候，搜索到关键词点击跳转到之后，从当前行右侧往左选中关键词那行的时候会滚动异常，突然右侧滚动条往上滚动很多，导致从关键词那一行一直选中了异常滚动位置的那一片
2. 期望结果：鼠标拖动选中不要异常滚动

Image

### 支持信息

DBX Version: v0.5.76
Runtime: Desktop
Operating System: macOS 26.6
Architecture: aarch64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Problem

Fixes #5520

When searching in the SQL editor (Cmd/Ctrl+F) with a long statement, jumping to a match and then dragging to select text on that line could cause the editor to suddenly scroll a large distance, with the resulting selection spanning far more than the user intended.

## Root cause

The search-jump dispatched `scrollIntoView: true`, which uses CodeMirror's default `"nearest"` alignment — it scrolls the *minimum* distance needed to bring the match into view, which frequently leaves the matched line sitting flush against the very top or bottom edge of the viewport. Starting a drag-selection from that edge-adjacent line then triggers CodeMirror's own built-in "autoscroll while dragging near the viewport edge" behavior, producing a large, unexpected scroll jump and an over-extended selection — even for a purely horizontal drag within one line.

Confirmed with an automated repro (headless browser driving real mouse events): before the fix, a 60px horizontal drag on the match's line caused `scrollTop` to climb continuously from 171 to 1227 (the whole document), with the selection jumping from the match's line to the very end of the document. After the fix, `scrollTop` stayed constant throughout the same drag.

## Fix

Scroll the match into view with `y: "center"` alignment instead of the default `"nearest"`, in both search paths:
- `EditorSearchPanel.vue`'s scoped `findInScope()` jump
- `QueryEditor.vue`'s full-document search (via the `scrollToMatch` config option on CodeMirror's `search()` extension, which backs the built-in `findNext`/`findPrevious` commands)

## Testing

- `pnpm typecheck` — passes
- `vitest run apps/desktop/src/components/editor/__tests__/EditorSearchPanel.spec.ts` — passes
- Automated before/after repro via Playwright driving the running dev app (search-jump → real mouse drag), confirming the scroll no longer runs away

### DIFF
--- a/apps/desktop/src/components/editor/EditorSearchPanel.vue
+++ b/apps/desktop/src/components/editor/EditorSearchPanel.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { ref, nextTick, onBeforeUnmount, watch } from "vue";
 import { useI18n } from "vue-i18n";
-import type { EditorView } from "@codemirror/view";
+import { EditorView } from "@codemirror/view";
 import { EditorSelection } from "@codemirror/state";
 import { setSearchQuery, openSearchPanel as cmOpenSearchPanel, findNext as cmFindNext, findPrevious as cmFindPrevious, replaceNext as cmReplaceNext, replaceAll as cmReplaceAll } from "@codemirror/search";
 import { ChevronUp, ChevronDown, ChevronRight, TextSelect, X } from "@lucide/vue";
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in b8d37774fd, will be released in the next version.

PR #6025, opened 2026-08-12
Feature：支持 MyBatis foreach 集合参数展开

zipg:
## 改动说明

- 识别 MyBatis `<foreach>` 标签，并在 SQL 参数弹窗中使用 `collection` 名称展示集合参数
- 支持 JSON 数组及逗号分隔值输入，按照所选的字符串、数字、布尔值等类型生成 SQL 字面量
- 保留 `open`、`separator`、`close` 与循环体语义，兼容循环体中的其他模板参数
- 空集合安全展开为 `NULL`，避免生成无效 SQL 或意外扩大查询范围
- 补全参数输入提示的多语言文案，并保持现有 `#{name}` 单值替换行为不变

## 验证

- `corepack pnpm check`
- `corepack pnpm build`
- Oracle 11g 实库验证：`IN (1,2,3)` 正常执行并返回预期结果，空集合生成的 `IN (NULL)` 返回 0 行

### DIFF
--- a/apps/desktop/src/components/editor/SqlParameterDialog.vue
+++ b/apps/desktop/src/components/editor/SqlParameterDialog.vue
@@ -47,6 +47,10 @@ const syntaxLabels: Record<SqlParameterSyntax, string> = {
   sqlserver: "@name",
 };
 
+function syntaxLabel(parameter: SqlParameterDescriptor): string {
+  return parameter.collection ? "<foreach>" : syntaxLabels[parameter.syntax];
+}
+
 const resolvedSql = computed(() => substituteSqlParameters(props.sql, values.value, { databaseType: props.databaseType, enabledSyntaxes: props.enabledSyntaxes }));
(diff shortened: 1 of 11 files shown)


t8y2 (commented):
Maintainer patch applied in `9cb97809`.

Root causes: MyBatis `foreach index` was treated like a global parameter, XML comparison entities were decoded inside SQL literals/comments/identifiers, and regex-based closing-tag matching could be confused by foreach-like text in SQL comments or strings.

Patch: foreach indices use their zero-based iteration position and stay local to the loop; entity decoding is SQL-aware; and foreach tag matching uses a linear scanner that skips SQL strings, identifiers, comments, and dollar-quoted bodies.

Validation: 90/90 complete SQL parameter tests passed, `pnpm typecheck`, `oxfmt --check`, and `git diff --check` passed. The scanner remains linear and adds no polling, I/O, or database requests.

t8y2:
Thanks for the contribution! Merged in d4d6501a, will be released in the next version.

PR #6029, opened 2026-08-12
fix(ssh): raise host-key prompt above the connection dialog

### LINKED ISSUE #5989: [Bug] 添加SSH连接时，新主机的确认框会被连接弹窗挡住
### 数据库类型和版本

PostgreSQL

### 问题描述

先保存一个数据库连接
然后编辑，编辑的时候进入 ssh 隧道中添加一个新的服务器，因为系统会询问指纹没有记录过，是选yes还是no ；如果在这个时候，点击测试连接；询问弹窗会被连接弹窗挡住；并且这时候，连接编辑弹窗也无法被点击了,只能退出整个软件
目前想要正常使用的话，必须要先点保存，然后直接打开才能正常看到询问框；

PS:刚刚又看到有一个版本更新手快已经更新好了，我复制出来的支持信息不一定对了；

### 支持信息

DBX Version: v0.5.82
Runtime: Desktop
Operating System: macOS 15.7.7
Architecture: aarch64

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

_No response_

q396921921:
Fixes #5989

## Problem

Editing a saved connection → SSH tunnel → add a jump host with an unknown fingerprint → click "Test connection" triggers the SSH host-key confirmation dialog. That dialog is invisible, buried behind the still-open connection editor dialog, and both dialogs stop responding to clicks — the only way out is to force-quit the app.

## Root cause

Every dialog portal teleports into `<body>` on mount, so paint order at the shared `z-50` layer follows *mount* order, not which dialog opened last. `SshHostKeyPromptDialog` is mounted unconditionally at app start (`AppDialogs.vue`), while the connection editor mounts on demand. So the editor's portal always lands after the prompt's in the DOM and paints over it — while the underlying dialog stacking library also demotes the editor to `pointer-events: none` as the lower modal layer. The prompt is invisible, the editor is inert, and the window is stuck.

Confirmed by inspecting real z-index/pointer-events values in a running instance before and after the fix (not just from reading the code).

## Fix

`SshHostKeyPromptDialog`'s `DialogContent` now gets an explicit `overlay-class="z-[200]"` / `portal-class="z-[200]"` (both existing props on the shared dialog component) so it always paints above every other dialog layer in the app — checked the rest of the codebase for existing z-index usage (up to `z-[140]`) to pick a value that stays clear of it.

Added a regression test asserting those classes stay on the component.

## Testing

- `pnpm typecheck`, `pnpm lint`, `pnpm exec oxfmt --check` on the changed files — all pass
- `pnpm test` — full suite passes (909 files / 8449 tests)
- Manually reproduced before/after in a running desktop dev build: before, the host-key prompt is buried and unclickable, and accepting/dismissing it is impossible without force-quitting; after, the prompt renders on top, is clickable, and the connection editor stays intact and usable afterward.

### DIFF
--- a/apps/desktop/src/components/ssh/SshHostKeyPromptDialog.vue
+++ b/apps/desktop/src/components/ssh/SshHostKeyPromptDialog.vue
@@ -238,7 +238,16 @@ function submitSecret() {
 
 <template>
   <Dialog v-model:open="visible">
-    <DialogContent class="sm:max-w-[460px]" :show-close-button="false" @interact-outside.prevent @escape-key-down.prevent>
+    <!-- Every dialog portal teleports into <body> when its component mounts, so
+    body order (and therefore paint order at the shared z-50) is fixed by mount
+    order, not by which dialog opened last. This prompt mounts once at app start,
+    so a dialog mounted on demand later (e.g. the connection editor) paints over
+    it while still being demoted to pointer-events:none as the lower modal layer
+    — the prompt is invisible, the dialog is dead, and the window is stuck. This
+    is a blocking prompt, so keep it above every other layer in the app (the
+    tallest today are the image preview at 80/81 and the sidebar overlays at
+    100). -->
+    <DialogContent class="sm:max-w-[460px]" overlay-class="z-[200]" portal-class="z-[200]" :show-close-button="false" @interact-outside.prevent @escape-key-down.prevent>
       <DialogHeader>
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in f88e3956, will be released in the next version.

PR #6036, opened 2026-08-12
fix(editor): resolve table structure preview for quoted identifiers with hyphens

### LINKED ISSUE #5983: [Bug] 编辑器预览表结构识别错误
### 数据库类型和版本

Postgresql 12

### 问题描述

编辑器预览表结构时，识别错误，如图所示：
Image

log."public-rate_kingdee", 识别成了public.rate_kingdee

### 支持信息

DBX 版本: v0.5.82
运行环境: 桌面端
操作系统: Windows 10.0.26200.8655
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

Hovering a schema-qualified table reference like `log."public-rate_kingdee"` in the SQL editor to preview its structure resolves the wrong table (or shows nothing at all if the mangled name happens not to collide with any real table).

## Root cause

The hover tooltip's `identifierRangeAt` in `QueryEditor.vue` scanned characters with a plain `/[\w$.]/` class, which has no notion of quoted identifiers. On a quoted identifier containing a hyphen, it truncates at the hyphen and drops the surrounding quotes — e.g. hovering `"public-rate_kingdee"` extracts just `rate_kingdee`, which then gets looked up against the connection's *default* schema instead of the actual `log` schema, producing a wrong (or missing) preview.

Ctrl+click "go to definition" navigation on the same identifier already works correctly because it goes through the quote-aware `extractQualifiedIdentifierAt` parser in `sqlNavigation.ts` — the hover tooltip path just never used it.

## Fix

`identifierRangeAt` now delegates to `extractQualifiedIdentifierAt`, re-quoting each returned part so the identifier still round-trips correctly through `splitQualifiedIdentifier` downstream.

## Testing

- Added a regression test in `sqlNavigation.spec.ts` covering the exact `log."public-rate_kingdee"` scenario, both for `extractQualifiedIdentifierAt` directly and for the re-quote round trip through `splitQualifiedIdentifier`.
- `pnpm typecheck` — clean
- `pnpm lint` — no new warnings
- `pnpm vitest run apps/desktop/src` — full suite passes (628 files / 5726 tests)
- Manually verified against a live KingbaseES connection: created `log."public-rate_kingdee"`, ran `select * from log."public-rate_kingdee"`, and confirmed the hover preview now correctly shows `create table "systems"."log"."public-rate_kingdee" (...)` instead of resolving to an unrelated table / showing nothing.

Fixes #5983

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -66,6 +66,7 @@ import {
 import { usesOracleSessionCompletionColumns as shouldUseOracleSessionCompletionColumns } from "@/lib/sql/oracleCompletionSession";
 import {
   extractIdentifierDetailsAt,
+  extractQualifiedIdentifierAt,
   isSqlKeyword,
   matchSqlObject,
   matchTable,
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in 0b5c0bdc, will be released in the next version.

PR #6041, opened 2026-08-12
fix(editor): normalize hover-DDL copy to strip alignment spaces

### LINKED ISSUE #5732: [Bug] 复制 SQL 粘贴到 IDEA 中出现多余空格
### 来源
QQ 群反馈 · **反馈人**：传奇

### AI 摘要
用户在复制 SQL 语句并粘贴到 IntelliJ IDEA 时，会出现一连串的不必要空格，影响查询格式。

### 原始描述
复制成 sql 直接粘贴到 IDEA 中会有一连串的空格

q396921921:
## Summary

Fixes #5732 — copying a SQL statement from dbx and pasting into IntelliJ IDEA produces long runs of extra spaces.

**Root cause**: the SQL editor's hover-DDL tooltip (shown when hovering a table name) vertically aligns column name/type/nullable/etc. fields using literal padding spaces (`alignColumnRows()` in `hoverTableSql.ts`). dbx's own tooltip renders this as `white-space: pre`, so it looks fine on screen, but selecting and copying that text via the native OS copy carries the literal padding spaces verbatim. Pasting into a plain-text editor like IDEA (which doesn't collapse whitespace) exposes dozens of literal spaces per line — exactly the reported "一连串的空格".

**Fix**: added `normalizeAlignedSqlWhitespace()` which collapses the alignment padding back to single spaces while preserving each line's leading indentation, and wired it into the hover tooltip's `copy` event handler in `QueryEditor.vue`. Only the clipboard payload is rewritten — the on-screen tooltip alignment is untouched.

## Test plan

- [x] Reproduced live: ran the dev-web frontend against a real PostgreSQL 17 container, hovered a table, selected the DDL tooltip text, did a native ⌘C, and read the clipboard — confirmed runs of 20+ literal spaces per line before the fix.
- [x] Verified the fix removes the extra spaces in the same live repro (clipboard now single-spaced; on-screen alignment unchanged).
- [x] Added unit tests for `normalizeAlignedSqlWhitespace` in `hoverTableSql.spec.ts`.
- [x] `npx vitest run` on `hoverTableSql.spec.ts` and `sqlHoverLayout.spec.ts` — all 39 tests pass.
- [x] `oxlint`/`oxfmt` clean on changed files.

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -2364,6 +2364,7 @@ function createHoverDom(title: string, detail: string, sqlContent?: string, rows
       if (!selection || selection.isCollapsed) return;
       if (!dom.contains(selection.anchorNode) && !dom.contains(selection.focusNode)) return;
       const text = selection.toString();
+      if (text !== sqlContent) return;
       const normalized = normalizeAlignedSqlWhitespace(text);
       if (normalized === text) return;
       event.clipboardData?.setData("text/plain", normalized);
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -80,7 +80,7 @@ import {
   sqlObjectNavigationTypeFromCompletionObjectType,
   type SqlObjectNavigationTarget,
 } from "@/lib/sql/sqlNavigation";
-import { buildHoverTableSql, ddlForHoverPreview, hoverTableMatchesScope, quoteIdentifier, quoteQualifiedName, reformatHoverDdl, scopeHoverTables, type HoverTableScope } from "@/lib/editor/hoverTableSql";
+import { buildHoverTableSql, ddlForHoverPreview, hoverTableMatchesScope, normalizeAlignedSqlWhitespace, quoteIdentifier, quoteQualifiedName, reformatHoverDdl, scopeHoverTables, type HoverTableScope } from "@/lib/editor/hoverTableSql";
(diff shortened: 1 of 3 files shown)


t8y2 (commented):
Maintainer patch applied in `c11ca591`.

Root cause: copied hover DDL normalized every repeated whitespace run, including whitespace inside SQL strings, quoted identifiers, comments, and dollar-quoted bodies.

Patch: whitespace normalization is now SQL-aware and only runs for a complete hover DDL selection; partial selections remain byte-for-byte unchanged.

Validation: 36/36 focused hover DDL tests passed, including quoted strings, identifiers, comments, dollar quotes, and escaped literals; `oxfmt --check` and `git diff --check` passed. No dependency changes and no additional runtime I/O, polling, or database requests.

t8y2 (commented):
Maintainer conflict resolution completed from `c11ca591e82bf8e4aa96ce167a8758bf81a520d1` to `f6e482e1f68602d18c12ebb5125b6f5ddba689af` on current main `0b5c0bdcc000d86c4b4821f864e0a90f13332286`.

The conflict was limited to the hover helper import after #6036 added quoted-identifier handling. The resolution preserves both behaviors: quoted identifiers continue to resolve correctly, and complete hover DDL copies retain SQL-aware alignment-whitespace normalization.

`git range-diff` shows the semantic-whitespace patch unchanged and only the first commit's import merged with the new base. The full resolved diff contains the same three PR-owned files, 36/36 focused hover tests passed, `oxfmt --check` and `git diff --check` passed, and there are no dependency changes.

t8y2:
Thanks for the contribution! Merged in de5c2cb2, will be released in the next version.

PR #6042, opened 2026-08-12
✨ feat(data-grid): 支持按数据类型为单元格着色

### LINKED ISSUE #3332: [Feature] Color-code Data Table Cells Based on Column Data Type
### 数据库类型和版本

General

### 支持信息

DBX Version: v0.5.55
Runtime: Desktop
Operating System: Windows 10.0.26200.8655
Architecture: x86_64

### 当前痛点

The data table currently displays color-coded data type indicators in the column headers, making it easy to identify each column's data type. However, the cell values themselves are displayed using the same default color regardless of their data type

### 期望方案

It would be a nice usability improvement if the data table cells also inherited the same color coding as their corresponding column's data type. This would make it much easier to visually scan large datasets and quickly distinguish between strings, numbers, booleans, dates, JSON, etc., without constantly referring back to the header

The coloring should be subtle enough to maintain readability while providing visual cues. Ideally, this could also be an optional setting for users who prefer a neutral table

**Benefits**
- Faster visual scanning of large result sets
- Improved readability when working with many columns
- Consistent UI by extending the existing header color-coding to the table data
- Helps quickly identify unexpected values or type mismatches

This enhancement would complement the existing colored data type indicators in the headers and make the overall data browsing experience more intuitive

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

Image

eryajf:
- 新增数据类型颜色开关，可在网格工具栏和编辑器设置中配置
- 覆盖普通、转置及 Canvas 渲染模式，并保留选中、搜索、编辑等状态样式优先级
- 补充多语言文案及统一类型颜色映射逻辑

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

关闭：

Image

打开：

Image

关闭：

Image

开启：

Image

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

close #3332


t8y2 (commented):
Maintainer patch applied in `147f73f4`.

Root causes: dialect-ambiguous types were classified without database context; the disabled setting still evaluated per-cell interaction state; and unlayered global type-color selectors overrode existing hover text utilities.

Patch: SQL Server `timestamp`/`rowversion` and PostgreSQL `bit` variants are classified with database context, disabled coloring returns before cell-state reads, and DataGrid type selectors now live in the components layer.

Validation: 48/48 focused DataGrid visual tests passed; `oxfmt --check` and `git diff --check` passed. The default-off path now avoids the added per-cell state work. No dependencies, polling, background I/O, or database requests were added.

t8y2:
Thanks for the contribution! Merged in 433626f5, will be released in the next version.

PR #6045, opened 2026-08-12
feat(xugu): report structured agent errors

mapan0424:
## Summary

This PR adds `structured_error_v1` support to the Xugu runtime agent so DBX can distinguish a reusable SQL failure from a damaged connection, protocol failure, timeout, or explicit cancellation.

The change is intentionally limited to `agents/drivers/xugu`. It does not modify shared core behavior, frontend behavior, or any other database driver.

## Why this is needed

Previously, Xugu failures were returned as undifferentiated RPC errors. That left the DBX core without the information required to make a safe recovery decision:

- a syntax error or missing object is a normal SQL failure and the same session should remain usable;
- an EOF, broken socket, or corrupted driver frame can leave the physical connection unusable and the session must be quarantined;
- an explicit cancellation and a watchdog timeout both interrupt an in-flight operation, but they must retain their real cause and must not silently return a potentially unsafe session to the pool.

This distinction matters especially for Xugu because the currently pinned `go-xugu-driver` v1.0.12 exposes server errors as plain text instead of a typed database error, and its blocking query/exec calls do not reliably surface the request context as the final error. A classifier based only on broad string matching would therefore be unsafe.

### 1. Advertise structured errors only on the v2 runtime protocol

- The Xugu runtime handshake now advertises `structured_error_v1` together with multi-session support.
- The legacy v1 handshake remains unchanged and does **not** advertise this capability.
- Runtime and legacy malformed-request/dispatch error paths now receive the request method and the exact `agentSessionId` when available.

Every structured error includes the contract-required fields:

- `contractVersion`
- `category`
- `retryable`
- `sessionDisposition`
- `stage`
- `operationOutcome`

The request session ID is echoed exactly without trimming or normalization, as required by the core contract.

### 2. Use conservative Xugu-specific classification

The classifier preserves the original user-facing message and extracts only high-confidence diagnostics:

| Failure source | Structured result | Recovery decision |
| --- | --- | --- |
| Stable Xugu server header such as `[E19260 L6 C29]` during execution | `sql`, non-retryable, `vendorCode=19260` | keep the session |
| Typed EOF, `driver.ErrBadConn`, closed socket, or network error | `connection` | quarantine during an active operation |
| Driver connection wrapper text where the original cause is no longer available | `connection` | conservative recovery based on the request stage |
| `parseMsg: unknown message type` | `connection` | quarantine because v1.0.12 can surface a lost initial socket read this way |
| Session capacity reached | `resource`, retryable | keep runtime/session state |
| Missing logical session | `protocol` | quarantine the missing/invalid session reference |
| Explicit user cancellation | `canceled` | quarantine |
| Agent watchdog expiration | `timeout` | quarantine |

Additional safeguards:

- typed network identities take precedence over text parsing;
- a valid Xugu `[E...]` server header takes precedence over incidental words such as `unexpected EOF` inside the server message;
- the first stable server error code is exposed as `vendorCode`, while the complete original message is retained;
- no SQLSTATE is fabricated;
- no permission/syntax/object category is guessed from broad numeric ranges, because observed Xugu codes do not support a reliable range-based mapping;
- the plain word `killed` is not guessed to mean cancellation or timeout;
- this PR never requests `replace_runtime`.

### 3. Preserve the real timeout/cancellation source

The query and exec paths now record whether interruption came from the agent watchdog or an explicit cancel request and wrap the final driver result with an internal sentinel.

This is necessary because the driver may return `killed`, EOF, a parse error, another transport error, or even a late result after the kill request. The final string alone is not sufficient to identify the cause reliably.

Explicit cancellation also stops and clears the watchdog while holding the same lock. This makes the first observed cause authoritative: a user cancellation cannot later be relabeled as a timeout when the original deadline passes.

## Compatibility and scope

- Only Xugu agent files are changed.
- Existing successful query, metadata, session, and object-management paths are unchanged.
- Ordinary Xugu SQL errors keep the logical session available, which was verified against a live database.
- Broken transport/protocol state, timeout, and cancellation are isolated instead of being reused.
- Legacy protocol v1 remains compatible because it does not advertise the new capability.
- The original Xugu message, including text such as `L6 C29`, is preserved. Automatic editor line/column highlighting is **not** added here because the shared DBX structured-error allowlist currently has no line/column fields and the shared frontend parser does not recognize that Xugu notation. That should be handled separately as a shared-layer enhancement rather than expanding this Xugu-only change.

### Automated Xugu agent validation

- `GONOSUMDB=gitee.com/XuguDB/go-xugu-driver go test -race -count=1 ./...` — passed
- `go vet ./...` — passed
- Xugu package coverage — 70.0% of statements
- Linux amd64 static agent build with `CGO_ENABLED=0` — passed
- `git diff --check` — passed

The regression suite covers:

- every structured category and required contract field;
- exact session ID echo, including leading/trailing whitespace;
- Xugu error codes from 1 to 9 digits, optional `L`/`C`, multiple codes, trailing NUL, and malformed/multiline headers;
- Xugu server errors containing EOF-like text without misclassifying them as transport failures;
- typed network errors and driver parser corruption;
- runtime v2 and legacy v1 handshake boundaries;
- all runtime/legacy malformed-request and dispatch-error response paths;
- missing sessions and session-capacity errors;
- blocking query and exec watchdog timeouts;
- explicit query and exec cancellation;
- the cancellation-versus-watchdog race, including waiting beyond the original deadline and verifying that only one kill is issued.

### Core contract validation

- `cargo test -p dbx-core --lib strict_structured_error --no-default-features --locked` — **5 passed, 0 failed**

This verifies that the emitted category/stage/outcome/disposition combinations satisfy DBX's strict `structured_error_v1` parser and recovery rules.

### Live XuguDB validation

An opt-in, read-only integration test was added for environments that provide Xugu connection settings:

1. open a v2 runtime session using a non-administrator account on a business database;
2. execute a guaranteed missing-object query;
3. verify `category=sql`, `sessionDisposition=keep`, a non-zero `vendorCode`, and the exact session ID;
4. execute `SELECT 1` through the same logical session.

The test passed against the local XuguDB environment, confirming that a normal server-side SQL error is structured correctly and does not unnecessarily discard a healthy session. Public CI skips this test unless `XUGU_LIVE_TEST=1` and the `XUGU_LIVE_*` settings are provided.

## Files changed

- `agents/drivers/xugu/main.go`
- `agents/drivers/xugu/main_test.go`
- `agents/drivers/xugu/protocol_error.go`
- `agents/drivers/xugu/protocol_error_test.go`
- `agents/drivers/xugu/protocol_error_live_test.go`

### DIFF
--- a/agents/drivers/xugu/main.go
+++ b/agents/drivers/xugu/main.go
@@ -225,11 +225,6 @@ type response struct {
 	Error   *rpcError       `json:"error,omitempty"`
 }
 
-type rpcError struct {
-	Code    int    `json:"code"`
-	Message string `json:"message"`
-}
-
 type connectParams struct {
 	Host             string `json:"host"`
 	Port             int    `json:"port"`
@@ -539,6 +534,7 @@ type server struct {
 	activeRows        map[*sql.Rows]context.CancelFunc
 	activeTimer       *time.Timer
 	activeTimedOut    bool
+	activeCanceled    bool
 	// killSession, if non-nil, is called to force-kill the current
 	// statement on the database server. Tests may replace it with a
 	// stub. The real implementation is set during connectWithControl.
@@ -619,14 +615,14 @@ func newRuntimeServer() *runtimeServer {
 func (r *runtimeServer) handleLine(line string) (response, bool) {
 	var req request
 	if err := json.Unmarshal([]byte(line), &req); err != nil {
-		return errorResponse(nil, err), false
+		return errorResponse(nil, "", "", err), false
 	}
 	if len(req.ID) == 0 {
 		req.ID = json.RawMessage("1")
 	}
 	result, shutdown, err := r.dispatch(req.Method, req.Params)
 	if err != nil {
-		return errorResponse(req.ID, err), false
+		return errorResponse(req.ID, req.Method, stringParam(req.Params, "agentSessionId"), err), false
 	}
 	return response{JSONRPC: "2.0", ID: req.ID, Result: result}, shutdown
 }
@@ -637,7 +633,7 @@ func (r *runtimeServer) dispatch(method string, params map[string]json.RawMessag
 		return map[string]any{
 			"protocolVersion":      multiSessionProtocolVersion,
 			"agentProtocolVersion": multiSessionProtocolVersion,
-			"capabilities":         []string{"connect", "test_connection", "metadata", "query", "ddl", "multi_session"},
+			"capabilities":         []string{"connect", "test_connection", "metadata", "query", "ddl", "structured_error_v1", "multi_session"},
 		}, false, nil
 	case "open_session":
 		agentSessionID := stringParam(params, "agentSessionId")
@@ -711,7 +707,7 @@ func (r *runtimeServer) openSession(agentSessionID string, params connectParams)
 	}
 	if len(r.sessions) >= maxAgentSessions {
 		r.mu.Unlock()
-		return fmt.Errorf("agent session limit reached: %d", maxAgentSessions)
+		return fmt.Errorf("%w: %d", errAgentSessionLimit, maxAgentSessions)
 	}
 	r.mu.Unlock()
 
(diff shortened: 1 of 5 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in a67bcb09d, will be released in the next version.

PR #6076, opened 2026-08-13
fix(editor): wrap SQL preview text instead of clipping at panel border

### LINKED ISSUE #5992: [Bug] 编辑页完整sql外框展示不完整
### 数据库类型和版本

mysql8.0

### 问题描述

编辑页完整sql外框展示不完整
Image

### 支持信息

DBX 版本: v0.5.82
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem
In the DataGrid row-edit flow, when the generated UPDATE/INSERT SQL is a long single-line statement, the bottom SQL preview panel (`SqlPreviewPanel.vue`) silently clips the text at the panel's right border instead of wrapping or scrolling — the tail of the statement (including the `WHERE` clause) becomes completely invisible.

Root cause: `highlightSql()` calls Shiki's `codeToHtml()` without `structure: "inline"`, so it emits a full `<pre>` block whose default `white-space: pre` doesn't wrap, and the container has no overflow handling for it.

## Fix
- Pass `structure: "inline"` to `codeToHtml()`, matching the existing pattern already used in `sqlHighlighter.ts` elsewhere in the app.
- Render the highlighted output inside a `<pre>` with `whitespace-pre-wrap break-words`, mirroring the analogous SQL preview in `TableStructureEditor.vue`.

## Testing
Reproduced and verified visually via a local dev build (`pnpm dev:web` + `dbx-web` backend) connected to a SQLite test database: edited a long-text cell to generate a long `UPDATE` statement.

- Before: SQL preview showed only the first line, clipped mid-statement, `WHERE` clause invisible.
- After: full statement wraps across multiple lines inside the border, nothing clipped.

Also verified short SQL statements and dark theme still render correctly (no regression).

Fixes #5992

### DIFF
--- a/apps/desktop/src/components/editor/SqlPreviewPanel.vue
+++ b/apps/desktop/src/components/editor/SqlPreviewPanel.vue
@@ -67,6 +67,7 @@ async function highlightSql() {
     highlightedHtml.value = highlighter.codeToHtml(displaySql.value, {
       lang: "sql",
       theme,
+      structure: "inline",
     });
   } catch {
     // fallback to plain text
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch applied on the current head.

The original change is correct: Shiki now emits inline highlighted markup and the surrounding preformatted block wraps long SQL instead of clipping it. The previous frontend failure was caused by three formatting differences already fixed on main, not by this PR.

Added a focused regression test that locks both the inline Shiki structure and the wrapping preformatted styles. Validation passed against the latest main merge: focused Vitest, formatting, `pnpm typecheck`, and `git diff --check`.

t8y2:
Thanks for the contribution! Merged in 33a3e0d, will be released in the next version.

PR #6090, opened 2026-08-13
fix(grid): extend exact query row bound to LIMIT/OFFSET dialects

### LINKED ISSUE #6089: [Bug] Pagination next-page stays enabled past a user LIMIT on MySQL/Postgres (same class as #6033, SQL Server only)
## Database type and version

MySQL 8.0 / PostgreSQL (any version routed through the standard LIMIT/OFFSET pagination strategy)

## Description

This is the same bug class as #6033 (fixed for SQL Server in e18d1dded, "fix(sqlserver): respect TOP row bound in pagination"), but for the standard `LIMIT`/`OFFSET` dialects (MySQL, Postgres, and every other database type that isn't given its own `pagination_strategy` arm — i.e. everything that falls through to `TablePaginationStrategy::LimitOffset`).

Reproduction:
1. Run a query whose own SQL already caps the result with `LIMIT`, where that value equals (or is a multiple of) the grid's page size — e.g. `select id from some_table limit 100;` with the default page size of 100.
2. The result grid shows page 1 correctly, but the "next page" control stays enabled.
3. Clicking it loads an always-empty page 2 (or, once the bound is exceeded, an empty page past it).

Root cause: `build_query_pagination_execution_plan` in `crates/dbx-core/src/query_result_sql.rs` only computed `exact_query_row_bound` for `DatabaseType::SqlServer` (via `top_level_top_row_count`). For every other dialect the frontend has no authoritative "no more data" signal once a page comes back full, so `canGoNextDataGridPage` falls back to its "full page, maybe more" heuristic and keeps offering a next page that the user's own `LIMIT` can never satisfy.

## Expected result

Once the user's own top-level `LIMIT` bounds the query below or at a page boundary, the pagination control should stop offering further pages past that bound, the same way it now does for SQL Server's `TOP`.

## Fix

A PR extends the existing `exact_query_row_bound` detection (added for SQL Server in e18d1dded) to the standard `LIMIT`/`OFFSET` dialects, reusing the same frontend consumption path — no frontend changes needed, since `canGoNextDataGridPage`/`resultTotalRowCount` handling was already dialect-agnostic.

/claim

q396921921:
## Problem

@t8y2's e18d1dded fixed the result-grid pagination "next page" control staying enabled past a user-written `TOP` bound for SQL Server (#6033) — when a query's own SQL already caps the result at exactly the grid's page size, the frontend had no authoritative "no more data" signal and fell back to a "full page, maybe more" heuristic, offering an always-empty next page.

That fix only computed `exact_query_row_bound` for `DatabaseType::SqlServer`. The standard `LIMIT`/`OFFSET` dialects — MySQL, Postgres, and every other database type routed through `TablePaginationStrategy::LimitOffset` — have the exact same gap: `select id from some_table limit 100;` at page size 100 still offers a next page that's always empty. Filed as #6089.

## Fix

Extends the existing `exact_query_row_bound` detection added in e18d1dded to also parse a literal top-level `LIMIT` on the standard dialects, reusing `top_level_limit_row_count` (already used elsewhere in the pagination planner). No frontend changes — the `resultTotalRowCount`/`canGoNextDataGridPage` consumption path from e18d1dded was already dialect-agnostic, so extending the backend detection is the entire fix.

## Testing

- Extended the existing `pagination_plan_exposes_only_safe_exact_top_bounds` test (renamed to `pagination_plan_exposes_only_safe_exact_row_bounds`) with MySQL/Postgres/no-dialect `LIMIT` cases, including an existing-`OFFSET` case and a no-bound negative case.
- Added two `dataGridPagination.test.ts` cases documenting the frontend behavior once the bound is known (equal-to-page-size stops pagination; larger-than-page-size pages normally and stops at the bound).
- `cargo test -p dbx-core --lib query_result_sql`: 131 passed, 0 failed.
- `pnpm vitest run packages/app-tests/dataGridPagination.test.ts packages/app-tests/queryStore.test.ts`: 193 passed, 0 failed (includes the pre-existing SQL Server `exactQueryRowBound` parametrized suite from e18d1dded — unaffected, still passing).
- `pnpm typecheck`, `cargo check --no-default-features -p dbx-core`: clean.
- Manually reproduced and verified against a local SQLite connection (routes through the same `LimitOffset` strategy as MySQL/Postgres): before the fix, `select id from t limit 100;` at page size 100 left "next page" clickable and loaded an empty page 2; after the fix, `resultTotalRowCount` resolves to the exact bound and "next page" is correctly disabled. Also verified a `limit 300` case (3× page size) pages normally through all 3 pages and stops there.

Fixes #6089

### DIFF
--- a/crates/dbx-core/src/query_result_sql.rs
+++ b/crates/dbx-core/src/query_result_sql.rs
@@ -1176,19 +1176,34 @@ pub(crate) fn top_level_top_row_count(sql: &str) -> Option<usize> {
 }
 
 fn top_level_limit_row_count(sql: &str) -> Option<usize> {
-    let token = top_level_sql_tokens(sql).into_iter().find(|token| token.text == "LIMIT")?;
-    parse_standard_limit_row_count(sql, token.start + token.text.len())
+    let tokens = top_level_sql_tokens(sql);
+    let limit_index = tokens.iter().position(|token| token.text == "LIMIT")?;
+    let token = &tokens[limit_index];
+    let (count, suffix_start) = parse_standard_limit_row_count(sql, token.start + token.text.len())?;
+    let suffix_start = skip_sql_whitespace(sql, suffix_start);
+    if sql.get(suffix_start..)?.starts_with('%') {
+        return None;
+    }
+    if tokens[limit_index + 1..].iter().enumerate().any(|(offset, token)| {
+        token.text == "BY"
+            || token.text == "PERCENT"
+            || (token.text == "WITH" && tokens.get(limit_index + offset + 2).is_some_and(|next| next.text == "TIES"))
+    }) {
+        return None;
+    }
+    Some(count)
 }
 
-fn parse_standard_limit_row_count(sql: &str, start: usize) -> Option<usize> {
+fn parse_standard_limit_row_count(sql: &str, start: usize) -> Option<(usize, usize)> {
     let mut cursor = skip_sql_whitespace(sql, start);
     let first = parse_usize_literal(sql, &mut cursor)?;
     cursor = skip_sql_whitespace(sql, cursor);
     if sql.get(cursor..)?.starts_with(',') {
         cursor = skip_sql_whitespace(sql, cursor + 1);
-        return parse_usize_literal(sql, &mut cursor);
+        let count = parse_usize_literal(sql, &mut cursor)?;
+        return Some((count, cursor));
     }
-    Some(first)
+    Some((first, cursor))
 }
 
 fn parse_usize_literal(sql: &str, cursor: &mut usize) -> Option<usize> {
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
I applied a small maintainer follow-up in `2c5fb67bb`.

Root cause: reading only the first integer after a top-level `LIMIT` can misclassify dialect-specific clauses as an exact total-row bound. Examples include ClickHouse `LIMIT ... BY` / `WITH TIES` and percentage limits such as DuckDB `LIMIT 10%` or `LIMIT 10 PERCENT`.

Patch: retain the parsed integer only when the following top-level tokens still describe a plain exact `LIMIT`/`OFFSET` clause. Added regression cases for `BY`, `WITH TIES`, `%`, and `PERCENT` forms while preserving ordinary exact limits.

Validation:
- `cargo test -p dbx-core --no-default-features --lib pagination_plan_exposes_only_safe_exact_row_bounds`
- `cargo fmt --check`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in a0354726b, will be released in the next version.

PR #6104, opened 2026-08-13
fix(schema): quote bare column defaults in generated DDL

### LINKED ISSUE #5482: [Bug] 比较架构功能，新增varchar类型字段，默认值不会自动添加引号导致部署失败
### 数据库类型和版本

MySQL 5.7.42

### 问题描述

1. 新增MySQL的varchar字段，设置默认值 例如 THE_VALUE，使用比较架构功能部署，构建出来的DDL，默认值无引号，导致部署失败

Image

Image

Image

2.希望支持自动补全默认值的引号

### 支持信息

DBX 版本: v0.5.76
运行环境: 桌面端
操作系统: Windows 10.0.26100.4652
系统架构: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

Dhevenddra:
## 变更说明

`schema_diff` interpolates the driver's `column_default` straight into the DDL it generates, at four
separate sites. MySQL's `information_schema` stores a string default **unquoted**, so adding a
`varchar` column whose default is `THE_VALUE` produces:

```sql
ALTER TABLE `t` ADD COLUMN `c` varchar(64) DEFAULT THE_VALUE
```

MySQL rejects that and the compare-schema deploy fails, which is what #5482 reports. PostgreSQL
returns the default already quoted and cast (`'new'::text`), so the raw passthrough looks correct
there. That is why the defect is MySQL-shaped rather than universal.

This adds `default_literal(default, data_type)`, which quotes **only** a bare value on a column whose
type takes a literal. Passed through untouched: values that are already quoted, prefixed literals
(`N'x'`, `b'x'`, `x'..'`, `e'..'`), function calls, numbers, `NULL`, and bare temporal keywords such
as `CURRENT_TIMESTAMP`. Embedded single quotes are doubled, so `it's` becomes `'it''s'`. An
empty-string default now emits `DEFAULT ''` rather than a bare `DEFAULT `, which was the same defect
in a quieter form.

Applied at all four sites that emitted a raw default: `column_def`, both `mapped_type` branches, and
the `ALTER COLUMN ... SET DEFAULT` path.

Two neighbouring modules already solve this for their own output
(`table_structure_sql::util::format_default_for_sql` and `transfer::format_mysql_default_literal`),
each private to its own module. The rule here follows that existing precedent rather than inventing a
new one. If a fourth site ever needs it, I would rather consolidate the three than add a fourth copy,
and I am happy to do that in a follow-up if you want it.

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

- [ ] 本 PR 涉及前端改动，已附截图/录屏

No frontend changes. The diff touches only `crates/dbx-core/src/schema_diff.rs`.

## 验证

- [ ] `make check` 通过
- [x] 相关测试通过

To be straight with you about the two unchecked boxes: I could not run them. This machine cannot
build the desktop app (the full-feature build fails on `fontconfig` via the `system-fonts` feature),
so I verified the crate directly instead:

```
cargo test -p dbx-core --no-default-features --lib
    4481 passed; 0 failed; 38 ignored

cargo fmt --check      clean
```

Rebased onto `a82ccde` today and re-ran the suite there, so the result above is against current
`main`, not the branch point.

New tests, both in `schema_diff.rs`:

- `added_varchar_column_quotes_a_bare_default_mysql` reproduces #5482 directly and asserts the
  generated DDL contains `DEFAULT 'THE_VALUE'`. It fails on `main` without this change.
- `default_literal_only_quotes_bare_values_that_need_it` pins the passthrough rules, so a later
  change cannot start double-quoting `N'guest'` or numeric and function defaults.

The existing PostgreSQL, SQLite and auto-increment cases in the same module are what convinced me the
rule is safe, since they exercise the paths that must **not** change.

## 关联 Issue

Close #5482

### DIFF
--- a/crates/dbx-core/src/schema_diff.rs
+++ b/crates/dbx-core/src/schema_diff.rs
@@ -3115,7 +3115,7 @@ fn column_def(col: &ColumnInfo, db_type: DatabaseType) -> String {
         definition.push_str(" NOT NULL");
     }
     if let Some(default) = &col.column_default {
-        definition.push_str(&format!(" DEFAULT {default}"));
+        definition.push_str(&format!(" DEFAULT {}", default_literal(default, &col.data_type)));
     }
     if profile.inline_column_comment {
         if let Some(comment) = &col.comment {
@@ -3247,6 +3247,54 @@ fn comment_literal(comment: &str) -> String {
     format!("'{}'", comment.replace('\'', "''"))
 }
 
+/// Bare temporal keywords that are defaults in their own right and must not be quoted.
+const TEMPORAL_DEFAULT_KEYWORDS: [&str; 8] =
+    ["current_timestamp", "current_date", "current_time", "now", "localtime", "localtimestamp", "getdate", "sysdate"];
+
+/// Prefixes that introduce an already-quoted literal: SQL Server / Sybase `N'x'`,
+/// MySQL `b'1'` and `x'1f'`, Postgres `e'\n'`.
+const QUOTED_LITERAL_PREFIXES: [&str; 4] = ["n'", "b'", "x'", "e'"];
+
+/// Render a column default as a SQL literal.
+///
+/// Drivers hand `column_default` back verbatim and they do not agree on its
+/// shape. MySQL's `information_schema` stores a string default unquoted
+/// (`THE_VALUE`), while Postgres returns it already quoted and cast
+/// (`'new'::text`). Emitting the raw value is right for the second and
+/// produces DDL that will not deploy for the first, so quote only a bare
+/// value on a column whose type takes a quoted literal.
+///
+/// `table_structure_sql::util::format_default_for_sql` and
+/// `transfer::format_mysql_default_literal` do the same job on their own
+/// paths; both are private to their modules.
+fn default_literal(default: &str, data_type: &str) -> String {
+    let trimmed = default.trim();
+    if trimmed.eq_ignore_ascii_case("NULL") {
+        return trimmed.to_string();
+    }
+    // Already a literal: `'x'`, `'new'::text`, or a prefixed form like `N'x'`.
+    let lowered = trimmed.to_ascii_lowercase();
+    if trimmed.starts_with('\'') || QUOTED_LITERAL_PREFIXES.iter().any(|prefix| lowered.starts_with(prefix)) {
+        return trimmed.to_string();
+    }
+    // Expressions and function calls: `uuid()`, `(uuid())`, `nextval('s'::regclass)`.
+    if trimmed.contains('(') {
+        return trimmed.to_string();
+    }
+
+    let base_type = data_type.split('(').next().unwrap_or(data_type).trim().to_ascii_lowercase();
+    let takes_text_literal = ["char", "text", "string", "clob", "enum"].iter().any(|kind| base_type.contains(kind));
+    let takes_temporal_literal = base_type.contains("date") || base_type.contains("time");
+
+    if takes_temporal_literal && TEMPORAL_DEFAULT_KEYWORDS.iter().any(|kw| trimmed.eq_ignore_ascii_case(kw)) {
+        return trimmed.to_string();
+    }
+    if takes_text_literal || takes_temporal_literal {
+        return format!("'{}'", trimmed.replace('\'', "''"));
+    }
+    trimmed.to_string()
+}
+
 fn column_comment_sql(
     table_name: &str,
     column_name: &str,
@@ -3356,7 +3404,7 @@ fn generate_create_table_sql(
                 }
                 if !skip_default {
                     if let Some(default) = &col.column_default {
-                        def.push_str(&format!(" DEFAULT {default}"));
+                        def.push_str(&format!(" DEFAULT {}", default_literal(default, &mapped_type)));
                     }
                 }
                 if profile.inline_column_comment {
@@ -3383,7 +3431,7 @@ fn generate_create_table_sql(
                 }
                 if !skip_default {
                     if let Some(default) = &col.column_default {
-                        def.push_str(&format!(" DEFAULT {default}"));
+                        def.push_str(&format!(" DEFAULT {}", default_literal(default, &mapped_type)));
                     }
                 }
                 if profile.inline_column_comment {
@@ -3889,7 +3937,10 @@ fn generate_schema_sync_sql_inner(
                                 }
                                 if column.changes.iter().any(|change| change.starts_with("default:")) {
                                     parts.push(if let Some(default) = &source.column_default {
-                                        format!("  ALTER COLUMN {name} SET DEFAULT {default}")
+                                        format!(
+                                            "  ALTER COLUMN {name} SET DEFAULT {}",
+                                            default_literal(default, &mapped.data_type)
+                                        )
                                     } else {
                                         format!("  ALTER COLUMN {name} DROP DEFAULT")
                                     });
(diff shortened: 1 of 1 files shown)


t8y2 (changes requested):
当前实现仍无法可靠区分“裸字符串默认值”和“裸 SQL 表达式”，需要调整后再合并。

1. `default_literal` 只豁免了少量时间关键字，之后会把其他文本类型裸值统一加引号。因此 PostgreSQL `text DEFAULT CURRENT_USER`、Oracle `varchar2 DEFAULT USER` 会被生成成固定字符串 `'CURRENT_USER'`、`'USER'`，改变原有语义。
2. 任何包含 `(` 的默认值都会被当作表达式。MySQL 元数据可能把 `varchar DEFAULT 'a(b)'` 返回为裸值 `a(b)`，当前实现仍会生成无效的 `DEFAULT a(b)`。
3. 相同裸 token 在不同数据库中可能分别代表字符串或表达式，无法仅依赖目标类型和字符串形态统一判断。生成链已经持有 source dialect，MySQL 元数据也保留了 `EXTRA` 等信息，建议把格式化规则绑定到来源数据库及其元数据语义。

DBeaver 的通用列 DDL 处理也不会简单地把所有字母开头的文本默认值统一加引号，而是保留 `USER`、`CURRENT_USER` 等表达式；这进一步说明这里需要数据库来源信息，而不是全局启发式。

请补充至少以下回归测试：MySQL 字符串 `a(b)`、PostgreSQL `CURRENT_USER`、Oracle `USER`，以及 `SET`/二进制字符串等当前类型识别边界。

Dhevenddra:
You are right on both counts, and I can trace both directly in the code as written.

1. `CURRENT_USER` on a text column. My keyword exemption only applies when the column is temporal, so PostgreSQL `text DEFAULT CURRENT_USER` falls straight through to the quoting branch and becomes `'CURRENT_USER'`. Oracle `varchar2 DEFAULT USER` behaves the same way. That changes the semantics rather than fixing anything.
2. `a(b)`. I treat any value containing `(` as an expression and pass it through untouched, so a MySQL varchar default of `a(b)` still produces the invalid `DEFAULT a(b)` that this PR was supposed to remove.

Both come from the same mistake. I keyed the decision on the target type and the shape of the string, and that can never separate a bare string from a bare expression.

Your third point is the actual answer, and the plumbing is already in place. `generate_create_table_sql` and `generate_schema_sync_sql_inner` both take `source_dialect: Option<DialectKind>`, and `ColumnInfo` carries `extra`. `transfer.rs` already does exactly what you describe: it reads `DEFAULT_GENERATED` out of `extra` to decide whether a MySQL default is an expression, covered by `mysql_create_table_keeps_current_timestamp_default_and_on_update`.

So I will rework it as:

- **MySQL**: `extra` containing `DEFAULT_GENERATED` means expression, pass through. Otherwise the value is a literal and is quoted even when it contains parentheses, which fixes `a(b)`.
- **PostgreSQL and Oracle**: the metadata already returns string defaults quoted or cast, so a bare token is an expression and is never quoted. `CURRENT_USER` and `USER` are left alone.
- `column_def` needs the source dialect threaded through from its callers in `generate_schema_sync_sql_inner`.

One question before I write it. That would be the third copy of this rule, alongside `transfer::format_mysql_default_literal` and `table_structure_sql::util::format_default_for_sql`. Would you rather I lift a single shared helper that all three use, or keep the change local to `schema_diff.rs` so this PR stays narrow? I would rather ask than move code across modules uninvited.

Regression tests will cover what you listed: MySQL `a(b)`, MySQL `CURRENT_TIMESTAMP` carrying `DEFAULT_GENERATED`, PostgreSQL `CURRENT_USER`, Oracle `USER`, and the `SET` and binary-string type boundaries.

Dhevenddra:
Reworked. Both cases you raised are fixed, and the rule now follows the source dialect instead of the value.

MySQL is the only dialect this repairs now, because it is the only one whose metadata strips the quotes off a string default. Postgres, Oracle and SQL Server return the value already quoted or wrapped, so a bare token there is an expression and gets passed through. `text DEFAULT CURRENT_USER` and `varchar2 DEFAULT USER` come out unchanged.

For MySQL the `EXTRA` column decides. 8.0.13 and later flag an expression default as `DEFAULT_GENERATED`, which `transfer.rs` already depends on, so a value is a literal unless the server says it was generated. The parenthesis guess is gone, and `a(b)` is quoted.

Removing that guess did break one existing test, `dropped_mysql_table_rollback_preserves_defaults_comments_indexes_and_fk`. It has a column defaulting to `(uuid())` with no `EXTRA` set. MySQL writes an expression default wrapped in parentheses and reports it back the same way, so the wrapping is the syntax rather than a coincidence. The check now asks whether the value is wholly wrapped, which is a different question from whether a parenthesis appears somewhere inside it. `a(b)` is not wrapped, so it still gets quoted.

`column_def` now takes the source dialect from its callers. Where none is declared the comparison is same-dialect, so the target stands in for it.

`cargo test -p dbx-core --no-default-features --lib` gives 4484 passed, 0 failed. `cargo clippy -p dbx-core --no-default-features --lib` is clean, and `cargo fmt --check` passes. The new tests cover what you listed: MySQL `a(b)`, PostgreSQL `CURRENT_USER`, Oracle `USER`, the SET boundary, and a binary column where a hex default is left alone while a bare string on the same column is quoted.

I kept this inside `schema_diff.rs` rather than lifting a shared helper, since that was still an open question. Say the word and I will do the consolidation as a follow-up.

t8y2 (changes requested):
新提交已经解决了上一轮提到的 source dialect、`a(b)`、`CURRENT_USER` / `USER` 和 SET/二进制边界，但 MySQL 时间默认值的精度形式仍会被错误加引号。

`default_literal` 对旧版 MySQL 无 `DEFAULT_GENERATED` 标记的兼容分支，只用完全相等匹配 `CURRENT_TIMESTAMP`、`NOW` 等关键字。因此合法的 `TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6)` 会落入时间类型字符串分支，生成带引号的字符串默认值。同类的 `NOW()`、`LOCALTIME(3)`、`LOCALTIMESTAMP(3)` 也有相同问题。

仓库现有 transfer 默认值格式化已经显式支持 `CURRENT_TIMESTAMP(...)`，table structure 路径也有 `CURRENT_TIMESTAMP(3)` 回归测试。请让旧版 MySQL temporal fallback 识别允许可选精度参数的时间表达式，并补充 schema diff 的 `CURRENT_TIMESTAMP(6)`（至少覆盖 `extra=None`）回归测试。不要恢复“任意包含括号即表达式”的旧启发式，以免再次漏掉字符串 `a(b)`。

Dhevenddra:
Good catch, and thank you for pointing at the existing precedent rather than just the symptom.

You are right. The pre-8.0.13 fallback compared the value against the temporal keywords with exact equality, so `TIMESTAMP(6) DEFAULT CURRENT_TIMESTAMP(6)` matched nothing, fell through to the temporal string branch and came out quoted. `NOW()`, `LOCALTIME(3)` and `LOCALTIMESTAMP(3)` behaved the same way.

The fallback now accepts a keyword on its own or followed by an argument list, which is what `transfer::is_mysql_function_default` already does for these same values:

```rust
fn is_temporal_keyword_default(value: &str) -> bool {
    let upper = value.to_ascii_uppercase();
    TEMPORAL_DEFAULT_KEYWORDS.iter().any(|keyword| {
        let keyword = keyword.to_ascii_uppercase();
        upper == keyword || upper.strip_prefix(&keyword).is_some_and(|rest| rest.starts_with('('))
    })
}
```

The match stays keyed on the known keywords rather than on the presence of a parenthesis, so the old heuristic does not come back. `default_literal_keeps_temporal_defaults_that_carry_a_precision` covers `CURRENT_TIMESTAMP(6)`, `NOW()`, `LOCALTIME(3)`, `LOCALTIMESTAMP(3)` and the lowercase form, all with `extra=None` as you asked. It also asserts the two boundaries in the other direction: `a(b)` on a varchar is still quoted, and so is `CURRENT_TIMESTAMPX(6)`, so a keyword prefix cannot be mistaken for the keyword itself.

`cargo test -p dbx-core --no-default-features --lib` gives 4485 passed, 0 failed, 38 ignored. `cargo fmt --check` is clean.

The question from the previous round is still open whenever you want to answer it: this is now the third copy of the default-formatting rule alongside `transfer::format_mysql_default_literal` and `table_structure_sql::util::format_default_for_sql`. Happy to lift a shared helper in a follow-up if you would prefer that to three separate implementations.

t8y2 (commented):
Maintainer patch pushed to `e76a1dfb` on top of `38bf4ec5`.

Root cause: default classification normalized the source text and then reused that normalized value when emitting a quoted MySQL string, losing significant whitespace and conflating the string `"NULL"` with SQL `NULL`.

Patch: use the normalized value only for expression classification and quote the original source value for MySQL string defaults. Existing non-MySQL and temporal-expression behavior is preserved. No dependencies or extra database requests were added.

Validation:
- `default_literal_` regressions — 5 passed

t8y2 (approved):
Maintainer patch resolves the requested changes; exact-head CI verified.

t8y2:
Thanks for the contribution! Merged in f6518ff3, will be released in the next version.

PR #6129, opened 2026-08-13
fix(jdbc): stop forcing Connector/J cursor fetching

wbean:
## Summary
- Stop automatically injecting Connector/J's `useCursorFetch=true` for generic JDBC/MySQL connections.
- Release the JDBC plugin as `0.1.30` and document explicit cursor-fetch opt-in through the connection URL.
- Add a regression test proving DBX does not add the vendor property to driver connection properties.

## Why
JDBC standardizes `ResultSet` traversal and `Statement#setFetchSize` as a driver hint; it does not standardize a MySQL server-cursor protocol. `useCursorFetch` is a Connector/J-specific property that changes driver/server behavior.

A `jdbc:mysql:` URL can target a MySQL-compatible server rather than MySQL itself. Forcing that property made even `SELECT 1` block until DBX's 30-second timeout against a compatible backend, with both the official Connector/J driver and the compatibility driver.

Cursor fetching is now an explicit driver/server capability: users who have verified support can configure `useCursorFetch=true` in the JDBC URL. Compatibility drivers can keep the property unset or normalize it for their server behavior.

## Validation
- `cd plugins/jdbc && ./gradlew --no-daemon --rerun-tasks test shadowJar`
- Direct JSON-RPC verification with the rebuilt plugin against the affected backend: `SELECT 1`, a table read, and a two-page table result all completed without the 30-second timeout.

### DIFF
--- a/plugins/jdbc/README.md
+++ b/plugins/jdbc/README.md
@@ -36,6 +36,18 @@ lib/dbx-jdbc-plugin.jar
 
 DBX does not bundle Java or JDBC drivers. Install Java locally and add database-specific driver JAR paths in the DBX JDBC connection form.
 
+## MySQL-compatible cursor fetching
+
+DBX uses standard JDBC result-set paging, but does not automatically set Connector/J's `useCursorFetch` property.
+That property enables a MySQL-specific server cursor protocol; it is not part of JDBC and may be unsupported by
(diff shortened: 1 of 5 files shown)


t8y2:
Thanks for the contribution! Merged in 1e0687a, will be released in the next version.

PR #6135, opened 2026-08-13
fix(editor): seed search box with full selection including newlines

### LINKED ISSUE #5435: [Bug] Ctrl+F/H 搜索框未自动填充换行符需手动输入
### 来源
QQ 群反馈 · **反馈人**：少年好久不见

### AI 摘要
编辑器内容包含换行符时，使用 Ctrl+F 或 Ctrl+H 打开搜索/替换框，搜索框显示为空白，不会自动填入换行符，用户必须手动输入 \n 才能匹配换行符，期望应自动识别选中内容中的换行符。

### 原始描述
如果编辑器内容存在换行符的情况，直接使用CTRL+F搜索或者CTRL+H，搜索框是空白的，不会自动填充换行符，需要手动输入换行符\n

q396921921:
## What

Fixes #5435

`openSearch()` in `EditorSearchPanel.vue` only seeded the search box (`searchText`) from the current editor selection when that selection did **not** contain a newline:

```js
if (selText && !selText.includes("\n")) {
  searchText.value = selText;
}
```

So selecting text that spans multiple lines and pressing Ctrl+F/Cmd+F (or Ctrl+H) left the search box completely empty and showed "no results", forcing the user to manually retype the query (including an escaped `\n`) to search across lines — exactly what the reporter described.

## Fix

Seed the search box from the selection whenever it's non-empty, regardless of whether it contains a newline. The existing multi-line "scope to selection" logic right below is untouched — seeding and scoping are two separate concerns and shouldn't be coupled by a shared guard.

Note: a plain HTML `<input>` cannot visually render an embedded newline (the browser strips it from the rendered value), so a two-line selection shows up in the box with the lines run together with no visible break. That's an inherent `<input>` limitation, not something this fix can change — but the underlying search query still receives the real newline character, so cross-line matching now works correctly (verified below).

## Testing

- Added `apps/desktop/src/components/editor/__tests__/EditorSearchPanel.newlineSeed.spec.ts`, which mounts the real component with a real CodeMirror `EditorView`, selects text spanning a newline, opens search, and asserts the match count renders `1/1` (was `noResults` before the fix).
- `pnpm vitest run` — full suite passes (8802 tests).
- `pnpm vue-tsc --noEmit --project apps/desktop/tsconfig.json` — passes.
- `pnpm oxlint --vue-plugin apps/desktop/src` on the changed files — clean.
- Manually verified in a running `pnpm dev:web` instance (SQLite connection, two-line SQL selected, Cmd+F): before the fix the search box was empty with "无结果" (no results); after the fix the match indicator shows "1/1" and the selection-scope indicator is active, confirming the cross-line query is now found.

### DIFF
--- a/apps/desktop/src/components/editor/EditorSearchPanel.vue
+++ b/apps/desktop/src/components/editor/EditorSearchPanel.vue
@@ -276,7 +276,7 @@ function openSearch(): boolean {
     cmOpenSearchPanel(v);
     const sel = v.state.selection.main;
     const selText = v.state.sliceDoc(sel.from, sel.to);
-    if (selText && !selText.includes("\n")) {
+    if (selText) {
       searchText.value = selText;
     }
     // Set scope when there's a multi-line selection
--- a/apps/desktop/src/components/editor/__tests__/EditorSearchPanel.newlineSeed.spec.ts
+++ b/apps/desktop/src/components/editor/__tests__/EditorSearchPanel.newlineSeed.spec.ts
@@ -1,0 +1,75 @@
+// @vitest-environment happy-dom
+
+import { EditorState } from "@codemirror/state";
+import { EditorView } from "@codemirror/view";
+import { createApp, defineComponent, h, nextTick, type App } from "vue";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("vue-i18n", () => ({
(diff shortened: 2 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in d263210, will be released in the next version.

PR #6146, opened 2026-08-13
fix(schema-diff): quote unquoted string/temporal defaults for MySQL family

### LINKED ISSUE #6139: [Bug] 比较架构生成 SQL 未对 enum/varchar 等字符串默认值做引号包裹，导致语法错误
### 数据库类型和版本

MySQL 9.7

### 🐛 Bug 描述

在后端 `schema_diff.rs` 的「比较架构」功能中，生成库表同步迁移 SQL（包括 `CREATE TABLE` / `ADD COLUMN` / `MODIFY COLUMN` / `CHANGE COLUMN` / `ALTER COLUMN SET DEFAULT`）时，所有列的 `DEFAULT` 值均被**原样裸拼接**，未根据列的数据类型进行引号感知处理。

这导致 `enum` / `set` / `varchar` / `char` / `text` 等字符串类型列的默认值生成为非法 SQL，例如：

```sql
-- ❌ 当前错误生成
DEFAULT active

-- ✅ 期望正确生成
DEFAULT 'active'
```

执行该迁移语句时数据库将报语法错误，阻断自动化部署流程。

### 🔍 根因分析

`schema_diff.rs` 在拼接 `DEFAULT` 子句时，直接将默认值作为原始字符串插入 SQL 模板，缺少以下关键逻辑：

-   未识别目标列是否为字符串/枚举/集合类型；
-   未对字符串类默认值自动添加单引号包裹；
-   未对值内部已有的单引号做转义处理（如 `'it''s'`）。

### ✅ 期望行为

生成迁移 SQL 时，应根据列类型智能处理 `DEFAULT` 值：

| 列类型 | 默认值示例 | 期望 SQL 输出 |
| :--- | :--- | :--- |
| `varchar` / `char` / `text` | `active` | `DEFAULT 'active'` |
| `enum` / `set` | `pending` | `DEFAULT 'pending'` |
| `int` / `bigint` / `decimal` | `0` | `DEFAULT 0` |
| `boolean` | `true` | `DEFAULT TRUE` |
| 含单引号的字符串 | `it's` | `DEFAULT 'it''s'` |

### 📋 复现步骤

1.  创建一个包含 `varchar` 或 `enum` 类型列且设有非空默认值的源表；
2.  触发「比较架构」功能生成同步迁移 SQL；
3.  检查生成的 SQL 中对应列的 `DEFAULT` 子句；
4.  在目标数据库执行该 SQL → 报语法错误。

### 支持信息

DBX 版本: v0.5.83
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Summary

MySQL's `INFORMATION_SCHEMA.COLUMNS.COLUMN_DEFAULT` (and `SHOW COLUMNS`) returns string/enum/set/date/datetime default literals **without** surrounding quotes — e.g. `active` instead of `'active'`. `schema_diff.rs` spliced this raw value directly into generated `CREATE TABLE` / `ADD COLUMN` / `MODIFY COLUMN` / `ALTER COLUMN SET DEFAULT` statements, producing invalid SQL such as:

```sql
-- generated (broken)
DEFAULT active

-- expected
DEFAULT 'active'
```

which fails to execute against the target database, blocking the "比较架构" (compare schema) sync flow.

- Add a `quote_unquoted_string_defaults` flag to `DdlDialectProfile`, set `true` only for the MySQL family (Mysql/Doris/StarRocks/Goldendb/Sundb/Databend/Gbase/ManticoreSearch) — the other dialect profiles already receive pre-quoted/pre-formatted defaults from their own introspection (verified: Postgres via `pg_get_expr`, SQL Server's `COLUMN_DEFAULT` parenthesizes+quotes, Oracle's `DATA_DEFAULT`, SQLite's `dflt_value` all already include quoting), so they opt out and `format_default()` is a no-op for them.
- Add `DdlDialectProfile::format_default(data_type, default_value)`, which quotes (with `''` escaping) bare string/enum/set/date/datetime literals, while leaving function-call expressions (`uuid()`) and bare keyword expressions (`CURRENT_TIMESTAMP`) untouched.
- Wire it into all four `DEFAULT`-emitting call sites in `schema_diff.rs`.

Duplicate of/related to #5482 (same root cause, reported earlier with a narrower repro).

## Test plan

Reproduced the bug first: temporarily disabled the new quoting flag and confirmed 5 new tests fail with exactly the broken output described in the issue (`DEFAULT active`, `DEFAULT pending`, `DEFAULT 2024-01-01`, unescaped `it's`), then re-enabled the fix and confirmed they pass.

- [x] `cargo test -p dbx-core --lib schema_diff::` — 197 passed (9 new: varchar/enum/date default quoting, `''` escaping, function-call/keyword defaults left alone, numeric defaults untouched, MODIFY COLUMN path, Postgres already-quoted defaults untouched)
- [x] `cargo test -p dbx-core --lib ddl_profile::` — 6 passed
- [x] `cargo clippy -p dbx-core --lib` — no new warnings on touched files
- [x] `cargo fmt -p dbx-core` — clean

Fixes #6139

### DIFF
--- a/crates/dbx-core/src/schema_diff.rs
+++ b/crates/dbx-core/src/schema_diff.rs
@@ -3108,14 +3108,14 @@ fn quote_id(name: &str, db_type: DatabaseType) -> String {
     profile_for(db_type).quote_ident(name)
 }
 
-fn column_def(col: &ColumnInfo, db_type: DatabaseType) -> String {
+fn column_def(col: &ColumnInfo, db_type: DatabaseType, source_dialect: Option<DialectKind>) -> String {
     let profile = profile_for(db_type);
     let mut definition = format!("{} {}", quote_id(&col.name, db_type), col.data_type);
     if !col.is_nullable {
         definition.push_str(" NOT NULL");
     }
     if let Some(default) = &col.column_default {
-        let default = profile.format_default(&col.data_type, default);
+        let default = profile.format_default(&col.data_type, default, col.extra.as_deref(), source_dialect);
         definition.push_str(&format!(" DEFAULT {default}"));
     }
     if profile.inline_column_comment {
@@ -3357,7 +3357,8 @@ fn generate_create_table_sql(
                 }
                 if !skip_default {
                     if let Some(default) = &col.column_default {
-                        let default = profile.format_default(&mapped_type, default);
+                        let default =
+                            profile.format_default(&mapped_type, default, col.extra.as_deref(), source_dialect);
                         def.push_str(&format!(" DEFAULT {default}"));
                     }
                 }
@@ -3385,7 +3386,8 @@ fn generate_create_table_sql(
                 }
                 if !skip_default {
                     if let Some(default) = &col.column_default {
-                        let default = profile.format_default(&mapped_type, default);
+                        let default =
+                            profile.format_default(&mapped_type, default, col.extra.as_deref(), source_dialect);
                         def.push_str(&format!(" DEFAULT {default}"));
                     }
                 }
@@ -3863,7 +3865,7 @@ fn generate_schema_sync_sql_inner(
                             };
                             parts.push(format!(
                                 "  ADD COLUMN {}{}",
-                                column_def(&convert_col(source), db_type),
+                                column_def(&convert_col(source), db_type, source_dialect),
                                 position
                             ));
                         }
@@ -3876,7 +3878,10 @@ fn generate_schema_sync_sql_inner(
                             let mapped = convert_col(source);
                             if profile.alter_uses_modify_column {
                                 if column.changes.iter().any(|change| !change.starts_with("order:")) {
-                                    parts.push(format!("  MODIFY COLUMN {}", column_def(&mapped, db_type)));
+                                    parts.push(format!(
+                                        "  MODIFY COLUMN {}",
+                                        column_def(&mapped, db_type, source_dialect)
+                                    ));
                                 }
                             } else {
                                 let name = quote_id(&column.name, db_type);
@@ -3892,7 +3897,12 @@ fn generate_schema_sync_sql_inner(
                                 }
                                 if column.changes.iter().any(|change| change.starts_with("default:")) {
                                     parts.push(if let Some(default) = &source.column_default {
-                                        let default = profile.format_default(&mapped.data_type, default);
+                                        let default = profile.format_default(
+                                            &mapped.data_type,
+                                            default,
+                                            source.extra.as_deref(),
+                                            source_dialect,
+                                        );
                                         format!("  ALTER COLUMN {name} SET DEFAULT {default}")
                                     } else {
                                         format!("  ALTER COLUMN {name} DROP DEFAULT")
@@ -3911,7 +3921,7 @@ fn generate_schema_sync_sql_inner(
                                     parts.push(format!(
                                         "  CHANGE COLUMN {} {}",
                                         old_name,
-                                        column_def(&mapped, db_type)
+                                        column_def(&mapped, db_type, source_dialect)
                                     ));
                                 }
                                 RenameColumnSyntax::RenameColumn => {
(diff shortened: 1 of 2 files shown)


t8y2 (changes requested):
The current default-value classifier still changes or breaks valid defaults:

1. An empty string is returned without quoting, so MySQL `DEFAULT ''` can still become a bare `DEFAULT`.
2. Any value containing `(` is treated as an expression, so a string such as `a(b)` becomes invalid `DEFAULT a(b)`.
3. Classification is based on the target MySQL profile rather than the existing source dialect and source metadata. Cross-dialect synchronization can turn expressions such as PostgreSQL `CURRENT_USER` or Oracle `USER` into fixed strings.

Please distinguish literals and expressions using the source dialect/metadata, and add coverage for empty strings, `a(b)`, PostgreSQL `CURRENT_USER`, Oracle `USER`, and precision-bearing temporal functions.

q396921921:
Thanks for the review — all three points are addressed in 05f8edb91.

I reproduced the classifier's actual behavior against a live MySQL 8.4 instance before rewriting it, and it confirmed the shape-based heuristic (`contains('(')`) genuinely cannot distinguish a literal from an expression:

```sql
CREATE TABLE t (
  id    CHAR(36)    DEFAULT (uuid()),   -- expression
  code  VARCHAR(20) DEFAULT 'a(b)'      -- literal
);
```

```
COLUMN_NAME  COLUMN_DEFAULT  EXTRA
id           uuid()          DEFAULT_GENERATED
code         a(b)                          <- empty
```

Both come back with the exact same shape (`ident(...)`); only `EXTRA` tells them apart. So the fix now uses `EXTRA = DEFAULT_GENERATED` (already captured in `ColumnInfo.extra`) as the primary signal, with a keyword-with-optional-precision fallback (`CURRENT_TIMESTAMP`, `CURRENT_TIMESTAMP(3)`, `NOW()`) for engines that don't populate `EXTRA` the same way — which also fixes the precision-bearing case, since the old all-alphabetic check rejected `CURRENT_TIMESTAMP(3)` due to the digits/parens.

1. **Empty string** — `mysql_default_needs_quoting` no longer bails out on an empty value; a string-type column with an empty default is now quoted to `''`. Test: `mysql_create_table_quotes_empty_string_default`.
2. **`a(b)`** — the `contains('(')` heuristic is gone. Literals are now quoted regardless of embedded parens; only an `EXTRA`/keyword-shape match skips quoting. Test: `mysql_create_table_quotes_literal_that_contains_parens`.
3. **Source dialect / cross-dialect expressions** — `format_default` now takes `source_dialect: Option<DialectKind>` (already threaded through `generate_create_table_sql`/`generate_schema_sync_sql_inner`) and only runs the MySQL-specific unquote fix-up when the source is MySQL-family (or unknown). Postgres `CURRENT_USER` / Oracle `USER` sourced into a MySQL target now pass through untouched instead of becoming string literals. Tests: `mysql_target_from_postgres_source_leaves_current_user_default_unquoted`, `mysql_target_from_oracle_source_leaves_user_default_unquoted`.

Also added `mysql_create_table_leaves_precision_timestamp_default_unquoted(_without_extra)` for the precision-bearing temporal function case.

One pre-existing test (`dropped_mysql_table_rollback_preserves_defaults_comments_indexes_and_fk`) assumed MySQL reports expression defaults as `(uuid())` with wrapping parens — the live-DB check above shows that's not accurate (MySQL strips the parens and signals the expression via `EXTRA` instead), so I corrected that fixture/assertion to match.

`cargo test -p dbx-core --lib schema_diff::` (203 passed, incl. 6 new tests from this round), `cargo test -p dbx-core --lib sql_dialect::ddl_profile::` (6 passed), `cargo clippy -p dbx-core --lib` (clean), `cargo fmt -p dbx-core` (clean), `cargo check --workspace` (clean).

q396921921:
Closing — this is a duplicate of #5482, which was already fixed and merged via #6104 (`default_literal()` in schema_diff.rs). I compared the two: #6104's implementation covers everything this PR addresses (empty-string quoting, the `a(b)` literal-vs-expression ambiguity via `EXTRA`, and source-dialect gating for `CURRENT_USER`/`USER`) plus additional cases we didn't handle (hex/binary literals, `N'x'`/`b'x'`/`x'..'`/`e'..'` prefixed literals, the `(uuid())` wrapped-expression form). No incremental value left to carry forward, so closing rather than asking for another review pass on a redundant change.

PR #6147, opened 2026-08-13
fix(desktop): keep downloading in background on accidental dialog dismiss

### LINKED ISSUE #6125: [Bug] 更新对话框点击外部关闭后下载进度丢失
### 来源
QQ 群反馈 · **反馈人**：柴头柴脑

### AI 摘要
更新过程中，点击更新对话框外部区域会导致对话框关闭；再次检查更新并下载时，无法恢复上次下载进度，只能重新下载。

### 原始描述
更新过程中，点击更新对话框外的区域，会关闭对话框，重新点击检查更新，并下载更新，只会重新下载，无上次下载进度

q396921921:
## Problem

Fixes #6125

While an update is downloading, clicking outside the update dialog (or pressing Escape) closes it via the same `handleOpenChange` path used by the explicit close/cancel buttons. That path emits `cancel-download` whenever `isDownloadingUpdate` is true, so an accidental outside click silently cancels the in-flight download. Since the backend has no resume/checkpoint support, reopening the dialog and clicking "Download & Install" again always restarts from 0%.

## Fix

`UpdateDialog.vue` already has an `isCloseBlocked` guard (installation only) that prevents `@interact-outside` / `@escape-key-down` from closing the dialog. This PR adds a second guard, `blocksImplicitDismiss` (`isInstallingUpdate || isDownloadingUpdate`), applied only to those two implicit-dismiss event handlers.

- Outside click / Escape while downloading: now a no-op — dialog stays open, download keeps running in the background.
- Explicit close (X) button / footer "Cancel" button: unchanged — still cancels the download, exactly as before (covered by the existing `cancels the background download when the close button is clicked` test).

## Testing

- Added two component tests to `UpdateDialog.spec.ts` that dispatch a real `pointerdown` outside the dialog and a real `Escape` keydown while `isDownloadingUpdate` is true, asserting `cancel-download` is not emitted and the dialog stays open. Verified these fail against the pre-fix code (real repro) and pass after the fix.
- `pnpm vitest run` — 931 files / 8803 tests passing, no regressions.
- `pnpm typecheck` and `oxlint` pass.

### DIFF
--- a/apps/desktop/src/components/layout/UpdateDialog.spec.ts
+++ b/apps/desktop/src/components/layout/UpdateDialog.spec.ts
@@ -119,6 +119,11 @@ async function pressEscape() {
   await flushDialog();
 }
 
+async function clickOutside() {
+  document.dispatchEvent(new PointerEvent("pointerdown", { bubbles: true, cancelable: true }));
+  await flushDialog();
+}
+
 afterEach(() => {
   for (const app of mountedApps.splice(0)) app.unmount();
   document.body.innerHTML = "";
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in fbf240e, will be released in the next version.

PR #6153, opened 2026-08-13
feat(redis): add decoded unicode view for the JSON editors

Abeautifulsnow:
## Summary

Adds a **Raw / Decoded** display toggle to the three Redis JSON editors (String JSON, native RedisJSON, and hash member JSON) so non-ASCII `\uXXXX` escapes render as their decoded characters while the raw source is preserved on save.

Decoded mode is a view-layer transform only:

- `decodeJsonUnicodeEscapes` decodes non-ASCII `\uXXXX` escapes inside JSON string tokens — escape-aware scanning means an escaped quote (`\"`) never terminates the token, surrogate pairs combine into a single code point, and unpaired surrogates stay escaped.
- `mapDisplayToRaw` maps a decoded-mode draft back onto the canonical raw source. Untouched escapes and the original whitespace/formatting are preserved; inserted display text is written back verbatim as JSON source, so `\\`, `\"`, and control escapes keep their exact value semantics (no double-escaping on save).
- The toggle persists per user in `localStorage`, defaults to `decoded`, and **preserves unsaved drafts** when switching modes — the active draft is converted through the mode transform (`decodeJsonUnicodeEscapes` ⇄ `mapDisplayToRaw`) instead of being reset.

Two review findings were folded into this PR:
1. Decoded-mode edits containing `\\`, `\"`, or control escapes are no longer re-escaped on save (JSON semantics preserved end-to-end).
2. An escaped quote inside a string token is consumed inside the token rather than treated as its terminator, so later `\uXXXX` escapes stay decodable.

## Change Type

- [x] New feature
## Change Type

- [x] New feature
- [ ] Bug fix
- [ ] Performance improvement
- [ ] Code refactor
- [ ] Documentation update
- [ ] CI / Build

## Frontend Changes

This PR changes UI: a two-segment **Raw / Decoded** toggle is added to the header of the String JSON editor, the native RedisJSON editor, and the hash member JSON editor, plus the new `redis.jsonViewRaw` / `redis.jsonViewDecoded` i18n keys.

- [x] This PR includes frontend changes and a screenshot/recording is attached (below)

image

image

## Verification

- [x] `vitest` — `safeJsonFormat.spec.ts` (44 tests), `RedisValueViewer.redisJson.spec.ts` + `redisValuePresentation.spec.ts` (40 tests); combined run of 4 related specs: 89 tests passed
- [x] `vue-tsc --noEmit --project apps/desktop/tsconfig.json` — clean
- [x] `oxlint --vue-plugin apps/desktop/src` — no new errors
- [ ] `make check` — not run (local environment)
- [ ] `make cargo-check-fast` — not run (no Rust changes)

## Related Issue

Related to #6074


t8y2 (commented):
Patched in `22c444490`.

The decoded JSON edit mapping now computes its diff on Unicode code points and converts the resulting ranges back to UTF-16 offsets. A supplementary character is therefore never split into independent high and low surrogates when mapping an edited draft back to the raw JSON source.

Added a regression covering an escaped `\ud83d\ude00` value changed from 😀 to 😁, then saved, parsed, and decoded again.

Checks run on the patched content:
- `vitest run apps/desktop/src/lib/common/__tests__/safeJsonFormat.spec.ts apps/desktop/src/lib/__tests__/redis/RedisValueViewer.redisJson.spec.ts` — 2 files, 60 tests passed (45 + 15)
- `oxfmt --check apps/desktop/src/lib/common/safeJsonFormat.ts apps/desktop/src/lib/common/__tests__/safeJsonFormat.spec.ts`
- `git diff --check`

This corrects the previous review body's inaccurate 89-test count; the earlier pre-patch run contained 59 tests, and this patched run contains 60.

t8y2 (commented):
Maintainer patch verified at `22c444490`.

The decoded JSON edit mapping now diffs Unicode code points and maps the resulting ranges back to UTF-16 offsets, so supplementary characters cannot be split into independent high and low surrogates during save.

Regression coverage changes an escaped `\ud83d\ude00` value from 😀 to 😁, then verifies the saved JSON parses correctly and decodes back to the edited draft.

Checks run on this exact patched content:
- `pnpm exec vitest run apps/desktop/src/lib/common/__tests__/safeJsonFormat.spec.ts apps/desktop/src/lib/__tests__/redis/RedisValueViewer.redisJson.spec.ts` — 2 files, 60 tests passed (45 + 15)
- `pnpm exec oxfmt --check apps/desktop/src/lib/common/safeJsonFormat.ts apps/desktop/src/lib/common/__tests__/safeJsonFormat.spec.ts` — passed
- `git diff --check` — passed

The previous review was submitted against `23f362443` and remains attached to that historical commit. This review records the completed maintainer patch against the current exact head.

t8y2:
Thanks for the contribution! Merged in a3eaac3, will be released in the next version.

PR #6154, opened 2026-08-13
fix(editor): wrap generated SQL templates instead of overflowing the pane

### LINKED ISSUE #6038: [Bug] 复现步骤：右键表生成SQL二级菜单查询语句、更新语句，自动生成的语句超出边框
### 数据库类型和版本

mssqlserver

### 问题描述

复现步骤：右键表生成SQL二级菜单查询语句、更新语句，自动生成的语句超出边框。
期望结果：生成的语句在边框内。
Image

### 支持信息

DBX 版本: v0.5.82
运行环境: 桌面端
操作系统: Windows 10.0.26100.9168
系统架构: x86_64

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

_No response_

q396921921:
## Summary
- Right-clicking a table → "生成SQL" (Generate SQL) → 查询语句/更新语句/插入语句/删除语句 (SELECT/UPDATE/INSERT/DELETE) opens a new tab whose generated SQL joins every column onto a single line. With word-wrap off by default, wide tables produced a line that overflowed the editor pane instead of wrapping (repro: mssql table with 20+ columns, matches the issue screenshot).
- `queryStore.createTab` now accepts a `forceWordWrap` option; `openSqlTemplateTab` (used by all "Generate SQL" template handlers) passes `forceWordWrap: true` when creating the tab.
- `ContentArea.vue` forwards `activeTab.forceWordWrap` into `QueryEditor`'s `forceWordWrap` prop — that prop already existed and is fully wired to a CodeMirror word-wrap compartment, but nothing was passing it before.
- Only these generated-template tabs wrap; the global word-wrap editor setting is untouched for normal editing.

Fixes #6038

## Test plan
- [x] Added `apps/desktop/src/stores/__tests__/queryStore.templateForceWordWrap.spec.ts`: confirmed it fails before the fix (reproducing the bug) and passes after — covers `createTab` defaulting to no forced wrap, `createTab` honoring `forceWordWrap`, `openSqlTemplateTab` requesting it, and `ContentArea` forwarding it to `QueryEditor`.
- [x] `pnpm vitest run` — full suite: 8811 tests, all passing (one unrelated `exportSmoke.spec.ts` cold-cargo-build timeout on first run, passed on retry).

### DIFF
--- a/apps/desktop/src/components/layout/ContentArea.vue
+++ b/apps/desktop/src/components/layout/ContentArea.vue
@@ -1060,6 +1060,7 @@ defineExpose({ focusSearch, refreshData, refreshQueryEditorCompletionCache, hand
               :statement-execution-markers="activeStatementExecutionMarkers"
               :initial-viewport="activeTab.editorViewport"
               :initial-selection="activeTab.editorSelection"
+              :force-word-wrap="activeTab.forceWordWrap"
               @update:model-value="emit('editorUpdate', activeTab.id, $event)"
               @selection-change="emit('editorSelectionChange', $event)"
               @send-selection-to-ai="emit('sendSelectionToAi', $event)"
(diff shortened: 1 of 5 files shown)


t8y2:
Thanks for the contribution! Merged in 68c22e8, will be released in the next version.

PR #6166, opened 2026-08-13
fix: use schema instead of database for GaussDB M mode view DDL

AndrewDi:
In GaussDB M mode (MySQL-compatible), the schema parameter received from the frontend is the actual schema name (e.g. dbe_perf), not the database name (e.g. pubdb). The previous code passed the database name to mysql_object_source_sql, generating SHOW CREATE VIEW `pubdb`.`view_name` which failed because pubdb is a database, not a schema.

Fix gaussdb_m_view_object_source_sql to use the schema parameter instead of the database parameter when building the SHOW CREATE VIEW SQL.

### DIFF
--- a/crates/dbx-core/src/schema.rs
+++ b/crates/dbx-core/src/schema.rs
@@ -2998,22 +2998,41 @@ mod tests {
         config.driver_profile = Some("gaussdb-m".to_string());
 
         assert_eq!(
-            gaussdb_m_view_object_source_sql(&config, "tenant`db", "active`users", &db::ObjectSourceKind::View)
-                .as_deref(),
-            Some("SHOW CREATE VIEW `tenant``db`.`active``users`")
+            gaussdb_m_view_object_source_sql(
+                &config,
+                "connection_db",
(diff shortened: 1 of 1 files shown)


AndrewDi:
@zipg 修复获取view的schema错误

t8y2 (commented):
Maintainer patch pushed to `4c5f0792` on top of `38bf4ec5`.

Root cause: the GaussDB M-mode view source path could qualify `SHOW CREATE VIEW` with the connection database instead of the tenant schema when those names differ.

Patch: keep database and schema inputs explicit, qualify the view with the schema, and add a regression where connection database and tenant schema intentionally differ. No dependencies or additional database requests were added.

Validation:
- `gaussdb_m_view_object_source_sql_is_qualified_and_gated` — 1 passed

t8y2 (approved):
Maintainer patch and exact-head CI verified.

t8y2:
Thanks for the contribution! Merged in 9c50cc78, will be released in the next version.

PR #6173, opened 2026-08-13
fix: GaussDB M模式bool列显示为true/false

AndrewDi:
GaussDB M模式的JDBC驱动将BIT/BOOLEAN类型的列以byte[]
(0x74=true, 0x66=false)形式返回，而非java.lang.Boolean。
readValue方法中byte[]检查优先于Boolean检查，导致
byte[]被转换为十六进制字符串显示。

修复：对BIT/BOOLEAN类型的列优先使用rs.getBoolean()
获取正确的布尔值。

### DIFF
--- a/plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
+++ b/plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
@@ -3028,10 +3028,7 @@ public final class DbxJdbcPlugin {
     ) throws SQLException {
         int columnType = meta.getColumnType(index);
 
-        // GaussDB M-mode JDBC driver may return boolean columns as byte[]
-        // (0x74 for true, 0x66 for false) instead of java.lang.Boolean.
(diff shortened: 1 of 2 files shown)


AndrewDi:
@zipg PTAL

t8y2 (commented):
Maintainer patch pushed to `49a4de5f` on top of `38bf4ec5`.

Root cause: treating every JDBC `BIT` value as a boolean collapsed multi-bit binary values, while `BOOLEAN` columns should continue to follow JDBC `getBoolean`/`wasNull` semantics.

Patch: keep JDBC boolean access for `BOOLEAN`; for `BIT`, convert only single-byte `t`/`f` values and preserve multi-byte values as hexadecimal. No dependencies or extra database requests were added.

Validation:
- focused `DbxJdbcPluginTest` boolean/bit cases — passed
- Gradle JDBC compile and test tasks — successful

t8y2 (approved):
Maintainer patch and exact-head CI verified.

t8y2:
Thanks for the contribution! Merged in b22cd907, will be released in the next version.

PR #6183, opened 2026-08-13
fix(objects): support ctrl/shift-click multi-select in database browser

### LINKED ISSUE #6152: [Bug] 按住ctrl或者shift键，然后使用鼠标点选，没办法去多选表进行操作
### 数据库类型和版本

mysql，DM，人大金仓

### 问题描述

1、复现方法：打开数据库，进入详情，或者就在左侧链接列表，按着ctrl或者shift，鼠标点选
2、期望结果：可以选择，不用使用右上角的多选功能

### 支持信息

DBX 版本: v0.5.83
运行环境: 桌面端
操作系统: Windows11 Beta预览版  10.0.26220.9022
系统架构: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Summary

The sidebar tree already supports ctrl/cmd-click toggle and shift-click range-select for tables (merged just before this issue was filed). The database detail view (`ObjectBrowser.vue`, opened via "浏览对象" / entering a database) only offered a checkbox-toggle mode: click a toolbar button to reveal a checkbox column, then click each row's checkbox individually — no way to ctrl/cmd-click or shift-click rows directly.

- Adds a pure helper `objectBrowserTableSelectionRange` (mirrors the sidebar's existing `sidebarTreeSelection.ts` range logic) that computes the contiguous TABLE-row range between an anchor and the clicked row
- Wires ctrl/cmd-click (toggle) and shift-click (range-select) into `onRowClick`, gated to `TABLE` rows since only tables are selectable
- The existing bulk-action toolbar (export/delete/clear selection) already reacts to the same selection state, so no other changes were needed

Fixes #6152

## Test plan

- [x] `objectBrowserSelection.spec.ts`: 6 new unit tests covering forward/backward range, non-table rows inside the span being dropped, missing-anchor fallback, and single-row collapse
- [x] Full suite: 8813 tests passing
- [x] Manual verification in a real dev build: created a throwaway SQLite connection with 6 tables, drove real clicks via CDP raw input (Cmd-click on macOS is the actual modifier — Ctrl+Click is intercepted as right-click by macOS/Chrome). Confirmed Cmd-click toggles individual tables and shift-click selects the contiguous range from the anchor, with the bulk-action toolbar updating correctly.

### DIFF
--- a/apps/desktop/src/components/objects/ObjectBrowser.vue
+++ b/apps/desktop/src/components/objects/ObjectBrowser.vue
@@ -120,7 +120,7 @@ import {
   type ObjectBrowserSortKey,
 } from "@/lib/table/objectBrowserRows";
 import { isSourceOnlyObjectBrowserRow, resolveRowClickAction, shouldDeferSingleClick, type ObjectBrowserRowAction } from "@/lib/table/objectBrowserRowAction";
-import { objectBrowserTableSelectionRange } from "@/lib/table/objectBrowserSelection";
+import { objectBrowserTableSelectionAnchor, objectBrowserTableSelectionRange } from "@/lib/table/objectBrowserSelection";
 import { customTypeCapabilities, supportsTypeObjectSource } from "@/lib/database/databaseObjectCapabilities";
 import { filterObjectBrowserTableColumns } from "@/lib/table/objectBrowserTableInfo";
 import { createSidePanelRequestGuard } from "@/lib/table/sidePanelRequestGuard";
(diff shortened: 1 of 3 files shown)


t8y2 (commented):
Maintainer patch applied in `3ff6655d`.

Root cause: when Shift-click started without a valid visible table anchor, the range helper fell back to the clicked row only for that calculation, but the component did not persist the fallback anchor. Consecutive Shift-clicks therefore continued to collapse to a single row.

Patch: resolve the effective visible table anchor through a small pure helper, persist that anchor before selecting the range, and cover both the valid-anchor and missing/filtered-anchor cases.

Checks run:
- focused Vitest: 1 file, 8 tests passed
- `pnpm typecheck`
- focused Oxlint: 0 warnings / 0 errors
- focused Oxfmt check
- `git diff --check`

t8y2:
Thanks for the contribution! Merged in 10b1ba4, will be released in the next version.

PR #6184, opened 2026-08-13
fix(editor): keep indenting a selection instead of letting a stray completion hijack Tab

### LINKED ISSUE #5419: [Bug] 编辑器里面按Tab键对齐问题
### 数据库类型和版本

PostgreSQL 11.12, compiled by Visual C++ build 1914, 64-bit

### 问题描述

1.编辑器选中几行语句，按tab键对齐，按第一次可以正常往右对齐，但同时会悬浮显示一些信息（不知道是怎么关联出来的），然后再按一次tab键，会将选中的语句替换为悬浮显示的第一条选项
2.期望按tab键可以一直往右对齐
3.实际请看gif动图

Image

### 支持信息

DBX 版本: v0.5.75
运行环境: 桌面端
操作系统: Windows 10.0.26200.8655
系统架构: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Problem

Pressing Tab to indent a multi-line selection in the SQL editor works the first time, but a completion popup can appear as a side effect of that edit (the doc-changed handler re-triggers SQL completion based on cursor context, regardless of what kind of edit just happened). The next Tab press then accepts that unrelated completion instead of indenting further, replacing the selected text with a suggestion.

## Fix

`acceptCompletionOrNextSnippetField` (used by the Tab keymap) now only goes through the completion-accept/wait path when the selection is empty. A non-empty selection always goes straight to block indent (`indentMore`), so a completion popup that appears mid-indent can never hijack a subsequent Tab press. Snippet-field navigation is unaffected either way.

## Testing

- Added two regression tests in `queryEditorCompletionKeymap.spec.ts` that extract the real `handleTab`/`acceptCompletionOrNextSnippetField` function bodies from `QueryEditor.vue` (same harness this file already uses) and assert that a multi-line selection keeps calling `indentMore` instead of `acceptCompletion`, both when completion is already `active` and when it's `pending` and resolves after the Tab press. Both fail against the pre-fix code and pass after.
- Full suite: `pnpm test` — 937 files / 8926 tests passing.

Fixes #5419

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -1973,9 +1973,14 @@ function extendQueryEditorSelectionForView(currentView: EditorViewType): boolean
 }
 
 function acceptCompletionOrNextSnippetField(view: EditorViewType): boolean {
-  const completionStatus = codeMirrorCompletionStatus?.(view.state) ?? null;
-  if (completionStatus === "active" && (codeMirrorAcceptCompletion?.(view) ?? false)) return true;
-  if (completionStatus) return waitForCompletionTab(view);
+  // A non-empty selection means Tab is being used for block indent, not word
+  // completion. A completion popup can still appear as a side effect of the
+  // indent edit itself, so it must never hijack this (or a following) Tab press.
+  if (view.state.selection.main.empty) {
+    const completionStatus = codeMirrorCompletionStatus?.(view.state) ?? null;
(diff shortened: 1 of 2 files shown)


q396921921:
Friendly bump — this has been sitting since 08-13 with green CI and no review yet. Happy to address any feedback whenever you get a chance to take a look. Thanks!

q396921921:
关闭后重新提交一个新 PR（分支内容已 rebase 到最新 main），方便重新触发 review 流程。

PR #6202, opened 2026-08-14
fix(structure): let charset/collation selects fill their resizable column

### LINKED ISSUE #6052: 编辑表结构里面 字符集和排序规则的显示长度不够
### 数据库类型和版本

mariadb 11.8.2

### 问题描述

1. 编辑表结构里面 字符集和排序规则的显示长度不够，例如utf8mb3_uca1400_ai_ci  这个规则长的话 就显示不全，把表格分割线拉长也无法显示出.
2. 5.77 更新日志里面说 有表头序号，但是看表没序号  是哪里设置或者开启.

Image
Image
Image

### 支持信息

DBX 版本: v0.5.82
运行环境: 桌面端
操作系统: Windows 10.0.19045.6466
系统架构: x86_64

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

_No response_

q396921921:
## Summary
- `TableStructureEditor.vue`'s charset/collation `SearchableSelect` triggers were hardcoded to `w-20`/`w-28` (~112px), capping their display width regardless of the actual (user-resizable) `<td>` width — so dragging the column divider wider had no effect on truncated values, exactly as reported.
- The sibling "data type" select in the same table already used `w-full` to correctly fill its column; charset/collation were simply inconsistent with that existing pattern.
- Changed both triggers to `w-full` so they respect the resizable column, matching the data-type select's behavior.

## Test plan
- [x] Added `TableStructureEditor.charsetCollation.spec.ts`, mounting the real component with `db_type: "mysql"` and asserting the charset/collation trigger classes — fails on unmodified code, passes after the fix.
- [x] Reproduced live against a local MySQL 8.4 instance: created a column with `COLLATE utf8mb4_ja_0900_as_cs_ks`. Before the fix, the structure editor showed `utf8mb4_j...` truncated even with the column dragged fully wide; after the fix it displays in full.
- [x] Full suite: `vitest run` — 6069/6070 tests pass (the 1 failure is a pre-existing, unrelated `exportSmoke` cold-compile timeout on first Rust build, not caused by this change).
- [x] `oxlint` and `vue-tsc --noEmit` both clean.

Fixes #6052

### DIFF
--- a/apps/desktop/src/components/structure/TableStructureEditor.charsetCollation.spec.ts
+++ b/apps/desktop/src/components/structure/TableStructureEditor.charsetCollation.spec.ts
@@ -1,0 +1,331 @@
+// @vitest-environment happy-dom
+
+import { createApp, nextTick, type App } from "vue";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const mocks = vi.hoisted(() => ({
+  connection: {
+    id: "structure-charset-test",
+    name: "MySQL",
+    db_type: "mysql",
+    driver_label: "MySQL",
+  },
+  ensureConnected: vi.fn(),
+  listDataTypes: vi.fn(),
+  buildTableStructureChangeSql: vi.fn(),
+  updateEditorSettings: vi.fn(),
+  loadObjectDdl: vi.fn(),
+  invalidateObjectDdl: vi.fn(),
+  loadObjectMetadataFacet: vi.fn(),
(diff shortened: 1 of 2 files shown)


t8y2 (approved):
Validated on exact head 38c13aeabbe5a2278089962d47f0293f2b376ae2.

t8y2:
Thanks for the contribution! Merged in 5b40487, will be released in the next version.

PR #6206, opened 2026-08-14
fix(sql): require whitespace after -- for MySQL line comments

### LINKED ISSUE #5382: [Bug] 1：编辑表结构，外键、触发器等切换过去不会显示，需要手动点刷新表结构才会显示
### 数据库类型和版本

mysql dm

### 问题描述

1：编辑表结构，外键、触发器等切换过去不会显示，需要手动点刷新表结构才会显示。
2：执行sql脚本时，如果脚本中带有 --  格式的注释信息时，无法正确执行，复制到查询执行器中能正常执行

### 支持信息

DBX 版本: v0.5.74
运行环境: 桌面端
操作系统: Windows 10.0.26100.8246
系统架构: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Summary

Fixes #5382 — specifically the second bug described in the issue body: running a `.sql` script/file with `--` comments doesn't execute correctly, while pasting the same SQL into the query editor works fine.

(The title bug in #5382 — foreign keys/triggers tab not refreshing until manual refresh — was already fixed upstream by `5325c4263`/`7be11f54c`/`560087256` before this PR; I verified that separately and left it alone here.)

## Root cause

MySQL requires the second dash of a `--` comment opener to be followed by whitespace (or end of input) — `5--1` is subtraction (`5 - -1`), not a comment. DBX's `SqlStatementSplitter` (used by both "Run Script" and "Execute SQL File") treated any `--` as a comment opener unconditionally, with no dialect check. So a script like:

```sql
INSERT INTO t VALUES (1, 5--1);
INSERT INTO t VALUES (2, 10);
```

had its first statement's terminating `;` swallowed into the (wrongly detected) comment, silently merging both statements into one multi-statement chunk sent to the driver in a single round trip — under-reporting the statement/row counts in the execution progress UI (3 statements / 1 affected row instead of 4 / 2), and would hard-fail with a syntax error on any connection where multi-statement execution isn't enabled.

The query editor's single-statement execute path sends raw SQL text straight to the server without client-side splitting, so it was never affected — matching the reporter's "works pasted into the editor, fails as a script" observation.

## Fix

Added a `requires_whitespace_after_line_comment_dashes` flag to `SqlDialectProfile`, set only for MySQL-compatible dialects (MySQL, Doris, StarRocks, ManticoreSearch, Goldendb). The splitter now only starts a line comment on `--` when the following character is whitespace or end-of-input, for dialects that require it. Postgres, Oracle/DM, SQL Server, etc. are unaffected — they already treat `--` as an unconditional comment opener, matching their own grammar.

## Test plan

- [x] New unit tests in `crates/dbx-core/src/sql.rs`: `5--1` is not treated as a comment for MySQL, doesn't merge adjacent statements, existing `-- comment` (with space) behavior is unchanged, non-MySQL dialects are unaffected, and the no-space opener correctly handles chunk-boundary splitting.
- [x] `cargo test -p dbx-core --lib sql::` — 865 passed, 0 failed.
- [x] `cargo clippy -p dbx-core --lib -- -D warnings` — clean.
- [x] Manual end-to-end repro against a real local MySQL 8.0 instance via "Execute SQL File": before the fix, a script with `INSERT ... VALUES (1, 5--1); INSERT ...;` reported 3 statements / 1 affected row; after the fix, 4 statements / 2 affected rows, matching the source file's actual statement count.

### DIFF
--- a/crates/dbx-core/src/sql.rs
+++ b/crates/dbx-core/src/sql.rs
@@ -296,14 +296,15 @@ fn sql_file_encoding_error() -> String {
 }
 
 /// Whether a `--` sequence starts a line comment. MySQL (and MySQL-wire-compatible
-/// dialects) requires the second dash to be followed by whitespace or end-of-input —
+/// dialects) requires the second dash to be followed by whitespace, a control character,
+/// or end-of-input —
 /// `5--1` is subtraction (`5 - -1`), not a comment. Other dialects (Postgres, Oracle/DM,
 /// SQL Server, ...) treat `--` as a comment opener unconditionally.
 fn dash_dash_starts_line_comment(profile: SqlDialectProfile, char_after_dashes: Option<char>) -> bool {
     if !profile.requires_whitespace_after_line_comment_dashes {
         return true;
     }
-    char_after_dashes.is_none_or(|ch| ch.is_whitespace())
+    char_after_dashes.is_none_or(|ch| ch.is_whitespace() || ch.is_control())
 }
 
 #[derive(Default)]
@@ -317,6 +318,7 @@ pub struct SqlStatementSplitter {
     dollar_quote_tag: Option<String>,
     postgres_dollar_quoted_routine: bool,
     previous: Option<char>,
+    pending_mysql_line_comment_dashes: bool,
     custom_delimiter: Option<String>,
     options: SqlParsingOptions,
 }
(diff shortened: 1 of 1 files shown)


t8y2 (commented):
Maintainer patch pushed in d100a05.

Root cause: MySQL accepts whitespace or control characters after the second dash, but the splitter only checked Rust whitespace. It also treated the end of each streamed chunk as the real SQL EOF, so a chunk ending in a complete `--` could incorrectly turn the next chunk into a comment.

Patch: accept control characters, preserve an undecided MySQL `--` state across chunks, resolve it only when the next character or real `finish()` is available, and add regressions for the complete-dash chunk boundary and control-character opener.

Checks: MySQL-filtered Rust tests passed; `cargo check --no-default-features`, formatting, and `git diff --check` passed. MySQL 8.4.2 probes confirmed `5--1` remains arithmetic while real EOF and `0x01` start comments. No dependency changes.

t8y2 (approved):
Validated on exact head d100a056faf917ef50d42eac70bf4903650b826c.

t8y2:
Thanks for the contribution! Merged in 880d635, will be released in the next version.

PR #6213, opened 2026-08-14
fix(grid): require drag distance before extending cell selection

q396921921:
## Summary

Refs #6203 — on macOS, a plain left-click in the query result grid was frequently misread as a drag, auto-selecting adjacent cells/rows instead of the one clicked.

- Clicking a data cell or the row-number gutter started extending the selection on the very next `pointermove`/`mouseenter` event, with zero minimum drag distance.
- Normal trackpad jitter during a click (a few pixels of movement between mousedown and mouseup) was enough to cross a cell/row boundary, turning a single click into a multi-cell or multi-row selection.
- The same class of bug was already fixed for the tab bar (`TAB_DRAG_HORIZONTAL_THRESHOLD` in `useTabDrag.ts`) but the grid's `useDataGridSelection.ts` never got an equivalent guard.

## Fix

Added a 12px drag-confirmation threshold (matching the tab bar's existing value) in `useDataGridSelection.ts`, gating both:
- the document-level `pointermove` handlers used while dragging (`handleSelectionPointerMove` / `handleRowSelectionPointerMove` / `finishRowSelection`), and
- the DOM `mouseenter`-driven extension path (`DataGrid.vue`'s `onCellMouseenter`, which calls `extendCellSelection` directly with no pixel-distance info of its own — exposed via a new `isCellSelectionDragConfirmed()` getter).

A real drag (movement past the threshold) still selects a range normally; only sub-threshold jitter is now ignored.

Scope note: the issue also reported similar-looking symptoms on the tab bar and the SQL editor. The tab bar turned out to already be fixed by an earlier commit. The editor's click-selects-a-word behavior comes from CodeMirror's native text-selection handling (not grid-specific), so it's left out of this PR to keep the change small and low-risk.

## Testing

- Reproduced first with a headless-browser script that simulates a real mousedown → small pointer jitter (a few px) → mouseup sequence against a live MySQL-backed grid, confirming the multi-cell/row selection bug before the fix and a clean single-cell/row click after it.
- Confirmed a genuine larger drag still produces a correct multi-cell/row selection (the fix only suppresses sub-threshold jitter).
- `apps/desktop/src/composables/__tests__/useDataGridSelection.spec.ts` (existing suite, 12/12 passing).
- Full data-grid test surface (`composables/__tests__`, `components/grid/__tests__`, `lib/__tests__/dataGrid`, `lib/dataGrid/__tests__`): 937/937 passing.
- `vue-tsc --noEmit` clean.

## Test plan for reviewers

- [ ] Open a table's data in the grid, click a single cell with a real trackpad — should stay a single-cell selection even with normal click jitter
- [ ] Click a row number in the gutter — should stay a single-row selection
- [ ] Drag-select across multiple cells/rows — should still work as before

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -4305,6 +4305,7 @@ const {
   selectionFocus,
   finishCellSelection,
   extendCellSelection,
+  isCellSelectionDragConfirmed,
   restoreCellSelectionState,
   cellIsSelected,
   columnIsSelected,
@@ -4482,7 +4483,7 @@ function onCellMouseenter(rowIndex: number, visibleColIdx: number, actualColIdx:
     quickDownloadMenuCell.value = retainBinaryCellDownloadMenuForHover(quickDownloadMenuCell.value, { rowIndex, col: actualColIdx });
     if (!isScrolling.value) hoveredDetailCell.value = { rowIndex, col: actualColIdx };
   }
-  extendCellSelection(rowIndex, visibleColIdx);
+  if (isCellSelectionDragConfirmed()) extendCellSelection(rowIndex, visibleColIdx);
 }
 
 function onCellMouseleave(rowIndex: number, actualColIdx: number) {
--- a/apps/desktop/src/composables/__tests__/useDataGridSelection.spec.ts
+++ b/apps/desktop/src/composables/__tests__/useDataGridSelection.spec.ts
@@ -39,6 +39,45 @@ function rowEvent(options: { meta?: boolean; shift?: boolean } = {}): MouseEvent
   } as MouseEvent;
 }
 
+function installPointerDocument() {
+  const originalDocument = globalThis.document;
+  const originalRequestAnimationFrame = globalThis.requestAnimationFrame;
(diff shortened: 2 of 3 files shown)


t8y2 (commented):
Maintainer follow-up in 8d9d172b.

Confirmed risk: the 12 px drag threshold protects both cell and row-gutter selection, while the grid's native `mouseenter` path relies on `isCellSelectionDragConfirmed()`. Without focused coverage, pointer jitter or a finish-path change could silently reintroduce accidental multi-cell or multi-row selection.

Patch: added regression tests for cell and row-gutter movement below 12 px, movement exactly at 12 px, the cell hover-extension gate, and unconfirmed mouseup paths. I also changed the PR body linkage from `Fixes #6203` to `Refs #6203`, because the SQL editor symptom documented in that issue remains outside this PR's scope.

Checks: `pnpm exec vitest run apps/desktop/src/composables/__tests__/useDataGridSelection.spec.ts` (1 file, 16 tests), `pnpm typecheck`, and `git diff --check`. No dependency changes.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 30d13bc7, will be released in the next version.

PR #6221, opened 2026-08-14
Feat/session credential store for no save password

Abeautifulsnow:
## Summary

Separate **persisted connection config** from **runtime session credentials** so that
`save_password = false` connections only prompt for the password once per session
instead of on every connect, while still never persisting it. This also fixes the
ORA-01005 root cause where a save overwrote the runtime password with an empty one
and tore down the pool, and a pool-equivalence bug that kept a stale pool alive after
a saved password changed.

## Problem

For connections with "don't save the database password locally" enabled:

- The frontend re-prompted for the password on **every** `connect()`; there was no
  way to enter it once and reuse it for the rest of the session.
- The backend `sync_connection_configs` path overwrote the runtime config (which
  carried the session password) with the persisted config (empty password), then
  judged the pool stale and destroyed it → subsequent AI/metadata/pool rebuilds read
  an empty password → Oracle **ORA-01005**.
- `connection_configs_pool_equivalent` cleared `password` unconditionally, so
  updating a `save_password = true` connection to a new password left its existing
  pool alive and authenticated with the **old** password.

## Solution

A new in-memory `SessionCredentialStore` on `dbx-core` `AppState`, shared by both the
Tauri and Web backends, keyed by `(owner_scope, connection_id)`:

- **Desktop (Tauri):** owner scope is the empty string (single user).
- **Web:** owner scope is the authenticated session token, injected at the request
  boundary via `tokio::task_local` (`with_credential_owner`) in the auth middleware,
  so deep pool-creation code doesn't need the owner threaded through every signature.
- Credentials live **only in process memory** — never written to SQLite, cloud-sync
  exports, logs, browser storage, or any disk path. `Debug` exposes only connection-id
  keys, never password values.

### Behavior

- First connect prompts as before; the password is recorded for the session.
- Subsequent connects, editor SQL, AI tools, metadata reads, and pool rebuilds reuse
  it without prompting.
- It is dropped on process exit / Web logout / session expiry / connection deletion.
- `save_password = false` connections resolve their password **only** from the store
  (no fallback to the persisted `FileSecretStore`); `save_password = true` keeps the
  existing hydrated-password path.

### Web multi-session isolation

- Credentials are owner-scoped; one login session can't see or reuse another's
  temporary password.
- No-save pools are owner-tracked (`pool_credential_owners`); a request from a
  different session tears down and rebuilds the pool instead of reusing a connection
  established with someone else's password.
- Logout calls `clear_owner(token)` and only clears that session's credentials.

### Frontend

- `hasSessionCredential` is resolved from a new backend status endpoint, so
  `connect()` / `ensureConnected` skip the prompt when a session credential exists.
- Disconnect menu now offers two actions for no-save connections: **Disconnect**
  (keeps the credential) and **Disconnect & forget this password** (drops it, next
  connect prompts again).
- `savePasswordHint` updated to reflect the per-session semantics across all 8 locales.

### Pool invalidation

`connection_configs_pool_equivalent` now ignores the `password` difference **only when
both configs have `save_password = false`** (the "session password vs persisted empty"
case). When either side has `save_password = true`, any password change is treated as a
real parameter change and the pool is destroyed.

## New commands / routes

- Tauri: `session_credential_status`, `forget_session_credential`
- Web: `POST /connection/session-credential-status`, `POST /connection/forget-session-credential`

## Out of scope

- Session-izing transport-layer sub-secrets (SSH / proxy / Redis Sentinel /
  `connection_string` / `init_script` / MQ / MQTT / Nacos) — these are unaffected by
  `save_password`.
- Web instance login passwords.
- Any change to the existing `FileSecretStore` encryption.

## Tests

- `dbx-core`: `SessionCredentialStore` owner isolation / clear_owner / debug
  redaction; `apply_session_credential` owner scoping; `pool_credential_owner_mismatch`;
  `connection_configs_pool_equivalent_ignores_password_differences` and
  `connection_configs_pool_equivalent_detects_saved_password_change`.
- `dbx-web` / `src-tauri`: `sync_connection_configs_ignores_password_only_changes`,
  `connection_final_proxy_port_sanitizes_no_save_passwords`.
- Frontend: `connectionStore.savePassword.spec.ts`.

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [x] `make check` 通过
- [x] 相关测试通过


t8y2 (commented):
已在 `a18c1631` 中补齐本轮维护者修复：

- 全局连接配置发生 host、user 或其他池相关参数变更、删除及 ID 复用时，按 `connection_id` 一次性清除所有 Web owner 的会话凭据和池 owner 状态，避免旧密码或旧池流入新的连接定义；缺失 pool owner 的 Web no-save 池现在按不可信处理。
- `SessionCredentialStore` 改用内部 tuple key，`Debug` 只输出凭据数量，不再暴露 owner/session token、连接 ID 或密码。
- `set` 返回 generation write token；连接失败只 compare-and-remove 当前 generation，旧并发 attempt 不会删除后来 attempt 写入的新密码。

回归测试覆盖双 owner 的编辑、删除与 ID 复用，Debug secret redaction，以及 barrier 控制的 A 写入 → B 覆盖 → A 失败回滚；同时保留 Desktop/Web 与 `save_password=true/false` 的既有行为。

验证通过：相关 Rust/Web/Tauri 测试 21 项、前端 no-save 测试 14 项、`cargo fmt`、workspace `cargo check --no-default-features`、严格 Clippy、前端 typecheck、lint（0 errors）和 `git diff --check`。未新增依赖、轮询、后台 I/O 或数据库请求。

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 1ff24aa, will be released in the next version.

PR #6228, opened 2026-08-14
fix(editor): keep scroll position when closing a view's DDL dialog

### LINKED ISSUE #6067: [Bug] Cmd+单击查看视图 DDL 关闭后，当前 SQL 编辑器会跳转到页面顶部
### 数据库类型和版本

通用（SQL 编辑器 / 视图对象导航，与具体数据库无关）

### 问题描述

1. 复现步骤：
   - 打开一个较长的 SQL 查询页签，将编辑器滚动到中部或底部（不要停在首行）
   - 在 SQL 中对**视图名**使用 `Cmd+单击`（macOS；Windows/Linux 对应 `Ctrl+单击`）打开该视图的 DDL
   - 查看完毕后关闭 DDL 对话框（点关闭、Esc 或右上角 X）
2. 期望结果：回到原来的 SQL 查询页签时，编辑器滚动位置、光标和选区保持打开 DDL 之前的状态
3. 实际结果：当前 SQL 编辑窗口会跳转到页面顶部，原先滚动位置丢失，需要重新找刚才编辑的位置
4. 复现稳定性：每次通过 `Cmd+单击` 查看视图 DDL 再退出都会发生

### 支持信息

DBX 版本: v0.5.82
运行环境: 桌面端
操作系统: macOS 27.0
系统架构: aarch64

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

- 触发路径是查询编辑器里对**视图**做 `Cmd+单击`，不是侧边栏右键「查看 DDL」，也不是执行 SQL。
- 当前实现里，视图的 `Cmd+单击` 会走 `App.vue` 的 `onClickTable`：识别为 view / materialized_view 后打开 `QueryEditorDdlViewDialog`（`DdlViewDialog.vue`），而不是新开结构编辑页签。
- 同类但不同触发的历史问题：[执行 SQL，编辑器会跳转到头部 #4825](https://github.com/t8y2/dbx/issues/4825)（已关闭，触发是执行 SQL，不是关闭 DDL 对话框）。
- 排查时可重点看：
  - 对话框关闭后的焦点还原 / body scroll-lock 是否把外层页面或编辑器滚回顶部
  - `QueryEditor.vue` 的 viewport / selection 恢复（`restoreEditorViewport`、`restoreEditorSelection`、`onActivated`）是否在对话框开关时被误触发
  - CodeMirror `dispatch({ selection })` 默认 `scrollIntoView` 是否在关闭对话框后把视口拉到文档开头
- 建议一并确认：对**表名** `Cmd+单击` 打开表结构 / DDL 后再回到原查询页签，是否也有同样的滚动跳动。

q396921921:
## Problem

Cmd/Ctrl+click on a view name inside the SQL editor opens `DdlViewDialog` to preview its DDL (`onClickTable` in `App.vue`). Closing the dialog snaps a long, scrolled-down query editor back to the very top, losing the user's scroll position, cursor, and selection — reported in #6067 with a stable, every-time repro.

## Root cause

Closing the dialog lets Radix restore focus to whatever element was focused before it opened, via a plain native `element.focus({ preventScroll: true })`. That's normally safe, but in **WebKit** (the actual renderer of the desktop app's webview on macOS) refocusing a `contenteditable` this way resets its native caret to the very start of the element. CodeMirror's DOM observer picks that up as a genuine selection change and scrolls to follow it — snapping the editor to the top.

This only reproduces in WebKit, not Chromium, which is why the codebase already has an established pattern of manually guarding focus restoration in several other Popover/Dialog/DropdownMenu usages (`@close-auto-focus.prevent` in `TemporalCellEditor.vue`, `AiAssistant.vue`, `QueryChart.vue`, `NacosAdminConsole.vue`).

## Fix

`DdlViewDialog.vue`:
- Record the `.cm-editor` ancestor of `document.activeElement` (if any) when the dialog opens.
- On `@close-auto-focus`, if that editor is still in the DOM, call `event.preventDefault()` and restore focus through CodeMirror's own `EditorView.focus()` (found via `EditorView.findFromDOM`) instead of letting the browser do it. `EditorView.focus()` suppresses CodeMirror's selection observer while it restores the DOM selection to match its actual (unmoved) internal state, which avoids the WebKit quirk entirely.
- When the dialog was opened from somewhere other than the query editor (e.g. no `.cm-editor` ancestor), the handler is a no-op and Radix's default behavior is unchanged — no regression there.

## Testing

Reproduced and verified with real browser automation (Playwright) against the actual dev app, connected to a real local MySQL 8.4 database with a genuine view, since this bug does not reproduce in jsdom/happy-dom (not a real browser engine) nor in Chromium (different focus/selection semantics than WebKit):

- **Chromium**: did not reproduce the bug at all (confirms this needed WebKit specifically, matching the desktop app's actual macOS webview).
- **WebKit** (Playwright's webkit engine): reliably reproduced pre-fix (`scrollTop` goes from `822` → `0` after closing via the Close button, and identically via Escape), and confirmed fixed post-fix (`scrollTop` unchanged across close, verified with both the Close button and Escape key).
- To rule out any measurement artifact, I reverted the fix in place, re-captured the broken behavior with screenshots, then reapplied the fix and re-captured the fixed behavior — a genuine A/B comparison, not two independently styled runs.
- `pnpm typecheck` is clean.
- Full `pnpm test`: 9069 passed, 1 pre-existing unrelated failure (`databaseUserAdmin.test.ts`, about `mysqlRenameUserHostSql`) that upstream's `3db119ca4` already fixes independently of this PR.

Fixes #6067

### DIFF
--- a/apps/desktop/src/components/objects/DdlViewDialog.vue
+++ b/apps/desktop/src/components/objects/DdlViewDialog.vue
@@ -51,6 +51,10 @@ const ddlEditorContainer = ref<HTMLDivElement>();
 const ddlSearchPanelRef = ref<InstanceType<typeof EditorSearchPanel>>();
 const ddlEditorView = shallowRef<EditorView | null>(null);
 
+// The CodeMirror editor (if any) that was focused when this dialog opened, so focus can be
+// restored to it on close. Not a ref: read/written outside of render, never needs reactivity.
+let editorRootToRestoreFocus: HTMLElement | null = null;
+
 async function loadDdl(force = false) {
   ddlError.value = "";
   ddlLoading.value = true;
@@ -81,11 +85,33 @@ watch(
   () => props.open,
   async (open) => {
     if (!open) return;
+    const active = document.activeElement;
+    editorRootToRestoreFocus = active instanceof HTMLElement ? active.closest(".cm-editor") : null;
     ddlContent.value = "";
     await loadDdl();
   },
   { immediate: true },
 );
+
+/**
+ * Restores focus through CodeMirror's own `EditorView.focus()` instead of the browser default.
+ *
+ * Radix's default close-auto-focus calls the plain DOM `.focus()` on whatever was focused before
+ * the dialog opened. In WebKit (the desktop app's webview on macOS), refocusing a contenteditable
+ * this way resets its caret to the very start of the document; CodeMirror then treats that as a
+ * real selection change and scrolls to follow it, snapping a long query editor to the top (#6067).
+ * `EditorView.focus()` avoids this by suppressing its own selection observer while it restores the
+ * DOM selection to match its actual (unmoved) internal state.
+ */
+function onDdlDialogCloseAutoFocus(event: Event) {
+  const target = editorRootToRestoreFocus;
+  editorRootToRestoreFocus = null;
+  if (!target || !target.isConnected) return;
+  event.preventDefault();
+  void import("@codemirror/view").then(({ EditorView }) => {
+    EditorView.findFromDOM(target)?.focus();
+  });
+}
 
 /**
  * Creates a lightweight read-only CodeMirror editor inside the dialog.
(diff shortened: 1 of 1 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in a578916, will be released in the next version.

PR #6235, opened 2026-08-14
fix(sql): quote MySQL reserved-word identifiers with backticks in completion

### LINKED ISSUE #6224: [Bug] mysql数据库，tab自动补全数据库和表的时候，没有 ` 标识符
### 数据库类型和版本

mysql8.3.0

### 问题描述

mysql数据库，tab自动补全数据库和表的时候，没有 ` 标识符

### 支持信息

DBX 版本: v0.5.83
运行环境: 桌面端
操作系统: Windows 10.0.26200.7171
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Summary

- SQL editor tab-completion for tables, schemas/databases, columns, routines, and join conditions only backtick/quote-escaped reserved-word identifiers for the PostgreSQL dialect (via `quoteSqlIdentifier`). MySQL identifiers fell through unquoted, so completing a table like `` `order` `` inserted `order` instead of `` `order` `` — a MySQL reserved word — producing a real SQL syntax error when the query ran.
- Adds `quoteCompletionApplyIdentifier`, used only at the SQL editor's completion-apply call sites, which backtick-quotes MySQL reserved-word identifiers — mirroring the existing backtick logic already used for `SELECT *` column expansion (`quoteSelectStarColumnIdentifier`) in the same file.
- The exported `quoteSqlIdentifier` keeps its original PostgreSQL-only behavior unchanged, since `dataGridConditionColumnOptions` (data grid condition editor) also depends on it and intentionally leaves MySQL/SQL Server/Oracle identifiers unquoted there, applying its own quoting at insertion time via a separate path (`dataGridConditionIdentifierQuote`). An early version of this fix that changed `quoteSqlIdentifier` directly broke that call site's locked-in test; this version only changes the SQL-editor completion call sites.

## Test plan

- [x] Reproduced against real MySQL 8.0 (Docker) + local `dbx-web` backend + `pnpm dev:web` frontend: created a table named `` `order` `` (MySQL reserved word), typed `select * from ord`, pressed Tab. Before the fix, completion inserted `order` unquoted and executing the query failed with `ERROR 1064 (42000)`. After the fix, completion inserts `` `order` `` and the query executes successfully.
- [x] Added `packages/app-tests/sqlCompletion.test.ts` cases mirroring the existing PostgreSQL reserved-identifier tests, asserting MySQL table and column completions are backtick-quoted when needed and left unquoted when safe.
- [x] Full suite: `pnpm run test` (9090 tests), `pnpm run typecheck`, `pnpm run lint` all pass.

Fixes #6224

### DIFF
--- a/apps/desktop/src/lib/sql/sqlCompletion.ts
+++ b/apps/desktop/src/lib/sql/sqlCompletion.ts
@@ -3091,6 +3091,18 @@ export function quoteSqlIdentifier(identifier: string, dialect?: "mysql" | "post
 
 const POSTGRES_IDENTIFIER_KEYWORDS = new Set(SQL_KEYWORDS.map((keyword) => keyword.toLowerCase()));
 
+// Unlike quoteSqlIdentifier (also used by the data grid condition editor, which
+// intentionally leaves MySQL identifiers unquoted and applies backticks itself
+// at insertion time), SQL editor completion apply/insertText needs the quoting
+// baked in up front, so MySQL reserved-word identifiers get backtick-quoted here.
+function quoteCompletionApplyIdentifier(identifier: string, dialect?: "mysql" | "postgres" | "sqlserver"): string {
+  if (dialect === "mysql") {
+    if (!requiresPostgresIdentifierQuote(identifier, POSTGRES_IDENTIFIER_KEYWORDS)) return identifier;
+    return `\`${identifier.replaceAll("`", "``")}\``;
+  }
+  return quoteSqlIdentifier(identifier, dialect);
+}
+
 function quoteSelectStarColumnIdentifier(identifier: string, dialect?: "mysql" | "postgres" | "sqlserver", databaseType?: DatabaseType): string {
   if (!requiresPostgresIdentifierQuote(identifier, POSTGRES_IDENTIFIER_KEYWORDS)) return identifier;
   if (databaseType) return quoteTableIdentifier(databaseType, identifier);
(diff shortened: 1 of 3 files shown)


t8y2 (commented):
Patched in `59b32529`.

Root cause:
- SQL completion reused the PostgreSQL/generic keyword checks for MySQL, so MySQL-only reserved identifiers such as `database`, `accessible`, and `use` could still be inserted without backticks.
- A routine's supplied `applyName` bypassed identifier quoting, including both current-database names and cross-database qualified names.

Fix:
- Reused the repository's existing MySQL-only reserved-word set for completion identifiers across schema, table, column, and JOIN insertion paths.
- Applied MySQL quoting to each unquoted part of routine `applyName`, while preserving already quoted spelling and existing behavior for other dialects.

Validation:
- `pnpm exec vitest run packages/app-tests/sqlCompletion.test.ts apps/desktop/src/lib/__tests__/table/tableSelectSql.spec.ts --reporter=dot` — 2 files, 244 tests passed.
- `pnpm typecheck` — passed.
- Focused `oxlint` — 0 warnings and 0 errors.
- Focused `oxfmt` verification for the source files and newly added test blocks — passed.
- No dependency manifests changed.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 1a10db4, will be released in the next version.

PR #6251, opened 2026-08-14
fix(sidebar): stop clearing search from reopening untouched groups

### LINKED ISSUE #6065: [Bug] 清空表名搜索框时，不用去刷新和打开其他菜单，比如视图、存储过程、序列等等
### 数据库类型和版本

DM8

### 问题描述

1. 清空搜索框
2. 自动加载并打开视图、存储过程、序列等等

### 支持信息

v0.5.70
桌面端
macOs26.4
aarch64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Summary

Clearing the sidebar search box was force-reloading and leaving expanded *every* object group that the search had touched while typing (views, procedures, sequences, ...) — including groups the user never expanded and that had zero matches.

Root cause: when a search matches something inside a collapsed group, the sidebar temporarily force-expands that group to check for matches. On clearing the search, the code unconditionally restored/reloaded every group it had touched during the search, regardless of whether it was actually open beforehand.

Fix: track which groups search auto-expanded purely on the user's behalf (`searchAutoExpandedNodeIds`), and collapse those back on clear without reloading them. Groups the user had genuinely expanded before searching still get restored/refreshed as before.

Fixes #6065

## Test plan

- Real repro: MySQL 8.0 container with a test DB (2 tables, 2 views, 2 procedures). Manually expanded only "Views", searched "widget" (matches across Tables/Views/Procedures), then cleared the search box.
  - Before fix: Tables, Views, Procedures, Functions, and Triggers all ended up expanded after clearing — including Functions/Triggers, which had zero matches and were never touched.
  - After fix: only "Views" (expanded before the search) remains expanded; the rest collapse back with no extra reload.
- `pnpm typecheck` — clean
- `pnpm test` — 950/950 test files, 9100/9100 tests passed


t8y2 (commented):
Maintainer patch applied at 7f5703188b: separated remote-filtered groups from search-owned expansions so clearing or changing search restores group state without losing ancestor matches. Verified 16 focused frontend tests, vue-tsc, formatting, and diff checks.

t8y2 (approved):
Maintainer validation completed for exact head 7f5703188b.

t8y2:
Thanks for the contribution! Merged in b9e63dc57e, will be released in the next version.

PR #6254, opened 2026-08-14
fix(elasticsearch): format Kibana-console-style REST requests

### LINKED ISSUE #6230: [Bug] ElasticSearch查询语句格式化操作无效
### 数据库类型和版本

ElasticSearch 8

### 问题描述

ElasticSearch查询语句格式化操作无效

Image

### 支持信息

DBX版本：0.5.82
桌面端
macOS 26.6

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

Fixes #6230.

The query editor's Format toolbar action did nothing for Elasticsearch/Meilisearch queries written in Kibana-console style (`METHOD /path {json body}`, e.g. `GET /master_base/_search {"query":{"bool":{"must":[...]}}}`).

Root cause: `QueryEditor.vue`'s `formatCurrentSql()` dispatches by database type, but had no case for this REST-request syntax. It fell through to `detectAndFormatStructured` (which only recognizes text starting with `{`/`[`/XML markers) and then to the generic SQL formatter, which threw a parse error on the non-SQL `GET ...` text. That error is caught by an existing guard (`formatSqlForEditing`, added to avoid corrupting incomplete SQL while typing) that silently returns the original text unchanged — so clicking Format produced no visible error and no change.

## Solution

Added `apps/desktop/src/lib/elasticsearch/elasticsearchFormatter.ts`:
- Reuses the existing `elasticsearchRestRequestRanges` statement splitter (already used for execution/danger-checks) to detect one or more Kibana-console-style requests in the document, handling multi-request documents and preserving separators/comments between them.
- For each request, normalizes the `METHOD /path` onto its own line and pretty-prints the JSON body via the existing `formatJsonSource` helper (same one used for plain-JSON auto-formatting).
- Falls back cleanly: non-REST-request text (e.g. bare JSON, or non-ES database types) is untouched and continues through the existing JSON/XML/SQL detection path unchanged.

Wired into `QueryEditor.vue`'s `formatCurrentSql()` ahead of the existing structured/SQL fallback. Zero behavior change for every other database type.

## Testing

- Added 7 unit tests in `apps/desktop/src/lib/__tests__/elasticsearch/elasticsearchFormatter.spec.ts` covering: single-line request with JSON body, request with no body, multiple requests separated by a blank line, invalid JSON body (falls back to "unsupported" toast instead of corrupting text), bare JSON without a method line, non-Elasticsearch database types, and Meilisearch (which shares the same REST syntax).
- Full suite: `pnpm test` — 951 files / 9107 tests passing.
- `pnpm typecheck` clean.
- Manually reproduced and verified against a real Elasticsearch 8.15 instance (Docker) via a real browser: before the fix, clicking Format on the query from the issue's screenshot left it unchanged; after the fix, it pretty-prints into the expected multi-line JSON. Also confirmed the formatted query still executes correctly and returns the expected document (fix doesn't corrupt the request).

## Breaking changes

None — the new formatting path is only reached for Elasticsearch/Meilisearch text that actually matches the `METHOD /path` request syntax; everything else is untouched.

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -27,6 +27,7 @@ import { buildSqlInConditionFromPasteSource, insertTextForSqlInCondition } from 
 import { resolveSqlSingleQuoteKeyAction } from "@/lib/sql/sqlQuoteCaret";
 import { convertSqlSelectionCase, type SqlSelectionCaseMode } from "@/lib/sql/sqlSelectionCase";
 import { formatMongoShellText } from "@/lib/mongo/mongoFormatter";
+import { detectAndFormatElasticsearchRequests } from "@/lib/elasticsearch/elasticsearchFormatter";
 import { useConnectionStore, COMPLETION_METADATA_CONCURRENCY } from "@/stores/connectionStore";
 import { useSettingsStore } from "@/stores/settingsStore";
 import { useTheme } from "@/composables/useTheme";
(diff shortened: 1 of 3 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 185ff68d, will be released in the next version.

PR #6280, opened 2026-08-14
fix(xugu): keep package member groups scoped

mapan0424:
## Summary

This change fixes two Xugu package-tree issues:

1. Expanding a package procedure/function group could replace the package members with every procedure or function from the containing schema.
2. After navigating between schema-level routines and package members, a package function group could show a non-zero count but fail to expand.

## Root cause

Xugu package members are represented with the existing `group-procedures` and `group-functions` tree node types. Those same node types are also used for schema-level object groups. The sidebar toggle path therefore treated a package-owned group as a database metadata group and invoked the generic object loader. That loader queried the whole schema and overwrote the package-scoped children. The resulting node state also made subsequent package-group expansion unreliable.

## Implementation

- Detect Xugu package-owned procedure/function groups using their `parentType: "package"` metadata.
- Toggle their already-loaded package members locally instead of routing them through the schema-level object-group loader.
- Keep the condition explicitly Xugu-specific so Oracle and other database providers retain their existing behavior.
- Add a regression test that expands and collapses a package function group twice while asserting that no schema-routine metadata load is issued.

## User impact

Package groups now remain scoped to the package specification. Standalone procedures/functions from the same schema are no longer mixed into a package, and package procedure/function groups can be expanded, collapsed, and expanded again consistently.

## Validation

- `go test -count=1 ./...` in `agents/drivers/xugu` — passed.
- `pnpm exec vitest run apps/desktop/src/components/sidebar apps/desktop/src/lib/__tests__/sidebar apps/desktop/src/stores/__tests__/connectionStore.completion.spec.ts` — 43 test files and 258 tests passed.
- `pnpm exec oxfmt --check apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue apps/desktop/src/components/sidebar/__tests__/SidebarTreeRuntimeHost.expansion.spec.ts` — passed.

The workspace-wide Vue typecheck still reports the pre-existing missing `@dbx-app/mongo-shell` workspace package; the changed files introduce no type errors.

## Scope

The change is limited to the desktop sidebar routing and its regression test. No Xugu SQL, metadata query, driver protocol, or non-Xugu database implementation was changed.

### DIFF
--- a/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
+++ b/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
@@ -691,7 +691,9 @@ async function toggle(requestId = beginNavigationRequest()) {
   // schema and replace the package's member list with unrelated objects.
   if (currentDatabaseType() === "xugu" && (node.type === "group-procedures" || node.type === "group-functions") && node.parentType === "package") {
     node.isExpanded = !node.isExpanded;
-    if (wasExpanded && !connectionStore.sidebarSearchQuery) connectionStore.releaseCollapsedTreeNodeChildren(node.id);
+    // Package member groups have no group-level reload path: their children
+    // are hydrated together with the owning package. Releasing a large group
+    // here would leave it empty on the next local-only expand.
     emitNodeToggled(node, wasExpanded);
     return;
   }
--- a/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
+++ b/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
@@ -684,6 +684,18 @@ async function toggle(requestId = beginNavigationRequest()) {
     return;
   }
 
+  // Xugu package members are represented with the same visual group types as
+  // schema-level procedures/functions, but their children are already loaded
+  // from the owning package specification.  Keep this scoped group local:
+  // routing it through loadSidebarObjectGroup would query every routine in the
+  // schema and replace the package's member list with unrelated objects.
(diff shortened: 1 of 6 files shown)


mapan0424:
Follow-up fix for the package-member release edge case:

- Xugu package procedure/function groups are hydrated together with their owning package and do not have a group-level reload path.
- The previous local toggle branch could release a large group (400+ descendants), clear its children, and then expand it locally without reloading, leaving an empty expanded node.
- The local Xugu package-member path now preserves its loaded children when collapsing. This keeps package scope intact and avoids routing the group through the schema-wide routine loader.
- Added regression coverage with 401 package function members: collapse, verify children remain, expand again, and verify all members are still present.

Validation performed:
- Sidebar test suite: 10 files, 37 tests passed.
- Targeted package expansion regression: passed.
- Oxfmt check: passed.
- Full TypeScript check is currently blocked by the repository's pre-existing missing `@dbx-app/mongo-shell` module errors, unrelated to this change.

t8y2 (commented):
Maintainer patch applied on reviewed head 8cdf59b41cc192150c6c14024b462c1629547728, producing 09da8b02857e0b46eed451a1a3154a75082e03b8.

- Package member groups now route refresh/reload back to the owning Xugu package instead of the generic schema object-group loader.
- Large collapsed package groups continue retaining their already-loaded children.
- Verified: 3 focused Vitest files passed (65/65), pnpm typecheck passed, and git diff --check passed.

The patch was pushed by fast-forward without rewriting contributor history.

t8y2:
Thanks for the contribution! Merged in 5060782, will be released in the next version.

PR #6281, opened 2026-08-14
fix(xugu): route type expansion through member metadata

mapan0424:
## Summary

Fixes Xugu object type expansion when the sidebar restores or loads a tree node through the generic metadata path.

## User-visible problem

Expanding a Xugu object type could issue a table-column lookup against the type name and return a vendor error indicating that the table or view did not exist. The type itself was valid; only the metadata route was incorrect.

## Root cause

The direct sidebar click path already recognized Xugu object types and loaded their attributes and methods through the scoped completion endpoint. However, the shared `loadTreeNodeChildren` path used by tree restoration and refresh handled every `type` node with the generic custom-type loader. That path treats the object name as a relation and can eventually request table columns, which is invalid for Xugu object types.

## Changes

- Route Xugu expandable type nodes through `loadXuguTypeMembers` from the shared tree loader.
- Keep the existing generic custom-type path unchanged for PostgreSQL-family and other supported databases.
- Add a regression test proving that generic tree restoration calls the Xugu member endpoint and never calls the generic custom-type details loader.

## Compatibility and scope

This is a narrow frontend metadata-routing fix. It changes behavior only when the connection is Xugu and the node is explicitly marked as an expandable object type. Other database types and existing custom-type handling are unchanged.

## Validation

- Xugu live metadata validation confirmed object type attributes can be read through the member endpoint and the session remains usable.
- Vitest: 59 tests passed across the connection metadata-loading and Xugu type-member suites.
- Formatting check passed with oxfmt.
- `git diff --check` passed.

### DIFF
--- a/apps/desktop/src/stores/__tests__/connectionStore.metadataLoading.spec.ts
+++ b/apps/desktop/src/stores/__tests__/connectionStore.metadataLoading.spec.ts
@@ -289,6 +289,47 @@ describe("connectionStore metadata loading", () => {
     expect(typeNode.children?.every((child) => child.parentName === "ADDRESS_T")).toBe(true);
   }, 15000);
 
+  it("routes generic tree restoration for Xugu types through the member loader", async () => {
+    const completionAssistantSearch = vi.fn(async (request: { object_kinds?: string[] }) => ({
+      candidates: request.object_kinds?.includes("column") ? [{ name: "street", kind: "column", data_type: "VARCHAR(120)" }] : [],
+      incomplete: false,
+      fallback_used: false,
+    }));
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch applied on reviewed head 50a3923b0e12a3c7752780a3c23930daaa0827cc, producing f34cfeaf02b120a83d5a78c0581d3e37c8173c6e.

- Forced tree restoration now reloads expanded Xugu type members instead of entering the user-click toggle/collapse branch.
- Normal click behavior remains unchanged.
- Verified: 2 focused Vitest files passed (59/59), pnpm typecheck passed, and git diff --check passed.

The patch was pushed by fast-forward without rewriting contributor history.

t8y2:
Thanks for the contribution! Merged in 90af026, will be released in the next version.

PR #6282, opened 2026-08-14
fix(sqlserver): focus real data result in multi-target execution too

### LINKED ISSUE #6189: 在执行诸如 spwholock 等返回多个结果集的 SQL 语句时，DBX 默认会将展示卡片自动切换并聚焦至最后一个结果集（例如非核心的“DBCC 执行完毕”消息）
### 数据库类型和版本

SqlServer 10.50.2500.0

### 支持信息

DBX 版本: v0.5.83
运行环境: 桌面端
操作系统: Windows 10.0.22621.4317
系统架构: x86_64

### 当前痛点

#5566 之前在这个issue里边，作者说已经在v0.5.83解决，但是更新之后，在执行诸如 spwholock 等返回多个结果集的 SQL 语句时，DBX 默认会将展示卡片自动切换并聚焦至最后一个结果集（例如非核心的“DBCC 执行完毕”消息）问题依然存在

### 期望方案

希望可以把结果集聚焦在第一个

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须做

q396921921:
## Problem

When a SQL Server batch ends in a `PRINT`/`DBCC`-style message with no rows of its own, the backend synthesizes a pseudo result with a single `Message` column (`crates/dbx-core/src/db/sqlserver.rs:519-567`). The frontend's generic "first result with columns" pick can land on that message instead of real data.

`doExecute()` (the normal single-connection SQL editor path) already corrects this — added in `be3336c1e` for #5967 — by redirecting focus to the first real data result whenever the message result ends up selected. However, `executeTargetSql()` (the "execute across multiple database targets" path used by the multi-DB execute dialog) never got the same correction, so running a multi-result-set batch through that path still leaves the message result focused instead of the real data — exactly the symptom reported in #6189.

Separately, `queryStore.setActiveResultIndex()` only resolved tabs via `tabs.value.find(...)`, while every other execution-tab lookup in the store (`getExecutionTab`, `executeTabSql`) uses `findExecutionTab()`, which also checks the `multiDbExecutionWorkers` map used by the multi-target path. Without fixing this too, the correction above would silently no-op for worker tabs.

## Solution

- Extracted the message-focus correction into a shared `focusSqlServerDataResult()` helper in `useSqlExecution.ts`, now called from both `doExecute()` and `executeTargetSql()`.
- Widened `setActiveResultIndex()` to resolve tabs via `findExecutionTab()` instead of only `tabs.value`, matching the existing lookup pattern used elsewhere in the store.

## Testing

- New test `apps/desktop/src/composables/__tests__/useSqlExecutionMultiDbFocus.spec.ts`: exercises the real (unmocked) `executeTargetSql`/`doExecute` code against realistic SQL-Server result shapes. Confirmed the fix is real by reverting it locally and re-running — the "focuses the real data result" test fails on unfixed code (`expected 0 to be 1`) and passes after the fix.
- New test `packages/app-tests/sqlserverResultFocus6189.test.ts`: drives the real `queryStore` end-to-end to rule out a "stale `activeResultIndex` carries over between executions" theory (disproven — `preserveActiveResultIndex` is never passed by either execution path).
- Full suite: `pnpm test` → 964/965 files, 9222/9223 tests passed (the 1 failure, `DocumentBrowserFieldSearch.spec.ts`, is a pre-existing/unrelated MongoDB document-browser UI flake — reproduces in isolation on this branch with no relation to the files touched here). `pnpm typecheck` and `pnpm lint` are clean.
- Note: the originally-reported repro (v0.5.83, plain single-connection SQL editor) is already fixed on `main`/v0.5.84 by `be3336c1e` — the reporter just hadn't updated yet. This PR fixes the identical, still-live gap in the multi-target execution path.

Fixes #6189

### DIFF
--- a/apps/desktop/src/composables/__tests__/useSqlExecutionMultiDbFocus.spec.ts
+++ b/apps/desktop/src/composables/__tests__/useSqlExecutionMultiDbFocus.spec.ts
@@ -1,0 +1,288 @@
+/**
+ * Issue #6189 — SQL Server "focus jumps to the message result".
+ *
+ * Background:
+ *  - `crates/dbx-core/src/db/sqlserver.rs:523-542` synthesizes a pseudo result with a
+ *    single "Message" column and `server_message: true` for any batch segment that
+ *    produced only server messages (PRINT / "DBCC execution completed" / ...).
+ *  - `stores/queryStore.ts:4727` then picks the first result that HAS COLUMNS as the
+ *    active one. The synthesized message grid has a column, so a message result can win.
+ *  - `composables/useSqlExecution.ts` (`focusSqlServerDataResult`, shared by both
+ *    `doExecute` and `executeTargetSql`) corrects that: when the selected result is a
+ *    `server_message` one it re-focuses the first real data result. Before this fix,
+ *    `executeTargetSql` (the multi-database execute path) had no such correction.
+ */
+import { computed, ref } from "vue";
+import { createPinia, setActivePinia } from "pinia";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { useSqlExecution } from "../useSqlExecution";
+import { useHistoryStore } from "@/stores/historyStore";
+import { useQueryStore } from "@/stores/queryStore";
+import type { ConnectionConfig, QueryTab } from "@/types/database";
+
+vi.mock("vue-i18n", () => ({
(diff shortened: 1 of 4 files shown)


t8y2:
Thanks for the contribution! Merged in 98df97561, will be released in the next version.

PR #6284, opened 2026-08-14
fix(kafka): enable Adjust Partitions button for single-partition topics

### LINKED ISSUE #6208: [Bug] kafka默认分区没法修改
### 数据库类型和版本

kafka

### 问题描述

Image
1.kafka默认分区没法修改
2.希望增加一个topic复制
3.

Image
long的字节没法展示

### 支持信息

DBX 版本: v0.5.83
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Summary

Fixes t8y2/dbx#6208

Kafka topics with exactly 1 partition did not show the Adjust Partitions button in the Topics panel, because the partitioned flag was only set when partitions > 1.

Since every Kafka topic inherently has partitions, this flag should be true whenever the partition count is known.

## Changes

- crates/dbx-core/src/mq/adapters/kafka.rs: Changed partitioned condition from p > 1 to p > 0
- Added unit test kafka_topic_partitioned_flag_true_for_any_partition_count covering single-partition, multi-partition, and missing-partition cases

## Before

Single-partition Kafka topics: partitioned: false, Adjust Partitions button hidden

## After

Single-partition Kafka topics: partitioned: true, Adjust Partitions button visible and functional

### DIFF
--- a/crates/dbx-core/src/mq/adapters/kafka.rs
+++ b/crates/dbx-core/src/mq/adapters/kafka.rs
@@ -1220,5 +1220,4 @@ mod tests {
         let partitions = missing.get("partitions").and_then(|v| v.as_u64()).map(|v| v as u32);
         assert_eq!(partitions.map(|p| p > 0).unwrap_or(false), false);
     }
-
 }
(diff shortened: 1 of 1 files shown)


t8y2 (commented):
Maintainer patch applied in `fc276beae` (previous head `2ce77260f`).

The partition behavior was already correct. The patch replaces the three boolean `assert_eq!` calls in the new regression test with idiomatic boolean assertions so the PR passes the workspace clippy gate.

Validation: targeted Kafka test, `cargo fmt --check`, `cargo clippy -D warnings`, and `git diff --check` passed.

t8y2:
Thanks for the contribution! Merged in 9fbbf70d4, will be released in the next version.

PR #6289, opened 2026-08-14
fix(ui): use tabular-nums for live progress counters

difagume:
## Summary

Live progress counters (download %, rows imported/exported, elapsed seconds, count badges) jitter because the font uses proportional numerals — `1` is narrower than `0`, so the text reflows as digits change.

This PR adds the `tabular-nums` utility (`font-variant-numeric: tabular-nums`) to all live counters for fixed-width digits. The repo already uses it elsewhere (elapsed timers, grid numeric cells, dashboard metrics); this makes the remaining counters consistent.

## Changes

All in `apps/desktop/src/components/`:

| File | Counter fixed |
| --- | --- |
| `layout/UpdateDialog.vue` | Download % |
| `import/TableImportDialog.vue` | Rows imported + % |
| `export/DatabaseExportDialog.vue` | Rows exported |
| `connection/ConnectionDialog.vue` | Agent install bytes + % |
| `editor/MultiDbExecuteDialog.vue` | Completed/total + count badges |
| `lineage/FieldLineageDialog.vue` | Loading done/total |
| `search/DatabaseSearchDialog.vue` | Table loading + result count |
| `grid/DataGrid.vue` | Elapsed seconds |
| `sql-file/SqlFileExecutionDialog.vue` | File progress + stat counters |

## Video

Side-by-side comparison, same counters cycling 0→100%: **left = without fix** (numbers jitter), **right = with fix** (stable). Same values on both sides, so the only difference is digit width.

https://github.com/user-attachments/assets/e98a4c81-d6b9-45fe-b45d-07bed7bcaaf5

## Testing

UI-only change (one utility class per element). Verified on a temporary dev preview page (`/preview.html`) showing all 9 counters with and without `tabular-nums`; the video above is from that page. No logic changed, no tests needed.

### DIFF
--- a/apps/desktop/src/components/connection/ConnectionDialog.vue
+++ b/apps/desktop/src/components/connection/ConnectionDialog.vue
@@ -8304,7 +8304,7 @@ function openExternalUrl(url: string) {
           <div class="flex items-center justify-between gap-3">
             <div class="min-w-0">
               <div class="truncate text-sm font-medium">{{ agentInstallLabel || agentInstallDriverKey }}</div>
-              <div class="mt-1 text-xs text-muted-foreground">{{ agentInstallProgressLabel }}</div>
+              <div class="mt-1 text-xs text-muted-foreground tabular-nums">{{ agentInstallProgressLabel }}</div>
             </div>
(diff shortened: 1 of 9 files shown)


t8y2:
Thanks for the contribution! Merged in 6a2000def, will be released in the next version.

PR #6304, opened 2026-08-15
fix(grid): re-measure column header widths once the data grid font finishes loading

### LINKED ISSUE #6271: [Bug] 查询结果集的列宽展示和配置有偏差
### 数据库类型和版本

MySQL 8.0

### 问题描述

1.复现步骤：
1、视图选项中默认列宽为“标准”
2、任意sql，查询结果集后，列宽展示效果不是按“标准”模式渲染的
3、在有查询结果后再变更列宽后，展示变为正常，比如做一次把列宽改为宽松再改回标准的动作

2.期望结果：
直接查询渲染的效果和配置一致

3.实际结果：
配置和渲染不一致

### 支持信息

DBX 版本：V0.5.84
运行环境：桌面端
操作系统：macOS 26.5.2
系统架构：aarch64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Problem

The data grid's custom "Geist Variable Tabular" font is loaded via `@font-face` with `font-display: swap`. Column header widths are measured with `canvas.measureText()` as soon as a query result mounts (`DataGrid.vue` calls `initColumnWidths()` synchronously during component setup, before the browser has necessarily finished loading that font).

Nothing in the codebase re-triggered this measurement once the font actually finished loading — the only things that recompute column widths are explicit density / font-size / font-family changes. So a result rendered while the font was still loading kept whatever fallback-font-derived widths it started with, and stayed that way until the user manually toggled the column width density setting (Standard → Comfortable → Standard), which happens to force a full recompute.

This matches the reported symptom exactly: default density is "Standard" the whole time, the very first query's column widths don't look right, and manually re-selecting Standard "fixes" it.

## Fix

Listen for the browser's `document.fonts` `loadingdone` event and fold a tick counter into the existing header measurement key (`columnHeaderMeasurementKey`). This reuses the already-existing, already-tested forced-recompute path (`useDataGridColumnResize`'s watcher on `[density, compactColumnHeaderActions, headerMeasurementKey]`) so it now also fires automatically once the grid font finishes loading — no manual toggle needed.

## Testing

- `pnpm typecheck` — clean
- `pnpm lint` — clean (no new warnings)
- `pnpm vitest run apps/desktop/src` — 675 files / 6319 tests passed
- Instrumented Playwright evidence (headless Chromium, real MySQL 8.0 container, a table with a long column name so header-text width dominates over cell-value width): reading the grid's `--col-w-N` CSS vars showed `351px` when the custom font resolves normally vs `344px` when the same font is permanently unavailable, confirming the header measurement is genuinely coupled to font-resolution state. I was not able to naturally force the exact mid-flight timing window in this headless macOS/Linux sandbox (tried natural timing and artificial network delays of up to 8s from page load — the fallback/real-font gap only showed up when the font was made permanently unavailable), likely because headless Chromium's font substitution differs from the reporter's Windows 10 WebView2 environment.

Fixes #6271

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -2097,8 +2097,19 @@ const columnWidthDensity = computed(() => settingsStore.editorSettings.columnWid
 const tableFontFamily = computed(() => settingsStore.editorSettings.tableFontFamily);
 const columnWidthCacheKey = computed(() => props.cacheKey?.trim() || undefined);
 const columnStructureSignature = computed(() => createDataGridColumnStructureSignature(props.result.columns, props.result.column_types));
-const columnHeaderMeasurementKey = computed(() => [tableFontSize.value, tableFontFamily.value]);
+// Bumped once a still-loading @font-face (e.g. the custom data grid font) finishes downloading, so
+// widths measured against a temporary fallback font before it loaded get re-measured with the real one.
+const dataGridFontsLoadedTick = ref(0);
+const columnHeaderMeasurementKey = computed(() => [tableFontSize.value, tableFontFamily.value, dataGridFontsLoadedTick.value]);
 let columnHeaderMeasureContext: CanvasRenderingContext2D | null | undefined;
+
+if (typeof document !== "undefined" && document.fonts) {
+  const onDataGridFontsLoadingDone = () => {
+    dataGridFontsLoadedTick.value++;
+  };
+  document.fonts.addEventListener("loadingdone", onDataGridFontsLoadingDone);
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch applied in `98ea4eac5` (previous head `aa7f484e6`).

Root cause: every mounted DataGrid subscribed to the global `document.fonts.loadingdone` event, so an unrelated font load caused all grids to invalidate their column measurements.

Patch: wait only for the configured grid header font through `document.fonts.load`, ignore stale font-setting requests, and update the architecture regression coverage.

Validation: focused app test (7/7), frontend typecheck, formatting, and `git diff --check` passed.

t8y2:
Thanks for the contribution! Merged in 8bf4485a7, will be released in the next version.

PR #6305, opened 2026-08-15
fix(grid): enable copy-as-SQL when right-clicking a selected cell

### LINKED ISSUE #6272: [Bug] 数据网格复制有问题
### 数据库类型和版本

通用

### 问题描述

1. 复制为select不知道为啥一直是禁用，不管怎么选择单元格，如下图

Image

Image
2.  复制为insert或者update有的时候是禁用有的时候不是禁用，如下图

Image

Image
个人测试好像是直接右键一个单元格就不是禁用的，不过复制出来的东西和右键行是一样的，如果先左键选中了单元格在这个单元格再右键就是禁用的了
3.  第二步两种选择单元格的方式，复制出来的where语句完全不一样

### 支持信息

DBX 版本: v0.5.84
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

个人认为直接右键未选择中单元格的时候应该是和右键行一样的处理了
我认为都是右键单元格 不应该出来这个差别

q396921921:
Fixes #6272

## Problem

The data grid right-click copy submenu has two bugs:

1. **"Copy as SELECT" is always disabled** regardless of how cells are selected. The `sourceName` check in `canCopyWithExtractor` lacked a `displayName` fallback, so when `sourceColumns` is undefined (common for ad-hoc queries) the check always fails. The INSERT path already used `sourceName ?? displayName` but SELECT did not.

2. **"Copy as INSERT/UPDATE" is inconsistently enabled** — works when right-clicking an unselected cell but disabled when left-selecting first then right-clicking. `buildRequest` only entered the full-row context branch when `contextSelectionIsSynthetic=true`. With a prior cell selection, `contextSelectionIsSynthetic=false`, so the matrix branch used only the selected cell(s) — selecting a PK cell alone caused the "has writable column" check to fail.

## Fix

- **Fix A**: `canCopyWithExtractor` sql-select path — `sourceName?.trim()` → `(sourceName ?? displayName)?.trim()` to match the INSERT fallback.
- **Fix B**: Extended `buildRequest` context-cell handling so INSERT/UPDATE extractors also use the full row when there is a non-synthetic right-click context on a single cell, matching the existing synthetic behavior. `sql-select` keeps using only the clicked cell via the matrix branch.

## Tests

- Updated 1 existing test to match the corrected behavior.
- Added 2 regression tests:
  - `enables SELECT copy when sourceColumns is undefined by falling back to display names`
  - `enables INSERT/UPDATE when right-clicking a previously selected primary-key cell`
- All 3 new/updated tests fail on unmodified code and pass with the fix.
- Full suite: 965 files / 9225 tests pass, `vue-tsc` clean.

### DIFF
--- a/apps/desktop/src/composables/__tests__/useDataGridExport.spec.ts
+++ b/apps/desktop/src/composables/__tests__/useDataGridExport.spec.ts
@@ -197,7 +197,7 @@ describe("useDataGridExport prepared row statements", () => {
     expect(state.canCopyRow.value).toBe(true);
   });
 
-  it("builds SQL UPDATE from only selected writable columns while retaining a hidden primary key", async () => {
+  it("builds SQL UPDATE using the full row when right-clicking a selected cell", async () => {
     const item = row([7, "Ada", true]);
     const matrix: CellSelectionMatrix = {
       rowIndexes: [0],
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in cd6f99fbc, will be released in the next version.

PR #6308, opened 2026-08-15
fix(transfer): widen gap between object list and dialog scrollbars

### LINKED ISSUE #6246: [Bug] 数据传输窗口滚动条重叠
### 数据库类型和版本

MySQL 8.0

### 问题描述

1. 复现步骤：打开数据传输，选择数据库
2. 期望：表滚动条不被数据传输窗口滚动条遮挡
3. 实际情况：表的滚动条选中窗口比较窄，如果稍微偏一点就会触发窗口的滚动条
4. 截图，下方补充了视频
Image

### 支持信息

DBX 版本: v0.5.83
运行环境: 桌面端
操作系统: Fedora Linux 44 (KDE Plasma Desktop Edition) 44
系统架构: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

#4409 关联issue，如果外部窗口和内部表的滚动条都保留的话，要保证不会互相遮挡重叠

https://github.com/user-attachments/assets/b241ff79-3416-4bd4-9d47-69fe8a8b58c6

q396921921:
## Problem

In the Data Transfer dialog, the object-selection tree ("Tables"/"Views"/etc. list) renders its own independently-scrollable box (`max-h-[240px] overflow-y-auto`) nested inside the dialog's own outer scrollable container (`overflow-y-auto pl-4 pr-1`). The two containers' right edges sat only ~4px apart, so on any platform with a real (non-overlay) scrollbar, the two scrollbars' hit-areas overlap or nearly touch — a slight mouse offset while trying to scroll the table list ends up dragging the dialog's own scrollbar instead.

## Root cause

Measured on the live DOM (via `getBoundingClientRect`) with a source database selected (30 tables, enough to force the inner box to scroll):
- Outer dialog scroll container right edge: `x=1243`, `scrollHeight=720 > clientHeight=433`
- Inner object-list box right edge: `x=1239`, `scrollHeight=1192 > clientHeight=238`
- Gap: **4px** — smaller than any realistic scrollbar width (8–18px depending on OS/theme), so the two scrollbar hit-areas necessarily collide.

## Fix

Added `mr-4` to the inner object-list scroll box in `ObjectSelectionTree.vue`, widening the gap from 4px to 20px — comfortably clearing common scrollbar widths (including the ~17px Windows classic scrollbar) so the two hit-areas no longer overlap.

## Testing

- Reproduced against a real MySQL 8.0 database (30 tables) through the actual running app (real `dbx-web` backend + `pnpm dev:web` frontend), measuring the two scroll containers' geometry before and after the fix.
- Could not capture a literal before/after screenshot of overlapping scrollbar thumbs: headless Chromium (this sandbox, macOS) always renders 0-width overlay scrollbars regardless of `--disable-features=OverlayScrollbar` / forced `::-webkit-scrollbar` CSS, so a classic OS scrollbar never paints pixels here. The DOM geometry measurement is the objective substitute.
- `pnpm typecheck` clean, `pnpm lint` (oxlint) clean, full test suite: 6319/6319 passed.
- Rebased cleanly onto latest `upstream/main`.

Fixes #6246

### DIFF
--- a/apps/desktop/src/components/transfer/ObjectSelectionTree.vue
+++ b/apps/desktop/src/components/transfer/ObjectSelectionTree.vue
@@ -202,8 +202,10 @@ function groupSelectionState(kind: TransferObjectKind, items: string[]): "none" 
       {{ t("transfer.noObjects") }}
     </div>
 
-    <!-- Independent scroll area for tree groups -->
-    <div v-else class="flex-1 min-h-0 max-h-[240px] flex flex-col gap-1.5 overflow-y-auto rounded-md border border-border/80 bg-muted/10 p-1.5 scrollbar-thin scrollbar-thumb-muted-foreground/20">
+    <!-- Independent scroll area for tree groups. mr-4 keeps this box's own
+         scrollbar clear of the dialog's outer scrollbar (only pr-1 away in
+         DataTransferDialog.vue) so the two hit areas don't overlap. -->
+    <div v-else class="flex-1 min-h-0 max-h-[240px] flex flex-col gap-1.5 overflow-y-auto rounded-md border border-border/80 bg-muted/10 p-1.5 mr-4 scrollbar-thin scrollbar-thumb-muted-foreground/20">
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks for the contribution! Merged in a407c77a1, will be released in the next version.

PR #6309, opened 2026-08-15
fix(editor): honor the configured accept-completion shortcut on Tab

### LINKED ISSUE #6236: [Feature] 在SQL编辑器中写SQL代码，会弹出代码补全框，但是我按TAB键或者ENTER键都会自动补全，有时候我不想要补上去，还要手动去删除
### 数据库类型和版本

通用

### 支持信息

DBX版本：v0.5.83
运行环境：桌面端
系统：Windows11 专业版

### 当前痛点

在SQL编辑器中写SQL代码，会弹出代码补全框，但是我按TAB键或者ENTER键都会自动补全，有时候我不想要补上去，还要手动去删除。在配置中设置了ENTER快捷键接受补全，但是我按TAB键还是会生效。

### 期望方案

可不可以统一一下，或者有个配置页让用户自己选择用什么快捷键实现代码补全

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须做

### 补充信息

_No response_

q396921921:
## Problem

The completion popup in the SQL editor always accepted on Tab, even when the user remapped the "Accept completion" shortcut in Settings to something else (e.g. Enter). That made the shortcut setting effectively decorative for the Tab key: users had no way to stop Tab from silently accepting a completion they didn't want.

## Root cause

`QueryEditor.vue` wires Tab up in two independent places:

1. A shortcut-aware binding driven by `settings.shortcuts.acceptCompletion` — this one correctly respects the user's configured key.
2. A separate `Prec.highest` **hardcoded** `{ key: "Tab", run: handleTab }` binding, used to handle plain indentation and snippet-field navigation when Tab is pressed.

`handleTab` unconditionally called `acceptCompletionOrNextSnippetField`, which accepts any currently-active completion popup — regardless of what the user configured as their accept-completion shortcut. Since this binding has the highest precedence, it silently overrode the user's setting.

## Fix

`handleTab` now only accepts an active completion via Tab when the configured `acceptCompletion` shortcut actually resolves to Tab (`tabKeyAcceptsCompletion()`). When it doesn't, Tab falls through to `handleTabWithoutAcceptingCompletion` (keeps snippet-field navigation, skips accepting completion) and then normal indentation, leaving the completion popup open so the user can accept it with their configured key instead.

## Testing

Reproduced against real production code with vitest (`QueryEditor.vue`'s `handleTab` extracted and exercised directly, per this repo's existing test convention for this file):

- Added a regression test that configures the accept-completion shortcut to `Enter` and presses Tab while a completion is active. Against the pre-fix code this test genuinely fails (`acceptCompletion` gets called once despite the config). Against the fix it passes.
- Added a second regression test confirming Tab still accepts completion when the shortcut is left at its default (`Tab`) — no behavior change for the default/unconfigured case.
- `pnpm exec vitest run`: 967 files / 9252 tests passing.
- `pnpm typecheck`: clean.
- `pnpm lint`: clean (no new warnings; pre-existing warnings are all in unrelated files).

Fixes #6236

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -748,7 +748,25 @@ function editorIndentUnit(): string {
 }
 
 function handleTab(view: EditorViewType): boolean {
-  return acceptCompletionOrNextSnippetField(view) || performNormalTab(view);
+  if (tabKeyAcceptsCompletion()) {
+    return acceptCompletionOrNextSnippetField(view) || performNormalTab(view);
+  }
+  return handleTabWithoutAcceptingCompletion(view) || performNormalTab(view);
+}
+
+// The Tab key is always wired up for indentation and snippet-field navigation,
+// but it must only accept an open completion popup when the user's configured
+// "accept completion" shortcut is actually Tab — otherwise a user who remapped
+// that shortcut (e.g. to Enter) would find Tab silently accepting completions
+// anyway, ignoring their setting (dbx#6236).
+function tabKeyAcceptsCompletion(): boolean {
+  const shortcuts = normalizeShortcutSettings(settingsStore.editorSettings.shortcuts);
+  return shortcutToCodeMirrorKey(shortcuts.acceptCompletion) === "Tab";
+}
+
+function handleTabWithoutAcceptingCompletion(view: EditorViewType): boolean {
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in 057b24b5e, will be released in the next version.

PR #6311, opened 2026-08-15
fix(nacos): force full syntax tree before first paint in config editor

### LINKED ISSUE #6302: [Bug] Nacos YAML 配置注释样式显示异常
### 来源
QQ 群反馈 · **反馈人**：以年

### AI 摘要
在 Nacos 的 YAML 配置编辑器中，使用 # 注释时，第一段 # 的样式显示异常。保存时显示正常，但保存成功后重新进入页面显示不正常。

### 原始描述
nacos yaml配置，#注释配置时，第一段#样式显示会有问题，保存时显示正常，保存成功后重新进去就不正常了

q396921921:
## Problem

In the Nacos YAML config editor, when a config's content starts with two or more consecutive `#` comment lines, reopening the config after a fresh page load can render the first comment line in plain/default styling (no comment color or italics) while the following comment lines render correctly. Editing/saving looks fine; the issue only shows up right after a fresh (re)load, and any subsequent edit self-heals it.

Fixes #6302

## Root cause

An isolated CodeMirror test confirms the Lezer syntax tree for the same YAML content is byte-identical whether built via `EditorState.create()` with the full document at once (mirrors the reload path) or via incremental typed-in updates (mirrors live editing) — so this is not a grammar/parsing bug.

`mountConfigEditor()` in `NacosAdminConsole.vue` builds `EditorState.create({...})` and immediately constructs a new `EditorView` from it, without forcing the syntax tree for the freshly loaded content to be fully computed first. This leaves a narrow window (worse under system load) where the initial paint can happen before syntax highlighting decorations for the start of the document have caught up.

## Fix

Call `ensureSyntaxTree(state, content.length, 500)` — the same helper already used the same way in `QueryEditor.vue` — right after building the `EditorState` and before constructing the `EditorView`, forcing a synchronous full parse so highlighting is complete before the first paint.

## Testing

- Reproduced in a real environment: `dbx-web` dev backend + a real Nacos server, created a YAML config with 3 leading `#` comment lines via the app's own Nacos config editor, saved, reloaded the page, reopened the config — the first comment line rendered in plain/default color while the following two rendered correctly (confirmed both visually and via `getComputedStyle` on the live DOM, not just a screenshot artifact).
- This is a genuine but low-probability mount-time race — after applying the fix I was not able to re-trigger the original race window in ~25 further attempts (including CPU-throttled and frame-by-frame DOM-sampling runs) to produce a clean isolated "after" capture, so this PR relies on the confirmed root cause plus full regression coverage rather than a repeated live repro.
- `vue-tsc --noEmit`: clean
- Full test suite: 9250/9250 tests passing across 967 files (including all existing Nacos-related specs)
- `oxlint` on the changed file: no new warnings (one pre-existing, unrelated warning at a different line)

### DIFF
--- a/apps/desktop/src/components/nacos/NacosAdminConsole.vue
+++ b/apps/desktop/src/components/nacos/NacosAdminConsole.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { computed, nextTick, onBeforeUnmount, onMounted, ref, shallowRef, useId, watch } from "vue";
 import { Compartment, type Extension } from "@codemirror/state";
-import { StreamLanguage } from "@codemirror/language";
+import { StreamLanguage, ensureSyntaxTree } from "@codemirror/language";
 import type { EditorView } from "@codemirror/view";
 import { Archive, ArrowLeftRight, CheckCircle2, ChevronDown, ChevronLeft, ChevronRight, Clipboard, Columns3, Download, FileClock, FileInput, FileText, Loader2, Network, Plus, RefreshCw, Save, Search, Send, Server, Trash2, X } from "@lucide/vue";
 import { Button } from "@/components/ui/button";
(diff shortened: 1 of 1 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in d782667b2, will be released in the next version.

PR #6312, opened 2026-08-15
fix(schema-diff): keep expression index key parts unquoted for postgres-family dialects

### LINKED ISSUE #6295: [Bug] 表结构同步错误
### 数据库类型和版本

highgo 4.5

### 问题描述

所有变更已回滚。: Agent RPC error (-1): com.highgo.jdbc.util.PSQLException: ERROR: column "COALESCE(height, '-1'::integer::double precision)" does not exist DBX_AGENT_ERROR_DATA:{"contractVersion":1,"category":"sql","retryable":false,"sessionDisposition":"keep","stage":"execute","operationOutcome":"unknown","agentSessionId":"89d8d6f648764d65aab111e77733aaf6","sqlState":"42703","vendorCode":0,"exceptionClass":"com.highgo.jdbc.util.PSQLException"}

错误原因,生成的ddl index有误
CREATE UNIQUE INDEX "uq_tankong_sta_type_time" ON "public"."tankong_data" USING btree (
    "sta_id", 
    "data_type", 
    "data_time", 
    "COALESCE(height, '-1'::integer::double precision)"   -- ❌ 双引号内被当作列名
);

解决方案
去掉双引号，使用裸表达式
CREATE UNIQUE INDEX "uq_tankong_sta_type_time" 
ON "public"."tankong_data" USING btree (
    "sta_id", 
    "data_type", 
    "data_time", 
    COALESCE(height, -1.0)   -- ✅ 无引号，视为表达式
);

### 支持信息

DBX 版本: v0.5.84
运行环境: 桌面端
操作系统: Manjaro Linux
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

Reported in #6295: schema sync/apply on Highgo fails with

```
ERROR: column "COALESCE(height, '-1'::integer::double precision)" does not exist
```

The generated DDL was:

```sql
CREATE UNIQUE INDEX "uq_tankong_sta_type_time" ON "public"."tankong_data" USING btree (
    "sta_id",
    "data_type",
    "data_time",
    "COALESCE(height, '-1'::integer::double precision)"
);
```

## Root cause

Expression/functional index key parts (as introspected from `pg_get_indexdef`) arrive as raw expression text, not a plain column name. `create_index_sql` (schema-diff sync path) and `build_create_index_statements` (table-structure-editor path) blindly double-quoted every index "column" entry for Postgres-family dialects (Postgres, Highgo, GaussDB, Kingbase, Vastbase, OpenGauss, Kwdb, Firebird, Vertica, Exasol, Uxdb). Wrapping the whole expression in double quotes turns it into a literal (and nonexistent) column identifier instead of an expression, which is exactly the reported error.

MySQL already special-cases this correctly for functional key parts (`mysql_index_column_sql`); the Postgres-family path had no equivalent.

## Fix

Leave index key parts that look like an expression (contain `(`, `::`, or whitespace) bare instead of quoting them, mirroring the existing MySQL functional-index handling already present in the same functions. Applied identically in both:
- `crates/dbx-core/src/schema_diff.rs` (schema-diff sync DDL)
- `crates/dbx-core/src/table_structure_sql/indexes.rs` (table-structure editor DDL)

## Testing

No live Highgo database was used; per this repo's established convention for DDL-generation bugs, regression tests call the real production functions directly with a constructed `IndexInfo`/index model and assert on the generated DDL text.

- Added `schema_diff::tests::preserves_bare_expression_in_postgres_family_unique_index_ddl` — confirmed failing on unfixed code (`sql.contains(...)` assertion fails, expression comes back double-quoted) and passing after the fix.
- Added `table_structure_sql::tests::preserves_bare_expression_in_postgres_create_index_statement` — same before/after confirmation for the table-structure-editor path.
- Full scoped module runs post-fix: `schema_diff::` 195 passed / 0 failed, `table_structure_sql::` 172 passed / 0 failed.
- `cargo fmt --check` and `cargo clippy -- -D warnings` both clean.

Fixes #6295

### DIFF
--- a/agents/common/src/main/java/com/dbx/agent/IndexInfo.java
+++ b/agents/common/src/main/java/com/dbx/agent/IndexInfo.java
@@ -13,13 +13,18 @@ public final class IndexInfo {
     private String index_type;
     private List<String> included_columns;
     private String comment;
+    // Parallel to `columns`: true at index i means columns[i] is a raw expression (e.g. sourced
+    // from pg_get_indexdef), not a plain column name. Empty when the introspection source
+    // doesn't track this (provenance unknown for that dialect/path). Mirrors
+    // crate::types::IndexInfo::key_is_expression on the Rust side (#6312 review).
+    private List<Boolean> key_is_expression;
 
     public IndexInfo() {
         this("", Collections.emptyList(), false, false);
     }
 
     public IndexInfo(String name, List<String> columns, boolean is_unique, boolean is_primary) {
-        this(name, columns, is_unique, is_primary, null, null, null, null);
+        this(name, columns, is_unique, is_primary, null, null, null, null, Collections.emptyList());
     }
 
     public IndexInfo(
@@ -32,6 +37,20 @@ public final class IndexInfo {
         List<String> included_columns,
         String comment
     ) {
+        this(name, columns, is_unique, is_primary, filter, index_type, included_columns, comment, Collections.emptyList());
+    }
+
+    public IndexInfo(
+        String name,
+        List<String> columns,
+        boolean is_unique,
+        boolean is_primary,
+        String filter,
+        String index_type,
+        List<String> included_columns,
+        String comment,
+        List<Boolean> key_is_expression
+    ) {
         this.name = name;
         this.columns = columns == null ? Collections.emptyList() : columns;
         this.is_unique = is_unique;
@@ -40,6 +59,7 @@ public final class IndexInfo {
         this.index_type = index_type;
         this.included_columns = included_columns;
         this.comment = comment;
+        this.key_is_expression = key_is_expression == null ? Collections.emptyList() : key_is_expression;
     }
 
     public String getName() {
@@ -74,6 +94,10 @@ public final class IndexInfo {
         return comment;
     }
 
+    public List<Boolean> getKey_is_expression() {
+        return key_is_expression;
+    }
+
     public void setName(String name) {
         this.name = name;
     }
@@ -106,6 +130,10 @@ public final class IndexInfo {
         this.comment = comment;
     }
 
+    public void setKey_is_expression(List<Boolean> key_is_expression) {
+        this.key_is_expression = key_is_expression;
+    }
+
     @Override
     public boolean equals(Object other) {
         if (this == other) return true;
@@ -118,12 +146,15 @@ public final class IndexInfo {
             && Objects.equals(filter, that.filter)
             && Objects.equals(index_type, that.index_type)
             && Objects.equals(included_columns, that.included_columns)
-            && Objects.equals(comment, that.comment);
+            && Objects.equals(comment, that.comment)
+            && Objects.equals(key_is_expression, that.key_is_expression);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(name, columns, is_unique, is_primary, filter, index_type, included_columns, comment);
+        return Objects.hash(
+            name, columns, is_unique, is_primary, filter, index_type, included_columns, comment, key_is_expression
+        );
     }
 
     @Override
@@ -136,6 +167,7 @@ public final class IndexInfo {
             + ", index_type=" + index_type
             + ", included_columns=" + included_columns
             + ", comment=" + comment
+            + ", key_is_expression=" + key_is_expression
             + ")";
     }
 }
--- a/agents/common/src/main/java/com/dbx/agent/PostgresLikeAgent.java
+++ b/agents/common/src/main/java/com/dbx/agent/PostgresLikeAgent.java
@@ -533,47 +533,87 @@ public abstract class PostgresLikeAgent extends AbstractJdbcAgent {
             return listIndexesFromCatalogArrays(schema, table, attributeCache);
         }
         return unchecked(() -> {
-            List<IndexInfo> result = new ArrayList<>();
+            // One row per index key position rather than array_agg/GROUP BY: avoids
+            // rs.getArray(...), whose JDBC support is flaky enough across these vendor drivers
+            // that readAttributeNumbers() below already has to defensively catch
+            // SQLException/UnsupportedOperationException/IncompatibleClassChangeError around it.
+            // LEFT JOIN pg_attribute (instead of an inner JOIN) so expression key parts
+            // (attnum = 0, e.g. functional index columns) survive instead of being silently
+            // dropped; their text comes from pg_get_indexdef instead of a.attname, and
+            // is_expression records which case applied so Rust never has to guess from
+            // characters in the text (#6312 review).
+            Map<String, IndexBuilder> byName = new LinkedHashMap<>();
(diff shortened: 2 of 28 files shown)


t8y2 (changes requested):
request changes: the new expression detector treats whitespace, `(`, or `::` as proof that an index key is an expression. Those characters can also belong to a quoted column identifier.

For example, PostgreSQL metadata can return the ordinary column name `order item` through `a.attname`. `looks_like_index_expression` then leaves it bare and generates `CREATE INDEX ... (order item)` instead of `CREATE INDEX ... ("order item")`.

Please preserve whether each index key came from a table column or from `pg_get_indexdef` in the metadata model, and quote only actual column keys. Please also add regressions for quoted identifiers containing spaces, parentheses, and `::`, plus a real PostgreSQL-family validation of the generated DDL.

q396921921:
Thanks for the catch — you're right, the character heuristic couldn't tell a genuinely odd but real column name (e.g. `order item`) apart from an actual expression key part, since both can contain spaces/`(`/`::`.

Pushed 5b55fa8, which drops the heuristic in favor of real provenance:

- `IndexInfo` now carries `key_is_expression: Vec<bool>`, populated during Postgres-family introspection from `a.attname IS NULL` at each key position — both in the modern `indnkeyatts`-aware query and the compat fallback for older servers without it.
- `create_index_sql` (schema-diff sync) and `build_create_index_statements` (table-structure editor) now trust that flag when it's present, and only fall back to the old character-based guess for dialects bridged through the generic Agent driver, where no real provenance is available.

Regressions added, matching each point raised:
- `postgres_index_metadata_tracks_expression_key_provenance` — the introspection SQL actually selects provenance for both query variants.
- `quotes_real_columns_whose_names_contain_expression_like_characters` — real columns named `order item` / `a(b)` / `a::b` stay quoted, while a genuine `COALESCE(...)` key part stays bare.
- `preserves_key_provenance_when_rebuilding_an_untouched_postgres_index` — regenerating an *unedited* existing index (only uniqueness changed) trusts the original snapshot's per-key provenance instead of re-guessing from characters.
- `real_postgres_round_trip_quotes_columns_and_leaves_expressions_bare` (`#[ignore]`, requires `DBX_TEST_POSTGRES_URL`) — introspects a live table that has both a weird real column and a genuine expression key, regenerates the DDL, and executes it against Postgres.
- A sqlparser-based syntax check on the generated DDL.

CI is green (rust-test / rust-fmt-clippy / frontend).

t8y2 (changes requested):
Request changes: the current fix does not cover the actual Highgo/Java Agent metadata path.

1. Java `IndexInfo` has no expression-key provenance, while `PostgresLikeAgent` joins `pg_attribute`; expression keys use `attnum=0` and are filtered out before Rust receives the index metadata.
2. The table-structure path reconstructs provenance by the first matching string position. A real column whose name equals an expression string can be assigned the wrong kind; provenance must remain ordinal rather than inferred from text.
3. The live test is ignored and exercises only the native Rust PostgreSQL path, while the exact-head `agents` job is skipped.

Please extend the Agent protocol/model and metadata query to preserve an ordinal `column vs expression` marker, consume it without string guessing, and add an Agent/Highgo-equivalent regression test covering mixed column/expression keys.

q396921921:
Addressed in 03a00a7c8. Point by point on the round-2 review:

1. **Java `IndexInfo` had no expression-key provenance; `PostgresLikeAgent` inner-joined `pg_attribute`, dropping `attnum = 0` key parts before Rust ever saw them.**
   Added `key_is_expression: List<Boolean>` to Java's `IndexInfo` (mirrors `crate::types::IndexInfo::key_is_expression`, flows through unchanged via the existing generic Agent JSON deserialization in `agent_driver.rs` — no Rust bridging code needed). Rewrote `PostgresLikeAgent.listIndexes()`'s query — the path Highgo actually uses — to `LEFT JOIN pg_attribute` instead of an inner join, and to source expression key parts from `pg_get_indexdef(ix.indexrelid, k.n, true)` instead of `a.attname`, mirroring `POSTGRES_INDEXES_SQL` in the native `db/postgres.rs` path. Switched from `array_agg`/`GROUP BY` to one row per key position, grouped into `IndexInfo` in Java, which also sidesteps the `java.sql.Array` fragility this file already defensively catches around elsewhere (`readAttributeNumbers`).

   The catalog-array branch (`listIndexesFromCatalogArrays`) has the same underlying gap, but no `PostgresLikeAgentProfile` currently calls `withCatalogAttributeArraysMappedInJava()` — it's unreachable by any driver today. Left it as-is (just updated the constructor call to keep compiling) with a comment noting a future profile enabling that path would need the same treatment, rather than expanding this fix onto dead code.

2. **`table_structure_sql/indexes.rs` reconstructed provenance by first textual match, which can misassign a real column whose name collides with another key's expression text.**
   Replaced `key_part_is_expression` (per-column, first-match lookup) with `key_expression_flags`, computed once per index: it walks the edited key list against the original snapshot with a consume-once match, so each original key part's provenance can only be claimed by one edited key part. A real column and an expression sharing identical text can no longer steal each other's classification. Added a regression test (`preserves_key_provenance_by_ordinal_position_not_first_text_match`) with two identically-texted key parts of different provenance, which fails under the old first-match logic and passes under the new ordinal one.

3. **The live test only exercises the native Rust path; the `agents` CI job is skipped.**
   Added `listIndexesPreservesRealColumnVsExpressionProvenance` in `PostgresLikeAgentTest`, using the same JDBC proxy fixtures already used elsewhere in that file (no live DB) — fakes a mixed index (plain column, a collision-prone real column name with a space — the round-1 review's own `order item` example — and a genuine expression key part) and asserts both `getColumns()` and `getKey_is_expression()` are correct and properly aligned. Added a companion test asserting the query text uses `LEFT JOIN`+`pg_get_indexdef` as a guard against regressing back to the inner-join bug. Since this touches `agents/**`, the `agents` CI job will now run on this PR (it's path-filtered) rather than being skipped.

Verified locally: `cargo test -p dbx-core schema_diff:: table_structure_sql::` (370 passed, 1 pre-existing `#[ignore]`d live-DB test unaffected), `cargo fmt --check` / `cargo clippy -- -D warnings` clean, and `cd agents && ./gradlew test` — all 30+ modules build and pass, including the new tests.

t8y2 (commented):
Patched in `e13a56604`.

Root cause: when per-key expression provenance was absent, the DDL paths guessed from characters such as whitespace, parentheses, and `::`. Valid quoted column names could therefore be emitted as bare expressions.

Fix: expression key parts are now emitted bare only when `key_is_expression` explicitly marks them as expressions. New or unknown key parts remain quoted identifiers, while existing expression-index provenance is preserved.

Validation:
- 4 exact `dbx-core` regression tests covering expression provenance and expression-like column names

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 4660a11f2, will be released in the next version.

PR #6313, opened 2026-08-15
fix(redis): keep loading key pages until the tree view actually overflows

### LINKED ISSUE #6022: [Bug] Data missing from Redis query
### 数据库类型和版本

Redis 8.6.2

### 问题描述

Data retrieved from Redis using DBX is significantly less than that retrieved by other tools such as RESP. For example, as shown in the image below, RESP retrieves 4 records, while DBX only retrieves 1. The key is a String, and the value is an int.

RESP:
Image

DBX:

Image

### 支持信息

DBX Version: v0.5.82
Runtime: Desktop
Operating System: Windows 10.0.26100.9106
Architecture: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Problem

On a large Redis database, DBX's key tree browser can show far fewer keys than actually exist, with no error and no indication anything is missing (#6022).

## Root cause

The tree view groups keys by their `:`-delimited namespace, so the rendered rows stay far fewer than the loaded key count once folders collapse (e.g. 1000 loaded keys folded into a handful of visible top-level groups). SCAN's cursor walks the keyspace in effectively random order, so the first page is an arbitrary slice of the whole DB — it will very often miss whatever the user is actually looking for.

Loading more keys only ever happened in response to the scroller's native `scroll` event. A container that doesn't overflow its own viewport never fires that event in a real browser, so once the visible (collapsed) tree fit inside the panel, the browser silently stopped pulling further SCAN pages — even though a "Load more" / "Fetch all" footer existed, most users never scroll past a seemingly-complete tree to notice it.

### Real-Redis confirmation

Seeded a Redis 8-alpine container with 4 "real" keys (matching the issue's `mt5:live01:...` shape) plus 200,000 filler keys, then called the exact production `scan_keys_batch` function with the same parameters the frontend sends on initial load (`pattern="*", count=1000, max_iterations=8`):

```
DBSIZE = 200004
first page: cursor=211584 keys_returned=1000 total_keys=200004
target keys found in first page: [] (0/4)
```

None of the 4 target keys land in the first page — this is exactly what the frontend renders and then stops on.

## Fix

`RedisKeyBrowser.vue`: after any load or folder expand, keep pulling pages (reusing the existing scroll-threshold logic) until the view is either genuinely scrollable or the keyspace/budget is exhausted — instead of relying solely on the scroll event.

## Testing

- New `RedisKeyBrowser.infiniteScroll.spec.ts`: mounts the real production component, stubs the scroller's `clientHeight`/`scrollHeight` to simulate a non-overflowing viewport, and asserts pages keep loading automatically with no scroll event, and that it correctly stops once the view genuinely overflows.
- Verified this test fails without the fix (`called 1 times` instead of `3`) and passes with it, confirming the fix addresses the reproduced defect.
- Full suite: `pnpm test` — 968 files / 9252 tests passing.
- `pnpm typecheck` / `pnpm lint` clean.

Fixes #6022

### DIFF
--- a/apps/desktop/src/components/redis/RedisKeyBrowser.infiniteScroll.spec.ts
+++ b/apps/desktop/src/components/redis/RedisKeyBrowser.infiniteScroll.spec.ts
@@ -1,0 +1,394 @@
+// @vitest-environment happy-dom
+
+import { createApp, nextTick } from "vue";
+import { createI18n } from "vue-i18n";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { RedisKeyInfo } from "@/lib/backend/api";
+
+const mocks = vi.hoisted(() => ({
+  redisScanKeysBatch: vi.fn(),
+  redisGetValue: vi.fn(),
+  redisSetString: vi.fn(),
+  redisJsonSet: vi.fn(),
+  redisHashSet: vi.fn(),
+  redisListPush: vi.fn(),
+  redisSetAdd: vi.fn(),
+  redisZadd: vi.fn(),
+  redisStreamAdd: vi.fn(),
+  redisSetTtl: vi.fn(),
+  redisSetExpireAt: vi.fn(),
+  redisCheckJsonModule: vi.fn(),
+  redisDeleteKey: vi.fn(),
+  redisDeleteKeys: vi.fn(),
+  redisExecuteCommand: vi.fn(),
+  saveHistory: vi.fn(),
+  toast: vi.fn(),
+  updateRedisDbKeyStats: vi.fn(),
+  listRedisCompletionCommandDocs: vi.fn(),
+  listRedisCompletionKeys: vi.fn(),
+  redisScanPageSize: 100,
+  infiniteScroll: true,
+  queryResultMaxRowsEnabled: true,
+  queryResultMaxRows: 5000,
+}));
+
+vi.mock("@/lib/backend/api", () => ({
+  redisScanKeysBatch: mocks.redisScanKeysBatch,
+  redisGetValue: mocks.redisGetValue,
+  redisSetString: mocks.redisSetString,
+  redisJsonSet: mocks.redisJsonSet,
+  redisHashSet: mocks.redisHashSet,
+  redisListPush: mocks.redisListPush,
+  redisSetAdd: mocks.redisSetAdd,
+  redisZadd: mocks.redisZadd,
+  redisStreamAdd: mocks.redisStreamAdd,
+  redisSetTtl: mocks.redisSetTtl,
+  redisSetExpireAt: mocks.redisSetExpireAt,
+  redisCheckJsonModule: mocks.redisCheckJsonModule,
+  redisDeleteKey: mocks.redisDeleteKey,
+  redisDeleteKeys: mocks.redisDeleteKeys,
+  redisExecuteCommand: mocks.redisExecuteCommand,
+  saveHistory: mocks.saveHistory,
+}));
+
+vi.mock("@/stores/connectionStore", () => ({
+  useConnectionStore: () => ({
+    ensureConnected: vi.fn().mockResolvedValue(undefined),
+    getConfig: () => ({ name: "Redis", redis_key_separator: ":", redis_scan_page_size: mocks.redisScanPageSize }),
+    updateRedisDbKeyStats: mocks.updateRedisDbKeyStats,
+    listRedisCompletionCommandDocs: mocks.listRedisCompletionCommandDocs,
+    listRedisCompletionKeys: mocks.listRedisCompletionKeys,
+    invalidateCompletionCache: vi.fn(),
+    refreshRedisDbKeyCounts: vi.fn(),
+  }),
+}));
+
+vi.mock("@/stores/settingsStore", () => ({
+  useSettingsStore: () => ({
+    editorSettings: {
+      infiniteScroll: mocks.infiniteScroll,
+      queryResultMaxRowsEnabled: mocks.queryResultMaxRowsEnabled,
+      queryResultMaxRows: mocks.queryResultMaxRows,
+    },
+  }),
+}));
+
+vi.mock("@/composables/useEditorFontFamilyStyle", () => ({
+  useEditorFontFamilyStyle: () => ({}),
+}));
+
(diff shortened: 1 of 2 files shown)


t8y2 (changes requested):
request changes: the automatic continuation currently introduces two unbounded-request paths.

1. `scheduleAutoLoadIfViewportShort` can repeatedly start new Redis SCAN budgets when pages are empty or contain only duplicate/non-matching keys. Because the stop limit tracks accumulated unique visible keys rather than scanned pages/requests, sparse searches can turn an ordinary tree load into a near full-database scan. Please add a bounded request/page budget for the whole automatic-fill operation and cover empty, duplicate, sparse-filter, and max-limit cases.

2. `loadMore` schedules another automatic load from `finally`. When a request fails, `hasMore` and the short-viewport condition remain unchanged, so it immediately retries again and can repeat indefinitely. Only schedule the next page after a page was successfully applied, and add a regression asserting that a failed request is not retried automatically.

q396921921:
Thanks for the review — both issues fixed in ec91b1bd7.

**1. Unbounded automatic continuation**

Added a hard page/request budget (`AUTO_LOAD_MAX_PAGES = 20`) for the automatic-fill chain in `maybeAutoLoadMoreRedisKeys`, tracked separately from the existing unique-visible-key stop condition. It's reset per operation in `invalidateScanRequests`, so a fresh search/load gets a fresh budget. Once the cap is hit, the automatic chain simply stops (`hasMore` stays true and the manual "Load more" / "Fetch all" actions remain unaffected and uncapped, since those are explicit user actions).

New tests cover:
- all-empty pages (cursor keeps advancing, no keys ever)
- all-duplicate pages (same key every time, unique count never grows)
- a sparse filter (new key only every 7th page)
- the pre-existing max-key limit still stopping well under the new page budget (regression check that the two stop conditions compose correctly)

**2. `loadMore` retrying after failure**

`loadKeys` and `loadMore` now only schedule the next automatic page after the current one actually applied — tracked via a `succeeded`/`applied` flag set right after the awaited call, so a thrown error skips it entirely (the flag is never reached on the throw path) rather than relying on `finally`, which ran regardless of success/failure before this fix. Added a regression test asserting exactly one failed request and no automatic retry, even after further ticks.

`pnpm typecheck` / `pnpm lint` clean. Full Redis component suite (`apps/desktop/src/components/redis`): 83/83 passing.

t8y2 (changes requested):
Request changes: the automatic Redis key loading is bounded by logical pages, not by actual backend work.

Each high-level page can still spend the full 50,000-COUNT scan budget. With the default 1,000-row page and twenty automatic pages, the path can reach roughly 147 DBX calls and about 1,050 Redis SCAN iterations. The new test uses an artificial single-call page size of 200,000 and therefore does not exercise this amplification.

Please enforce one total backend-call/SCAN-iteration budget across the complete automatic-fill operation, stop deterministically when that budget is exhausted, and add request-count tests using the production page size. The fix must not add polling or unbounded retries to the sidebar path.

q396921921:
Thanks for the detailed math — you're right, a page-count cap wasn't the right unit; the real cost was the retry loop *inside* each page. Fixed in 3cc806e0c.

**One shared total-iteration budget, not a page count**

`fetchScanPage`'s own COUNT-budget retry loop was the source of the amplification: each "automatic page" could already spend up to `ceil(50,000 / pageSize)` iterations internally while it kept coming back empty, so 20 automatic pages at the default 1,000-row size could still reach the ~147-call / ~1,050-iteration amplification you calculated — the page cap never saw that internal work.

Replaced `AUTO_LOAD_MAX_PAGES` with a single `AUTO_LOAD_TOTAL_SCAN_ITERATIONS = 50` budget (same unit as the backend's `max_iterations`), shared across the *entire* automatic-fill operation regardless of how many logical pages/backend calls it spans:

- `fetchScanPage` now takes an optional `iterationBudget` object it decrements in place on every real backend call, and caps its own per-call iteration count to whatever's left of it.
- The automatic path (`maybeAutoLoadMoreRedisKeys` → `loadMore`) passes the shared budget through; once `remaining <= 0` it returns immediately, before making any further backend call — deterministic stop, zero extra work.
- A manual "Load more" click or a real scroll event still calls `loadMore()` with no budget, so it keeps its own uncapped per-call retry (a single explicit user action); only the automatic follow-up it hands off to afterward is constrained by the shared budget. No polling, no unbounded retry added to the sidebar path.

Worst case is now bounded at ~`ceil(50/8) = 7` backend calls for the whole automatic-fill chain, independent of page size, pattern, or how the pages split.

**Tests now use the production default page size**

Dropped the artificial 200,000-row single-call size. `RedisKeyBrowser.infiniteScroll.spec.ts` now runs at `pageSize = 1000` (`REDIS_SCAN_PAGE_SIZE_DEFAULT`) and asserts both:
- total `redisScanKeysBatch` call count stays at `1 + ceil(50/8) = 8` (not the ~147 the unfixed amplification would produce), and
- the sum of every `iterations` (`max_iterations`) argument actually sent to the backend stays within the 50-iteration budget

across all-empty, all-duplicate, and sparse-filter scenarios, plus the pre-existing max-key-limit regression check.

Also caught and fixed a bug the refactor exposed while testing: the "Load more" button was bound as `@click="loadMore"` (bare reference), so Vue auto-passed the click `MouseEvent` as the new optional budget argument, corrupting the manual retry loop's iteration math. Bound explicitly as `loadMore()` now — full suite (968 files / 9257 tests), `pnpm typecheck`, `pnpm lint` all clean, including the button-click regressions this caught.

t8y2 (commented):
Patched in `4a7d8b776`.

Root cause: automatic Redis SCAN budget was charged only after an awaited request returned, so failed requests escaped the shared budget. A viewport resize also did not re-check whether bounded auto-loading was needed.

Fix: reserve each request's SCAN iterations before dispatch and trigger the same bounded fill check on scroller resize. The shared budget remains finite across page sizes and failures.

Validation:
- 12/12 focused Redis key browser tests passed
- `pnpm typecheck`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in e671b3da7, will be released in the next version.

PR #6314, opened 2026-08-15
fix(nacos): allow overriding target group when copying configs

### LINKED ISSUE #6298: [Bug] Nacos 配置同步时无法修改 Group，直接同步源 Group
### 来源
QQ 群反馈 · **反馈人**：以年

### AI 摘要
在 Nacos 配置同步过程中，用户无法修改 Group，系统直接同步了源配置的 Group，导致同步后的配置 Group 与源一致，不符合预期。

### 原始描述
nacos 配置同步，同步配置的时候，修改不了group，直接同步了源配置的group，这是不对的

q396921921:
## Problem

When syncing/copying Nacos configs to another namespace via the "Copy configs" dialog, there was no way to change the Group — the copy always silently reused the source Group, even when copying across environments that use different Group naming conventions.

Root cause: `NacosConfigTransferRequest` only carried a `targetNamespace`. The `transfer_configs` function (`crates/dbx-core/src/nacos/batch.rs`) always copied `config.group` straight from the source config, and the copy dialog (`NacosConfigBatchDialog.vue`) never exposed a field to override it.

Fixes #6298

## Solution

- Added an optional `target_group` field to `NacosConfigTransferRequest` (Rust), used to override the Group during transfer when non-empty; falls back to the existing behavior (keep source Group) when left blank, so this is fully backward compatible.
- Added a "Target group (optional)" input to the Copy dialog, threaded through preview/apply requests.
- Added i18n strings for all 8 locales.

## Testing

- `cargo test -p dbx-core --lib nacos::` — 164/164 passed (3 new tests covering: no override keeps source group, override replaces it, blank/whitespace override is treated as "no override").
- `NacosConfigBatchDialog.spec.ts` — 5/5 passed (1 new test verifying the target-group input is trimmed and threaded into the preview payload).
- `vue-tsc --noEmit` clean, `oxlint` clean.
- Verified end-to-end against a real local Nacos instance: copying a config without a target group keeps the source Group (reproducing the reported bug), and copying with a target group correctly lands the config under the new Group.

## Breaking changes

None — the new field is optional and defaults to the existing behavior.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/apps/desktop/src/components/nacos/NacosAdminConsole.vue
+++ b/apps/desktop/src/components/nacos/NacosAdminConsole.vue
@@ -1245,7 +1245,7 @@ async function exportConfigArchive(scope: NacosConfigSelectionScope) {
 
 const batchTransferRequest = shallowRef<NacosConfigTransferRequest | null>(null);
 
-async function previewBatch(payload: { scope: NacosConfigSelectionScope; targetConnectionId: string; targetNamespace: string; policy: NacosConflictPolicy }) {
+async function previewBatch(payload: { scope: NacosConfigSelectionScope; targetConnectionId: string; targetNamespace: string; targetGroup: string; policy: NacosConflictPolicy }) {
   batchLoading.value = true;
   batchError.value = "";
   batchPreview.value = null;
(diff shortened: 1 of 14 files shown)


t8y2 (commented):
Maintainer patch `fed42c038` closes the target-key collision found during review.

Root cause: overriding the Group could map configurations with the same Data ID from different source Groups onto one target namespace/Group/Data ID. Preview could report both as creates, while apply silently overwrote the first value.

Patch: validate the fully mapped target keys once in memory and reject duplicates before preview or apply performs any target request. Added a regression with the same Data ID in two source Groups.

Checks run:
- `cargo test -p dbx-core --no-default-features nacos::batch::tests::transfer_` (4 passed)

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in bdf5f6a87, will be released in the next version.

PR #6316, opened 2026-08-15
fix(mcp): fail closed when read-only policy and risk classifier disagree

### LINKED ISSUE #6053: [Bug] windows11上 mcp虽然设置了只读，但是在实际使用中，还是会出现写入操作
### 数据库类型和版本

mysql8.0.36

### 问题描述

1、在设置中的mcp板块设置了指定连接（一个连接）
2、MCP执行权限设置为只读
3、按照MCP配置的示例进行AI Agent的MCP配置
4、使用过程中，大部分的时候是只读的，极少部分是直接写入操作了

### 支持信息

DBX版本：v0.5.82
运行环境：桌面端
操作系统：Windows11专业版 25H2  26200.9168
系统架构：64 位操作系统, 基于 x64 的处理器

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

MCP's read-only enforcement gate (`validate_sql_policy` in `crates/dbx-mcp/src/server.rs`) computed the SQL risk via the AST-based `SqlRisk` classifier, but then discarded that result and decided whether to block a write purely from a separate keyword-only helper (`is_write_sql_for_database`) that has no awareness of row-locking clauses.

MySQL's legacy `LOCK IN SHARE MODE` syntax (still valid in 8.0.x, functionally equivalent to `FOR SHARE`) parses as a plain `SELECT` to the keyword scan, so it reached the database even when the connection's MCP policy was set to read-only — despite it taking real shared row locks (a write-adjacent side effect).

Fixes #6053.

## Solution

- `crates/dbx-core/src/sql_risk.rs`: teach the locking-clause detector `LOCK IN SHARE MODE` (it already recognized `FOR UPDATE` / `FOR SHARE` / `FOR KEY SHARE` / `FOR NO KEY UPDATE`).
- `crates/dbx-mcp/src/server.rs`: the read-only gate now fails closed on either signal — `risk != SqlRisk::ReadOnly || is_write_sql_for_database(...)` — so the AST classifier's determination can no longer be silently overridden by the weaker keyword-only check.

## Testing

- New unit test `classify_mysql_legacy_shared_lock_reads_like_for_share` in `sql_risk.rs` (asserts the locking read classifies as `Write`, and that plain reads with `mode`/`share` as ordinary identifiers still classify correctly).
- New integration test `read_only_policy_blocks_write_capable_sql_the_keyword_scan_misses` in `crates/dbx-mcp/tests/protocol.rs`: spins up a real `DbxMcpServer`, calls `dbx_execute_query` with the locking SQL under a read-only policy, and asserts it's rejected with `MCP_READ_ONLY`.
- Full suites: `cargo test -p dbx-core` (4812/4812 passed), `cargo test -p dbx-mcp` (41/41 lib + 6/6 integration passed).
- `cargo clippy -p dbx-core -p dbx-mcp --all-targets` clean.

No breaking changes — this only tightens an existing safety check.

### DIFF
--- a/crates/dbx-core/src/sql_risk.rs
+++ b/crates/dbx-core/src/sql_risk.rs
@@ -872,6 +872,10 @@ fn sql_contains_top_level_locking_clause(sql: &str, dialect: &dyn sqlparser::dia
         || words.windows(4).any(|window| {
             matches!(window, [r#for, no, key, update] if r#for == "FOR" && no == "NO" && key == "KEY" && update == "UPDATE")
         })
+        // MySQL's pre-8.0.1 spelling of `FOR SHARE`, still accepted by 8.0.x.
+        || words.windows(4).any(|window| {
+            matches!(window, [lock, r#in, share, mode] if lock == "LOCK" && r#in == "IN" && share == "SHARE" && mode == "MODE")
+        })
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


t8y2 (commented):
Maintainer patch `5ab8306c7` closes both read-only policy gaps found during review and fixes the formatter failure.

Root causes:
- valid read-only `SHOW` AST variants such as `SHOW COLLATION` and PostgreSQL `SHOW search_path` fell through to the conservative write catch-all;
- the fallback locking-clause scan only inspected depth zero, so legacy `LOCK IN SHARE MODE` inside subqueries or CTEs was missed.

Patch: classify all sqlparser read-only `SHOW` variants explicitly, scan contiguous unquoted locking keywords in every query block without matching strings/quoted identifiers, and add core plus MCP regressions for allowed SHOW statements and nested legacy locking reads.

Checks run:
- `cargo test -p dbx-core --no-default-features sql_risk::tests::`
- `cargo test -p dbx-mcp --test protocol read_only_policy`
- `cargo clippy -p dbx-core -p dbx-mcp --no-default-features -- -D warnings`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 8f3351f72, will be released in the next version.

PR #6317, opened 2026-08-15
fix(postgres): include partition children when generating table DDL

### LINKED ISSUE #6037: [Bug] postgresql 生成DDL问题及 MYSQL生成DDL问题
### 数据库类型和版本

postgresql 16.2

### 问题描述

1、postgresql 生成含有分区表的DDL时缺少当前已有分区的语句。不知道外键、触发器是否也有类似问题。
2、mysql 5.7 生成表DDL时，没有索引语句，不知道外键、触发器是否也有类似问题。
3、生成ddl语句时弹出来的界面建议是生成在一个新的SQL查询窗口，不建议现在这种弹出窗口，原因是显示不完整，也不支持格式化，修改之后的直接执行（比如创建一个同结构，只是表名不一样的表）
4、目前在服务器上存在以下问题，即：同一windows服务器，有多个用户连接服务器时，只能有一个用户能使用dbx，另外的人无法打开。

### 支持信息

dbx版本：0.5.82
运行环境：桌面端
操作系统：windows server 2019
系统架构：x64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Summary

Fixes #6037 — **point 1 only** (see scope note below).

Generating DDL for a partitioned PostgreSQL table only rendered its own `CREATE TABLE ... PARTITION BY` line and silently omitted every existing partition's `CREATE TABLE ... PARTITION OF` statement, so re-running the generated DDL against a fresh database would produce the parent table but none of its partitions.

## Root cause

`pg_ddl()` in `crates/dbx-core/src/schema.rs` fetched a table's own columns/indexes/partition-info and rendered just that one table. There was no lookup of a partitioned table's children.

## Fix

- Added `list_table_partitions()` in `crates/dbx-core/src/db/postgres.rs`, querying `pg_inherits` for a table's direct partitions.
- Made `pg_ddl` recurse into each child and append its DDL (boxed future, since an `async fn` can't self-recurse). This reuses all existing per-child rendering logic, so multi-level partition hierarchies (a partition that is itself further partitioned) are handled correctly too, not just one level.

## Scope note — issue has 4 points, only #1 is addressed here

1. ✅ Postgres partition DDL missing children — fixed in this PR.
2. Could not reproduce: "MySQL 5.7 DDL missing index statements". Traced `mysql_ddl()` — it's a pure passthrough of `SHOW CREATE TABLE` with no index-stripping anywhere in the Rust or frontend layers. Verified against a real MySQL 5.7 container with `KEY`/`UNIQUE KEY` indexes — the server's native `SHOW CREATE TABLE` output includes both, so dbx would pass them through unmodified. Happy to take another look if a repro case (e.g. via a proxy/compat layer) turns up.
3. UX suggestion (open DDL in a new query tab instead of a popup) — not a bug, out of scope for this PR.
4. A separate, unrelated issue (multi-user Windows Server session contention) — out of scope for this PR.

## Testing

Reproduced against a live PostgreSQL 16 container (parent → partition → further-partitioned sub-partition → grandchild), calling the real `pg_ddl()` function directly:

**Before:**
```sql
CREATE TABLE "public"."events" (
  "id" bigint NOT NULL,
  "created_at" timestamp without time zone NOT NULL,
  "payload" text
) PARTITION BY RANGE (created_at);
```
(all partitions silently missing)

**After:**
```sql
CREATE TABLE "public"."events" ( ... ) PARTITION BY RANGE (created_at);

CREATE TABLE "public"."events_2026" PARTITION OF "public"."events" FOR VALUES FROM ('2026-01-01 00:00:00') TO ('2027-01-01 00:00:00');

CREATE TABLE "public"."events_2027" PARTITION OF "public"."events" FOR VALUES FROM ('2027-01-01 00:00:00') TO ('2028-01-01 00:00:00') PARTITION BY RANGE (created_at);

CREATE TABLE "public"."events_2027_h1" PARTITION OF "public"."events_2027" FOR VALUES FROM ('2027-01-01 00:00:00') TO ('2027-07-01 00:00:00');
```

- `cargo test -p dbx-core --lib schema::ddl_tests::` — 33/33 pass (all pre-existing partition-render tests pass unchanged; new SQL-shape test added for the child-lookup query)
- `cargo test -p dbx-core --lib db::postgres::` — 178 pass
- `cargo fmt --check` clean
- `cargo clippy -p dbx-core --lib -- -D warnings` clean
- Rebased on latest `upstream/main`, no conflicts


t8y2 (changes requested):
当前分区 DDL 拼接会在 PostgreSQL 的多个合法分区场景中生成不等价或不可执行的 DDL，请补齐语义后再合并。

父表和子分区的 trigger 可能被重复输出；foreign partition 不能按普通本地分区生成；分区本地的约束、默认值或其他只属于子表的属性也可能在当前合并路径中丢失。仅把 partition children 追加到现有表 DDL 不能保证恢复后的对象与源库一致。

请按 PostgreSQL 对普通分区、foreign partition、父级继承对象和 partition-local 对象分别生成，明确去重规则，并用真实 PostgreSQL 覆盖：父/子同类 trigger、子分区本地 CHECK/default、foreign partition，以及导出后重新执行 DDL 的等价性验证。

q396921921:
Thanks for the detailed review — you were right on all four points. Reproduced each against a real PostgreSQL 16 container and fixed:

**1. Duplicate parent/child triggers.** PostgreSQL auto-clones a partitioned table's triggers onto every partition (tracked via `pg_trigger.tgparentid` since PG13). The trigger query only filtered `tgisinternal`, so a partition's own DDL re-declared the same trigger the parent's DDL already installs — replaying the generated DDL failed with `trigger "x" for relation "y" already exists`. Now filtered with `tgparentid = 0`, with a fallback query for PG < 13 (which never clones partition triggers, so `tgisinternal` alone is correct there).

**2. Foreign partitions rendered as ordinary local tables.** This turned out to be broader than just the partition case: `relkind = 'f'` (foreign table) was excluded from the partition-detection queries' `IN (...)` lists entirely, so *any* foreign table — partition or not — silently produced a wrong `CREATE TABLE` instead of `CREATE FOREIGN TABLE ... SERVER ... OPTIONS (...)`. Fixed by including `'f'` in those filters and branching the render on relkind, with a live test using `postgres_fdw` covering a foreign partition alongside a local one on the same parent.

**3. Partition-local CHECK constraints and defaults were lost.** Turns out CHECK constraints weren't rendered at all before this PR, for any Postgres table — added that from scratch. For partitions, local-vs-inherited filtering uses `coninhcount = 0`, since CHECK constraints propagate via the legacy inheritance-count mechanism rather than the `conparentid` tracking PK/FK constraints use. Also added detection of partition-local column `DEFAULT` overrides (by diffing the partition's own `pg_attrdef` against the parent's for the same column) rendered via `column WITH OPTIONS DEFAULT ...`.

**4. Equivalence verification.** While reproducing, I found the existing round-trip replay test was itself broken: it replayed both the parent DDL (which already embeds every child, that's the point of this PR) *and* each child DDL again separately, so every partition got double-created and the test was giving false confidence rather than actually verifying anything. Fixed that and extended it to cover the trigger/CHECK/default scenarios above; added a second live-Postgres test specifically for the foreign-partition case.

All changes verified against a live PostgreSQL 16 container (not just unit tests):
- `postgres_partition_metadata_renders_replayable_children_and_subpartitions` — parent with a trigger, a parent-level and a partition-local CHECK constraint, a partition-local DEFAULT override, sub-partitions, and a separate non-partition inheritance hierarchy; replays the full generated DDL into a fresh schema and asserts no duplicate `CREATE TRIGGER`.
- `postgres_partition_ddl_uses_foreign_table_syntax_for_foreign_partitions` — a parent with one local and one `postgres_fdw`-backed foreign partition; replays and asserts `CREATE FOREIGN TABLE ... SERVER ... OPTIONS (...)`.

`cargo test -p dbx-core --lib` (full offline suite + the two live-DB tests above), `cargo fmt --check`, and `cargo clippy -p dbx-core --lib --tests -- -D warnings` are all clean, and the full workspace (`dbx-core`, `dbx-cli`, `dbx-web`, `dbx-mcp`) builds.

Ready for another look whenever you have time.

t8y2 (changes requested):
Request changes: recursively appending the full partition tree to the shared single-table DDL API breaks export equivalence and request bounds.

1. Database export still iterates both parent and child relations and calls this same API, so the parent emits each partition and the child is emitted again later; restore can fail with `relation already exists`.
2. A child that explicitly executes `ALTER TABLE ONLY ... DROP DEFAULT` is indistinguishable from an inherited default because the metadata query inner-joins `pg_attrdef`; restore reintroduces the parent's default.
3. The recursive path serially performs the complete metadata query chain for every relation, approximately `9 × (N+1)` requests for a parent with N leaf partitions.

PostgreSQL documents partition-specific defaults and partition-tree index/DDL behavior here: https://www.postgresql.org/docs/current/ddl-partitioning.html

Please keep single-relation DDL separate from an explicit include-partitions mode, model inherited/overridden/dropped defaults, and add a full database-export replay plus bounded-request regression test.

t8y2 (commented):
Maintainer conflict resolution and patch pushed.

- old PR head: `2fce668b8216`
- latest base head: `aa6d60533c55`
- resolved PR head: `5cd17b8e525b`

Conflict cause: the PR introduced partition-tree DDL options while current `main` independently added Oracle portable-export plumbing and owned-serial DDL handling in the same schema/PostgreSQL paths.

Resolution:

- preserve the contributor's partition-tree options and complete child DDL behavior;
- preserve current `main` Oracle portable-export behavior and owned-serial rendering;
- add owned-serial detection to both modern and PostgreSQL 10 compatible partition-column batch queries;
- use `pg_constraint.conislocal` for local CHECK semantics;
- add PostgreSQL 10/OpenTenBase compatible bounded batch tiers for columns and indexes instead of per-relation fallback.

Validation:

- 4 focused Rust regression tests passed: owned-serial metadata, local CHECK semantics, modern/PG10 batch tiers, and serial DDL rendering;
- `cargo check -p dbx-core --no-default-features` passed;
- `cargo fmt --all -- --check`, `git diff --check`, conflict-marker scan, complete resolved diff inspection, and range-diff passed;
- original behavior commits were preserved by the range-diff, with one maintainer patch commit added;
- no manifest or lockfile changes.

Flags: `small-conflict=true`; no new dependency; not a large UI change. The explicit full-partition-DDL path now uses a bounded number of batch metadata requests and adds no polling, listener, background I/O, or normal query-path database request.

t8y2 (commented):
Maintainer mainline sync pushed: `5cd17b8e5` -> `e031b630f` (base `e6fa850a3`).

The previous exact-head Rust failure was the unrelated MiniMax baseline regression on the older base, not this partition-DDL change. The PR now contains current `main` without changing the reviewed partition behavior.

Validation after the sync: four focused PostgreSQL partition regressions passed (local CHECK semantics, bounded compatibility tiers, owned-serial replay, and nested partition rendering), `cargo check -p dbx-core --no-default-features` passed, and `git diff --check` is clean.

t8y2 (commented):
Mainline sync pushed after `#6535` fixed the unrelated MiniMax incremental replay regression that failed this PR's previous Rust CI.

The partition DDL patch is unchanged. The owned-serial metadata/rendering tests, local CHECK semantics, bounded compatibility tiers, `cargo fmt --check`, and the formerly failing MiniMax regression pass on this exact local head.

t8y2 (approved):
The partition DDL semantic blockers and conflict-resolution gaps are fixed on the exact head. Focused metadata/rendering tests and Rust checks pass; the remaining Windows bundle is still running.

t8y2:
Thanks for the contribution! Merged in 1345e3f, will be released in the next version.

PR #6318, opened 2026-08-15
fix(query): allow editing joined results with multiple qualifying source tables

### LINKED ISSUE #6269: [Bug] 多表join查询，结果集显示“只读结果”，不可编辑
### 数据库类型和版本

mysql 8.0

### 问题描述

1.复现步骤：
SELECT u.id, u.nickname, p.paper_id, p.paper_name from fd_user u join fd_paper p on (u.id = p.user_id) where u.id = 1;

这个语句中，u.id、p.paper_id，分别是2表的主键索引，应该是能推导出对应字段和需要变更的表
在dbeaver中，同样的查询语句，查出的结果集可编辑，生成的sql也是准确的，在dbx中不可编辑

2.期望结果：
多表关联时，有索引的结果集可编辑，可以参考dbeaver的实现效果

3.实际结果：
结果集无法编辑

### 支持信息

DBX 版本：V0.5.84
运行环境：桌面端
操作系统：macOS 26.5.2
系统架构：aarch64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## 变更说明

Fixes two independent bugs that combined to lock every multi-table JOIN result as read-only in the query result grid, even when the query had a perfectly safe editable target.

1. **Parenthesized `ON`/`USING` clauses broke the SQL-shape analysis entirely** (`crates/dbx-core/src/sql_editability.rs::parse_from_sources`, mirrored in `apps/desktop/src/lib/sql/sqlAnalysis.ts`). The function bailed out with an empty source list — reported as `"complex-source"` — whenever the `FROM` clause contained *any* parenthesis, including an ordinary `JOIN ... ON (a = b)` predicate (a very common style, and exactly the issue's repro SQL) or `JOIN ... USING (col)`. This check was redundant: `parse_table_source_at` already rejects derived tables (`FROM (SELECT ...) t`) per source position, and `find_top_level_keyword` is already parenthesis-depth-aware, so removing the blanket bail keeps subquery rejection intact while no longer misfiring on ordinary join predicates.

2. **Once analysis succeeded, `buildQueryMetadataPatch` (`apps/desktop/src/stores/queryStore.ts`) forced the whole result read-only whenever more than one source table qualified as an editable candidate** (i.e. had its full primary key + at least one writable column returned in the result — exactly the reported `u.id, u.nickname, p.paper_id, p.paper_name` case, where both `fd_user` and `fd_paper` qualify). Now it binds editing to the first-appearing source (typically the driving `FROM` table), the same way the single-candidate path already worked. Columns from the other source(s) simply fall out of `querySourceColumns`, which `useDataGridEditor.ts`'s `canEditColumn` already treats as individually read-only per cell — the same mechanism that already makes expression/computed columns read-only in an otherwise-editable grid. No changes were needed to the edit-apply path (`useDataGridEditor.ts`, `data_grid_sql.rs`).

This intentionally does **not** implement full DBeaver-style simultaneous multi-table editing (routing each column's UPDATE to its own source table) — that would require per-source `tableMeta`/save-statement plumbing through `data_grid_sql.rs`. This PR takes the smaller, lower-risk step: the join result is no longer fully locked, and the primary source's columns become editable.

## 变更类型

- [x] Bug 修复

## 涉及前端

- [x] 本 PR 涉及前端改动，已附验证说明（见下方）

Verified end-to-end in a real dev session (not mocked): an isolated MySQL 8.4 Docker container was seeded with the issue's exact schema (`fd_user`/`fd_paper`), a real `dbx-web` dev backend and real Vite frontend were started, and the issue's exact SQL was run through the real Pinia `query` store in a live browser:

```sql
SELECT u.id, u.nickname, p.paper_id, p.paper_name
FROM fd_user u JOIN fd_paper p ON (u.id = p.user_id)
WHERE u.id = 1
```

- Before the fix: `tab.queryEditabilityReason === "complex-source"` — grid fully locked.
- After the fix: `tab.queryEditabilityReason` is unset, `tab.tableMeta.tableName === "fd_user"`, `tab.querySourceColumns === ["id","nickname",undefined,undefined]` — `fd_user` columns editable, `fd_paper` columns individually read-only — matching the real rows returned from MySQL.

A full pixel screenshot of the editable grid itself wasn't obtainable in this environment: the app's virtualized sidebar/toolbar resists synthetic pointer events from headless browser automation (the same limitation already noted in #6298's PR). The result panel's target-table indicator pill was captured showing `dbx.fd_user` once bound, confirming the fix operates correctly through the real UI reactivity layer, not just the store in isolation.

## 验证

- [x] `pnpm typecheck` 通过
- [x] `pnpm lint` 通过（无新增告警）
- [x] `pnpm exec vitest run` 全量 964 文件 / 9260 测试通过
- [x] `cargo fmt --check -p dbx-core` 通过
- [x] `cargo clippy -p dbx-core --lib -- -D warnings` 通过
- [x] `cargo test -p dbx-core --lib sql_editability::` 28/28 通过（含 2 个新回归测试，均在 revert 修复后真实失败、重新应用后通过）
- [x] 端到端：真实 MySQL 8.4 容器 + 真实 dev 后端 + 真实浏览器 Pinia store，issue 原始 SQL 复现并验证修复

## 关联 Issue

Fixes #6269

### DIFF
--- a/apps/desktop/src/lib/__tests__/sql/editableQueryHiddenKeys.spec.ts
+++ b/apps/desktop/src/lib/__tests__/sql/editableQueryHiddenKeys.spec.ts
@@ -201,6 +201,24 @@ describe("editable query hidden primary keys", () => {
     expect(analyzeEditableQueryEditability("select id from jobs union select id from archived_jobs")).toEqual({ editable: false, reason: "set-operation" });
   });
 
+  it("disambiguates schema-qualified columns across cross-schema same-name tables", () => {
+    // Regression for PR #6318 review: two joined tables with the same table
+    // name in different schemas cannot be told apart by table name alone.
+    const result = analyzeEditableQueryEditability("select s1.foo.id, s1.foo.name, s2.foo.total from s1.foo join s2.foo on s1.foo.id = s2.foo.id");
+
+    expect(result.editable).toBe(true);
+    if (!result.editable) return;
+    expect(result.analysis.columns.map((column) => column.sourceKey)).toEqual(["foo:0", "foo:0", "foo:1"]);
+  });
+
+  it("leaves ambiguous unqualified columns unbound across cross-schema same-name tables", () => {
+    const result = analyzeEditableQueryEditability("select foo.id, foo.name from s1.foo join s2.foo on s1.foo.id = s2.foo.id");
+
+    expect(result.editable).toBe(true);
+    if (!result.editable) return;
+    expect(result.analysis.columns.every((column) => column.sourceKey === undefined)).toBe(true);
+  });
+
   it("resolves appended aliases to result indexes", () => {
     expect(
       hiddenResultColumnIndexes(
--- a/apps/desktop/src/lib/sql/sqlAnalysis.ts
+++ b/apps/desktop/src/lib/sql/sqlAnalysis.ts
@@ -46,6 +46,11 @@ export interface EditableQuerySource {
   tableName: string;
   tableNameQuoted?: boolean;
   alias?: string;
+  // True when this source sits on the nullable side of an outer join (or
+  // transitively behind one): it is not guaranteed to correspond to a real
+  // row of that table for every returned result row. A nullable source must
+  // never be treated as a safe, unique write target.
+  nullable?: boolean;
 }
 
 const POSTGRES_FOLDED_IDENTIFIER_TYPES = new Set(["postgres", "redshift", "gaussdb", "highgo", "uxdb", "vastbase", "kwdb", "opengauss", "questdb"]);
@@ -226,8 +231,13 @@ function parseSelectColumn(col: string, sources?: EditableQuerySource[]): Editab
   const alias = parseColumnAlias(source.rest);
   if (alias === null) return parseComputedSelectColumn(col);
   const sourceName = source.parts[source.parts.length - 1];
-  const qualifier = source.parts.length >= 2 ? source.parts[source.parts.length - 2] : undefined;
-  const sourceKey = qualifier ? sourceKeyForQualifier(sources, qualifier) : undefined;
+  // Everything before the column name (e.g. ["s1", "foo"] for "s1.foo.id").
+  // Two joined tables that share a table name across schemas ("s1.foo" vs
+  // "s2.foo") are only distinguishable by the full qualifier chain, not just
+  // the identifier immediately before the column.
+  const qualifierParts = source.parts.slice(0, -1);
+  const qualifier = qualifierParts.at(-1);
+  const sourceKey = qualifierParts.length ? sourceKeyForQualifierParts(sources, qualifierParts) : undefined;
   return {
     sourceName,
     sourceNameQuoted: source.quoted[source.quoted.length - 1],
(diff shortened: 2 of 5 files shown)


t8y2 (changes requested):
当前“从多个可编辑来源中选择目标表”的策略会在部分 JOIN 查询中选错写入目标，请修正来源绑定后再合并。

仅按可编辑性筛选并选择一个候选，无法表达 RIGHT/FULL JOIN 中哪一侧真实对应当前结果行；quoted alias、跨 schema 同名表以及同一表重复 join 时，也不能只靠规范化后的表名/别名可靠绑定。选错目标后，更新或删除可能落到另一张表或另一 schema，这是高风险数据修改。

请把结果列的来源信息与解析出的具体 relation identity（schema、表、alias、join side）一一绑定；存在多个可能目标或无法唯一证明时应保持只读，而不是猜选。请补充 RIGHT JOIN、FULL JOIN、quoted alias、跨 schema 同名表和 self-join 的回归测试。

q396921921:
Thanks for the review — you were right, `candidates[0]` as a "first in FROM/JOIN order" heuristic doesn't express join side at all, and picking any table among several structurally-qualifying candidates without proof was a guess, not a bound. Pushed a fix (7d815041d) that addresses each point:

**Join side is now tracked and enforced.** The parser (both the Rust engine and its mirrored TS twin) now walks the FROM/JOIN chain left-to-right and marks each source `nullable` when it sits on the optional side of an outer join: INNER/CROSS/comma joins guarantee both sides, LEFT/RIGHT guarantee only one side, FULL guarantees neither, and a RIGHT/FULL following an earlier LEFT resets those earlier guarantees (left-associative evaluation, so `(A LEFT JOIN B) RIGHT JOIN C` makes A and B both nullable relative to C). `queryStore.ts` now excludes nullable sources from candidacy outright — a source that isn't guaranteed present for every result row is never a safe write target, regardless of whether its primary key columns are structurally present in the projection.

**No more guessing among multiple targets.** Editing now binds only when exactly one non-nullable, structurally-qualifying source remains. If zero or more than one source could be "the" target (e.g. an INNER JOIN where both sides independently qualify, or a FULL JOIN where neither side is guaranteed), the query stays read-only.

**Qualifier binding now uses the full relation identity, not just the last identifier.** Column/star qualifiers used to resolve to a source using only the single identifier immediately before the column (e.g. `foo` in `s1.foo.id`), which silently drops the schema. Two joined tables sharing a table name across schemas (`s1.foo` vs `s2.foo`) were therefore ambiguous and could bind to either — I reproduced this and confirmed it was a real, live bug, not hypothetical. Qualifier resolution now walks the entire chain (schema.table, or catalog.schema.table) and matches unambiguously, falling back to unbound/read-only when it still can't prove a unique source. Quoted aliases and self-joins were already resolved correctly via each source's positional key (`alias:index`) — added regression coverage to lock that in.

**Regression tests added** (Rust `sql_editability.rs` + TS `sqlAnalysis.ts`/`queryStore.test.ts`):
- LEFT/RIGHT/FULL join nullability tracking, including the LEFT-then-RIGHT reset case
- RIGHT JOIN and FULL JOIN at the `queryStore` level (nullable side excluded / stays read-only)
- INNER JOIN with two independently-qualifying candidates now stays read-only instead of guessing
- Quoted alias + JOIN, self-join, cross-schema same-table-name (both the disambiguated and the still-ambiguous-when-unqualified case)

All existing tests still pass; `cargo clippy -D warnings` and `oxlint` are clean. Happy to adjust further if there's a scenario this still misses.

t8y2 (changes requested):
Request changes: the original joined-result editing requirement is still not implemented safely.

1. The Issue's INNER JOIN has two non-nullable source tables with complete keys, so the code produces two candidates and returns `complex-source`; the new test explicitly preserves this read-only behavior. The PR therefore cannot currently close the Issue.
2. DBX still stores one global `tableMeta` and one source-column set. Safe multi-table editing requires per-result-column source/identity routing, not selection of one global target.
3. Exact-head frontend typecheck fails because the current TypeScript target does not support the two uses of `Array.prototype.at`.

The current source analysis also performs one serial metadata load per source on a cold cache. Please either implement per-column/per-source editing with bounded metadata loading, or narrow the PR scope and remove the closing claim; also fix the typecheck failure.

q396921921:
关闭本 PR，交还 issue #6269 认领。

Review 中指出当前方案在架构层面有根本性局限：全局单一 tableMeta / 单一 source-column 集合无法支撑按结果列做 per-column 来源路由，导致 issue 描述的原始场景（INNER JOIN 双非空唯一键源表）在测试里被保留为只读、并未真正解决；同时 frontend typecheck 因 Array.prototype.at 目标版本不支持而失败。这需要重新设计支持按列/按源精确路由并限界元数据加载的方案，而不是在当前实现上小修小补，因此选择关闭并让其他贡献者或后续重新设计的 PR 接手。

PR #6319, opened 2026-08-15
fix(transfer): fail fast on incompatible target structure before truncating data

### LINKED ISSUE #6248: [Bug] 数据传输功能,选择了结构和数据覆盖模式,如果两表字段不一致,传输失败.
### 数据库类型和版本

mysql8.0

### 问题描述

1.正式库表A和测试库表A字段不一致的情况,使用数据传输,报错,表结构不能覆盖

### 支持信息

DBX 版本: v0.5.83
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Summary

- "Structure + data" transfer always requests `create_table`, but the target's create/sync-structure DDL is silently skipped once the target table already exists — its columns are left untouched.
- In Overwrite mode, DBX then **TRUNCATEs the target's existing data first**, and only afterward discovers via a raw MySQL insert error (`ER_BAD_FIELD_ERROR`) that the untouched target is missing a column the source needs — so the target's data is destroyed even though the transfer ultimately fails.
- Adds a pre-flight column-compatibility check (`missing_transfer_target_columns`) that runs before the TRUNCATE when the target table pre-existed and `create_table` was requested, so a structure mismatch now fails fast with a clear message, before any data is touched.

Fixes #6248

## Test plan

- [x] Added `live_mysql_transfer_structure_overwrite_rejects_incompatible_target_columns` (real MySQL 8.4, not mocked): reproduced the bug against unfixed code first — target's pre-existing row was destroyed by TRUNCATE even though the transfer errored; after the fix, the same test passes (transfer fails fast, target row survives).
- [x] Added 3 unit tests for the new pure `missing_transfer_target_columns` helper.
- [x] `cargo test -p dbx-core --lib transfer::` — 198/198 passed.
- [x] `cargo test -p dbx-core --test live_mysql_transfer` against real MySQL — new test passes; existing `live_mysql_transfer_preserves_spatial_values_and_modes` (covers Overwrite mode) still passes, no regression.
- [x] `cargo fmt --check -p dbx-core` / `cargo clippy -p dbx-core --lib --tests -- -D warnings` — clean.
- [x] Rebased onto latest `upstream/main`, re-ran tests post-rebase.

### DIFF
--- a/crates/dbx-core/src/transfer.rs
+++ b/crates/dbx-core/src/transfer.rs
@@ -1407,6 +1407,15 @@ fn writable_transfer_columns(
         .collect()
 }
 
+/// Columns DBX is about to write into `target_columns` (via `col_names`) that
+/// don't exist in the target table. MySQL (and most other targets) compare
+/// unquoted identifiers case-insensitively, so the check does too.
+fn missing_transfer_target_columns(target_columns: &[db::ColumnInfo], col_names: &[String]) -> Vec<String> {
+    let existing: std::collections::HashSet<String> =
+        target_columns.iter().map(|column| column.name.to_lowercase()).collect();
+    col_names.iter().filter(|name| !existing.contains(&name.to_lowercase())).cloned().collect()
+}
+
 fn transfer_key_columns(columns: &[db::ColumnInfo], db_type: &DatabaseType) -> Vec<String> {
     let uses_unique_key_model = matches!(db_type, DatabaseType::Doris | DatabaseType::StarRocks);
     columns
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Patched in 90699299f after validating the transfer path against DBeaver's target-column mapping behavior.

Root cause: the preflight treated every target column name as case-insensitive, only checked source columns missing from the target, swallowed target metadata failures, and could fetch the same target metadata twice before upsert/identity handling.

Patch: apply database-aware identifier matching, reject target-only NOT NULL columns without a default/generated value before overwrite, propagate metadata errors with context, and reuse one target-column read across validation and later transfer logic. This adds no metadata request; create-table + upsert now uses one fewer target metadata lookup.

Checks:
- cargo test -p dbx-core --lib transfer:: --no-default-features (199 passed)
- cargo check -p dbx-core --no-default-features
- cargo clippy -p dbx-core --lib --tests --no-default-features -- -D warnings
- cargo fmt -p dbx-core -- --check
- real MySQL 8.4.2 live transfer regression: both missing source columns and target-only required columns fail before TRUNCATE, with pre-existing rows preserved

t8y2 (commented):
Rebased the maintainer patch from 90699299f to f8045460e on current base bd4cfd6b9 and resolved the small transfer.rs conflict.

Conflict cause: main added Kyuubi beside Hive/Impala in the unsupported-upsert path while the patch moved the target-column metadata read earlier for pre-TRUNCATE validation. The resolution preserves both behaviors: Kyuubi still skips upsert metadata handling and uses its new cursor path, while compatible targets reuse the single metadata read for validation and later insert handling. Kyuubi was also added to the case-insensitive column matching group.

Dependency flags: unchanged; no dependency or lockfile changes.

Revalidated on the resolved exact head:
- cargo test -p dbx-core --lib transfer:: --no-default-features (200 passed, including Kyuubi transfer coverage)
- cargo check -p dbx-core --no-default-features
- cargo clippy -p dbx-core --lib --tests --no-default-features -- -D warnings
- cargo fmt -p dbx-core -- --check
- real MySQL 8.4.2 regression: both mismatch directions fail before TRUNCATE and preserve pre-existing rows
- git range-diff confirms the original PR behavior and maintainer patch are preserved, with only the Kyuubi base integration added

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in e97a5a5b8, will be released in the next version.

PR #6321, opened 2026-08-15
fix(mongodb): reconcile execution summary status for Mongo commands

### LINKED ISSUE #6085: [Bug] mongdb 查询页面摘要不显示
### 数据库类型和版本

MongoDB 3.6

### 问题描述

执行结果页面->摘要都没有统计出来，比如执行状态，执行完成，返回行等。并且没有查询出来数据的时候无法点击数据表
Image

### 支持信息

DBX 版本: v0.5.82
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

_No response_

q396921921:
## Problem

Running a MongoDB command from the query editor (e.g. `db.users.find()`) leaves the **Summary** panel stuck showing "Execution finished 0/1", 0%, status "Not run" (i.e. skipped), and rows "—" — even though the query actually succeeded and the Table Data tab shows the correct rows.

Root cause is two-fold:
1. `createBatchSqlExecution` always split the editor text with the **SQL** statement splitter regardless of database type, instead of the Mongo-aware command splitter already used elsewhere (`statementRanges()` in `tabPresentation.ts`).
2. The MongoDB execution branch in `executeTabSql` never fed its per-command results back into the batch execution state (`reconcileBatchSqlResults`) the way the SQL path does. So every Mongo command's summary item stayed at its initial `pending`/`running` status and got downgraded to `skipped` ("Not run") once execution finished, regardless of whether it actually succeeded or failed.

## Fix

- Split with `splitMongoCommandRanges` when the connection's database type is `mongodb`, so the batch item list matches the commands that will actually run.
- Call `reconcileBatchSqlResults` with the real per-command results after the Mongo execution loop finishes, mirroring the existing SQL streaming path. Mongo results don't carry `statement_index`, so this reconciles positionally, which is safe because each parsed Mongo command pushes exactly one result (success or error) in order.

## Testing

- Added `apps/desktop/src/stores/__tests__/queryStore.mongoExecutionSummary.spec.ts` (3 new tests: successful find, failing command, multi-command batch). Confirmed they fail against the pre-fix code (reproducing the exact "skipped"/`completed: 0` bug) and pass after the fix.
- Real end-to-end repro: local MongoDB 7.0 + `dbx-web` dev backend + `pnpm dev:web`, ran `db.users.find()` against a 3-document collection.
  - Before: Summary showed "Execution finished 0/1", status "Not run", rows "—".
  - After: Summary shows "Execution finished 1/1", status "Success", rows "3".
- Full suite: `pnpm vitest run apps/desktop/src` — 682 files / 6361 tests pass.
- `pnpm typecheck` and `pnpm lint` clean.

Fixes #6085

### DIFF
--- a/apps/desktop/src/stores/__tests__/queryStore.mongoExecutionSummary.spec.ts
+++ b/apps/desktop/src/stores/__tests__/queryStore.mongoExecutionSummary.spec.ts
@@ -1,0 +1,121 @@
+import { createPinia, setActivePinia } from "pinia";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { splitMongoCommandRanges } from "@/lib/mongo/mongoShellCommand";
+
+const mocks = vi.hoisted(() => ({
+  ensureConnected: vi.fn(),
+  getConnectionConfig: vi.fn(),
+  mongoFindDocuments: vi.fn(),
+  mongoParseShellCommand: vi.fn(),
+}));
+
+vi.mock("@/lib/backend/api", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@/lib/backend/api")>();
+  return {
+    ...actual,
+    mongoFindDocuments: mocks.mongoFindDocuments,
+    mongoParseShellCommand: mocks.mongoParseShellCommand,
+  };
+});
+
+vi.mock("@/stores/connectionStore", () => ({
+  useConnectionStore: () => ({
+    ensureConnected: mocks.ensureConnected,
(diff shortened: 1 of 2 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in ad7a66c, will be released in the next version.

PR #6324, opened 2026-08-15
fix(redis): keep JSON tree key labels from collapsing vertically

### LINKED ISSUE #6293: [Bug]
### 数据库类型和版本

redis json视图 字段名有时候会竖着显示

### 问题描述

Image

### 支持信息

v0.5.84 桌面端 windows 10

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

In the Redis JSON tree view (used for member/value detail dialogs), field/key labels sometimes render vertically, one character per line, instead of horizontally — see #6293.

Root cause: `.json-tree-row` in `apps/desktop/src/components/common/JsonTree.vue` is a flex row, and the row's key/index/punctuation/bracket spans had no `flex-shrink` guard. The ancestor `.json-tree` sets `overflow-wrap: anywhere`, which lowers a flex item's automatic minimum size down to roughly one character. When a row's total content (indent + key + a long unbroken value) exceeds the available width, the flex-shrink algorithm can squeeze whichever sibling has the largest content-size mismatch down toward that ~1-character minimum instead of wrapping the long value normally — collapsing the key into a vertical single-character column.

Reproduces with a narrow container width + deep nesting + a long value with no spaces (e.g. embedded HTML/URL), matching the intermittent nature described in the report.

## Solution

Give the row's label spans (`.json-tree-key`, `.json-tree-index`, `.json-tree-punctuation`, `.json-tree-bracket`) `flex: 0 0 auto` so they keep their natural width, and give the value/summary spans (`.json-tree-string`, `.json-tree-number`, `.json-tree-boolean`, `.json-tree-null`, `.json-tree-summary`) `flex: 1 1 auto; min-width: 0` so only the value content wraps within the remaining row width.

## Testing

- Reproduced live in the running app: connected to a real Redis instance, opened a Set member's JSON view at a narrow viewport (500px) with 8 levels of nesting and a long unbroken string value — the key collapsed into a vertical single-character column before the fix, and stayed horizontal (value wrapping normally) after.
- `pnpm lint` — clean (no new warnings on the changed file)
- `pnpm typecheck` (`vue-tsc --noEmit`) — clean
- `pnpm test` (`vitest run`) — 974 files / 9291 tests passing

Fixes #6293

### DIFF
--- a/apps/desktop/src/components/common/JsonTree.vue
+++ b/apps/desktop/src/components/common/JsonTree.vue
@@ -317,9 +317,13 @@ defineExpose({ expandAll, collapseAll, resetExpansion, refresh });
 .json-tree-string,
 .json-tree-number,
 .json-tree-boolean,
-.json-tree-null,
-.json-tree-summary {
+.json-tree-null {
   flex: 1 1 auto;
+  min-width: 0;
+}
+
+.json-tree-summary {
+  flex: 0 1 auto;
   min-width: 0;
 }
 
--- a/apps/desktop/src/components/common/JsonTree.vue
+++ b/apps/desktop/src/components/common/JsonTree.vue
@@ -307,6 +307,22 @@ defineExpose({ expandAll, collapseAll, resetExpansion, refresh });
   font-weight: 650;
 }
 
+.json-tree-key,
+.json-tree-index,
+.json-tree-punctuation,
+.json-tree-bracket {
+  flex: 0 0 auto;
+}
+
+.json-tree-string,
+.json-tree-number,
+.json-tree-boolean,
+.json-tree-null,
+.json-tree-summary {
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
已在 15edd50 中补齐布局修复：折叠容器的 summary 改为仅收缩、不占用剩余空间，右括号会紧跟摘要而不是被推到行尾。

验证：`pnpm exec vitest run packages/app-tests/jsonTreeLayout.test.ts`（1 个文件、1 条测试通过），`git diff --check` 通过。

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in ed8726777, will be released in the next version.

PR #6325, opened 2026-08-15
fix(editor): resync caret after paste to fix WebKit comma-jump bug

### LINKED ISSUE #4647: [Bug] 在in条件的()里,逗号会自动往前跑
### 数据库类型和版本

mysql8.0和阿里的polar

### 问题描述

复现步骤:
1.在一堆in条件的()里,输入条件之后，点个逗号，然后粘贴，再点个逗号，第二个逗号会跑到第一个逗号后边。

### 支持信息

DBX 版本: v0.5.68
运行环境: 桌面端
操作系统: macOS 26.5
系统架构: aarch64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Problem

Pasting text inside a SQL `IN (...)` list and then typing another
character (e.g. a comma) could insert it in the wrong place. Repro
from the issue: type `IN (1`, type `,`, paste `2`, type `,` — expected
`IN (1,2,)` but got `IN (1,,2)` (the second comma jumps in front of
the pasted text).

## Root cause

On WebKit (macOS WKWebView, which the desktop app runs on), after
CodeMirror replaces DOM text nodes to apply a paste,
`document.getSelection()` reports the correct caret position, but
WebKit's *native* typing-insertion anchor doesn't follow — it silently
stays at the pre-paste position, so the next typed character lands
there instead.

Confirmed with a real browser (not a code-reading guess): the exact
same Playwright script produces the correct result on Chromium and
the bug on WebKit. The trigger is simply "any character sitting right
after the cursor at paste time" — it's unrelated to auto-closed
brackets/parens (reproduces the same way with a plain trailing
letter). A genuine subsequent native selection change (e.g. an arrow
key) forces WebKit to re-anchor correctly.

## Fix

On any `input.paste` transaction, replicate that "genuine selection
change" nudge on the next animation frame via two harmless CodeMirror
selection-only dispatches (move the caret away one position, then
back). This is a no-op on Chromium/Firefox, whose insertion point
already tracks the real selection, so no browser sniffing is needed.

The pure position math is extracted into a small testable module
(`queryEditorPasteCaretResync.ts`) rather than left inline.

## Testing

- New regression tests (`queryEditorPasteCaretResync.spec.ts`): 4/4
  passing, covering the nudge-target calculation and that the paste
  handler is wired to call it.
- Editor test suite: 212/212 passing.
- Full frontend suite (`pnpm exec vitest run apps/desktop`): 6361/6361
  passing.
- `pnpm typecheck` / `pnpm lint`: clean.
- Manually verified end-to-end with Playwright driving a real local
  dev instance (real MySQL connection) in both WebKit and Chromium,
  reproducing the exact steps from the issue before and after the fix.

Fixes #4647

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -1492,11 +1492,12 @@ async function pasteClipboardAsSqlInCondition(): Promise<boolean> {
 function resyncCaretAfterPaste(view: EditorViewType) {
   const EditorSelection = codeMirrorEditorSelection;
   if (!EditorSelection) return;
-  const pos = view.state.selection.main.head;
-  const nudged = computePasteCaretResyncTarget(pos, view.state.doc.length);
+  const selection = view.state.selection;
+  const pos = selection.main.head;
+  const nudged = computePasteCaretResyncTarget(selection, view.state.doc.length);
   if (nudged === null) return;
   requestAnimationFrame(() => {
-    if (!view.dom.isConnected || view.state.selection.main.head !== pos || !view.state.selection.main.empty) return;
+    if (!view.dom.isConnected || view.state.selection.ranges.length !== 1 || view.state.selection.main.head !== pos || !view.state.selection.main.empty) return;
     view.dispatch({ selection: EditorSelection.cursor(nudged) });
     view.dispatch({ selection: EditorSelection.cursor(pos) });
   });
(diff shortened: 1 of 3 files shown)


t8y2 (commented):
已在 9d87e1f 中把 WebKit paste caret nudge 限定为单一空光标。CodeMirror 多光标和非空选区不再进入 workaround，因此不会被压缩为单光标。

验证：`pnpm exec vitest run apps/desktop/src/lib/__tests__/editor/queryEditorPasteCaretResync.spec.ts`（1 个文件、6 条测试通过，包含真实 CodeMirror 多光标 selection），`pnpm typecheck` 和 `git diff --check` 通过。

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 030627cc1, will be released in the next version.

PR #6326, opened 2026-08-15
fix(mysql): keep the edit-view draft pretty-printed instead of MySQL's flattened definition

### LINKED ISSUE #5057: [Bug] mysql查看视图闪退，编辑视图，所有视图代码会变成1行，注释丢失
### 数据库类型和版本

MySQL  5.6.24

### 问题描述

问题一：
1、点击连接
2、点击数据库
3、点击视图，运行几秒后闪退

问题二：
4、右键视图，编辑视图，所有视图代码变成一行，且所有注释丢失。

Image

### 支持信息

DBX 版本: v0.5.71
运行环境: 桌面端
操作系统: Windows 10.0.28020.2539
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
Fixes #5057 (the second half of it — see scope note below).

## Root cause

MySQL's `SHOW CREATE VIEW` always returns the view definition already flattened by the server: comments stripped, everything collapsed onto effectively one line. Verified against a real MySQL 8.0 instance:

```sql
CREATE VIEW v_active_users AS
-- this is a leading comment
SELECT id, name -- inline comment
FROM users
WHERE active = 1;

SHOW CREATE VIEW v_active_users;
-- CREATE ... VIEW `v_active_users` AS select `users`.`id` AS `id`,`users`.`name` AS `name` from `users` where (`users`.`active` = 1)
```

`ObjectSourceDialog.vue` already runs this raw definition through `formatSqlForDisplay(...)` to pretty-print it for the read-only "view source" preview (`content`). But when opening the editor — both via the "Edit" button (`editSource()`) and when a view is opened directly in edit mode — it seeded the editable draft from the *raw* `editableText` instead of the pretty-printed text, so the editor showed everything squashed onto one line.

## Scope

The original issue reports two problems:
1. Views list crashing after a few seconds — a separate report (`onenewcode`) says this is already fixed in a later release; not reproduced here.
2. Editing a view collapses all code to one line and loses comments.

For (2), the "collapses to one line" part is a genuine dbx bug and is what this PR fixes. The "comments lost" part is not fixable at the application layer — MySQL's server never stores the original view text, so the comments are gone before dbx ever sees the definition (same `SHOW CREATE VIEW` output shown above). No tool can recover them.

## Fix

Reuse the already-computed pretty-printed text (`formatted`/`content.value`) to seed the edit draft instead of the raw single-line `editable`/`editableText.value`, via a small helper `resolveObjectSourceEditDraft` (falls back to the raw text if formatting produced nothing usable).

## Testing

- `apps/desktop/src/lib/__tests__/table/objectSourceEditor.spec.ts`: new tests for `resolveObjectSourceEditDraft`, including a real round-trip through `formatSqlForDisplay` on the exact single-line text MySQL returns.
- `apps/desktop/src/components/objects/__tests__/ObjectSourceDialog.editDraft.spec.ts`: new wiring test confirming both draft-seeding call sites use the pretty-printed text. Reverted the `.vue` fix and confirmed these tests genuinely fail against the old code, then reapplied and confirmed they pass.
- Full suite: `pnpm exec vitest run` — 9294/9295 passing; the sole failure (`DocumentBrowserFieldSearch.spec.ts`) is a pre-existing unrelated MongoDB UI flake (also seen and documented in earlier unrelated PRs), not touched by this change.
- `pnpm typecheck` clean, `oxlint` clean on changed files.
- Rebased onto latest `upstream/main`, no conflicts.

### DIFF
--- a/apps/desktop/src/components/objects/ObjectSourceDialog.vue
+++ b/apps/desktop/src/components/objects/ObjectSourceDialog.vue
@@ -104,7 +104,7 @@ async function loadSource(nextEditing = props.initialEditing && canEdit.value) {
     const formatted = await formatSqlForDisplay(editable, props.formatDialect ?? props.dialect, settingsStore.editorSettings.sqlFormatter);
     editableText.value = editable;
     content.value = formatted;
-    draft.value = nextEditing && canEdit.value ? resolveObjectSourceEditDraft(formatted, editable) : "";
+    draft.value = nextEditing && canEdit.value ? resolveObjectSourceEditDraft(props.databaseType, resolvedType, formatted, editable) : "";
     editing.value = nextEditing && canEdit.value;
     if (nextEditing && !canEdit.value) {
       toast(t("objects.sourceReadOnly"), 3000);
@@ -131,7 +131,7 @@ function editSource() {
     if (!canEdit.value) toast(t("objects.sourceReadOnly"), 3000);
     return;
   }
-  draft.value = resolveObjectSourceEditDraft(content.value, editableText.value);
+  draft.value = resolveObjectSourceEditDraft(props.databaseType, resolvedObjectType.value, content.value, editableText.value);
   saveError.value = "";
   editing.value = true;
 }
--- a/apps/desktop/src/components/objects/ObjectSourceDialog.vue
+++ b/apps/desktop/src/components/objects/ObjectSourceDialog.vue
@@ -7,7 +7,7 @@ import { useConnectionStore } from "@/stores/connectionStore";
 import { useSettingsStore } from "@/stores/settingsStore";
 import { copyToClipboard } from "@/lib/common/clipboard";
 import { formatSqlForDisplay, type SqlFormatDialect } from "@/lib/sql/sqlFormatter";
-import { buildEditableObjectSource, buildExecutableObjectSourceStatements, executeObjectSourceSave, formatObjectSourceSaveError } from "@/lib/table/objectSourceEditor";
+import { buildEditableObjectSource, buildExecutableObjectSourceStatements, executeObjectSourceSave, formatObjectSourceSaveError, resolveObjectSourceEditDraft } from "@/lib/table/objectSourceEditor";
 import { loadObjectSourceWithRoutineFallback } from "@/lib/table/objectSourceLoad";
 import { xuguRoutineMetadataFromDefinition, type XuguRoutineMetadata } from "@/lib/table/routineParameters";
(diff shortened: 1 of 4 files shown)


t8y2 (commented):
已在 4d1323a 中把格式化 draft 限定为 MySQL VIEW；其他数据库和其他对象类型继续使用后端返回的 raw editable source，避免格式化文本改变原有编辑/保存语义。

验证：`pnpm exec vitest run apps/desktop/src/lib/__tests__/table/objectSourceEditor.spec.ts apps/desktop/src/components/objects/__tests__/ObjectSourceDialog.editDraft.spec.ts`（2 个文件、17 条测试通过），`pnpm typecheck` 和 `git diff --check` 通过。

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 2f2303e1d, will be released in the next version.

kenvqin:
From #5057 

> For (2), the "collapses to one line" part is a genuine dbx bug and is what this PR fixes. The "comments lost" part is not fixable at the application layer — MySQL's server never stores the original view text, so the comments are gone before dbx ever sees the definition (same `SHOW CREATE VIEW` output shown above). No tool can recover them.

Regarding the statement: "The 'comments lost' part is not fixable at the application layer — MySQL's server never stores the original view text, so the comments are gone before dbx ever sees the definition... No tool can recover them."

I would respectfully challenge the absolute conclusion that no tool can recover them. In my own experience, both HeidiSQL and Navicat are able to correctly display the original comments inside view definitions on MySQL 5.6.

Here's why: In MySQL 5.6 (and earlier versions), the server does in fact store the original view definition, including comments and formatting, inside the corresponding .frm file located in the server's data directory. Both HeidiSQL and Navicat use the LOAD_FILE() function to directly read that .frm file from the server, thereby retrieving the original SQL text that contains the comments.

In contrast, the standard SHOW CREATE VIEW command returns a normalized version of the view definition that discards comments and formatting, which is what the original author was referring to. But that does not mean no tool can recover them — it simply means tools that rely solely on SHOW CREATE VIEW cannot, while tools that read the .frm file directly can.

Important caveat: This method works for MySQL 5.6 and earlier. Starting from MySQL 8.0, the .frm file has been removed as the storage mechanism changed to a new data dictionary, so this recovery method will no longer work. But at least for MySQL 5.6, the statement that "no tool can recover them" is factually incorrect.

PR #6327, opened 2026-08-15
fix(backup): create destination directory before writing backup file

### LINKED ISSUE #6109: [Bug] 备份总是报错怎么搞
### 数据库类型和版本

mysql 8.0

### 问题描述

基本上备份不能用

### 支持信息

当前版本为 0.5.82。
macos 26.2 (25C56)

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

Scheduled/manual database backups fail unconditionally if the configured destination folder is missing at the moment a run executes (deleted, unmounted external/network drive, cloud-sync placeholder, etc.). `export_database_sql_core` opened the output file with `std::fs::File::create` and never ensured the parent directory existed, so every run against a since-removed destination fails with a raw OS error:

```
Failed to write file: No such file or directory (os error 2)
```

This matches the report that backups "basically don't work" (issue provided no further detail — no error message, no repro steps).

## Fix

Create the destination directory (`create_dir_all`) before opening the backup output file in `crates/dbx-core/src/database_export.rs`, mirroring the pattern already used elsewhere in this crate for file-writing paths (e.g. `agent_service.rs`, `agent_manager.rs`).

## Testing

Real reproduction against a real, independent MySQL 8.0 docker container (not the shared dev container) and the actual production `export_database_sql_core` function (not mocked):

- Added `live_mysql_database_export_creates_missing_destination_directory` to `crates/dbx-core/tests/live_mysql_database_export.rs`, pointing the export at a destination directory that is deliberately never created.
- Before the fix: test fails with `Failed to write file: No such file or directory (os error 2)`.
- After the fix: test passes — directory is created and the exported SQL file contains the seeded data.
- `cargo test -p dbx-core --lib database_export::` — 62/62 passing (no regressions).
- `cargo fmt --check` / `cargo clippy -D warnings` — clean.
- Rebased onto latest `upstream/main`, no conflicts.

Note: an unrelated pre-existing test in the same file (`live_mysql_database_export_restores_dependent_views`, a spatial-type JSON assertion) fails independently of this change on this environment — untouched code path, out of scope for this fix.

Fixes #6109

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1784,6 +1784,7 @@ dependencies = [
  "utoipa",
  "uuid",
  "webpki-roots 0.26.11",
+ "windows-sys 0.61.2",
  "winnow 1.0.4",
  "zip",
  "zstd",
--- a/agents/drivers/duckdb/Cargo.lock
+++ b/agents/drivers/duckdb/Cargo.lock
@@ -1263,6 +1263,7 @@ dependencies = [
  "unicode-width",
  "uuid",
  "webpki-roots 0.26.11",
+ "windows-sys 0.61.2",
  "winnow",
  "zip 4.6.1",
  "zstd",
--- a/apps/desktop/src/components/backup/ScheduledDatabaseBackupSettings.vue
+++ b/apps/desktop/src/components/backup/ScheduledDatabaseBackupSettings.vue
@@ -203,16 +203,12 @@ async function submitSchedule() {
   if (!canSave.value) return;
   saving.value = true;
   try {
+    await api.recordDatabaseExportDestination(draft.value.destinationDirectory);
     saveSchedule({
       ...draft.value,
       databases: allDatabases.value ? [] : [...selectedDatabases.value],
       tablePatterns: draft.value.tableFilterMode === "all" ? [] : normalizeDatabaseBackupTablePatterns(tablePatternsInput.value),
     });
-    // Best-effort: records the destination's filesystem identity as of
-    // configuration time so a mount that disappears before this schedule's
-    // very first run is refused instead of silently recreated elsewhere.
-    // Must not block saving the schedule itself on failure.
-    await api.recordDatabaseExportDestination(draft.value.destinationDirectory).catch(() => {});
     scheduleDialogOpen.value = false;
     toast(t(editingScheduleId.value ? "databaseBackup.scheduleUpdated" : "databaseBackup.scheduleCreated"), 2500);
   } catch (error: any) {
--- a/apps/desktop/src/components/backup/ScheduledDatabaseBackupSettings.vue
+++ b/apps/desktop/src/components/backup/ScheduledDatabaseBackupSettings.vue
@@ -208,6 +208,11 @@ async function submitSchedule() {
       databases: allDatabases.value ? [] : [...selectedDatabases.value],
       tablePatterns: draft.value.tableFilterMode === "all" ? [] : normalizeDatabaseBackupTablePatterns(tablePatternsInput.value),
     });
+    // Best-effort: records the destination's filesystem identity as of
+    // configuration time so a mount that disappears before this schedule's
+    // very first run is refused instead of silently recreated elsewhere.
+    // Must not block saving the schedule itself on failure.
+    await api.recordDatabaseExportDestination(draft.value.destinationDirectory).catch(() => {});
     scheduleDialogOpen.value = false;
     toast(t(editingScheduleId.value ? "databaseBackup.scheduleUpdated" : "databaseBackup.scheduleCreated"), 2500);
   } catch (error: any) {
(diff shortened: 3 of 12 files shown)


t8y2 (changes requested):
在写备份前无条件 `create_dir_all` 会给外接盘或挂载目录带来数据落错位置的风险，请不要按当前方式合并。

当配置的目标位于外接盘、网络盘或其他挂载点，而该挂载临时消失时，自动创建父目录可能直接在本机根文件系统上重建同名目录，随后备份会“成功”写入错误磁盘，用户也不容易察觉。这个风险比原来的明确失败更严重。

请区分首次配置的普通本地目录与已经配置的挂载目标：挂载目标不可用时应明确报错并停止，而不是自动创建占位目录。请补充模拟目标父路径消失/挂载不可用的测试，证明不会把备份写到错误位置。

q396921921:
Good catch, thanks for the detailed explanation — that's a real risk I hadn't considered. Pushed a follow-up commit that addresses it.

The new commit replaces the unconditional `create_dir_all` with `ensure_export_destination_dir`, which distinguishes the two cases you described:

- **A directory `export_database_sql_core` has never used before** (normal first-time configuration of a local folder) — still safe to auto-create, same as before.
- **A directory it has used successfully before** — its device id is recorded (via the existing `Storage` `state_store` KV table, keyed by the destination path) after every successful export. On the next run, if that same directory is now missing, or now resolves to a different filesystem device than last time, the export is refused with a clear error instead of being recreated:
  - `Backup directory <path> is missing. It existed during a previous export to this location, so dbx will not recreate it automatically -- if this is on a removable or network drive, reconnect it and try again.`
  - `Backup directory <path> now resolves to a different filesystem than the last successful export to this location. Refusing to write here automatically -- ...`

So a destination is only ever auto-created the very first time it's used; any run after that treats "directory missing" (or "directory now on a different device") as a hard stop, never a silent recreate.

Tests added:
- Two fast unit tests in `crates/dbx-core/src/database_export.rs` (`ensure_export_destination_dir_creates_a_never_before_seen_local_directory`, `ensure_export_destination_dir_refuses_to_recreate_a_destination_that_disappeared`) that simulate the destination directory disappearing after a prior successful use and assert both that the export call fails and that the directory is **not** resurrected on disk.
- A new live-MySQL integration test, `live_mysql_database_export_refuses_to_recreate_a_destination_that_disappeared`, exercising the same scenario through the public `export_database_sql_core` API end-to-end (export once successfully, delete the destination to simulate an unmounted drive, export again, assert failure and that the directory stays gone).

Ran all of these against a real local MySQL instance to confirm the behavior, in addition to the existing `live_mysql_database_export_creates_missing_destination_directory` test which still covers the legitimate first-time-use case from #6109.

t8y2 (changes requested):
Request changes: the backup destination can still silently move to the wrong filesystem.

1. Scheduled plans are stored in frontend localStorage, and the backend records directory identity only on the first export. If a configured mount disappears before that first run, `load_state` is empty and the code recreates the path on the underlying disk; the new live test currently treats this unsafe behavior as success.
2. Directory/device validation and `File::create` are separate operations. The mount can disappear or be replaced between them, so the checked path is not necessarily the opened destination.
3. The non-Unix path records no device identity at all, so replacing a Windows drive or mount at the same path is not detected.

Please record a stable destination identity when the schedule is configured, verify the actually opened target against that identity, and cover first-run mount loss, replacement between check/open, and Windows-equivalent behavior.

t8y2 (commented):
Maintainer patch pushed in f8813ffeb.

Root cause: the schedule was persisted before destination identity recording, recording failures were swallowed, missing destinations were treated as a successful no-op, and an opened file with an unknown identity was allowed even when the expected destination identity was known. Those paths could still save or write a backup without proving that the selected mount was the one actually opened.

Patch:
- record and persist the destination identity before saving the schedule, propagating failures to the existing error toast;
- reject a scheduled destination that disappeared or is no longer a directory before its identity is recorded;
- fail closed when a known expected identity cannot be obtained from the opened file handle;
- add frontend and Rust regression coverage for the failure paths.

Checks run:
- focused Vitest: 1 file, 6 tests passed;
- focused dbx-core tests: 3 tests passed;
- `cargo fmt --all -- --check`;
- `cargo clippy -p dbx-core --no-default-features --lib -- -D warnings`;
- `pnpm typecheck`;
- `git diff --check`.

t8y2 (commented):
Final maintainer patch pushed in f8d1fcf91, superseding the earlier intermediate patch head.

Root cause: the schedule was persisted before destination identity recording, recording failures were swallowed, missing destinations were treated as a successful no-op, and an opened file with an unknown identity was allowed even when the expected destination identity was known. Those paths could still save or write a backup without proving that the selected mount was the one actually opened.

Patch:
- record and persist the destination identity before saving the schedule, propagating failures to the existing error toast;
- reject a scheduled destination that disappeared or is no longer a directory before its identity is recorded;
- fail closed when a known expected identity cannot be obtained from the opened file handle;
- add frontend and Rust regression coverage for the failure paths;
- update DuckDB's independent Cargo lockfile for the PR's existing Windows dependency so the `agents` job can run with `--locked`.

Checks run:
- focused Vitest: 1 file, 6 tests passed;
- focused dbx-core tests: 3 tests passed;
- DuckDB CI-equivalent locked test: 35 unit tests and 10 worker-process tests passed;
- `cargo fmt --all -- --check`;
- `cargo clippy -p dbx-core --no-default-features --lib -- -D warnings`;
- `pnpm typecheck`;
- `git diff --check`.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 4ce27f8ae, will be released in the next version.

Issue #5905, opened 2026-08-11
[Bug] 设置自定义快捷键，更新后失效


Jasper-lc:
### 数据库类型和版本

MySQL 8.0

### 问题描述

1.复现步骤，将查看的DDL快捷键设置自己使用的，有了新版本的，点击更新版本，快捷键被改成默认了。 导航  -> 侧边栏打开方式   修改成双击打开，更新版本后，同样变成了默认 单机打开
2.希望改成，更新后，这些设置后的东西，不改成默认

### 支持信息

DBX 版本: v0.5.81
运行环境: 桌面端
操作系统: Windows 10.0.26100.4652
系统架构: x86_64

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

q396921921:
Following up here after investigating this further in #6329.

I audited the settings persistence path on the current `main` branch (SQLite storage, startup load, and the settings dialog's save path) and could not reproduce "an already-applied setting reverting to default after an update" on current code — including with a real desktop build (real Tauri process, real SQLite file) and a production-code-level reproduction of the closest race condition I could find, which turned out not to actually lose data.

Two historical bugs that match this exact symptom were fixed in v0.5.79 and v0.5.80 — both before v0.5.81, the version you reported. Would you be able to share:

- What version you were on when you first customized the "view DDL" shortcut / set sidebar activation to double-click?
- Whether this has happened again on a more recent version (v0.5.85+), and if so, the exact sequence of actions (e.g., did you close the settings dialog without clicking Apply at some point before the update)?

That would help pin down whether this is still an active bug or was fixed by the v0.5.79/v0.5.80 changes before you filed this issue.

zipg:
当前 main 已核对设置持久化、启动加载及设置保存路径，并通过真实桌面端与 SQLite 配置文件验证，未再出现升级后自定义快捷键或侧边栏打开方式恢复默认。该症状对应的历史持久化问题已在 v0.5.79/v0.5.80 修复，因此按当前版本已修复关闭；如 v0.5.95+ 仍能复现，请附上升级前后版本及完整操作步骤后重新打开。

PR #6329, opened 2026-08-15
fix(settings): confirm before discarding unsaved editor settings

q396921921:
## Problem

While investigating #5905 ("view DDL" shortcut and sidebar double-click activation reverting to defaults after a DBX update), I found that the Editor Settings dialog silently discarded any unapplied draft whenever it was closed without going through **Apply**/**Apply and Close** — Escape, clicking outside, the dialog's X button, or a tab's plain **Close** footer button all called `emit("update:open", false)` directly, with no check for unsaved changes.

**Scope note:** this is a real, independently-reproducible bug, but it is not confirmed to be the root cause of #5905's "settings reset after an update" report — see the investigation section below and the review thread.

## Root cause

The Editor Settings dialog holds every setting (shortcuts, sidebarActivation, and dozens of other fields tracked in `EDITOR_SETTINGS_DRAFT_KEYS`) in a local draft. The draft is only committed to the store — and thus persisted to disk — via the explicit **Apply** / **Apply and Close** buttons.

Closing the dialog any other way (Escape, clicking outside, the dialog's X button, or a tab's plain **Close** footer button) called `emit("update:open", false)` directly, with no check for unsaved changes. The entire draft was silently discarded, with zero warning.

Because this loss is invisible until the app restarts and the "reverted" values become visible, users who restart around the time they install an update could plausibly attribute the reset to the update itself — though this is not confirmed to be the actual cause of #5905 (see below).

## Fix

Route every close path (Escape/outside-click/X via `onSettingsRootOpenChange`, and every footer "Close" button via `closeSettings`) through a single guard, `shouldConfirmEditorSettingsDialogClose`. When there's an unapplied draft, closing now shows a confirmation dialog instead of discarding silently — the user can go back to editing or explicitly discard. This mirrors the unsaved-changes pattern already used elsewhere in this codebase (SQL tab close confirmation in `AppTabBar.vue`, the data transfer dialog's discard confirmation).

## Testing

- `apps/desktop/src/lib/settings/__tests__/editorSettingsDraft.spec.ts`: new tests for the extracted `shouldConfirmEditorSettingsDialogClose` decision function, including the exact shortcut/sidebarActivation scenario from the issue. Verified real failure against the pre-fix baseline (temporarily reverted the function to prove the regression tests catch it), then confirmed pass after restoring the fix.
- `apps/desktop/src/components/editor/__tests__/EditorSettingsDialog.unsavedClose.spec.ts`: new tests verifying the component actually wires all four close paths through the shared guard and renders the confirmation dialog. Verified real failure against the pre-fix component (temporarily reverted the wiring), then confirmed pass after restoring.
- Full suite: `pnpm typecheck` clean, `pnpm lint` clean (no new warnings), `pnpm vitest run apps/desktop/src` — 682/682 files, 6366/6366 tests passing.
- i18n keys added to all 8 locale files (en, es, it, ja, ko, pt-BR, zh-CN, zh-TW) to keep them in parity.

## Investigation into #5905's "lost after an update" report

I audited the full persistence path (SQLite `app_state` table in `crates/dbx-core/src/storage.rs`, the startup load in `settingsStore.ts`'s `initEditorSettings()`, and the draft-vs-base diff patch this dialog uses to commit changes) and could not reproduce settings being lost across a restart/update on current `main`:

- Two historical bugs matching this exact failure mode were already fixed before the reporter's stated version (v0.5.81): a startup read-failure being treated as "no saved settings" (`6c2ca2ead`, v0.5.79), and a startup write race where `updateEditorSettings()` calls landing before `initEditorSettings()` resolved could persist in-memory defaults over real data (`39399622e`, v0.5.80, guarded today by a `pendingEditorSettingsPatches` queue).
- I also traced whether this dialog's own local-draft snapshot (seeded from the store at dialog-open time) could cause the same class of loss on Apply. It can't: `persistSettings()` submits a diff between the draft and a same-moment base snapshot, so an untouched field is never included in the write patch even when the dialog opens mid-race.
- I built and ran the real desktop app end-to-end (real Tauri process, real SQLite file, isolated data dir), and separately reproduced the exact dialog-open-mid-race timing against the real `settingsStore.ts`/`editorSettingsDraft.ts` production code with only the IPC boundary mocked. In both cases the persisted setting survived.

I couldn't find a live reproduction of "already-applied setting lost after an update" on current code, so I'm scoping this PR down to the independent close-confirmation fix and removing the closing relationship to #5905. Left a comment on the issue asking the reporter for their version history to help pin this down further.

Related to #5905 (does not close it — see investigation notes above and the review thread)


t8y2 (changes requested):
这个改动可以防止用户主动关闭设置窗口时丢弃未应用草稿，但还没有解决原 Issue 描述的“升级后已经应用的设置丢失”。

当前实现只在对话框关闭路径增加二次确认，没有证明升级/启动期间持久化设置为何会被覆盖，也没有覆盖设置已保存、应用重启或版本升级后的恢复链路。因此现在使用关闭关键词会误表示原问题已经修复。

请先复现升级后丢失的路径，定位持久化读取、默认值初始化或迁移阶段的根因，并添加对应的重启/升级回归测试；如果本 PR 只希望保留关闭确认这一独立改进，请移除对原 Issue 的关闭关系并明确缩小 PR 范围。

q396921921:
Thanks for the review — you're right that this only addresses a real-but-separate bug, not confirmed as the root cause of the update/restart loss described in #5905.

I went back and audited the actual persistence/startup path on current `main` (SQLite `app_state` table, `initEditorSettings()`'s startup load, and the draft-vs-base diff-patch this dialog uses to commit):

- Two historical bugs matching this exact failure mode (startup read-failure treated as "no saved settings"; a pre-init write race that could persist in-memory defaults over real data) were already fixed in v0.5.79 (`6c2ca2ead`) and v0.5.80 (`39399622e`) — both before the reporter's stated v0.5.81.
- I traced whether this PR's own dialog could reintroduce the same class of loss via its local-draft snapshot, and confirmed it can't: `persistSettings()` submits a draft-vs-base diff, so an untouched field is never written even if the dialog opens before settings finish loading.
- I built and ran the real desktop app end-to-end (real Tauri process, real SQLite file, isolated data dir) and separately reproduced the exact dialog-open-mid-race timing against the production `settingsStore.ts`/`editorSettingsDraft.ts` code with only the IPC boundary mocked. In both cases the persisted setting survived a restart.

I couldn't find a live reproduction of "already-applied setting lost after an update" on current code. I've scoped this PR down to the independent close-confirmation improvement and removed the closing relationship to #5905 (updated PR description accordingly), and left a comment on #5905 asking the reporter for more detail on their version history, since the most likely explanation is that the corruption happened on a build older than v0.5.79 and only became visible on a later restart.

t8y2 (commented):
Rebased this branch onto the latest `main` and resolved the merge conflict mechanically.

The only conflict was in the `EditorSettingsDialog.vue` import block. The resolution preserves both sides without changing product behavior: `main` keeps the custom UI color symbols, while this PR keeps `shouldConfirmEditorSettingsDialogClose`. Both original PR commits remain separate; no squash or unrelated changes were introduced.

Validation on the rebased head:
- `editorSettingsDraft.spec.ts` and `EditorSettingsDialog.unsavedClose.spec.ts`: 64 tests passed
- `pnpm typecheck`: passed
- `git diff --check`: passed
- dependency/manifest scan: no dependency files changed
- `git range-diff`: only the expected `main` custom-theme import context differs

t8y2:
Thanks for the contribution! Merged in 6a25bd1909, will be released in the next version.

PR #6330, opened 2026-08-15
fix(ai): treat null account as unknown auth state for CodeBuddy Code CLI

### LINKED ISSUE #6253: [Bug] codebuddy code客户端已经登录后正常使用了，在dbx客户端的还是获取不到对应的模型
### 数据库类型和版本

mysql

### 问题描述

Image
codebuddy code已经在正常使用了，但是dbx客户端却测试失败

### 支持信息

Image
Image

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

Reported in #6253: a user with a working, already-authenticated CodeBuddy Code CLI still gets "can't get models" in DBX's AI provider settings.

## Root cause

Verified live against the real `@tencent-ai/codebuddy-code@2.136.0` CLI (installed and probed directly, not guessed): its `initialize` control response always reports `"account": null`, whether or not the session is actually signed in. Real "not authenticated" only ever surfaces later, on an actual chat turn (`"Authentication required. Please use /login command to sign in..."`), never in this handshake.

`parse_account_auth_state` in `crates/dbx-core/src/ai_codebuddy_cli.rs` hard-mapped `account: null` to "definitely not authenticated" (`Value::Null => Some(false)`), so `list_codebuddy_models`/`test_codebuddy_connection` refused to return any models — for every user, logged in or not. Every other CLI-based provider in this codebase (Claude Code, Cursor, etc.) instead falls back gracefully on discovery ambiguity; CodeBuddy was the one hard-failing outlier, apparently cloned from the Claude Code integration without validating the null-account case against the real protocol.

## Fix

Treat `account: null` as an unknown auth signal (`None`) instead of a hard `false`, letting the existing `unwrap_or(!models.is_empty())` fallback decide based on whether the model catalog was actually returned — matching the rest of the codebase's convention.

## Evidence

- New unit test `treats_null_account_as_authenticated_when_models_are_present`, built from a real captured response of the actual CLI. Reverted the fix and reran: test failed (`assertion failed: result.authenticated`), confirming a real repro. Reapplied: passes.
- End-to-end: ran DBX's actual `list_codebuddy_models` against the real installed CLI binary (unauthenticated sandbox). Before the fix: `[codeBuddyNotAuthenticated]`. After the fix: returns the model list successfully. The downstream effort-resolution call (which actually sends a turn) still correctly reports "Authentication required" in the same unauthenticated environment — proving this fix doesn't mask genuine auth failures, only the bogus one at the handshake step.
- `crates/dbx-core/src/ai_codebuddy_cli.rs` scoped test suite: 15 passed (3 pre-existing `#[ignore]`d live tests unaffected).
- `cargo fmt --check` / `cargo clippy -p dbx-core --lib -- -D warnings`: clean.

Fixes #6253

### DIFF
--- a/crates/dbx-core/src/ai_codebuddy_cli.rs
+++ b/crates/dbx-core/src/ai_codebuddy_cli.rs
@@ -728,7 +728,12 @@ fn parse_codebuddy_effort_level_strings(model: &Value) -> Vec<String> {
 
 fn parse_account_auth_state(account: Option<&Value>) -> Option<bool> {
     match account? {
-        Value::Null => Some(false),
+        // The real CodeBuddy Code CLI (verified against @tencent-ai/codebuddy-code
+        // 2.136.0) always reports `account: null` in this control response, whether
+        // or not the session is signed in — it's not a login signal here. Treat it
+        // as unknown so callers fall back to the model-catalog-based heuristic
+        // instead of hard-blocking authenticated users (t8y2/dbx#6253).
+        Value::Null => None,
         Value::Bool(value) => Some(*value),
         Value::Object(fields) => fields
             .get("authenticated")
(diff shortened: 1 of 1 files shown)


t8y2 (commented):
已在 316b704 中把 `account` 改为三态：`null` 仅表示未知，因此模型发现仍可使用公开目录；“测试连接”则会发起一次最小真实模型请求，只有实际鉴权成功才返回成功，不能再由模型目录推断登录状态。

验证：CodeBuddy 模块 16 条测试通过（3 条 live 测试按预期忽略），覆盖“null account 可发现模型但鉴权探测失败”和已登录探测 marker；`cargo check --no-default-features`、格式检查和 `git diff --check` 通过。性能边界：只在用户显式点击测试连接时增加一次有界请求，不影响模型列表、启动或常规消息路径。

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 6988f9878, will be released in the next version.

PR #6332, opened 2026-08-15
fix(ai): cancel in-flight AI stream before clearing or switching chat

### LINKED ISSUE #5941: [Bug] AI助手使用时，当mcp工具调用失败时，无法继续聊天，且之前的聊天记录会清除
### 数据库类型和版本

mysql8.4

### 问题描述

1.打开AI助手
2.输入：查询库xxx中表xxx字段值为xxx的记录
3.然后会显示 mcp工具调用失败，然后，聊天窗口卡住，所有记录清除了

### 支持信息

Version 0.5.77 (0.5.77)
Desktop MacOS

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Summary

Fixes #5941 — when an AI-assistant request is still in flight (e.g. a stalled MCP tool call / slow provider response), clicking the clear-chat trash icon, "New Chat", deleting the active conversation, or switching to a different saved conversation wiped `messages`/`conversationId` **without** checking `isGenerating` or cancelling the in-flight request. `isGenerating` is only ever reset in `send()`'s own `finally` block, so the send box stayed disabled until the abandoned request eventually settled on its own — for a stalled response, that could be indefinite. This matches the report exactly: chat history gets cleared, but the user still can't continue chatting.

- `clearMessages()` and `selectConversation()` now cancel any in-flight stream first (`if (isGenerating.value) cancelStream();`). `startNewChat()` / `deleteConversation()` already funnel through `clearMessages()`, so they're covered automatically.
- Also guarded the assistant-message lookup in `send()`'s `catch` block the same way its `finally` block already does — an unguarded `messages.value[assistantIdx].content = ...` would throw if the array had already been replaced mid-flight by one of the actions above, silently swallowing the real error.

## Root cause

`messages` / `isGenerating` are local component refs in `AiAssistant.vue`, not a store. Nothing outside `send()`'s own `finally` resets `isGenerating`, so any action that abandons the in-flight turn without cancelling it strands the send box disabled.

## Reproduction

Real end-to-end repro: isolated `dbx-web` dev backend + `pnpm dev:web` + a throwaway SQLite connection + a custom/OpenAI-compatible AI provider pointed at a local stub server that accepts the request and sends real SSE headers but then stalls mid-stream (matching a stuck MCP tool call). Driven via headless Chromium.

**Before fix**: sending a message → stuck "thinking" (stop button shown) → clicking clear → chat history empties, but the stop button *stays* instead of the send button, and typing+sending a new message does nothing (silently dropped, since `send()` bails on `isGenerating`).

**After fix**: same steps → clicking clear cancels the stream, the send button reappears within ~200ms, and the next message is actually accepted and sent.

## Test plan

- [x] New regression test `apps/desktop/src/components/editor/__tests__/AiAssistant.clearAndCancel.spec.ts` (source-text assertions, matching this component's existing `AiAssistant.messageCopy.spec.ts` convention — the component pulls in too many stores/composables to cheaply full-mount). Verified revert→fail (3/4 fail against pre-fix code) → reapply→pass (4/4).
- [x] Updated a pre-existing test (`packages/app-tests/aiMessageLayout.test.ts`) that was pinned to the old unguarded line.
- [x] Full suite: `pnpm exec vitest run` — 975/975 files, 9295/9295 tests passed.
- [x] `pnpm run typecheck` (vue-tsc) — clean.
- [x] `pnpm run lint` (oxlint) — clean (13 pre-existing warnings, none in touched files).

### DIFF
--- a/apps/desktop/src/components/editor/AiAssistant.vue
+++ b/apps/desktop/src/components/editor/AiAssistant.vue
@@ -57,6 +57,7 @@ import { useToast } from "@/composables/useToast";
 import { useNavigationTargets } from "@/composables/useNavigationTargets";
 import { buildAiContext, resolveAiDatabaseTarget, resolveAiNamespaceSelection, resolveDefaultAiSchema, runAgentStream, isVectorDbType, isValidActionForMode, defaultActionForMode, type AiAction, type AiAssistantMode, type AiSqlFileContext, type CustomPromptContext } from "@/lib/ai/ai";
 import { isAiConfigModelCandidate } from "@/lib/ai/aiConfigCandidates";
+import { deleteConversationWithCancellation, stopAiGenerationWithFallback } from "@/lib/ai/aiConversationLifecycle";
 import { AiGenerationGuard } from "@/lib/ai/aiGenerationGuard";
 import { isTauriRuntime } from "@/lib/backend/tauriRuntime";
 import { addConfiguredAiModel, aiModelOptions } from "@/lib/ai/aiConfigList";
@@ -2034,64 +2035,23 @@ function waitForGenerationToClear(timeoutMs: number): Promise<void> {
 }
 
 async function cancelStream() {
-  // Plain "stop generating" (button click): the SAME conversation stays
-  // current, so send()'s own finally — once the backend actually acknowledges
-  // the cancellation — still owns cleanup normally (write final content,
-  // build agent steps, persistConversation()).
-  if (!isGenerating.value) return;
-  // Snapshot the generation (not currentSessionId) this click is stopping.
-  // Reviewed on PR #6332: gating everything below on `sessionId` truthiness
-  // used to make this whole function a no-op if Stop was clicked before
-  // send() reached `currentSessionId.value = sessionId` — which sits after a
-  // real await, so it's a reachable window, not a theoretical one. That left
-  // isGenerating stranded exactly like issue #5941's original bug, just via
-  // the Stop button instead of clear/switch. The generation id is set before
-  // send()'s first await, so peek()/isCurrent() work regardless of whether a
-  // session has been registered with the backend yet.
-  const myGeneration = aiGenerationGuard.peek();
-  const assistantIdx = currentAssistantMessageIndex;
-  const sessionId = currentSessionId.value;
-  if (sessionId) {
-    await aiCancelStream(sessionId).catch(() => {});
-  }
-  // The backend cancel RPC only *requests* cancellation (a tokio::sync::Notify
-  // that a hung tool call isn't listening for while it's actually executing —
-  // see crates/dbx-core/src/agent_loop.rs), so a genuinely stuck request
-  // (issue #5941's original repro: a hung MCP tool call) may never let
-  // send()'s finally run on its own. Give the backend a bounded grace period
-  // to stop normally; if it hasn't by then, force the same abandon path
-  // clear/switch uses, so Stop — the most visible "unstick" affordance in the
-  // UI — can never leave isGenerating stranded, even though the backend may
-  // keep running the hung call in the background. (The real fix is backend
-  // -side: agent_loop.rs's tool-execution await isn't raced against the
-  // cancellation notify the way ai_cli_agent.rs's subprocess path already is.)
-  await waitForGenerationToClear(STOP_FORCE_ABANDON_MS);
-  // Re-check the generation, not just isGenerating: a newer send() may have
-  // legitimately started (and be currently generating) during the grace
-  // period if the user cancelled and immediately sent a new message — that
-  // request must not be torn down by this stale Stop click.
-  if (isGenerating.value && aiGenerationGuard.isCurrent(myGeneration)) {
-    // The backend never acked in time. Flush and finalize the placeholder
-    // assistant message BEFORE abandoning: abandonInFlightRequest() only
-    // resets shared request state, it has no notion of "this generation's
-    // message" (it's also used by clear/switch/unmount, where messages.value
-    // is discarded/replaced right after, so there's nothing to finalize
-    // there). Without this, the message send() pushed stays permanently
-    // stuck mid-"thinking" in the still-current conversation — persisted as
-    // empty content on the next turn and fed to the LLM as empty history.
-    if (assistantDeltaFrame !== null) cancelAnimationFrame(assistantDeltaFrame);
-    flushAssistantDeltas();
-    const msg = messages.value[assistantIdx];
-    if (msg) {
-      msg.isThinking = false;
-      if (!msg.content) msg.content = t("ai.requestCancelled");
-    }
-    // Pass the session id already RPC'd above (possibly "") so
-    // abandonInFlightRequest() doesn't fire a redundant second cancel RPC for
-    // the same session — but still fires one if send() only registered a
-    // session AFTER this click (sessionId was "" here but isn't anymore).
-    abandonInFlightRequest(sessionId);
-  }
+  await stopAiGenerationWithFallback({
+    isGenerating: () => isGenerating.value,
+    currentGeneration: () => aiGenerationGuard.peek(),
+    isGenerationCurrent: (generation) => aiGenerationGuard.isCurrent(generation),
+    currentSessionId: () => currentSessionId.value,
+    cancelSession: aiCancelStream,
+    waitForGenerationToClear: () => waitForGenerationToClear(STOP_FORCE_ABANDON_MS),
+    flushPending: () => {
+      if (assistantDeltaFrame !== null) cancelAnimationFrame(assistantDeltaFrame);
+      flushAssistantDeltas();
+    },
+    currentAssistantMessageIndex: () => currentAssistantMessageIndex,
+    messageAt: (index) => messages.value[index],
+    cancelledMessage: () => t("ai.requestCancelled"),
+    abandon: (sessionId) => abandonInFlightRequest(sessionId),
+    persistConversation,
+  });
 }
 
 // Neutralizes all per-request transient state that must never survive into a
@@ -2291,9 +2251,17 @@ function selectConversation(conv: AiConversation) {
 }
 
 async function deleteConversation(id: string) {
-  await deleteAiConversation(id).catch(() => {});
-  conversations.value = conversations.value.filter((c) => c.id !== id);
-  if (conversationId.value === id) clearMessages();
+  await deleteConversationWithCancellation({
+    id,
+    currentConversationId: () => conversationId.value,
+    isGenerating: () => isGenerating.value,
+    abandon: () => abandonInFlightRequest(),
+    deletePersisted: () => deleteAiConversation(id).catch(() => {}),
+    afterDelete: () => {
+      conversations.value = conversations.value.filter((c) => c.id !== id);
+      if (conversationId.value === id) clearMessages();
+    },
+  });
 }
 
 function startNewChat() {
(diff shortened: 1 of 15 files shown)


t8y2 (changes requested):
当前修复还不能可靠隔离被取消的旧请求和新会话，请补充后再合并。

`cancelStream()` 依赖当前流已经完成注册；如果用户在流 ID 注册前清空或切换会话，取消仍可能成为空操作。旧 `send()` 随后继续执行时，也仍可通过增量回调、`catch` 或 `finally` 更新已经切换后的共享 `messages` / `isGenerating` 状态，从而污染新会话或提前结束新请求的生成状态。

请为每次发送引入独立的 generation/request identity，并让所有异步回调、增量刷新、错误处理和 `finally` 只更新仍属于自己的那一代状态；清空、切换和新建会话时使旧 generation 失效。请补充至少覆盖“流注册前取消”和“取消后立即发起新请求，旧请求随后完成”的竞态测试。

q396921921:
Thanks for the detailed review — you're right, `cancelStream()` alone can't guarantee isolation. Pushed a follow-up commit that adds a per-`send()` generation identity:

- **`lib/ai/aiGenerationGuard.ts`** (new): `AiGenerationGuard` — `begin()` claims a new generation id, `invalidate()` supersedes whatever is current, `isCurrent(id)` lets an async callback check it still owns shared state.
- **`send()`** claims `myGeneration` right after setting `isGenerating = true` (before the first `await`). Every place it later touches shared state — after the `ensureLoaded()` await, inside the agent-event callback, in `catch`, and in `finally` — now checks `aiGenerationGuard.isCurrent(myGeneration)` first and bails/no-ops otherwise.
- **`abandonInFlightRequest()`** (new, used by `clearMessages()`/`selectConversation()`, and transitively by `startNewChat()`/`deleteConversation()`): invalidates the generation and resets `isGenerating`/`currentSessionId`/the streaming delta buffers **synchronously**, before the best-effort `aiCancelStream()` RPC even runs. This is what actually closes the "cancel before stream registration" gap — invalidation no longer depends on the backend call having a session id to act on.
- `cancelStream()` (the plain stop-generating button / unmount) is intentionally left as just the backend RPC — that path doesn't switch conversations, so `send()`'s own `finally` (still the *same*, still-current generation) keeps owning cleanup once the backend acks, same as before.

This directly covers both races you called out:
1. **Cancel before registration**: `abandonInFlightRequest()` invalidates unconditionally and before the RPC call, so a `send()` that hasn't set `currentSessionId` yet is still cut off.
2. **Cancel → immediately send new → old resolves later**: the old generation's id becomes stale the instant `abandonInFlightRequest()`/a newer `begin()` runs; when the old `runAgentStream()` eventually settles, its `catch`/`finally` see they're no longer current and skip all writes.

Test coverage added:
- `lib/__tests__/ai/aiGenerationGuard.spec.ts` — exercises both races directly with real async/Promise ordering (not string matching): invalidating mid-await before "registration", and an old generation resolving after a newer one has already started.
- `components/editor/__tests__/AiAssistant.clearAndCancel.spec.ts` — extended to pin that `AiAssistant.vue` actually wires the guard at every call site (claim point, post-await recheck, event-callback guard, catch/finally guards, `abandonInFlightRequest()`'s ordering, and that the plain stop button intentionally does *not* force-invalidate).

Verified locally: `pnpm run typecheck` clean, `oxlint` clean (same 12 pre-existing warnings, none in touched files), `oxfmt --check` clean, full `vitest run` — 976/976 files, 9304/9304 tests passed.

t8y2 (changes requested):
Request changes: the generation guard prevents some stale writes, but the old request can still start or leak state after the chat is cleared/switched.

1. After awaiting referenced-file and AI-context loading, the code calls `runAgentStream()` without rechecking the generation. Clearing or switching during those awaits sends a cancel before a stream exists, then the stale request can still start and execute tools or authorized SQL.
2. `pendingCompaction` is not reset by the abandon/clear path. A compaction produced by the old conversation can therefore be inserted into the next conversation.
3. `onUnmounted()` calls only `cancelStream()` and does not invalidate the generation. If the backend stream is not registered yet, the request can still start and later update retained component state.

Please invalidate and recheck before stream creation, clear all request-owned pending state, and make unmount follow the same generation-invalidating path. Add races for clear/switch/unmount during context preparation and after `context_compacted`.

q396921921:
Thanks — pushed a follow-up commit that closes all 3 gaps:

1. **`runAgentStream()` recheck**: `send()` now rechecks the generation right after `loadReferencedSqlFiles()` and again right after `buildAiContext()`, in addition to the existing check after `ensureLoaded()`. A clear/switch/unmount that fires during either await now bails before `runAgentStream()` is ever called, instead of only bailing before writing the results.
2. **`pendingCompaction` leak**: consolidated all per-request transient state (`pendingCompaction`, the streaming delta buffers) into one `resetPendingRequestState()`, and `abandonInFlightRequest()` now calls it. A stale compaction summary can no longer get spliced into a newly-started conversation.
3. **`onUnmounted()`**: now calls `abandonInFlightRequest()` (same invalidating path as clear/switch) instead of the plain `cancelStream()` RPC-only call, so a request still mid-await when the component unmounts can't keep writing into the now-gone instance's closures.

While auditing these same code paths I found and fixed two closely related issues in scope:

- The `ensureLoaded()` supersede-check in `send()` was returning without clearing `allowWriteSqlForNextRun` — a confirmed write-SQL grant could leak into an unrelated `send()` in the next conversation. Now cleared on that path too.
- `cancelStream()` (the Stop button) previously relied entirely on the backend acking the cancel RPC before `isGenerating` cleared — under the exact failure mode from issue #5941's original repro (a hung MCP tool call), Stop could get stuck the same way clear/switch used to. It now waits a bounded grace period (`STOP_FORCE_ABANDON_MS` = 5s) and falls back to the same abandon path if the backend hasn't stopped by then, re-checking `currentSessionId` so it doesn't tear down a newer request the user started during that window.

Extended `AiAssistant.clearAndCancel.spec.ts` with cases for each of the above (recheck-after-each-await ordering, `resetPendingRequestState()`/`pendingCompaction` wiring, write-grant clearing, `onUnmounted()` using `abandonInFlightRequest()` and not `cancelStream()`, and the Stop-button grace-period-then-abandon sequence).

Verified locally (Node 22): `pnpm run typecheck` clean, `oxlint` clean (same 13 pre-existing warnings, none in touched files), `oxfmt --check` clean, full `vitest run` — 976/976 files, 9308/9308 tests passed.

q396921921:
Thanks — all 3 confirmed and fixed in a follow-up commit:

1. **Stuck placeholder message** (`AiAssistant.vue:2073`): `abandonInFlightRequest()` never touched `messages.value` on purpose (it's also used by clear/switch/unmount, where the array is discarded/replaced right after, so there's nothing to finalize there). But `cancelStream()`'s forced-abandon path keeps the same conversation current, so the empty "thinking" placeholder `send()` pushed was left stuck forever — never marked done, later persisted with empty content, and fed to the LLM as empty history on the next turn. Added a `currentAssistantMessageIndex` (tracked alongside `currentSessionId`) so `cancelStream()` can flush any pending delta and finalize that specific message — clearing `isThinking`, and only falling back to a new `ai.requestCancelled` string if nothing ever streamed (so partial content that was already showing isn't clobbered) — right before calling `abandonInFlightRequest()`.

2. **Stop no-op before session registration** (`AiAssistant.vue:2031`): confirmed — `send()` only assigns `currentSessionId` after `ensureLoaded()`'s await, so gating the whole function on `sessionId` truthiness made Stop a complete no-op in that window: no RPC, no wait, no abandon, reproducing #5941's stuck-`isGenerating` bug via the Stop button itself instead of clear/switch. Added `AiGenerationGuard.peek()` so `cancelStream()` can snapshot and later recheck the generation independent of whether a session id exists yet, and switched the entry guard to `isGenerating.value` instead of `sessionId`.

3. **Duplicate cancel RPC** (`AiAssistant.vue:2094`): `abandonInFlightRequest()` now takes the session id the caller already sent `aiCancelStream` for (if any) and skips re-firing it — while still firing if `send()` only registered a session *after* the Stop click (so the "not yet registered at click time" case from #2 above still gets cancelled at the backend once it exists).

Extended `AiAssistant.clearAndCancel.spec.ts` and `aiGenerationGuard.spec.ts` with cases for all three (finalizing the placeholder, `peek()`-based generation check replacing the `sessionId`-gated no-op, and the RPC dedup).

Verified locally (Node 22): `pnpm run typecheck` clean, `oxlint` clean (same 12 pre-existing warnings, none in touched files), `oxfmt --check` clean, i18n autofill check clean (added `ai.requestCancelled` to all locales), full `vitest run` — 976/976 files, 9315/9315 tests passed.

t8y2 (commented):
Maintainer patch pushed in `2aa557a8881b`.

Root causes:

- deleting the active conversation waited for persistence deletion before abandoning the running generation;
- forced Stop captured the assistant-message index before a placeholder necessarily existed;
- forced abandonment invalidated the generation before the normal `finally` persistence path, so the finalized stopped turn could be lost.

Patch:

- abandon the active generation synchronously before the first delete await;
- read the latest assistant index only after confirming generation ownership in the forced-stop path;
- persist the finalized conversation after forced abandonment;
- add behavioral asynchronous race coverage for delete-during-generation, early Stop, and forced-stop persistence.

Checks:

- 2 focused Vitest files, 18/18 tests passed;
- Oxlint passed with zero issues;
- Oxfmt and `git diff --check` passed;
- no manifest or lockfile changes.

Flags: `large-ui-change=true` because this affects shared AI send/stop/delete lifecycle; no new dependency; no polling, listener, background-I/O, or database-request increase.

q396921921:
这次的重构（`2aa557a88`）验证清单里提到了 vitest/oxlint/oxfmt，但漏跑了 typecheck——CI 的 `CI / frontend` job 在 typecheck 阶段挂了：

```
apps/desktop/src/components/editor/AiAssistant.vue(2043,5): error TS2322:
  Type '(sessionId: string) => Promise<boolean>' is not assignable to type '(sessionId: string) => Promise<void>'.
```

根因：`aiCancelStream()`（`lib/backend/http.ts:1458` / `tauri.ts:527`）返回 `Promise<boolean>`，但新抽的 `aiConversationLifecycle.ts` 里 `StopAiGenerationOptions.cancelSession` 字段类型是 `(sessionId: string) => Promise<void>`，`cancelStream()` 里直接把 `aiCancelStream` 传给这个字段，类型不匹配。

已在 `2bfe172dd` 修复：在 `cancelStream()` 调用点把 `cancelSession` 包一层 `.then(() => undefined)`，没有改动 `aiConversationLifecycle.ts` 的公共类型契约（影响面最小）。

验证（Node 22.13.0）：
- 修复前先本地复现了同一个 typecheck 报错，逐字比对与 CI 日志一致。
- 修复后 `pnpm check`（format/lint/typecheck/test 一起跑，同 CI 命令）：format ok、lint ok（12 条同样的 pre-existing 警告，不在改动文件内）、typecheck ok、`pnpm exec vitest run` 977/977 文件、9315/9315 测试通过（首次全量跑时 `DocumentBrowserFieldSearch.spec.ts` 里一条无关用例偶发失败，单独重跑和整体重跑均通过，确认是 flaky，与本次改动无关）。

t8y2 (approved):
The previously requested generation-isolation and cancellation fixes are complete on the current head.

The active generation is invalidated before destructive conversation operations, stale callbacks are guarded from shared state, forced Stop handles the placeholder race, and the final `Promise<void>` contract correction is present. Exact-head frontend CI passed.

t8y2:
Thanks for the contribution! Merged in 3e6e43c57, will be released in the next version.

PR #6334, opened 2026-08-15
fix(xugu): place public synonyms after schemas

mapan0424:
## Summary

This change refines the Xugu public-synonym schema-tree presentation.

## Why

Xugu exposes database-global public synonyms through a reserved metadata scope rather than through an owning schema. The existing tree placed that synthetic scope according to normal name sorting, which could make `Public synonyms` appear between real schemas and make it look like an ordinary schema. In addition, expanding that scope reused the complete database object profile, so it could show empty or irrelevant groups such as tables and views.

The result was confusing for users and could also reuse persisted children created before the public-synonym scope had its dedicated routing rules.

## What changed

- Keep real schemas on the normal natural-name ordering and place the Xugu `Public synonyms` scope after all real schemas.
- Preserve a real `GUEST` schema as an ordinary schema; it is not used as a public-synonym alias.
- Restrict the synthetic Xugu scope to the `SYNONYM` object type.
- In grouped mode, render only the Synonyms group for that scope.
- In simple mode, skip the table-list request for the synthetic scope and load only its synonym objects.
- Bump the schema/object cache namespaces so stale children cannot restore the old ordering or unrelated object groups.

## Compatibility and scope

The special handling is guarded by the Xugu database type and the exact reserved public-synonym scope. Ordinary Xugu schemas and all non-Xugu database providers retain their existing sorting, object discovery, and cache behavior.

## Validation

- `pnpm exec vitest run apps/desktop/src/lib/sidebar/__tests__/xuguPublicSynonyms.spec.ts apps/desktop/src/lib/__tests__/table/tableTree.spec.ts`
  - 30 tests passed.
- `pnpm typecheck` passed.
- Targeted `oxfmt --check` and `git diff --check` passed.
- `go test ./...` in `agents/drivers/xugu` passed.
- Project check completed format, lint, and typecheck successfully; the broad test run passed 9,251 tests. The only unrelated failures were the existing documentation-export smoke test timing out during Rust example startup and one pre-existing document-browser test environment failure.

## User impact

Users will see `Public synonyms` as a clearly separated database-global namespace at the bottom of the schema list, and expanding it will show only public synonyms. No changes are made to Xugu agent SQL or to other database providers.

### DIFF
--- a/apps/desktop/src/lib/__tests__/table/tableTree.spec.ts
+++ b/apps/desktop/src/lib/__tests__/table/tableTree.spec.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { appendTableTreeLoadMoreNode, buildGroupedObjectTreeNodes, buildSimpleObjectTreeNodes, buildTableTreeNodes, mergeTableInfosIntoObjects, mergeTableTreePageChildren, tablePartitionGroups, withoutTableTreeLoadMoreNodes } from "@/lib/table/tableTree";
+import { appendTableTreeLoadMoreNode, buildGroupedObjectTreeNodes, buildObjectGroupPlaceholderNodes, buildSimpleObjectTreeNodes, buildTableTreeNodes, mergeTableInfosIntoObjects, mergeTableTreePageChildren, tablePartitionGroups, withoutTableTreeLoadMoreNodes } from "@/lib/table/tableTree";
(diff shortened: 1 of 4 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 5bfb07e, will be released in the next version.

PR #6336, opened 2026-08-15
fix(redis): echo console command immediately instead of after response

### LINKED ISSUE #5856: [Bug] redis 的命令行界面只展示一句“欢迎使用 Redis 命令行控制台”但是输入的命令不展示，实际是输入了的，点击回车后命令才展示出来
### 数据库类型和版本

redis

### 问题描述

redis 的命令行界面只展示一句“欢迎使用 Redis 命令行控制台”但是输入的命令不展示，实际是输入了的，点击回车后命令才展示出来

Image

### 支持信息

版本：v0.5.81
系统：macOS 14.6.1 aarch64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

In the Redis command console (Command line tab), `runRedisCommand()` only appended the prompt/command/output to the terminal history *after* the backend response resolved. On any connection with real round-trip latency, pressing Enter would clear the input immediately but nothing appeared in the terminal until the response came back — making it look like the typed command was lost. On a fast local Redis this gap is imperceptible, which is likely why it wasn't caught earlier.

## Reproduction

Real end-to-end repro: real `dbx-web` dev backend + a real Redis instance, driven via Playwright with CDP `Network.emulateNetworkConditions` (800ms latency) to simulate a real round trip.

- Type `SET reproTestKey helloWorld`, press Enter.
- Screenshot ~50ms later (before the response resolves): the terminal shows only the static welcome line — the command is completely invisible.
- After the fix: the command echoes to the terminal immediately on submit; the result (`OK`) fills into the same entry once the response arrives (no duplicate line).

## Solution

- `appendCommandHistory()` now returns the new entry's id.
- `runRedisCommand()` echoes the command to the terminal synchronously on submit (before the `await`), then patches that same entry's `output`/`error` via a new `updateCommandHistory()` helper once the result or error arrives, instead of only pushing a history entry after the request resolves.

No behavior change on fast/local connections, since both the echo and the result land within the same tick there.

## Testing

- Added 2 regression tests in `RedisKeyBrowser.expiry.spec.ts` using a deferred (controllable) mock of `redisExecuteCommand`, asserting the command text is present in the terminal DOM *before* the mock resolves, and that success/error results patch the same entry rather than adding a second one.
- Reverted the fix and reran: both new tests genuinely failed (`expected 'redis.commandWelcome' to contain 'PING'`). Reapplied and reran: both passed.
- Full suite after fix: `apps/desktop/src` — 681 files / 6360 tests passing.
- `pnpm typecheck` clean, `pnpm lint` clean (no new warnings in the touched file).
- Rebased onto latest `upstream/main`, no conflicts, retested clean.

Fixes #5856

### DIFF
--- a/apps/desktop/src/components/redis/RedisKeyBrowser.expiry.spec.ts
+++ b/apps/desktop/src/components/redis/RedisKeyBrowser.expiry.spec.ts
@@ -844,6 +844,58 @@ describe("RedisKeyBrowser command completion", () => {
   });
 });
 
+describe("RedisKeyBrowser command console echo", () => {
+  it("echoes the submitted command to the terminal before the response arrives, then fills in the result", async () => {
+    const pending = deferred<{ value: unknown }>();
+    mocks.redisExecuteCommand.mockReturnValueOnce(pending.promise);
+    mountBrowser();
+    await settle();
+    await openCommandPanel();
+    await setCommandInput("PING");
+
+    const input = requiredElement<HTMLInputElement>("[data-redis-command-input]");
+    input.dispatchEvent(new KeyboardEvent("keydown", { key: "Enter", bubbles: true }));
+    await settle();
+
+    // The command must show up in the terminal right away — while the
+    // request is still in flight — not only once the response resolves.
+    const terminal = requiredElement<HTMLElement>(".redis-command-terminal");
+    expect(terminal.textContent).toContain("PING");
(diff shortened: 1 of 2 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 60e685d, will be released in the next version.

PR #6337, opened 2026-08-15
fix(sqlserver): fold connection-lock wait time into reported query duration

### LINKED ISSUE #6097: [Bug] SQL Server 查询结果展示缓慢，耗时显示不一致
### 来源
QQ 群反馈 · **反馈人**：	☆DJH☆

### 数据库类型和版本
SQL Server

### AI 摘要
查询耗时约10秒，但结果摘要显示仅6-8毫秒，结果展示明显缓慢，耗时统计与实际不符。

### 原始描述
经常出现 sql server 查询结果展示很慢，查询 10s，结果摘要显示耗时 6-8ms。见图。

q396921921:
Fixes #6097

## Problem

Users report SQL Server queries that visibly take ~10 seconds, while the Summary panel reports a duration of only 6-8ms.

## Root cause

Unlike MySQL/Postgres, which hand out independent connections from a real pool (`sqlx`/`deadpool_postgres`), SQL Server connections in dbx are serialized behind a single `Arc<Mutex<SqlServerClient>>` per pool key (`crates/dbx-core/src/connection.rs`). That same connection is also shared with metadata operations (autocomplete, schema-tree listing) on native SQL Server connections, since `pool_key_for_session_role` only splits off a separate metadata pool key for agent-type drivers, not for native SQL Server.

The per-query `execution_time_ms` timer in `crates/dbx-core/src/db/sqlserver.rs` (`execute_query_with_max_rows` / `execute_batch_with_max_rows_metadata`) only starts *after* the connection's mutex is acquired. So if a query has to queue behind another in-flight operation on the same connection (e.g. the sidebar populating table/column metadata), that queueing time is real wall-clock latency the user experiences, but is completely invisible to the reported duration — producing exactly the "10s felt, 6-8ms shown" mismatch reported in the issue.

I verified with a live MySQL container that the general duration-measurement → Summary-panel UI wiring itself is sound and DB-agnostic (a genuinely slow query correctly reports a large duration), which narrows the bug specifically to this SQL-Server-only lock-before-timer ordering.

## Fix

Added `lock_shared_client_with_wait` in `crates/dbx-core/src/query.rs`, a small helper that times the connection-lock acquisition and returns the wait alongside the guard. Both SQL Server call sites (single-statement execute and the multi-statement batch path) now use it and fold the measured wait into `execution_time_ms`, so the reported duration reflects what the user actually waited for.

## Testing

- Added 2 regression tests exercising the new helper directly:
  - a contended case (another holder keeps the mutex locked for 150ms) asserts the wait is captured (`wait_ms >= 150`)
  - an uncontended case asserts the reported wait stays negligible (`wait_ms < 50`)
- Verified before/after via `git apply -R` / `git apply` on the same patch: without the fix, `lock_shared_client_with_wait` doesn't exist and the new tests can't build; with the fix, both pass.
- `query::` module: 140/140 tests pass, no regressions.
- `cargo fmt --check` / `cargo clippy -p dbx-core --lib -- -D warnings`: clean.
- Full-crate `cargo test -p dbx-core --lib`: one pre-existing, unrelated failure (`nacos::http::tests::v3_connection_check_...`, a `502 Bad Gateway` from a real Nacos endpoint in this environment) confirmed unrelated to `query.rs`/`sqlserver.rs`.

I wasn't able to reproduce the exact reported scenario against a live SQL Server in this environment (arm64 Docker images for SQL Server have been unreliable here), so this fix is based on tracing the actual connection-pooling/locking code path rather than an end-to-end repro. Happy to adjust if maintainers can verify against a real SQL Server instance under concurrent load.

### DIFF
--- a/crates/dbx-core/src/query.rs
+++ b/crates/dbx-core/src/query.rs
@@ -1521,6 +1521,22 @@ where
     }
 }
 
+/// Locks a mutex-guarded shared connection (e.g. SQL Server's single connection
+/// per pool key) and reports how long the caller waited for the lock alongside
+/// the guard. Callers should fold the returned wait time into any execution-time
+/// metric they report, since a driver-level timer that only starts once the lock
+/// is held cannot see time spent queued behind another operation on the same
+/// connection.
+async fn lock_shared_client_with_wait<'a, T>(
+    client: &'a Arc<tokio::sync::Mutex<T>>,
+    cancel_token: Option<CancellationToken>,
+    timeout_duration: Option<Duration>,
+) -> Result<(tokio::sync::MutexGuard<'a, T>, u128), String> {
+    let started_at = std::time::Instant::now();
+    let guard = wait_for_value_opt(cancel_token, timeout_duration, client.lock()).await?;
+    Ok((guard, started_at.elapsed().as_millis()))
+}
+
 async fn sqlserver_pool_is_current(
     state: &AppState,
     pool_key: &str,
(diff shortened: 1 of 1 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 6ec3ff0, will be released in the next version.

PR #6338, opened 2026-08-15
fix(tabs): stop touch input from arming tab-drag reorder

### LINKED ISSUE #5854: [Bug] 对触控使用很不友好，单击标签或者选中的内容的时候，标签会跟着鼠标走
### 数据库类型和版本

Oracle

### 问题描述

Image

### 支持信息

DBX 版本：v0.5.81
运行环境：桌面端
操作系统：macOS 26.5.2
系统架构：x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Summary

Fixes #5854 — on touch input, tapping a tab could get misread as the start of a drag-to-reorder gesture, leaving the drag ghost stuck following the pointer.

- `useTabDrag.ts` only listened to `mousedown`/`mousemove`/`mouseup` and used a 12px horizontal-jitter threshold to tell "click jitter" apart from a real drag. Many touchscreen digitizers report to the OS/webview as a plain HID mouse with no distinguishable input type, and their tap contact-point drift routinely exceeds that 12px threshold, arming a drag from an ordinary tap.
- Because the drag only resets on `mouseup`, and the tab strip's native horizontal scroll competes for the same gesture on touch, the ghost element can be left following later pointer movement — matching the "标签会跟着鼠标走" (tab keeps following the pointer) symptom and screenshot in the issue.

## Fix

- Migrated `useTabDrag.ts` from `mousedown`/`mousemove`/`mouseup` to Pointer Events (`pointerdown`/`pointermove`/`pointerup`), matching the existing pattern already used in this codebase by `useTabScroll.ts` and `useDataGridColumnLayout.ts`. When `event.pointerType === "touch"`, the drag is never armed, letting the tab strip's native horizontal scroll handle the gesture instead.
- Raised `TAB_DRAG_HORIZONTAL_THRESHOLD` from 12px to 24px as defense-in-depth for touch input that the OS/webview reports as a generic mouse pointer (no distinguishable `pointerType`).
- Updated the comment in `useDataGridSelection.ts` that claimed exact parity with this threshold, since the two now intentionally diverge (that composable's own click-jitter threshold, from #6213, is unrelated to touch).

## Testing

- Reproduced against the real running app with Playwright's **webkit** engine (matching dbx desktop's WKWebView on macOS): a press-and-18px-drift-then-release on a tab reliably reproduced the floating drag ghost shown in the issue's screenshot; confirmed gone after the fix (same steps, isolated dev backend/frontend, not a production instance).
- Also verified in WebKit that calling `preventDefault()` on the touch-synthesized `mousedown` does not suppress the following `mouseup`/`click`, ruling that out as a contributing factor.
- Added unit tests in `useTabDrag.spec.ts`: 18px of jitter no longer arms a drag, and touch `pointerType` never arms a drag regardless of movement distance.
- Full suite: 9298 tests passing. `typecheck` and `lint` clean.

## Test plan for reviewers

- [ ] Open enough tabs to overflow the tab strip; on a touch-capable display, tap and lightly drag-scroll across the tabs — the strip scrolls natively, no ghost tab appears
- [ ] With a real mouse/trackpad, click a tab normally — activates immediately, no spurious drag
- [ ] With a real mouse/trackpad, deliberately drag a tab to reorder it — still works as before

### DIFF
--- a/apps/desktop/src/components/layout/AppTabBar.vue
+++ b/apps/desktop/src/components/layout/AppTabBar.vue
@@ -544,10 +544,13 @@ function handleTabClick(tab: QueryTab) {
   activateTab(tab.id);
 }
 
-function handleTabMouseDown(event: MouseEvent, tabId: string) {
+function handleTabMouseDown(event: PointerEvent, tabId: string) {
   if (event.button === 0) {
     dispatchBeforeTabSwitch(tabId);
-    event.preventDefault();
+    // Don't preventDefault touch pointerdowns: that would cancel the tab
+    // strip's native horizontal scroll, which is how touch users browse an
+    // overflowing tab bar.
+    if (event.pointerType !== "touch") event.preventDefault();
   }
   tabDrag.startDrag(event, tabId);
 }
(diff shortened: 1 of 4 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in b6c8383, will be released in the next version.

PR #6339, opened 2026-08-15
fix(export): keep MySQL database export table order alphabetical

q396921921:
## Summary

`export_database_sql_core` unconditionally re-sorted a database's tables by foreign-key dependency (topological order, parent-before-child) after the already-alphabetical table listing. For MySQL specifically, this reorder is pure noise: MySQL exports always emit `SET FOREIGN_KEY_CHECKS = 0` up front, so neither the generated DDL nor the INSERT statements ever need dependency-safe ordering. The result was that table structure exports came out in a foreign-key-dependency order that didn't match the alphabetical order tables are shown in everywhere else in the app, which reads as "shuffled"/"unordered" to users.

This PR skips the dependency sort for MySQL and keeps tables in their already-alphabetical listing order. Other database types (e.g. Postgres, which does not disable constraint checking during import) keep the existing dependency-ordering behavior unchanged.

I also investigated the ClickHouse part of the issue (same complaint) but could not reproduce it: ClickHouse has no foreign key support, so `list_foreign_keys_core` already returns an empty list for it and the dependency sort is already a no-op — a live-container test confirms ClickHouse table structure export is already alphabetical. Left ClickHouse untouched; likely the reporter conflated observations from both databases.

The issue's secondary point (MySQL EVENT objects missing from the export/object-type options) is a separate, unrelated gap and is out of scope for this PR.

## Root cause

- `crates/dbx-core/src/db/mysql.rs` `list_tables_sql` already has `ORDER BY TABLE_NAME` — alphabetical, deterministic.
- `crates/dbx-core/src/database_export.rs` (`export_database_sql_core`) then re-sorts via `crate::transfer::sort_tables_by_fk_dependency` regardless of database type, moving FK-bearing tables after the tables they reference.
- For MySQL this reorder has no correctness purpose, since `SET FOREIGN_KEY_CHECKS = 0` is always written for MySQL exports (`database_export.rs`), covering both DDL and DML ordering safety already.

## Fix

`database_export.rs`: gate the dependency-sort block on `db_type != DatabaseType::Mysql`.

## Test plan

- [x] New regression test `crates/dbx-core/tests/live_mysql_export_table_order.rs` against a real disposable MySQL 8.4 container: 6 tables including cross-table foreign keys chosen so FK order and alphabetical order diverge. Reverting the fix makes the test fail with the FK-topological order (`alpha,bravo,mango,zoo,orders,order_items`); with the fix it passes with alphabetical order (`alpha,bravo,mango,order_items,orders,zoo`). Also asserts two consecutive exports of the same unchanged schema produce identical output (already deterministic — the perceived "inconsistent between runs" was really "doesn't match the alphabetical order shown elsewhere").
- [x] `cargo test -p dbx-core --lib database_export::` — 62/62 passed
- [x] `cargo test -p dbx-core --lib transfer::` — 195/195 passed
- [x] `cargo fmt --check` / `cargo clippy -- -D warnings` — clean
- [x] Rebased onto latest `upstream/main`, no conflicts
- [ ] Full `cargo test -p dbx-core --lib` — attempted but the run stalled for ~2 hours with near-zero CPU usage in my environment (unrelated to this change, likely a local environment issue); killed and relied on the scoped module tests above instead.

Refs #6229

### DIFF
--- a/crates/dbx-core/src/database_export.rs
+++ b/crates/dbx-core/src/database_export.rs
@@ -1776,8 +1776,12 @@ pub async fn export_database_sql_core(
     };
 
     // Sort tables by foreign key dependency so referenced (parent) tables are
-    // exported before referencing (child) tables.
-    if tables.len() > 1 {
+    // exported before referencing (child) tables. MySQL exports always emit
+    // `SET FOREIGN_KEY_CHECKS = 0`, so both the generated DDL and any INSERTs
+    // already tolerate child-before-parent ordering; reordering away from the
+    // alphabetical listing there only makes the output look shuffled to users
+    // without providing any correctness benefit.
+    if tables.len() > 1 && db_type != DatabaseType::Mysql {
         let table_names: Vec<String> = tables.iter().map(|t| t.name.clone()).collect();
         match crate::transfer::sort_tables_by_fk_dependency(
             state,
--- a/crates/dbx-core/tests/live_mysql_export_table_order.rs
+++ b/crates/dbx-core/tests/live_mysql_export_table_order.rs
@@ -1,0 +1,137 @@
+use dbx_core::connection::AppState;
+use dbx_core::database_export::{export_database_sql_core, DatabaseExportRequest};
+use dbx_core::models::connection::{ConnectionConfig, DatabaseType};
+use dbx_core::query::execute_sql_statement;
+use dbx_core::storage::Storage;
+
+fn live_mysql_config(id: &str) -> ConnectionConfig {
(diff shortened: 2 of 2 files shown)


t8y2 (commented):
PR 描述已将 `Fixes #6229` 调整为 `Refs #6229`：当前改动补齐了通用对象备注读取，但尚未实现 MySQL EVENT，不能自动关闭该 Issue。代码和现有检查保持不变。

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in bd4cfd6, will be released in the next version.

PR #6341, opened 2026-08-15
fix(mqtt): add pause/resume control for the message stream

### LINKED ISSUE #5615: [Bug] 关于mqtt，查看某个订阅的topic信息时，无法暂停数据的自动滚动
### 数据库类型和版本

MQTT5.0

### 问题描述

使用mqtt连接成功后，订阅topicA，查看消息内容时无法将页面内容暂停。当消息刷新速度太快时，无法停下来详细查阅想看的那条消息内容

### 支持信息

DBX 版本: v0.5.76
运行环境: 桌面端
操作系统: Windows 10.0.22631.6199
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

The MQTT admin console polls for new messages every 3 seconds and unconditionally replaces the message list. When messages arrive quickly, there is no way to freeze the view to inspect a particular message — it keeps getting overwritten by the next poll.

## Fix

Add a Pause/Resume toggle next to the payload encoding selector. Toggling it gates the polling tick (`if (messagesPaused.value) return;`), so the message list stops being overwritten while paused and resumes streaming when unpaused. This mirrors the existing pause/resume pattern already used by the SQL Server activity trace panel (`Pause`/`Play` icons, same interaction).

## Testing

- Added `MqttAdminConsolePause.spec.ts`, which mounts the real `MqttAdminConsole.vue` component with fake timers and a mocked API, clicks the actual Pause/Resume buttons, and asserts the polling interval stops/resumes calling `mqttGetMessages`.
- Verified it's a genuine regression test: temporarily reverted just the polling guard → test failed (`expected mock to not be called, but called 2 times`) → restored the fix → test passed.
- `pnpm typecheck` — clean
- `pnpm oxlint --vue-plugin` — clean
- `pnpm vitest run apps/desktop/src/components/mqtt` — 3 files / 6 tests passed

Note: the MQTT admin console only runs in the desktop (Tauri) build — the web/dev-server backend explicitly rejects MQTT calls (`MQTT 功能暂不支持 Web/Docker 模式`), so a live end-to-end UI screenshot wasn't obtainable in my environment. The component-level regression test above exercises the exact same reactive logic the real UI runs on.

Fixes #5615

### DIFF
--- a/apps/desktop/src/components/mqtt/MqttAdminConsole.vue
+++ b/apps/desktop/src/components/mqtt/MqttAdminConsole.vue
@@ -9,7 +9,7 @@ import TopicTreeNode from "./TopicTreeNode.vue";
 import MqttPublishPanel from "./MqttPublishDialog.vue";
 import { decodePayload, PAYLOAD_ENCODINGS, PAYLOAD_ENCODING_LABELS, type PayloadEncoding } from "@/lib/mqtt/mqttPayloadCodec";
 import { safeLocalStorageGet, safeLocalStorageSet } from "@/lib/backend/safeStorage";
-import { ChevronDown, ChevronUp } from "@lucide/vue";
+import { ChevronDown, ChevronUp, Pause, Play } from "@lucide/vue";
 
 interface Props {
   connectionId: string;
@@ -27,6 +27,7 @@ const selectedTopic = ref<string>(props.initialTopic ?? "");
 const loading = ref(true);
 const error = ref<string | null>(null);
 const pollingTimer = ref<ReturnType<typeof setInterval> | null>(null);
+const messagesPaused = ref(false);
 const displayEncoding = ref<PayloadEncoding>("plaintext");
 const topicSearch = ref("");
 const showSubscriptionDialog = ref(false);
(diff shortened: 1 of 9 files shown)


t8y2 (commented):
已在 fbee40c 中补齐暂停竞态：轮询请求返回后会再次检查暂停状态，暂停前已经发出的在途请求不再覆盖当前消息列表。

验证：`pnpm exec vitest run apps/desktop/src/components/mqtt/__tests__/MqttAdminConsolePause.spec.ts`（1 个文件、2 条测试通过，包含 deferred 在途请求场景），`pnpm typecheck` 和 `git diff --check` 通过。

t8y2 (commented):
Updated the maintainer patch to 0ecc47d56 after the exact-head frontend check reported one formatting issue in the new MQTT regression test.

The only additional change removes the formatter-rejected blank line; runtime behavior is unchanged.

Revalidated:
- pnpm exec oxfmt --check apps/desktop/src/components/mqtt/__tests__/MqttAdminConsolePause.spec.ts
- pnpm exec vitest run apps/desktop/src/components/mqtt/__tests__/MqttAdminConsolePause.spec.ts (2 passed)
- pnpm typecheck
- git diff --check

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 7f1a98349, will be released in the next version.

PR #6348, opened 2026-08-15
fix(editor): bound SQL autocomplete context scanning to avoid O(document) per keystroke

### LINKED ISSUE #5968: [Bug] 大文件查看卡顿且无法编辑，一旦编辑就卡死
### 数据库类型和版本

通用问题（与具体数据库类型无关，主要影响编辑器/大文件查看场景）。复现场景覆盖大 SQL 文件与存储过程源码。

### 问题描述

1. **复现步骤**：
   - 在 DBX 中打开 **万行以上** 的大 SQL 文件，或打开/查看 **存储过程源码**（同样为万行级大文本）
   - 进行滚动、浏览查看
   - 尝试在编辑器中输入、修改内容

2. **期望结果**：
   - 大文件可流畅滚动查看
   - 可正常编辑，输入/删除不卡顿、不卡死

3. **实际结果**：
   - 查看大文件时界面明显卡顿
   - 几乎无法正常编辑
   - 一旦开始编辑，应用直接卡死（无响应）

4. **影响**：
   - 无法在 DBX 中处理中大型 SQL 脚本或较长存储过程源码
   - 严重时需强制退出应用

5. **复现规模与场景**：
   - 文件规模：万行以上
   - 场景：大 SQL 文件、存储过程源码（两者均复现）

### 支持信息

- DBX 版本：v0.5.82
- 运行环境：桌面端
- 操作系统：macOS 27.0
- 系统架构：aarch64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

- 建议排查方向：编辑器是否对超大文档做了虚拟滚动/分片渲染；语法高亮、自动补全、lint、历史栈等是否在全文变更时做了同步重计算导致主线程阻塞。
- 相关可能交叉问题：存储过程源码查看无响应等大文本场景（如 #5072）。

q396921921:
## Problem

Fixes #5968

Opening a SQL file or stored-procedure source with 10,000+ lines makes the editor nearly unusable: scrolling is janky and typing a single character can block the UI for a noticeable, growing amount of time.

Root cause, found by reproducing with a real `dbx-web` backend + real `vite` dev frontend under Playwright/CDP (not from reading the code): the SQL autocomplete engine runs on every keystroke (`autocompletion({ activateOnTyping: true, ... })`). `provideSqlCompletions` passes the **entire document text** to several lexical-context helpers:

- `activeSqlCompletionStatementSpan` → `tokenizeSqlSemantic(fullDoc, ...)` tokenizes the whole document from scratch
- `getSqlLexicalContext` (called first, on every keystroke, via `isSqlCompletionSuppressedContext`) scans from index `0` to the cursor

Both cost grows with total document size, so every keystroke on a large file pays an O(document) tax. This is the same class of problem `expandToSqlStatementWindow` (in `insertValueHints.ts`) was already written to avoid for the semantic highlighter ("large scripts do not pay O(document) on every keystroke"), but the autocomplete path never got the same treatment.

## Solution

Bound both `getSqlLexicalContext` and `activeSqlCompletionStatementSpan` to the statement window around the cursor using the existing `expandToSqlStatementWindow` helper (32KB lookback/lookahead, quote/comment-aware), translating any returned offsets back to absolute document positions. No new abstractions — reuses the utility already established in this codebase for this exact tradeoff.

## Testing

Measured with a real 15,000-line / 2.28MB SQL document, real keyboard events (`Input.dispatchKeyEvent`, not a synthetic paste), main-thread block time measured via a `page.evaluate` ping immediately after each keystroke:

Before (3 consecutive real keystrokes):
```
round-trip 0.57s / 0.73s / 0.72s
```
CPU profile (single keystroke, 1188ms sampled): `tokenizeSqlSemantic` 23.5%, `getSqlLexicalContext` chain ~10%+.

After (same script, same document, fresh browser session):
```
round-trip 0.25s / 0.34s / 0.43s
```
CPU profile: total sampled time dropped from 1188ms to 530ms (**-55%**); `tokenizeSqlSemantic`/`getSqlLexicalContext` no longer appear among the top offenders.

- `vitest run --root apps/desktop`: 684 files / 6384 tests passing
- `vue-tsc --noEmit`: clean
- `oxlint`: clean

## Scope note

Profiling also showed a separate, smaller (~20%) remaining cost from `currentStatementFrameLayer` (the current-statement highlight), which re-parses the full document on every keystroke via a cache shared with two other call sites (execute-statement navigation). Fixing that safely requires restructuring how that shared cache is used and is a larger, separate change — left out of this PR to keep it focused and minimal.


t8y2 (changes requested):
Request changes: the 32 KiB autocomplete window starts at an arbitrary document offset but resets lexical state to code/default.

If the cursor is more than 32 KiB into a quoted string, block comment, dollar-quoted body, or nested statement, the truncated prefix no longer contains the opening token. The scanner can then enable completion inside non-code text or compute the wrong statement span. The comment claiming that the window starts at a clean lexical boundary is not guaranteed by the implementation.

Please use cached/incremental lexical state, a syntax-tree boundary, or a conservative fallback when the starting state cannot be proven. Add regressions with the cursor beyond 32 KiB inside a string, block comment, dollar quote, and nested statement while preserving the intended per-keystroke performance bound.

q396921921:
Thanks for the detailed review — you're right, the clean-state assumption at the truncated window start wasn't actually guaranteed.

Pushed a fix: `resolveStatementStart` now verifies the backward-scan boundary instead of trusting it — it widens the lookback window until two consecutive scans agree on the same boundary, or until it reaches the true document start (index 0), where the clean-state assumption is always correct. This is the "conservative fallback when the starting state cannot be proven" option you suggested; realistic documents still resolve in one or two bounded scans (preserving the per-keystroke bound), and only a construct spanning tens of megabytes forces wider scans, with the total cost still bounded (a doubling series).

Also added dollar-quote (`$$` / `$tag$`) awareness, which the scanners were missing entirely — without it, a Postgres dollar-quoted function body containing `;` characters could desync the lexer regardless of window size.

Added regressions with the cursor more than 32 KiB into an unterminated string, block comment, dollar-quoted body, and nested parens (all four confirmed to reproduce the wrong boundary on the pre-fix code), plus a fast-path perf test confirming typical documents still avoid a full-document scan.

q396921921:
Thanks for the thorough follow-up review — all four confirmed and fixed:

1. **Dialect-agnostic `#`** — `stepLexState` (and `getSqlLexicalContext`'s own scanner) now take a `dialectId` and only treat `#` as a MySQL-style line-comment starter when `dialectId === "mysql"`, matching `tokenizeSqlSemantic`'s own dialect check. `activeSqlCompletionStatementSpan` now passes its already-resolved `dialectId` into `expandToSqlStatementWindow` instead of leaving it dialect-agnostic, so the window boundary and the tokenizer agree. Verified your exact repro (`SELECT 5 # 3; SELECT 1;`): under `"postgres"` the window now correctly isolates `SELECT 1`; under `"mysql"`/no dialect it still merges into one statement, matching `#` genuinely starting an unterminated comment there. Threaded the resolved dialect through `buildSqlCompletionItems`, `shouldAutoOpenSqlCompletion`, and both `QueryEditor.vue` call sites so this reaches the live completion path, not just the primitive.
2. **Per-`$` allocation** — replaced `sql.slice(index)` + regex with `matchDollarQuoteTag`, a sticky (`y`-flag) regex matched directly against the original string via `lastIndex`, no copy.
3. **Triplicated dollar-quote regex** — `matchDollarQuoteTag` is now exported from `semantic/tokens.ts` (the canonical tokenizer) and reused by both `insertValueHints.ts` and `sqlCompletion.ts`; fixed `tokens.ts`'s own copy of the same slice-allocation issue as a side effect.
4. **Duplicate window computation** — `expandToSqlStatementWindow` now wraps its computation in a single-entry memo keyed by `(sql, from, to, dialectId)`. String `===` is content equality in JS and the function is pure, so a cache hit is provably correct, not a staleness risk.

Added regressions: `#` under postgres vs. mysql/default, a 20k-`$`-marker perf guard, and a same-args memo-hit check that asserts reference equality (proving the cache actually short-circuits recomputation, not just returning an equal-shaped object).

All 6177 tests + typecheck + lint pass.

q396921921:
这一批推送汇总了两轮审查（一次人工 + 一次自动化 code-review）发现的问题，用中文完整说明一下现在的状态，方便你判断。

## 修好的问题

**方言不一致的另外两个漏点**：上一轮只把 `#` 的方言判断修到了自动补全路径，`QueryEditor.vue` 的语义高亮和 `codemirrorInsertValueHints.ts` 的 INSERT VALUES 提示这两处仍然没传 `dialectId`，postgres 下静默退化成 mysql 行为。这次把方言解析逻辑统一搬到 `semantic/dialect.ts` 的 `resolveSqlDialectId`（避免第三份重复实现），补齐了这两处调用。

**反斜杠转义未处理**：`stepLexState`（负责语句窗口边界扫描）完全没处理 `\'`/`\"` 转义，但 `getSqlLexicalContext`（负责窗口内的具体词法判断）一直都处理。两者不一致会导致 `SELECT 'it\'s a test; end' FROM t;` 这类输入里，字符串内部被转义的 `;` 被误判成语句边界。已对齐两者的处理方式，并用旧代码复现验证过（旧代码给出的窗口是错误截断的 `end' FROM t`，修复后是完整的语句）。

**单条超大语句被静默截断**：找语句终点原来只往前扫一个 32KB lookahead，找不到 `;` 就直接在这个位置截断，不会再往后找——超大的单条语句（比如几万行的批量 INSERT）会被静默解析成不完整的窗口。现在改成"找不到就翻倍继续扫，从上次断点续扫而不是重新扫"，保证要么找到真正的结尾，要么如实扫到文档末尾，不会再悄悄给错误答案。用一个 10 万字符、真正结尾在很远处的单语句验证过：旧代码在 32805 处截断，新代码正确给到真正的结尾。

**`\r`（无 `\n`）行注释未正确终止**：`tokens.ts` 和 `getSqlLexicalContext` 都同时判断 `\n`/`\r` 结束行注释，`stepLexState` 只判断了 `\n`，属于三处独立实现产生的真实漂移。已对齐。

**顺手清理**：`computeSqlStatementWindow` 里两处恒等于自身的 `Math.min`/`Math.max` 死代码删掉了。

## 复现验证但判断不修的问题（这里是我们唯一不确定维护者会不会认可的地方）

自动化审查还指出：`resolveStatementStart` 用来验证回退窗口起点词法状态干净的机制——"扩大窗口直到连续两次扫描结果一致就采信"——**这个启发式不是数学证明，是可以被骗过的**。我们用具体例子复现了：一个约 83 万字符的 `CREATE FUNCTION ... AS $$ ... $$`，`$$` 内部塞了 6 万条格式统一的 `SELECT n;`（没有任何引号/括号/注释字符），光标放在 body 深处，`expandToSqlStatementWindow` 给出的窗口是函数体中间一个 12 字符的碎片（`SELECT 36507`），而不是正确答案（整个 `CREATE FUNCTION` 语句，`from: 0`）。

深入分析后的结论是：**这类"大段无标点重复文本"的场景，和"这段文本前面藏着几十万字符前打开的未闭合结构"这两种情况，从只看局部内容的扫描器角度看是完全无法区分的**——不管扫描起点选在哪里，两次扫描都会收敛到同一个（错误的）答案，跟真实状态干不干净没有关系，纯属内容巧合。我们试过一个加强版本（要求两次扫描之间的空白区间必须出现至少一个引号/括号/注释类特殊字符才采信），确实能挡住这个复现用例，**但代价是会连带误伤"大量普通小语句、无特殊字符"这个最常见的场景**（比如批量 SQL 导入文件），让它退化成从文档开头全量扫描——正好是这个 PR 系列（issue #5968）最初要修的那个"大文件卡顿"场景，属于拆东墙补西墙，所以没有采用。

真正彻底解决这个问题需要维护者最初就提到的另外两个方案之一：增量缓存的词法状态，或者接入 `@codemirror/lang-sql` 已有的语法树——这两个都是比这轮 review 循环大得多的架构改动，而且会影响到现在这些函数被纯字符串直接调用的测试写法（`packages/app-tests/insertValueHints.test.ts` 里几十个用例都是这么写的）。所以这次选择的处理方式是：**把这个已知边界如实写进代码注释**（说明具体失败场景和为什么局部启发式修不了），**加一个"documents a known limitation"的回归测试把当前行为钉住**（用上面那个 6 万条语句的复现用例），这样这个边界不会被后续改动悄悄改坏，也不会被悄悄"修复"却没人注意到——真要修的时候，这条测试会提醒去更新它的断言。

这是我们主动做出的权衡，不是遗漏。**唯一的不确定性是这个权衡你是否认可**——如果你希望在合并前就把这个边界彻底解决（增量缓存或语法树方案），我可以另开一个 issue 跟进，这次 PR 先维持当前的、有明确文档和测试兜底的状态；如果你觉得这个已经足够（毕竟触发条件相当苛刻：需要几十万字符量级、完全没有标点的重复内容，且藏在一个同样量级的未闭合结构里），那这轮就到这里。看你怎么判断。

全部改动：6183 个测试 + typecheck + lint 全部通过。

t8y2 (changes requested):
Request changes: the latest head still accepts an unverified statement boundary inside long strings/comments, so SQL completion can start with a clean lexer state from the wrong location.

`apps/desktop/src/lib/sql/insertValueHints.ts:404` treats the boundary as trustworthy when two enlarged scans find the same semicolon. If both scan starts are still inside the same long lexical construct, they deterministically agree on the same false boundary. The regression case remains reproducible with an unterminated string containing repeated semicolons: the expected window start is `0`, but the implementation returns `100000`. Long block comments and line comments have the same issue.

`apps/desktop/src/lib/sql/sqlCompletion.ts:1756` then scans from that incorrect position as clean SQL, allowing completion inside content that is actually a string or comment. The new assertion at `packages/app-tests/insertValueHints.test.ts:362` explicitly preserves the non-zero false boundary rather than proving lexical correctness.

The fallback also remains `O(document)` across keystrokes because the memo key includes the complete changing SQL string. A durable fix needs incremental lexical checkpoints/parser state or another reliable boundary source, not another fixed scan window. The exact-head frontend CI also currently fails on the environment-sensitive hard-coded `<50ms` assertion.

q396921921:
Thanks for the sharper counterexample — you're right that the "known limitation" from the last round wasn't actually a rare/extreme edge case; an unterminated string with repeated `;` is exactly the state a user is in mid-typing an open quote, and the widen-and-agree heuristic converging on the same wrong boundary both times is a real bug, not a documented tradeoff.

Rather than patch the scanner a fourth time, this round replaces it as the boundary source for live editing with a syntax-tree boundary — one of the two options you suggested. `@codemirror/lang-sql`'s `sql()` extension is already mounted on the editor (`QueryEditor.vue`'s `buildSqlLanguageExtension`) and is a real, incrementally-parsed Lezer grammar with proper `Statement`/`String`/`LineComment`/`BlockComment` node types; it was just never consulted for this. New `sqlSyntaxTreeWindow.ts` resolves the statement window and string/comment context from that tree instead of re-deriving lexical state locally, and only ever reads whatever CodeMirror has already parsed in the background (`syntaxTree`/`syntaxTreeAvailable`, never `ensureSyntaxTree`) — so it never blocks, and when the parse hasn't caught up to the cursor yet it returns `null` and falls back to the existing scanner unchanged, i.e. never worse than today's behavior.

This directly answers all three things you flagged:
- **False boundary at `insertValueHints.ts:404`**: fixed for this class of input — the tree resolves your exact repro (unterminated string, repeated `;`) to the correct `from: 0` in one step, no heuristic. New test: `sqlSyntaxTreeWindow.spec.ts`.
- **Completion activating inside string/comment content at `sqlCompletion.ts:1756`**: fixed at the public API — `sqlCompletion.syntaxTree.spec.ts` reproduces the bug with the pure-string call (no live state) and shows it's fixed with a live `EditorState`, which is what an actual editor session always has.
- **O(document) across keystrokes via the memo cache**: no longer relevant on this path — CodeMirror maintains the tree incrementally across edits already (it's parsing in the background regardless, for syntax highlighting), so reading it isn't a new per-keystroke cost.

One gap I want to flag explicitly rather than let you find it: this app disables `doubleDollarQuotedStrings` (see `a067ee8d7`, issue #788) so PL/pgSQL `$$ ... $$` bodies highlight as nested SQL instead of one opaque string. That means the tree does **not** recognize a dollar-quoted body as a single unit, and can still misparse `;` inside it as real statement boundaries — the tree does not silently fix the original round-1 dollar-quote case. `resolveStatementWindowFromSyntaxTree` guards against this with a bounded backward scan for unmatched dollar-quote markers and returns `null` (safe fallback, same as today) rather than trusting a wrong boundary — but that guard is itself bounded (32KiB, same budget as the old scanner), so a tag opened further back is invisible to it. This is a real, disclosed tradeoff: fully solving it would mean either regressing #788's highlighting or maintaining a second parser instance solely for boundary-safety, which felt like a materially bigger change than this PR. Happy to open a follow-up issue for it if you'd rather not carry it forward silently — let me know if you want it solved now instead.

Also fixed the CI-flaky hard-coded `elapsedMs < 50ms` assertion (and two sibling wall-clock ones) — replaced with a self-calibrating ratio (elapsed at 10x scale vs 1x scale, generous multiplier, plus a loose absolute backstop against an actual hang) so it stops depending on the runner's absolute speed.

Full monorepo suite (`vitest run`): 979 files / 9343 tests passing. `vue-tsc --noEmit`: clean. `oxlint`: clean.

t8y2 (commented):
Maintainer patch pushed: `27d6019a3` -> `54ba69fc1` (base `e6fa850a3`).

Completion already found a bounded statement window, but the semantic model could still parse outside that active statement. The patch restricts semantic parsing to the active window, requires exact tagged dollar-quote delimiter matches, and falls back to the scanner when the syntax-tree statement is partial.

Validation: `sqlSyntaxTreeWindow.spec.ts` and `sqlCompletion.syntaxTree.spec.ts` passed all 15 tests, including the large-document regression. The final base-to-head diff is clean.

t8y2 (commented):
Mainline sync pushed after the shared frontend formatting fix entered `main`.

The bounded active-statement semantic parsing patch is unchanged. `oxfmt --check` passes and all 15 syntax-tree/completion regressions pass again.

t8y2 (approved):
The active-statement parsing boundaries and dollar-quote handling are fixed on the exact head. Focused regressions and required CI pass.

t8y2:
Thanks for the contribution! Merged in 41ffc969a, will be released in the next version.

Issue #6112, opened 2026-08-13
[Bug] 可以正常执行的sql语句，使用默认驱动连接可以正常使用执行计划功能，使用ojdbc11-21.9.0.0连接报错ORA-00905:缺失关键字


digitalzhao309-hash:
### 数据库类型和版本

Oracle 11g

### 问题描述

可以正常执行的sql语句，使用默认驱动连接可以正常使用执行计划功能，使用ojdbc11-21.9.0.0连接报错ORA-00905:缺失关键字
 #6050 也做对应修复，显示当前数据库暂不支持执行计划

### 支持信息

DBX 版本: v0.5.83
运行环境: 桌面端
操作系统: Windows 10.0.17763.316
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

q396921921:
Thanks for the report. I've been digging into this and set up a real Oracle instance + `ojdbc11-21.9.0.0` to test the custom-JDBC explain-plan path added in #6050's fix (`DbxJdbcPlugin.getOracleExplainInfo`, which builds `EXPLAIN PLAN SET STATEMENT_ID='...' FOR <your sql>`).

I did find one confirmed, reproducible asymmetry between the default driver and the custom JDBC driver: if the SQL contains an Oracle bind variable (`:1`, `:B1`, etc., which is common when pasting SQL copied from `V$SQL`/AWR reports) or a literal `?`, the built-in agent driver (`agents/drivers/oracle-go`) explicitly detects these placeholders and null-binds them before running EXPLAIN PLAN, while the custom JDBC plugin does none of that and fails with `ORA-17041: Missing IN or OUT parameter`.

However, that's `ORA-17041`, not the `ORA-00905: missing keyword` you reported — I tried a dozen other variations (trailing `;`, trailing `/`, CTEs, comments, `RETURNING INTO`, etc.) and couldn't reproduce `ORA-00905` specifically through this path.

Could you share:
- The exact SQL statement you ran through the execution-plan button (with any sensitive identifiers redacted/replaced)
- Whether it contains any `:name` style bind variables

That'll let me match the fix to what you're actually hitting rather than guessing further. I'll fix the bind-variable issue I found regardless, since it's a real bug in the same code path, but want to make sure it also addresses your case.

digitalzhao309-hash:
[...]

PR #6353, opened 2026-08-15
fix(oracle): null-bind explain plan placeholders for custom JDBC connections

q396921921:
## Summary

Related to #6112.

While investigating #6112 ("execution plan works with the default driver but ojdbc11 custom JDBC connections throw `ORA-00905`"), I found and confirmed a real, distinct bug in the same code path added by #6050's fix (`DbxJdbcPlugin.getOracleExplainInfo`): if the SQL going through the execution-plan button contains an Oracle bind variable (`:1`, `:B1`, or a literal `?` — common when SQL is copied from `V$SQL`/AWR reports), the custom-JDBC explain path wraps it in a `PreparedStatement` and calls `execute()` without ever binding those placeholders, so Oracle rejects it with `ORA-17041: Missing IN or OUT parameter`, even though the exact same statement runs fine when executed directly.

The built-in Oracle agent driver (`agents/drivers/oracle-go/main.go`, `getExplainInfo`) already handles this — it scans the SQL for bind markers and explicitly null-binds each one before requesting the plan, since `EXPLAIN PLAN` never needs real parameter values. The custom JDBC plugin added later in #6050 never got the same treatment, so default-driver vs. custom-JDBC explain behavior diverges for any statement containing bind variables.

## Fix

`getOracleExplainInfo` now calls `PreparedStatement.getParameterMetaData().getParameterCount()` after preparing the `EXPLAIN PLAN` statement and calls `setNull(index, Types.VARCHAR)` for each detected placeholder before `execute()`. This is simpler than re-implementing the Go driver's manual bind-marker scanner, and Oracle's own driver already tells us exactly how many distinct positions it needs bound (including de-duplicating repeated named binds).

## Note on #6112

I could not reproduce the exact `ORA-00905` from #6112 through this path — I tried a dozen SQL variations (trailing `;`/`/`, CTEs, comments, `RETURNING INTO`, etc.) against a real Oracle instance and only this bind-variable case reliably breaks the custom-JDBC explain path (with `ORA-17041`, not `ORA-00905`). I've asked the reporter in #6112 for the exact SQL to confirm whether this fix addresses their case or whether there's a second, still-unidentified bug. Submitting this now since it's a confirmed, real defect in the same code area regardless.

## Test plan

- [x] New unit test `oracleExplainNullsBindPlaceholdersInsteadOfFailingWithMissingParameter` in `DbxJdbcPluginTest.java`, exercising the fix through the mock JDBC driver harness (asserts `setNull` is called for each detected placeholder and the RPC call succeeds).
- [x] Verified against a real Oracle 23c instance (`gvenzl/oracle-free`) + real `ojdbc11-23.5.0.24.07` driver, invoking the actual compiled plugin over its JSON-RPC protocol:
  - Before fix: `SELECT * FROM T WHERE ID = :B1 AND NOTE = :B2` → `{"error":{"message":"ORA-17041: Missing IN or OUT parameter at index: 1..."}}`
  - After fix: same request → returns a real execution plan (`TABLE ACCESS BY INDEX ROWID` / `INDEX UNIQUE SCAN`, with visible `filter("NOTE"=:B2)` / `access("ID"=TO_NUMBER(:B1))` predicates)
- [x] Ran the plugin's full JUnit suite (97 tests) standalone (this sandbox's Gradle couldn't resolve unrelated pre-release `org.apache.maven:*:4.0.0-rc-5` metadata over the network — confirmed pre-existing/unrelated to this change by reproducing the same failure on a clean `upstream/main` checkout) — all 97 pass, including the existing `oracleExplainUsesPlanTableOnTheSharedConnectionAndCleansUp` regression test.

### DIFF
--- a/plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
+++ b/plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
@@ -1155,13 +1155,14 @@ public final class DbxJdbcPlugin {
 
     private static String getOracleExplainInfo(Connection connection, String sql, int timeoutSecs) throws SQLException {
         String statementId = "DBX_" + UUID.randomUUID().toString().replace("-", "").substring(0, 26);
+        String statementSql = trimStatementSql(sql);
         StringBuilder plan = new StringBuilder();
         try {
             try (PreparedStatement explain = connection.prepareStatement(
-                "EXPLAIN PLAN SET STATEMENT_ID = '" + statementId + "' FOR " + trimStatementSql(sql)
+                "EXPLAIN PLAN SET STATEMENT_ID = '" + statementId + "' FOR " + statementSql
             )) {
                 applyExplainTimeout(explain, timeoutSecs);
-                nullBindExplainParameters(explain);
+                nullBindExplainParameters(explain, statementSql);
                 explain.execute();
             }
             try (PreparedStatement read = connection.prepareStatement(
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch applied on the contributor branch.

- Old head: `fefe55ff39da36f7e926c51150036ace07e40fd4`
- New head: `fdf4bf7c3b05314c41f2dd41c96c9d67dbae5c35`
- Base: `bd4cfd6b995c2decc8b9497bb2512e11db5616ec`
- Root cause: the custom JDBC explain path relied exclusively on `PreparedStatement.getParameterMetaData()`. When a driver reported parameter metadata as unsupported with `SQLException`, the helper returned without binding any placeholders, so Oracle could still fail with `ORA-17041`.
- Fix: keep JDBC parameter metadata as the primary path, and only fall back when it is unavailable. The fallback mirrors the Oracle Go agent's bind-marker scanning boundaries, recognizes named, numeric, and `?` placeholders, and ignores markers inside quoted text, quoted identifiers, Oracle alternative-quoted literals, and comments before null-binding each discovered parameter.
- Validation: reproduced the missing-bind behavior with a driver that throws `SQLFeatureNotSupportedException`; the focused Oracle explain tests pass; `./gradlew clean test shadowJar` passes; `git diff --check` passes.
- Dependencies: no new dependencies.

t8y2 (approved):
Reviewed maintainer-patched exact head `fdf4bf7c3`. The placeholder fallback is bounded, tested, and no blocking issue remains.

t8y2:
Thanks for the contribution! Merged in 1d8771a84, will be released in the next version.

Issue #6285, opened 2026-08-14
dbx_execute_query parameter parsing fails: connection_name/connection_id nested as object


wordgao:
### Environment
- **MCP Server Version**: `@dbx-app/mcp-server@0.4.62`
- **MCP Client**: OpenCode (opencode.ai)
- **OS**: Windows 10 x64
- **Node.js**: v24.18.1

### Problem

When calling `dbx_execute_query` tool with `connection_name` or `connection_id` parameter, the parameter is not properly expanded but instead wrapped into a nested object, causing JSON parsing failure.

### Steps to Reproduce

1. Call `dbx_list_connections` — succeeds and returns connection list (confirms connection exists)
2. Call `dbx_execute_query` with parameters:
   ``json
   {
     "connection_name": "test_conn",
     "sql": "SELECT 1"
   }
   ``

### Actual Result

Error: `Invalid input for tool dbx_dbx_execute_query: JSON parsing failed`

The error shows the parameter was nested into an object:
``json
{
  "connection_name": {
    "tool": "dbx_dbx_execute_query",
    "error": "Invalid input..."
  }
}
``

### Expected Result

SQL query executes normally and returns results.

### Additional Notes

- `dbx_list_connections` tool works fine
- Same issue occurs when using `connection_id` (UUID format) parameter
- The parameter schema defines `connection_id` and `connection_name` as `["string", "null"]` union type — possibly a type resolution compatibility issue

t8y2:
Fixed in b521d037, will be included in the next release. This issue will be closed.

Root cause: The MCP schema generator exposed the optional connection selectors as a nullable string union. Some MCP clients mis-handle that schema and send a nested object, which DBX correctly rejects as a non-string value.

Fix: The public schema now advertises `connection_id` and `connection_name` as optional string fields while preserving the existing omitted, null, string, and validation behavior.

Please verify again after upgrading to the next release. If the issue still occurs, open a new issue instead of commenting here. Include the current version, reproduction steps, relevant logs, and a reference to this issue.

wordgao:
## Follow-up after upgrading to @dbx-app/mcp-server@0.4.63

Thanks for the fix! I upgraded to .4.63 and retested.

### Partially fixed

| Tool | Status |
|------|--------|
| dbx_list_connections | ✅ Works |
| dbx_execute_query | ✅ Works |
| dbx_get_schema_context | ✅ Works |
| dbx_list_tables | ❌ Still fails |
[...]

PR #6354, opened 2026-08-15
fix(mcp): narrow remaining optional field schemas to concrete types

### LINKED ISSUE #6344: [0.4.63] dbx_list_tables and dbx_describe_table parameter parsing still fails
## Description

After upgrading to @dbx-app/mcp-server@0.4.63 (which includes the fix from #6285), dbx_list_tables and dbx_describe_table still have the same parameter parsing issue.

## Partially Fixed

| Tool | Status |
|------|--------|
| dbx_list_connections | ✅ Works |
| dbx_execute_query | ✅ Works |
| dbx_get_schema_context | ✅ Works |
| dbx_list_tables | ❌ Still fails |
| dbx_describe_table | ❌ Still fails |

## Reproduction for dbx_list_tables

**Parameters:**
`json
{
  "connection_name": "slave",
  "database": "webapp"
}
`

**Error:**
`
Invalid input for tool dbx_dbx_list_tables: JSON parsing failed
`

The error shows the same nested object pattern:
`json
{
  "connection_name": {
    "tool": "dbx_dbx_list_tables",
    "error": "Invalid input..."
  }
}
`

## Reproduction for dbx_describe_table

**Parameters:**
`json
{
  "connection_name": "slave",
  "table": "users",
  "database": "webapp"
}
`

**Error:**
`
Invalid input for tool dbx_dbx_describe_table: JSON parsing failed
`

## Environment

- @dbx-app/mcp-server@0.4.63
- OpenCode (opencode.ai)
- Windows 10 x64
- Node.js v24.18.1

## Analysis

It seems the fix in 521d037 only partially addressed the issue. Tools with fewer parameters (execute_query, get_schema_context) now work correctly, but tools with more optional parameters (list_tables, describe_table) still have the schema parsing issue.

Reference: #6285

q396921921:
## Problem

b521d0377 (fixing #6285) narrowed `ConnectionSelector.connection_id`/`connection_name` from the schemars-default `"type": ["string", "null"]` union to a concrete `"type": "string"`, because some MCP clients (e.g. OpenCode, per #6344) can't resolve that union shape and instead wrap the argument in a nested object before calling the tool.

That fix only touched `ConnectionSelector`. Every other optional field on the same request structs was left publishing the identical union shape — most notably `ListTablesRequest.database`/`.schema` and `DescribeTableRequest.database`/`.schema`, which is exactly what #6344 reports still failing after upgrading to the version containing the partial fix.

## Fix

Apply the same `#[schemars(extend("type" = ...))]` narrowing consistently to every remaining optional field across the MCP request structs in `crates/dbx-mcp/src/server.rs`:

- `ListTablesRequest`, `DescribeTableRequest`: `database`, `schema`
- `ExecuteQueryRequest`: `database`, `session_id`, `cell_char_offset`, `cell_char_limit`
- `OpenSessionRequest`, `OpenTableRequest`, `ExecuteAndShowRequest`: `database` (+ `schema` on `OpenTableRequest`)
- `AddConnectionRequest`: `port`, `database`, `driver_profile`
- `RemoveConnectionRequest`: `connection_id`
- `ExecuteRedisCommandRequest`: `db`
- `SchemaContextRequest`: `database`, `schema`, `max_tables`

Every field stays optional (still omitted from `required`); only the published `type` is narrowed from a `["T", "null"]` union to a single `"T"`, matching the precedent set by b521d0377.

## Testing

- Added `optional_fields_never_publish_nullable_union_types`, which dumps the real generated JSON Schema for every affected tool and asserts each field publishes a single concrete type. Verified as a real regression test: reverted just the new `#[schemars(extend(...))]` annotations (keeping the test), reran — it failed with `dbx_list_tables.database must publish a single concrete type, not a union: Array [String("string"), String("null")]`; reapplied the annotations, reran — passes.
- `cargo test -p dbx-mcp --lib server::` — 23/23 pass.
- `cargo fmt --check -p dbx-mcp` / `cargo clippy -p dbx-mcp --lib --tests -- -D warnings` — clean.
- Full `cargo test -p dbx-mcp --lib` has 4 pre-existing failures (`backend::tests::web_mongo_*`, `web_backend_loads_connection_group_paths_from_sidebar_layout`, all `502 Bad Gateway` from a real HTTP call) — confirmed present identically on a clean `upstream/main` checkout with no changes, unrelated to this PR.

### Honesty note on end-to-end verification

I installed the real `opencode-ai` CLI (v1.18.18) and ran both the pre-fix and post-fix `dbx-mcp` binary through `opencode serve`, connected it as a real local MCP server, and had `opencode/big-pickle` (OpenCode's free hosted model) actually call `dbx_list_tables`/`dbx_describe_table` with the exact arguments from #6344's repro (`connection_name=slave, database=webapp`). **Both the pre-fix and post-fix binary returned the same clean `Error [CONNECTION_NOT_FOUND]: Connection "slave" not found.`** — neither reproduced the `JSON parsing failed` / nested-object symptom described in #6344 with this client version + model combination. So this PR is not a confirmed end-to-end fix for the exact symptom in #6344's environment — it's a code-level fix for a real, verifiable schema defect that is the same class of bug as the one already fixed (and acknowledged) by the maintainer in b521d0377, extended to close the gap that #6344 points at. The original reporter's specific provider/model wasn't stated in the issue, so I couldn't match their exact setup. Happy to have the reporter or maintainer re-verify against their original repro environment.

Fixes #6344

### DIFF
--- a/crates/dbx-mcp/src/server.rs
+++ b/crates/dbx-mcp/src/server.rs
@@ -44,8 +44,10 @@ pub struct ListTablesRequest {
     #[serde(flatten)]
     pub selector: ConnectionSelector,
     #[schemars(description = "Database name")]
+    #[schemars(extend("type" = "string"))]
     pub database: Option<String>,
     #[schemars(description = "Schema name")]
+    #[schemars(extend("type" = "string"))]
     pub schema: Option<String>,
 }
 
@@ -56,8 +58,10 @@ pub struct DescribeTableRequest {
     #[schemars(description = "Table name")]
     pub table: String,
     #[schemars(description = "Database name")]
+    #[schemars(extend("type" = "string"))]
     pub database: Option<String>,
     #[schemars(description = "Schema name")]
+    #[schemars(extend("type" = "string"))]
     pub schema: Option<String>,
 }
 
@@ -66,20 +70,24 @@ pub struct ExecuteQueryRequest {
     #[serde(flatten)]
     pub selector: ConnectionSelector,
     #[schemars(description = "Database name")]
+    #[schemars(extend("type" = "string"))]
     pub database: Option<String>,
     #[schemars(description = "SQL query to execute")]
     pub sql: String,
     #[schemars(
         description = "Session ID from dbx_open_session. When set, the query runs on the session's pinned connection, preserving USE/SET and other session state across calls."
     )]
+    #[schemars(extend("type" = "string"))]
     pub session_id: Option<String>,
     #[schemars(
         description = "Start character offset for every string cell (default 0, max 1000000). Use the next offset reported by a truncated result to slide through a long value; narrow the query to the target row and column first."
     )]
+    #[schemars(extend("type" = "integer"))]
     pub cell_char_offset: Option<u64>,
     #[schemars(
         description = "Maximum characters returned per string cell (default 200, max 4000). Increase only for an explicit long-value expansion."
     )]
+    #[schemars(extend("type" = "integer"))]
     pub cell_char_limit: Option<u64>,
 }
 
@@ -88,6 +96,7 @@ pub struct OpenSessionRequest {
     #[serde(flatten)]
     pub selector: ConnectionSelector,
     #[schemars(description = "Database name")]
+    #[schemars(extend("type" = "string"))]
     pub database: Option<String>,
 }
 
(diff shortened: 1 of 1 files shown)


t8y2 (approved):
Reviewed exact head `ac8b91501`. No blocking issues remain.

t8y2:
Thanks for the contribution! Merged in 044379da7, will be released in the next version.

PR #6358, opened 2026-08-15
🐛 fix(grid): show column comments for joined (multi-source) query results

### LINKED ISSUE #2129: 支持在SQL编辑器中悬停显示字段/表注释，并修复查询结果中字段注释不显示的问题
**来源**: QQ 群反馈 (by 自由自在)

## AI 摘要
用户希望将光标放在SQL语句中的字段或表名上时，能够显示对应的注释。同时，在多表联查的查询结果中，即使设置了在字段名下方显示注释，注释也未能显示，期望修复该显示问题。

## 原始描述
希望光标放在SQL语句字段或表名上能显示对应的注释；多表联查时，查询结果中的字段列设置了在字段名下方显示注释但是也没有显示，希望可以显示出来

0verme:
Fixes #2129 (first half: result-set column comments for joined queries).

## Problem

Executing a JOIN / multi-source query hid every column comment in the result grid. Single-table queries worked because the data grid builds its comment map from `tableMeta.columns`; multi-source results are not editable so `tableMeta` was `undefined`, and the source tables' metadata — already loaded during editability analysis — was discarded.

## Changes

- **`QueryTab`** (`types/database.ts`) gains `resultColumnComments` (merged comments from every JOIN source) and `queryDisplaySourceColumns` (display-only result→source mapping).
- **`queryStore.ts`** `buildQueryMetadataPatch` now populates both fields for multi-source branches via `mergeEditableSourceColumnComments` / `mergeEditableSourceDisplayColumns`; `applyQueryMetadataPatch`, tab clone and result-cache snapshot carry them.
- **`DataGrid.vue`** merges `resultColumnComments` into `columnCommentMap` and resolves header/tooltip comments through `queryDisplaySourceColumns` first for precise matching.
- **`ContentArea.vue`** passes the new props.

## Tests

- `queryStore.multiSourceColumnComments.spec.ts`: JOIN result merges comments from both sources + display mapping; single-source regression stays free of the new fields.
- `multiSourceColumnMapping.spec.ts`: sqlAnalysis per-source column mapping (incl. duplicate result names).
- `DataGridColumnComments.spec.ts`: extended source assertions.

## Validation

- `vue-tsc` typecheck ✅, `oxlint` ✅, `oxfmt --check` ✅
- 45 related test files / 983 tests pass; the 4 failing specs (Redis completion, grid popover/condition/readonly) were confirmed failing before these changes (pre-existing).

## Follow-up

The second half of #2129 — hovering a column/table name in the SQL editor to show its comment — will be a separate PR.

## Review follow-up

This update addresses the requested changes from the latest review:

- Resolve result-column comments by result ordinal instead of a global physical-column-name map.
- Preserve the source identity for each result column with `sourceKey` + `sourceColumn`, so same-named columns from joined sources do not collide.
- Reuse the existing metadata-aware source-column binder for aliases and source resolution; ambiguous or calculated columns intentionally receive no comment rather than a potentially incorrect one.
- Preserve quoted identifier semantics instead of globally lowercasing identifiers.
- Restore the new column metadata correctly across tab result-cache serialization / reload paths.

Regression coverage includes:
- same-named columns across joined tables
- alias resolution and ambiguous columns
- quoted mixed-case identifiers
- duplicate result names
- result-cache round trips
- single-table fallback behavior

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -335,6 +335,19 @@ interface DataGridProps {
   context?: "results" | "table-data";
   autoTransposeSingleRow?: boolean;
   sourceColumns?: Array<string | undefined>;
+  /**
+   * Column comments merged from every source table of a multi-source query
+   * (e.g. JOIN). Populated even when the result is not editable, so joined
+   * results still show comments. Keyed by physical column name (and its
+   * lower-cased form), same keys as the `tableMeta`-derived map.
+   */
+  resultColumnComments?: Record<string, string>;
+  /**
+   * Display-only result-column -> source-column mapping for multi-source
+   * results that are not editable. Used to resolve column comments precisely;
+   * never used for row identity or editing.
+   */
+  queryDisplaySourceColumns?: Array<string | undefined>;
   initialWhereInput?: string;
   initialOrderByInput?: string;
   sortColumn?: string;
@@ -539,6 +552,11 @@ const columnCommentMap = computed(() => {
       if (!map.has(normalizedName)) map.set(normalizedName, col.comment);
     }
   }
+  if (props.resultColumnComments) {
+    for (const [name, comment] of Object.entries(props.resultColumnComments)) {
+      if (comment && !map.has(name)) map.set(name, comment);
+    }
+  }
   return map;
 });
 const dataGridTopbarWidth = ref(0);
@@ -575,7 +593,9 @@ const readonlyTextCell = ref<{
 } | null>(null);
 
 function resolvedColumnComment(column: string, actualColIdx: number): string | undefined {
-  return dataGridColumnCommentFor(columnCommentMap.value, column, props.sourceColumns?.[actualColIdx]);
+  const displaySource = props.queryDisplaySourceColumns?.[actualColIdx];
+  const sourceName = displaySource ?? props.sourceColumns?.[actualColIdx];
+  return dataGridColumnCommentFor(columnCommentMap.value, column, sourceName);
 }
 
 function headerColumnComment(column: string, actualColIdx: number): string {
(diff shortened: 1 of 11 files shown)


t8y2 (changes requested):
Request changes: the result-column comment mapping loses source identity and therefore returns incorrect metadata.

1. Comments from all source tables are merged into one map keyed only by physical column name with first-source-wins behavior. Two tables with `id` columns can display the first table's comment for both result columns.
2. The result mapping stores a source column name but not its source key, and it uses unbound SQL analysis. A uniquely resolvable unqualified alias such as `name AS username` cannot map back to the physical column even though the existing metadata binder can resolve it.
3. The new fields are written to the result cache snapshot but are not restored when the cached result is reloaded.
4. Lowercase aliases in the global map also collapse valid quoted mixed-case identifiers.

Please store comments per result ordinal with source identity, reuse database-aware binding/canonicalization, return no comment for ambiguity, restore the fields from cache, and cover same-name sources, aliases, quoted case, duplicate result names, and cache restore.

0verme:
Addressed the review feedback in `ed2c71e10`.

1. Result column comments are now resolved by result ordinal and retain `sourceKey + sourceColumn`, avoiding collisions for same-named columns across joined sources.
2. Source-column resolution now reuses the existing metadata-aware binder instead of the previous unbounded SQL analysis, including alias resolution and ambiguity handling.
3. The new metadata is restored correctly from result cache snapshots / runs.
4. Identifier matching no longer globally lowercases quoted identifiers, preserving database-specific quoted identifier semantics.

I also added regression coverage for duplicate source column names, aliases / ambiguity, quoted mixed-case identifiers, duplicate result names, and cache round trips.

t8y2 (commented):
维护者补丁 `622eef73e` 已推送。

根因：复制查询 Tab 时对数组 `resultColumnComments` 使用对象展开，运行时会得到普通对象，破坏按列序号读取注释的数组语义。

修复：改为数组展开，并增加复制 Tab 时保持数组类型、内容及独立引用的回归测试。

验证：字段注释、来源映射、结果缓存和 Tab 复制相关 5 个测试文件共 42 项通过；`pnpm typecheck`、格式检查及 `git diff --check` 通过。未新增依赖。

t8y2 (approved):
Reviewed the maintainer-patched exact head `622eef73e`; column-comment regressions, typecheck, and frontend CI pass.

t8y2:
Thanks for the contribution! Merged in 64e5e88c2, will be released in the next version.

PR #6361, opened 2026-08-15
feat(postgres): support concurrent index creation

### LINKED ISSUE #6303: [Feature] 建索引的时候，用的是CREATE INDEX，在生产环境下建索引的话会导致锁表，如果长时间执行未完成的话会引起生产事故
### 数据库类型和版本

pg/金仓

### 支持信息

DBX 版本: v0.5.84
运行环境: 桌面端
操作系统: Windows 10.0.19045.7663
系统架构: x86_64

### 当前痛点

建索引的时候，用的是CREATE INDEX，在生产环境下建索引的话会导致锁表，如果长时间执行未完成的话会引起生产事故

### 期望方案

可以考虑建索引时使用CREATE INDEX CONCURRENTLY ，这样不会锁表

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

0verme:
## feat(postgres): support concurrent index creation

Fixes #6303

### Scope

This revision intentionally limits PostgreSQL concurrent index creation to:

- **new indexes**
- **non-partitioned tables**

### Safety guarantees

- Existing index edits cannot use Concurrent: the editor disables the option and the core SQL builder rejects the request with a hard error instead of silently falling back to blocking index DDL.
- Partitioned parent tables cannot use Concurrent: same UI disable + core rejection (`CREATE INDEX CONCURRENTLY is not supported on PostgreSQL partitioned parent tables...`).
- Core SQL generation (both `build_table_structure_change_sql` and `build_create_table_sql`) rejects unsupported Concurrent requests up front: empty statement list + explicit error, so the SQL preview shows the error, Save stops, and no partial DROP/CREATE is ever executed.
- Concurrent creation uses a dedicated application-level 30-minute timeout (`CONCURRENT_INDEX_QUERY_TIMEOUT_SECS = 1800`) instead of the generic 30s query timeout, because concurrent builds may scan large tables and wait for concurrent transactions. The timeout reaches the backend over both HTTP and Tauri; `Some(0)` disables it.
- Same-name INVALID indexes (`pg_index.indisvalid = false`, leftovers of cancelled builds) are detected before a concurrent save is applied; the save is blocked with explicit drop-first guidance. No automatic deletion.
- If the partition-status probe fails, Concurrent is disabled (fail closed) rather than assumed safe.

### Non-goals

1. Concurrent replacement of existing indexes (`DROP INDEX CONCURRENTLY` + `CREATE INDEX CONCURRENTLY`).
2. `DROP INDEX CONCURRENTLY` replacement workflow.
3. Partition child concurrent build + `ALTER INDEX ... ATTACH PARTITION`.
4. Automated deletion/reindex of INVALID indexes.
5. Behavior guarantees for PG-compatible non-PostgreSQL engines (Kingbase, GaussDB, etc.).

### Tests

- Rust unit: new-index CIC emission; existing-index concurrent request rejected (no DROP/CREATE); partitioned-parent request rejected (no CREATE); plain paths unchanged; serde `partitioned` defaults to false; MySQL stale flag ignored.
- Frontend: pure `getConcurrentIndexAvailability` behavior tests (new normal table → enabled, existing index → disabled, partitioned parent → disabled, non-PG → unavailable, probe failure → fail closed); queryTimeout constant tests.
- Live PostgreSQL (ignored): concurrent build succeeds and leaves a valid index; transaction-block rejection; INVALID leftover detection; partitioned-parent request is refused at the application layer and never reaches the server.


t8y2 (changes requested):
Request changes: `CREATE INDEX CONCURRENTLY` is exposed through a workflow that does not preserve its safety guarantees.

1. Editing an existing index still performs a normal `DROP INDEX` before `CREATE INDEX CONCURRENTLY`. The normal drop can block on table users, and a failed build leaves the old index already removed.
2. The table-structure save path still applies the default 30-second query timeout. PostgreSQL can leave a same-name `INVALID` index when a concurrent build is cancelled, after which retries fail until it is handled explicitly.
3. The option is enabled for partitioned parent tables even though PostgreSQL rejects concurrent index creation on the partitioned parent. The supported approach is to build child indexes concurrently and attach them.

Official references:
- https://www.postgresql.org/docs/current/sql-createindex.html
- https://www.postgresql.org/docs/current/sql-dropindex.html
- https://www.postgresql.org/docs/current/ddl-partitioning.html

Please either scope the option to new, non-partitioned indexes with a dedicated long-running timeout and INVALID-index recovery, or implement the full safe replacement workflow including `DROP INDEX CONCURRENTLY` constraints.

0verme (commented):
Thanks for the review. Both fallback paths you flagged are now removed — unsupported concurrent requests fail closed instead of degrading to blocking DDL.

**1. Editing an existing index (normal DROP before CIC).**
The scope guard no longer downgrades. `build_table_structure_change_sql` / `build_create_table_sql` now run a new `validate_concurrent_index_scope` check first: an existing index with `concurrently: true` returns an empty statement list plus an explicit error (`CREATE INDEX CONCURRENTLY is only supported for newly created indexes...`), so no DROP INDEX and no plain CREATE INDEX can be generated behind the caller's back — even from stale drafts, hand-built payloads, or non-UI callers. The SQL preview shows the error, Save stops, nothing executes.

**2. Partitioned parent tables.**
Same treatment: `partitioned: true` + `concurrently: true` is rejected up front with `CREATE INDEX CONCURRENTLY is not supported on PostgreSQL partitioned parent tables...` instead of downgrading to a plain CREATE INDEX. The previous downgrade branch (and the live test that asserted the downgrade) was replaced with an assertion that no SQL reaches the server at all.

**3. Timeout.**
Concurrent saves use a dedicated application-level 1800s budget (`CONCURRENT_INDEX_QUERY_TIMEOUT_SECS`) that bypasses the 30s default; it is threaded through both the HTTP execute-batch route and the Tauri command, and `Some(0)` still disables it. PostgreSQL does not define a build deadline; a generic 30s budget could cancel a legitimate concurrent build and leave an INVALID index.

**4. INVALID leftovers.**
A preflight (`list_invalid_indexes`, `pg_index.indisvalid = false`) blocks the save with explicit drop-first guidance; failures keep the original DB error and only append a short hint. No string-parsed state machine and no auto-delete.

Also fixed: a regex capture-group bug in the frontend INVALID preflight that silently prevented it from triggering (`match[2]` → `match[1]`), stale `concurrently: true` values on restored existing-index drafts, and fail-closed behavior when the partition-status probe itself fails. The previous source-string frontend spec was replaced with behavior tests on a small pure helper.

t8y2 (changes requested):
Thanks for the update. The latest head closes the earlier SQL-generation and partitioned-parent guards, but three correctness issues still block this revision:

1. `apps/desktop/src/components/structure/TableStructureEditor.vue:2702` always replaces the configured query timeout with `1800` whenever the batch contains a concurrent index build. This overrides both a user value of `0` (timeout disabled) and values greater than 1800, despite the PR description stating that `Some(0)` remains unlimited. Please preserve `0` and avoid shortening a larger configured timeout.

2. A partition-status probe failure only sets `partitionStatusKnown = false` at `TableStructureEditor.vue:1504`. The checkbox becomes disabled, but an already-selected or restored new index can still retain `concurrently: true` (`:3636`) and be submitted. Please normalize or reject stale concurrent flags before SQL generation whenever availability becomes unknown/unsupported, and add a regression covering the transition from enabled to failed probe.

3. The INVALID-index live test constructs `CREATE INDEX "{schema}".idx_users_email ...` at `crates/dbx-core/tests/live_postgres_concurrent_index.rs:133`. PostgreSQL does not accept a schema-qualified index name in `CREATE INDEX`; the index inherits the table schema. As written, the ignored live test fails during setup rather than validating INVALID-index detection. Please create the index with an unqualified quoted name and keep the schema qualification on the table.

t8y2 (commented):
Maintainer patch pushed in `a62acc3c8d`.

Root cause: when the PostgreSQL partition probe temporarily failed, the draft was invalidated and its `concurrently` flag was normalized to `false`. A later successful probe automatically lifted the save blocker and regenerated SQL from that downgraded draft, silently producing a blocking `CREATE INDEX` instead of the user's original concurrent request.

Fix: preserve the Concurrent intent only while availability is transiently unknown, keep save/preview fail-closed during that state, and clear the flag only for definitive unsupported states. When a later probe succeeds, the preview now recovers as `CREATE INDEX CONCURRENTLY`.

Validation:
- focused Vitest: 2 files, 12 tests passed, including success → failure → success recovery
- `pnpm typecheck`
- Oxfmt checks for all four changed files
- `git diff --check`
- no dependency changes

t8y2:
Thanks for the contribution! Merged in aa6d60533c, will be released in the next version.

PR #6363, opened 2026-08-16
fix(sidebar): make initial fuzzy table search consistent

### LINKED ISSUE #6190: [Feature] 在sqlserver查询某个表时，表名称模糊匹配有待增强
### 数据库类型和版本

sqlserver

### 支持信息

DBX 版本: v0.5.83
运行环境: 桌面端
操作系统: Windows 10.0.22621.4317
系统架构: x86_64

### 当前痛点

例如表名称是“T_Erp_Nc_SuPlan_List”，第一次输入"erpncs"不能匹配到表名，

需要输入“terpncs”，这样才能匹配到目标表名，

在上一次输入“terpncs”可以匹配到表名之后，再次输入"erpncs"也可以匹配表名

### 期望方案

希望可以增加表名模糊匹配强度

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须做

### 补充信息

_No response_

0verme:
## Summary

Fixes #6190.

When searching tables in the sidebar on SQL Server, the fuzzy query `erpncs` could not find `T_Erp_Nc_SuPlan_List` on the **first** search, but after searching `terpncs` once, `erpncs` matched. The fuzzy matcher and the SQL Server fuzzy WHERE clause were already correct — the bug was that the **first search only had a partial candidate set** to match against.

## Root cause

Two paths fed search results from an incomplete candidate set:

1. **Local-mode search (default: `sidebarTableSearchLocal`)** — `loadLocalTableSearchResults` only *read* the persisted table search index (`loadSidebarTableSearchIndex`). When the index had never been built (it is built only by the manual refresh button), the UI fell back to matching against the **currently loaded first page of children** (page size 500, name-ordered). `T_Erp_Nc_SuPlan_List` sorts after hundreds of `A_Erp_*`/`T_Bas_*` tables, so it was not in the candidate set and `erpncs` missed. Once the index was built (e.g. via the refresh button), every subsequent search — `terpncs`, then `erpncs` again — matched against the complete table set and succeeded.
2. **Remote-mode search** — `loadPagedTableGroupChildren`/`loadPagedSimpleTableChildren` sent `limit = pageSize` for searches, and the SQL Server backend applied `SELECT TOP (500)` (or `filter_table_infos` `take(limit)`), truncating the fuzzy result set to the first 500 name-ordered matches. The `T_`-prefixed target sorted beyond the cutoff for the broad `erpncs` pattern (which matches far more tables than the `t`-constrained `terpncs`), so it was dropped from the first search.

## Why it only failed on the first search

- **Local mode:** the missing index was the only state difference. Before any index existed, the first search was limited to loaded children; `terpncs`/repeat `erpncs` succeeded only because the index had been built in between (manually or via children expansion). After the fix the index is built on first use, so the very first search covers the complete table set.
- **Remote mode:** every search re-queried the backend, so this was deterministic per query — the broad fuzzy pattern was truncated, the narrow one was not. The fix removes the truncation so both succeed identically on the first try.

## Fix

- `lib/sidebar/sidebarTableSearchIndex.ts` (new): `loadOrBuildSidebarTableSearchIndex` builds the persisted index when it is missing instead of returning `null`.
- `ConnectionTree.vue`: `loadLocalTableSearchResults` uses the helper, so the first local search builds the index.
- `stores/connectionStore.ts`: searches no longer truncate the fuzzy result set to the first page (`fetchLimit = undefined` for filtered loads; unfiltered loads keep the page+1 probe).

## Regression tests

- `T_Erp_Nc_SuPlan_List` + `erpncs` succeeds on the **first** local search (index build)
- Remote search returns the complete fuzzy result set (no page-size truncation), first try
- Repeated searches are order-independent: `erpncs` → `terpncs` → `erpncs`
- Case variants (`ERPnCS`/`ERPNCS`) behave identically
- Matcher behavior locked: separator-blind tiers still match `erpncs`/`terpncs`; per-word subsequence tiers and single-character handling unchanged; no new cross-word fuzzy looseness

## Validation

- New + existing vitest suites: 175 tests pass (sidebar search/control/regex-index/large-tree + connectionStore metadata-loading)
- `vue-tsc --noEmit` clean; `oxlint` clean for changed files; `oxfmt` applied
- Rust side untouched (SQL Server `sqlserver_list_tables_sql` already generates the fuzzy `%e%r%p%n%c%s%` pattern; this fix is purely the search/state path)

### DIFF
--- a/apps/desktop/src/components/sidebar/ConnectionTree.vue
+++ b/apps/desktop/src/components/sidebar/ConnectionTree.vue
@@ -40,6 +40,7 @@ import { sidebarTreeContextKey } from "@/lib/sidebar/sidebarTreeContext";
 import { createSidebarTreeRuntime, sidebarTreeRuntimeKey, type SidebarTreeRuntimeHostInstance } from "@/lib/sidebar/sidebarTreeRuntime";
 import { createSidebarPasteHandlerRegistry } from "@/lib/sidebar/sidebarPasteHandlerRegistry";
 import { insertSidebarTableSearchControls, isSidebarTableSearchControlNode } from "@/lib/sidebar/sidebarTableSearchControl";
+import { loadOrBuildSidebarTableSearchIndex } from "@/lib/sidebar/sidebarTableSearchIndex";
 import TreeItem from "./TreeItem.vue";
 import SidebarTreeRuntimeHost from "./SidebarTreeRuntimeHost.vue";
 import SidebarTreeItemDialogs from "./SidebarTreeItemDialogs.vue";
@@ -609,7 +610,17 @@ function filterGloballyIndexedRegexTables(nodes: TreeNode[]): TreeNode[] {
 
 async function loadLocalTableSearchResults(parentNodeId: string, refresh = false, focusRestore?: TableSearchFocusRestore) {
   try {
-    const entries = refresh ? await store.refreshSidebarTableSearchIndex(parentNodeId) : await store.loadSidebarTableSearchIndex(parentNodeId);
+    // A missing persisted index (null) means this scope was never indexed, so
+    // only the currently loaded first page of children is searchable. That
+    // silently misses alphabetically-late tables (e.g. "T_Erp_Nc_SuPlan_List"
+    // for "erpncs") until an explicit index refresh happens. Build the index
+    // on the first search so results always cover the complete table set,
+    // matching the behavior of an explicit refresh.
+    const entries = await loadOrBuildSidebarTableSearchIndex(
+      () => store.loadSidebarTableSearchIndex(parentNodeId),
+      () => store.refreshSidebarTableSearchIndex(parentNodeId),
+      refresh,
+    );
     localTableSearchResults.value = { ...localTableSearchResults.value, [parentNodeId]: entries };
   } finally {
     if (focusRestore) restoreTableSearchInput(focusRestore);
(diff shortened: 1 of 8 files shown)


t8y2 (changes requested):
Request changes: the correctness fix introduces unbounded work on the object-tree search hot path.

1. With local table search enabled by default, input, clear, and mode changes can immediately start building the full table index. An empty query can still scan all pages, and rapid typing can start multiple concurrent full builds because there is no in-flight deduplication.
2. Remote fuzzy search removes `limit`/`offset` for every non-empty query, sending every match through the database, IPC, store, and tree rendering path. The new test already demonstrates 801 returned objects, with no bound for larger schemas.
3. The tests do not assert request counts for empty input, rapid consecutive input, or a bounded remote result budget.

Please keep first/subsequent search behavior consistent while adding one deduplicated index build and an explicit, tested result/request budget. This needs a product/performance boundary rather than an unbounded default implementation.

t8y2 (commented):
Patched in `8136a73f6`.

Root cause: local-index and remote sidebar table-search modes owned independent debounce queues, so switching modes could leave stale work scheduled. External Doris/StarRocks catalog searches also used the normal page-size limit instead of the bounded fuzzy-search result budget.

Fix: scheduling one mode now cancels the pending job for the other mode, and both queues are cleared when search is disabled or the tree unmounts. External catalog search requests and metadata scope now consistently use the 2,000-result search budget.

Validation:
- 70/70 focused sidebar/search tests passed
- `pnpm typecheck`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 71136377e, will be released in the next version.

PR #6368, opened 2026-08-16
fix(import): merge imported connection groups into the existing sidebar

### LINKED ISSUE #6365: [Bug] 导出导入链接DBX格式 不能1:1还原
### 数据库类型和版本

MySQL

### 问题描述

1. 从Windows的一台机器导出dbx-connections.json文件
2. 到Mac电脑上导入dbx-connections.json文件
3. 结果不能导入的链接数量比之前多出了很多，同时很多链接也没有正确的放到文件夹里
4. 原本已有的其他文件夹也会被自动删掉

期望：
1. 导出的结构和链接是怎样的，导入还原时，应该1:1还原，包括目录结构和数量
2. 应该智能覆盖合并，不应该动其他无关的链接配置和文件夹以及结构

### 支持信息

DBX 版本: v0.5.85
运行环境: 桌面端
操作系统: macOS 15.7.5
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Summary

Fixes #6365 — exporting connections on one machine and importing them on another did not round-trip 1:1: folders that only existed on the target machine were dropped, and their connections got dumped at the sidebar root.

**Root cause**: `applySidebarLayout` (`connectionStore.ts`) replaced the entire sidebar layout with the imported one instead of merging. `reconcileLayout` only keeps groups present in the layout it's handed and appends any unlisted connection at the top level — so the target machine's own folders were silently discarded.

**Fix**: new `mergeSidebarLayout(current, imported)` in `sidebarLayout.ts`. It keeps the current tree, removes the imported connections from their current spots (so they move rather than duplicate), then grafts the imported tree onto it — matching folders by name at each level (so repeated imports reuse the same folder) and minting fresh ids for new groups to avoid id collisions. `applySidebarLayout` now calls `mergeSidebarLayout(sidebarLayout.value, layout)` before reconciling. This function is only reached from the import-layout confirm dialog, so the fix covers DBX-format imports as well as DBeaver/DataGrip/Navicat imports that go through the same path.

**Scope note — one reported symptom not addressed**: the issue also reports "the number of imported connections is much higher than expected." I could not reproduce this as an actual duplicate-creation bug — re-importing the same export is already deduplicated by name+host+port on unfixed code. My best read is that this was a perceived side effect of the same flattening bug (connections previously nested inside collapsed folders all surfacing at the root), which this fix addresses, but I don't have direct evidence of a separate duplication path and did not want to guess at a fix for something unreproduced. Separately (and out of scope for this PR): the existing dedup key is name+host+port only, so two connections that differ solely by database/username will merge into one on import — that's a real but different bug, happy to file a follow-up issue if useful.

## Test plan

- [x] New regression test `packages/app-tests/connectionImportRoundTrip.test.ts` — round-trips through the real `exportConnectionsToFile` (real AES-GCM encryption) and the real `importConnectionsFromFile` + `applySidebarLayout`. Fails on unfixed code (target-only folders missing, connections relocated to root); passes after the fix.
- [x] Scoped suite (`connectionImportRoundTrip` + `connectionStoreGroupedConnect` + `sidebarLayout`): 33/33 passing.
- [x] Full frontend suite: 985/985 files, 9352/9352 tests passing.
- [x] `pnpm typecheck` / `pnpm lint` clean.

### DIFF
--- a/apps/desktop/src/composables/useDialogSources.ts
+++ b/apps/desktop/src/composables/useDialogSources.ts
@@ -2,6 +2,7 @@ import { ref, watch } from "vue";
 import { useI18n } from "vue-i18n";
 import { useConnectionStore } from "@/stores/connectionStore";
 import { useToast } from "@/composables/useToast";
+import { hasSidebarLayoutEntries } from "@/lib/sidebar/sidebarLayout";
 import type { SidebarLayout } from "@/types/database";
 
 const showTransferDialog = ref(false);
@@ -310,7 +311,7 @@ export function useDialogSources() {
             : t("configExport.importNone"),
           4000,
         );
-        if (layout && count > 0) {
+        if (hasSidebarLayoutEntries(layout)) {
           pendingImportLayout.value = layout;
           showImportLayoutConfirm.value = true;
         }
@@ -325,7 +326,7 @@ export function useDialogSources() {
       const { count, layout } = await connectionStore.importConnectionsFromFile(pendingImportContent.value, passphrase);
       showConfigPassphraseDialog.value = false;
       toast(count > 0 ? t("configExport.importSuccess", { count }) : t("configExport.importNone"), 2000);
-      if (layout && count > 0) {
+      if (hasSidebarLayoutEntries(layout)) {
         pendingImportLayout.value = layout;
         showImportLayoutConfirm.value = true;
       }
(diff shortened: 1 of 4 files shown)


t8y2 (commented):
Patched in `d97c5a5bc`.

Root cause: the layout-apply confirmation was gated by `count > 0`. Re-importing an export whose connections already existed correctly returned `count = 0`, but its non-empty sidebar layout was silently skipped.

Fix: show the layout confirmation whenever the imported layout contains groups or ordering entries, independently of how many new connections were inserted.

Validation:
- 39/39 focused import round-trip tests passed
- `pnpm typecheck`

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 2f3a32955, will be released in the next version.

PR #6407, opened 2026-08-17
feat(gaussdb): add query_dop hint to count SQL for GaussDB

AndrewDi:
When building the count SQL for data grid total row calculation, add /*+ set(query_dop 16) */ optimizer hint for GaussDB databases to improve COUNT(*) performance.

## Summary

Add configurable `query_dop` optimizer hint for GaussDB `COUNT(*)` queries in the data grid. The degree of parallelism (DOP) is now configurable per connection via the connection dialog, instead of being hard-coded.

### Backend (Rust)
- **`crates/dbx-core/src/data_grid_sql.rs`**: Added `count_hint: Option<String>` field to `DataGridCountSqlOptions`. When provided, the hint is injected between `SELECT` and `COUNT(*)`. No hint is injected when the field is `None`.

### Frontend (TypeScript/Vue)
- **`apps/desktop/src/lib/database/jdbcDialect.ts`**: Added `GaussdbCountQueryDop` type (`1 | 2 | 4 | 8 | 16`), getter/setter functions (`gaussdbCountQueryDop`, `setGaussdbCountQueryDop`), and `gaussdbCountQueryDopHint()` helper that returns the hint string only when DOP > 1.
- **`apps/desktop/src/components/connection/ConnectionDialog.vue`**: Added a "COUNT(*) parallelism" Select dropdown in the GaussDB connection settings section. Options: `1 (Disabled)`, `2`, `4`, `8`, `16`. Default is `1` (no hint).
- **`apps/desktop/src/components/grid/DataGrid.vue`**: `buildCurrentCountTarget()` now reads the DOP from the connection config via `gaussdbCountQueryDopHint()`.
- **`apps/desktop/src/stores/queryStore.ts`**: `dataCountTarget` now reads the DOP from the connection config via `gaussdbCountQueryDopHint()`.

### i18n
- Added `common.disabled` key to all 8 locale files (en, zh-CN, zh-TW, ja, ko, es, it, pt-BR).
- Added `gaussdbCountQueryDop` and `gaussdbCountQueryDopHint` keys to en and zh-CN.

### Tests
- `builds_grid_count_sql` — verifies all non-GaussDB types remain unchanged (Postgres, Doris, StarRocks, Kingbase, GaussDB without hint, Postgres custom quote).
- `builds_grid_count_sql_with_optimizer_hint` — verifies GaussDB with hint output (with and without WHERE clause).
- `iris_data_grid_count_queries_the_table_without_wrapping_top_sql` — verifies Iris type unaffected.

## Behavior

| DOP Value | Generated SQL |
|-----------|---------------|
| 1 (default) | `SELECT COUNT(*) AS cnt FROM ...` |
| 2 | `SELECT /*+ set(query_dop 2) */ COUNT(*) AS cnt FROM ...` |
| 4 | `SELECT /*+ set(query_dop 4) */ COUNT(*) AS cnt FROM ...` |
| 8 | `SELECT /*+ set(query_dop 8) */ COUNT(*) AS cnt FROM ...` |
| 16 | `SELECT /*+ set(query_dop 16) */ COUNT(*) AS cnt FROM ...` |

## Notes

- The hint is only applied to GaussDB connections.
- When DOP is set to 1 (default), no optimizer hint is injected, preserving existing behavior for all other database types.
- The DOP value is stored in the connection\'s `external_config` as `gaussdbCountQueryDop`.

### DIFF
--- a/apps/desktop/src/components/connection/ConnectionDialog.vue
+++ b/apps/desktop/src/components/connection/ConnectionDialog.vue
@@ -121,13 +121,16 @@ import {
 import {
   GAUSSDB_M_JDBC_DRIVER_CLASS,
   gaussdbConnectionMode,
+  gaussdbCountQueryDop,
   gaussdbIdentifierQuoteStyle,
   gaussdbTargetServerType,
   setGaussdbConnectionMode,
+  setGaussdbCountQueryDop,
   setGaussdbIdentifierQuoteStyle,
   setGaussdbTargetServerType,
   supportsGaussdbIdentifierQuoteStyle,
   type GaussdbConnectionMode,
+  type GaussdbCountQueryDop,
   type GaussdbIdentifierQuoteStyle,
   type GaussdbTargetServerType,
 } from "@/lib/database/jdbcDialect";
@@ -547,6 +550,14 @@ const gaussdbTargetServerTypeComputed = computed<GaussdbTargetServerType>({
   get: () => gaussdbTargetServerType(form.value),
   set: (value) => {
     setGaussdbTargetServerType(form.value, value);
+    resetTestState();
+  },
+});
+
+const gaussdbCountQueryDopComputed = computed<GaussdbCountQueryDop>({
+  get: () => gaussdbCountQueryDop(form.value),
+  set: (value) => {
+    setGaussdbCountQueryDop(form.value, value);
     resetTestState();
   },
 });
@@ -8007,6 +8018,26 @@ function openExternalUrl(url: string) {
                     </Select>
                     <p class="text-xs leading-5 text-muted-foreground">
                       {{ t("connection.gaussdbTargetServerTypeHint") }}
+                    </p>
+                  </div>
+                </div>
+                <div v-if="showGaussdbConnectionMode" class="grid grid-cols-4 items-start gap-4">
+                  <Label :class="connectionLabelSmallPaddedClass">{{ t("connection.gaussdbCountQueryDop") }}</Label>
+                  <div class="col-span-3 grid gap-1">
+                    <Select v-model="gaussdbCountQueryDopComputed">
+                      <SelectTrigger class="h-9">
+                        <SelectValue />
+                      </SelectTrigger>
+                      <SelectContent>
+                        <SelectItem :value="1">1 ({{ t("common.disabled") }})</SelectItem>
+                        <SelectItem :value="2">2</SelectItem>
+                        <SelectItem :value="4">4</SelectItem>
+                        <SelectItem :value="8">8</SelectItem>
+                        <SelectItem :value="16">16</SelectItem>
+                      </SelectContent>
+                    </Select>
+                    <p class="text-xs leading-5 text-muted-foreground">
+                      {{ t("connection.gaussdbCountQueryDopHint") }}
                     </p>
                   </div>
                 </div>
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -3294,6 +3294,7 @@ async function buildCurrentCountTarget(): Promise<{ sql: string; schema?: string
       schema: props.tableMeta.schema,
       tableName: props.tableMeta.tableName,
       whereInput: currentWhereInput(),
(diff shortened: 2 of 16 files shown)


AndrewDi:
@t8y2 PTAL

t8y2 (changes requested):
Thanks for the contribution. The optimizer-hint syntax is valid, but two issues block applying it automatically to every GaussDB data-grid count query:

1. `build_data_grid_count_sql()` unconditionally forces `query_dop 16` for every GaussDB `COUNT(*)`. This overrides the instance/session tuning policy even for small tables, resource-constrained instances, or concurrent users. GaussDB documents `query_dop` as a resource-sensitive tuning parameter and recommends retaining the default unless CPU, memory, I/O, and network capacity have been evaluated. Please provide representative GaussDB measurements and define an explicit product boundary: opt-in/configurable DOP, adaptive DOP, or a demonstrated safe version/instance scope. A globally hard-coded value cannot be merged based only on the assumption that it improves performance.

2. The existing `builds_grid_count_sql` regression test currently fails because its GaussDB expectation was not updated. Please add focused coverage for the intended GaussDB output and confirm that all other database types remain unchanged.

Please also include the tested GaussDB version and before/after execution plans or timings for representative small and large filtered counts.

t8y2 (changes requested):
Thanks for the follow-up. The SQL builder is now generic and its focused tests pass, but the original resource-boundary blocker is still present at the two callers.

`apps/desktop/src/components/grid/DataGrid.vue` and `apps/desktop/src/stores/queryStore.ts` still pass `/*+ set(query_dop 32) */` for every GaussDB data-grid count. This changes the hard-coded degree from 16 to 32, but it does not make the behavior configurable or default-off, and it still applies the hint automatically without considering the instance's CPU/resource policy or measured benefit.

Please keep the existing no-hint behavior by default and only pass `countHint` through an explicit user/connection setting with a documented safe boundary. Add caller-level tests proving both the disabled/default path and the explicitly enabled path. If the intent is to enable it automatically, provide representative real-instance benchmarks and a resource-aware selection rule rather than a fixed global value.

t8y2 (commented):
Maintainer patch applied on the latest head.

- Old head: `697519e2e`
- New head: `3753528f8`
- Base checked: `4ce27f8ae`
- Root cause: `DataGrid.vue` passed its optional `connectionId` directly to `getConfig()`, so the new opt-in count-hint path failed TypeScript validation. The default-off and explicitly enabled caller behavior also lacked regression coverage.
- Patch: guard the optional connection lookup, keep GaussDB count parallelism disabled unless the connection explicitly selects a bounded DOP, and add helper plus query-store caller tests for both disabled and enabled paths.
- Validation: 2 Vitest files / 75 tests passed; `pnpm typecheck` passed; Oxfmt passed; Oxlint reported 0 errors and one pre-existing unrelated warning in `DataGrid.vue`; `git diff --check` passed.
- Dependency/UI flags: no dependency changes; the connection dialog addition is small and scoped to GaussDB.

t8y2 (commented):
Final maintainer patch review for `b2de9f4e1`.

- Starting contributor head: `697519e2e`
- Final patched head: `b2de9f4e1`
- Base checked: `4ce27f8ae`
- Behavior: GaussDB `COUNT(*)` parallelism remains disabled by default and only emits a bounded `query_dop` hint after an explicit per-connection selection.
- Fixes: guarded the optional DataGrid connection lookup that broke TypeScript validation; added helper and query-store caller regressions for disabled/enabled paths; manually completed all locale entries after the i18n autofill service returned empty responses twice.
- Validation: 2 Vitest files / 75 tests passed; `pnpm typecheck` passed after both patches; i18n autofill dry-run reports no missing new keys; Oxfmt passed; Oxlint reported 0 errors and one pre-existing unrelated warning in `DataGrid.vue`; `git diff --check` passed; merge simulation against the recorded base was clean.
- Flags: no dependency changes; small GaussDB-only connection-dialog addition, not a large UI change.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in f6b91e93d, will be released in the next version.

PR #6445, opened 2026-08-17
fix(editor): suggest MongoDB collection methods after db.<collection>.

jthanh8144:
## Description

Fixes MongoDB autocomplete not offering collection methods after typing `db.<collection>.` in the query editor.

The completion engine was returning the right items all along — the editor layer was discarding them:

1. **The editor filtered every option out.** For `db.users.`, the context is `collectionOrMethod` and `from` points at the start of the collection name (it has to, so that `db.audit.` can be rewritten into `db.getCollection("audit.logs")`). CodeMirror then filters the returned options against the document text between `from` and the cursor — `"users."` — and matches it against each `option.label`, which are bare method names (`find`, `findOne`, …). `FuzzyMatcher.match` returns `null` immediately when `word.length < pattern.length`, so all options were dropped and no popup appeared. Same for `db.users.fi`.
2. **`validFor` spanned dots.** `/["']?[\w_$.-]*$/` kept a result built for `db.` "still valid" while the user typed `users.`, so the editor went on re-filtering collection names instead of asking for methods.

Screenshot 2026-08-17 at 11 34 11
Screenshot 2026-08-17 at 11 34 57

### Changes

- `apps/desktop/src/lib/mongo/mongoCompletion.ts` — collection-qualified method items now carry `filterText` (`users.find`). The editor's existing `completionLabelPresentation` helper turns that into a matchable label while `displayLabel` still renders the bare `find`. This is the same mechanism already used for quoted collection names inside `getCollection("…")`.
- `apps/desktop/src/lib/mongo/mongoCompletion.ts` — `getMongoCompletionResultValidFor` now takes the completion context. For the `db.…` chain modes (`root`, `collection`, `collectionOrMethod`, `method`, `cursorMethod`) it allows only as many dot-separated segments as the current prefix has, so a newly typed `.` forces a fresh query. Field paths and quoted collection names keep the old dot-permissive behaviour, where a dot is part of the identifier rather than a scope change.
- `apps/desktop/src/components/editor/QueryEditor.vue` — passes the context through to `validFor`.
- `packages/app-tests/mongoCompletion.test.ts` — two regression tests.

## Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Performance improvement
- [ ] Refactor
- [ ] Documentation
- [ ] CI / build

## Frontend Changes

- [ ] This PR includes frontend changes, and a screenshot/recording is attached (below)

## Verification

Two regression tests were added in `packages/app-tests/mongoCompletion.test.ts`; both fail on `main` and pass with this change:

- `makes collection-qualified methods matchable by the prefix they replace` — asserts that for `db.users.`, `db.users.fi` and `db.audit.`, every returned item's matchable text (`filterText ?? label`) starts with the prefix it replaces. This is exactly the invariant CodeMirror's filter requires, and the one the old code violated.
- `stops reusing a completion result once the typed text gains a dot` — asserts the `validFor` window closes on a new dot in the `db.…` chain, and stays open for dots inside a quoted collection name.

Manual steps:

1. Open a query tab on a MongoDB connection.
2. Type `db.` → the collection list appears (unchanged).
3. Type a collection name, then `.` → **before:** no popup; **after:** `find`, `findOne`, `aggregate`, … appear.
4. Keep typing (`fi`) → the list narrows instead of disappearing.
5. Accept `find` → inserts `db.users.find({})`; for a name needing quoting, `db.order-items.` + `find` inserts `db.getCollection("order-items").find({})`.

- [x] `make check` passes (`format`, `lint`, `typecheck`, `test` — all ok)
- [x] `make cargo-check-fast` passes (no Rust changed; run for completeness)
- [x] Related tests pass

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -3384,7 +3384,7 @@ async function provideMongoCompletions(currentState: import("@codemirror/state")
   return {
     from: completionContext.from,
     options: items.map((item) => completionOptionForItem(item)),
-    validFor: getMongoCompletionResultValidFor(),
+    validFor: getMongoCompletionResultValidFor(completionContext),
   };
 }
 
--- a/apps/desktop/src/lib/mongo/mongoCompletion.ts
+++ b/apps/desktop/src/lib/mongo/mongoCompletion.ts
@@ -158,6 +158,9 @@ const METHOD_ARG_ROLES: Record<string, readonly MongoArgRole[]> = {
 
 const CALL_METHOD_PATTERN = new RegExp(`\\.(${Object.keys(METHOD_ARG_ROLES).join("|")})\\s*\\(`, "g");
 
+/** Modes reached by walking the `db.collection.method` chain, where a `.` switches item source. */
+const DOT_SCOPED_MODES = new Set<MongoCompletionMode>(["root", "collection", "collectionOrMethod", "method", "cursorMethod"]);
+
 /** Stages whose body is a fixed set of option keys rather than a field map. */
 const OPTION_STAGES = new Set(Object.keys(STAGE_OPTION_KEYS));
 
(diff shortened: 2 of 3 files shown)


t8y2:
Thanks for the contribution! Merged in a290948ad9, will be released in the next version.

PR #6484, opened 2026-08-17
fix(dameng): fall back to catalog views when DBMS_METADATA is unavailable

GIWTO:
## Summary
DM instances without an installed/authorized `DBMS_METADATA` system package fail every DDL fetch with `[-3325]` package/object resolution errors, so double-clicking a function/object source hard-errored. This PR makes object-source and table-DDL retrieval degrade gracefully:

- `DBMS_METADATA.GET_DDL` stays the primary path (`getObjectSource`, `getTableDdl`);
- on metadata-unavailable errors (error codes `-3325`/`-2207`, or catalog/permission message heuristics):
  - routines/packages/types: reassembled from `ALL_SOURCE` by line, falling back to `SYS.SYSOBJECTS` + `SYS.SYSTEXTS`
  - views / materialized views: `ALL_VIEWS.TEXT` (read-only body, not editable)
  - triggers: `ALL_TRIGGERS.TRIGGER_BODY` (read-only body, not editable)
  - sequences: `CREATE SEQUENCE` rebuilt from `ALL_SEQUENCES`
- when every catalog source is unavailable, a read-only placeholder explains `SP_CREATE_SYSTEM_PACKAGES(1)` instead of failing;
- connection-level errors (disconnects/timeouts) are still propagated unchanged.

## Review follow-ups included
- `TYPE` / `TYPE BODY` now map through `damengDdlObjectType`, so the catalog fallback for user-defined types is actually reachable
- blank lines are preserved when reassembling `ALL_SOURCE` / `SYS.SYSTEXTS`
- swallowed catalog-tier errors are logged at FINE level for diagnosability
- `getObjectSource` error handling unified into a single catch

## Verification
- `DamengAgentMetadataTest`: 49 tests, all green (includes new cases for function/view/trigger/sequence/type/package-body fallbacks, SYSTEXTS tier, unavailable placeholder, connection-reset propagation, and no-fallback for unrelated DBMS_METADATA errors)
- compiled and tested offline with JDK 17 against a full source rebuild of `common`/`test-support` (CI will run the real JDK 21 toolchain)


t8y2 (commented):
Maintainer patch pushed: `6c7db63d6` -> `668918557` (base `e6fa850a3`).

The fallback catalog views can return a bare routine declaration rather than executable DDL. The patch wraps recognized declarations with `CREATE OR REPLACE` and keeps unrecognized source read-only instead of producing unsafe editable DDL.

Validation: `./gradlew :dameng:test --tests com.dbx.agent.dameng.DamengAgentMetadataTest` completed successfully, and the final base-to-head diff is clean.

t8y2 (approved):
Maintainer patch verified on the exact head. Dameng fallback metadata tests and required CI pass.

t8y2:
Thanks for the contribution! Merged in 99066319c, will be released in the next version.

PR #6494, opened 2026-08-17
fix(settings): report apply failures and keep Apply & Close from stalling

### LINKED ISSUE #6485: [Bug] 设置里面修改编辑器字体后，无法点击应用和应用并关闭
### 数据库类型和版本

MySQL 8.0

### 问题描述

设置里面修改编辑器字体后，无法点击应用和应用并关闭

### 支持信息

DBX 版本: v0.5.86
运行环境: 桌面端
操作系统: Windows 10.0.22631.5909
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

Image

0verme:
## Summary

Fixes #6485.

The Settings dialog's **Apply** and **Apply & Close** buttons could appear to "do nothing" after editing the editor font: clicking them had no visible effect and **Apply & Close** would not close the dialog, with no error shown.

## Root cause

Both buttons call `persistSettings()`, which has several awaited persistence steps (editor-settings save, desktop-settings save) and **no error handling**. When any persistence step rejects (a failed save):

- **Apply** fails silently — no toast, no feedback, so it looks unresponsive.
- **Apply & Close** only runs `closeSettings()` *after* `await persistSettings()`, so on a rejected save the dialog is never closed and there is no feedback at all.

That is exactly the "click Apply / Apply & Close and nothing happens / the dialog won't close" behavior reported when changing the editor font.

## Changes

- `apps/desktop/src/components/editor/EditorSettingsDialog.vue`
  - Extract a single shared apply entrypoint (`applySettingsForResult()`) used by **both** **Apply** and **Apply & Close** (single source of truth — no duplicated save logic).
  - Wrap `persistSettings()` in try/catch; on failure show a `settings.applyFailed` toast instead of silently swallowing it.
  - **Apply & Close** now only calls `closeSettings()` once persistence succeeds, so it can no longer leave the dialog stuck open. On failure it stays open so the user can retry (and sees the error).
- Add the `settings.applyFailed` message to all 8 locale files (en, zh-CN, zh-TW, es, it, ja, ko, pt-BR).
- Add regression coverage: `EditorSettingsDialog.applyError.spec.ts` asserts both buttons share the guarded entrypoint and that close only happens after a successful apply.

## Testing

- New regression spec: 3 tests **fail before** the fix, **pass after**.
- Editor + settings unit suites: 98 tests pass.
- `pnpm typecheck` passes; `pnpm lint` introduces no new findings; `git diff --check` clean; files formatted with `oxfmt`.

Fixes #6485

### DIFF
--- a/apps/desktop/src/components/editor/EditorSettingsDialog.vue
+++ b/apps/desktop/src/components/editor/EditorSettingsDialog.vue
@@ -984,13 +984,34 @@ async function persistSettings() {
   }
 }
 
+function applySettingsErrorToast(error: unknown) {
+  toast(t("settings.applyFailed", { error: error instanceof Error ? error.message : String(error) }), 5000);
+}
+
+// Shared apply entrypoint used by both "Apply" and "Apply & Close". Persists the
+// current draft and reports failure explicitly instead of silently swallowing it
+// (persistSettings() has several awaited persistence steps with no try/catch, so a
+// rejected save used to leave "Apply" silent and made "Apply & Close" unclosable with
+// no feedback). Returns whether persistence succeeded so the close path only proceeds
+// once the settings are actually saved.
+async function applySettingsForResult(): Promise<boolean> {
+  try {
+    await persistSettings();
+    return true;
+  } catch (error) {
+    applySettingsErrorToast(error);
+    return false;
+  }
+}
+
 async function applySettings() {
-  await persistSettings();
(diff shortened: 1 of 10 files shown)


t8y2:
Thanks for the contribution! Merged in cddc6a2, will be released in the next version.

PR #6497, opened 2026-08-17
fix(transfer): normalize Oracle CHAR length semantics for MySQL

### LINKED ISSUE #6479: [Compatibility] ORACLE 12C 到MySQL8.0 数据传输时脚本转换执行报错，期望可以正常转换执行
### 数据库类型和版本

ORACLE 12C、MySQL8.0 

### 问题描述

ORACLE 12C 到MySQL8.0 数据传输时脚本转换执行报错，期望可以正常转换执行

### 错误信息

```shell
Failed to create table: Server error: `ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'char),
  `ID` VARCHAR(6 char),
  `DWMC` VARCHAR(50 char),
  `JOB` VARCHAR(20 cha' at line 2'
```

### 支持信息

DBX 版本: v0.5.86
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 补充信息

_No response_

0verme:
## Summary

Fix Oracle → MySQL table transfer when Oracle character columns use an explicit
length unit, such as `VARCHAR2(50 CHAR)`.

Previously the Oracle-specific length qualifier leaked into the generated MySQL
DDL as `VARCHAR(50 char)`, which MySQL rejects with:

```
ERROR 1064 (42000): ... near 'char),
  `ID` VARCHAR(6 char),
  `DWMC` VARCHAR(50 char),
  `JOB` VARCHAR(20 cha' ...
```

The type rewrite path now emits valid target types such as `VARCHAR(50)`.

## Root cause

Oracle character columns can declare their length in characters or bytes.
The Oracle agent reads `ALL_TAB_COLUMNS.CHAR_USED` and, when a column uses
`CHAR` semantics (`VARCHAR2(50 CHAR)`), reports the data type with the unit
qualifier attached:

```
agents/drivers/oracle-go/main.go  oracleColumnTypeDDL / oracleCharacterLengthUnit
```

During a data transfer that creates the target table, the column type flows
through:

```
Oracle agent columnInfo.data_type  →  ColumnInfo.data_type  →  transfer
generate_create_table_ddl  →  postgres_column_type_sql  →  map_column_type
```

In `map_column_type` (transfer.rs) the `varchar2` branch took the parenthesized
length slice verbatim and forwarded it to the target, so `VARCHAR2(50 CHAR)`
became `VARCHAR(50 char)` for MySQL. The `char`/`nchar` branch had the same
problem (`CHAR(1 CHAR)` → `CHAR(1 char)`).

## Fix

Added a shared normalization helper `normalize_len_params` in
`sql_dialect/type_rewrite.rs` that strips an Oracle trailing `CHAR`/`BYTE`
length-unit qualifier from a single parenthesized length parameter while
preserving the numeric length, and used it in `map_column_type` for the
character-type branches.

- `VARCHAR2(50 CHAR)` → `VARCHAR(50)` (the issue scenario)
- `CHAR(10 CHAR)` → `CHAR(10)`
- `CHAR(10)` → `CHAR(10)` (untouched — a real `CHAR` type is not harmed)
- `NUMBER(10,2)` / `DECIMAL(10,2)` → unchanged (precision/scale not disturbed)
- `(10,2)` and `('a','b')` parameter lists are never touched

The fix is applied at the character-length normalization layer, not as
`source == Oracle && target == Mysql` string hacks, so it works for any
target that receives `VARCHAR(n)` / `CHAR(n)`.

## Tests

Executed and passing locally (`cargo test -p dbx-core --lib`):

- `sql_dialect::type_rewrite::tests::normalize_len_params_strips_oracle_char_unit`
- `sql_dialect::type_rewrite::tests::normalize_len_params_leaves_non_unit_params_untouched`
- `sql_dialect::type_rewrite::tests::normalize_len_params_handles_malformed_input`
- `transfer::tests::map_column_type_oracle_char_semantics_to_mysql`
- `transfer::tests::map_column_type_oracle_plain_varchar2_to_mysql`
- `transfer::tests::map_column_type_keeps_real_char_type_untouched`
- `transfer::tests::map_column_type_oracle_byte_semantics_locked`
- `transfer::tests::map_column_type_preserves_numeric_precision_scale`
- `transfer::tests::oracle_to_mysql_create_table_ddl_strips_char_length_units`
  (issue end-to-end DDL regression)

Plus the full `transfer::`, `sql_dialect::`, and `map_column_type` unit suites
(all green).

## Notes on Oracle BYTE semantics

MySQL has no byte-based length semantics; `VARCHAR(n)` is always character
length. This PR conservatively normalizes the unit qualifier so the generated
DDL stays valid (`VARCHAR2(50 BYTE)` → `VARCHAR(50)`). It does not attempt to
claim BYTE and CHAR are equivalent — that behavior is locked in by a
dedicated test (`map_column_type_oracle_byte_semantics_locked`). Preserving
byte semantics exactly would require a broader metadata/length-conversion
change that is out of scope for this issue.

Fixes #6479


t8y2 (commented):
Maintainer patch pushed: `757f5a912` -> `c2ba0fcce` (base `e6fa850a3`).

The original normalization correctly handled transfers to MySQL-compatible targets, but it also removed Oracle `CHAR`/`BYTE` length semantics when the target remained Oracle-family. The patch now preserves those semantics for Oracle-family targets and keeps normalization for non-Oracle targets.

Validation: `cargo test -p dbx-core --no-default-features map_column_type_` passed (15 tests), and the final base-to-head diff is clean.

t8y2 (commented):
CI formatting follow-up pushed after `c2ba0fcce`.

`cargo fmt --check` required the Oracle-family target test expression to use repository formatting. This is formatting-only; the transfer behavior is unchanged.

Validation: `cargo fmt --check` passed and the focused `map_column_type_` suite passed again.

t8y2 (commented):
Mainline sync pushed after `#6535` fixed the unrelated MiniMax incremental replay regression that failed this PR's previous Rust CI.

The Oracle transfer patch is unchanged. `cargo fmt --check`, all 15 `map_column_type_` regressions, and the formerly failing MiniMax regression pass on this exact local head.

t8y2 (approved):
The Oracle-family length semantics patch, maintainer regression, formatting, and exact-head Rust checks are complete.

t8y2:
Thanks for the contribution! Merged in 4d32bc498, will be released in the next version.

PR #6524, opened 2026-08-18
fix(editor): let global shortcuts win over CodeMirror's built-in Mod-[/Mod-]

### LINKED ISSUE #6418: [Bug] 切换侧边栏的快捷键在SQL库的查询标签中失效，但是鼠标点一下侧边栏才可以正常使用
### 数据库类型和版本

MySQL8.0

### 问题描述

Image

### 支持信息

DBX版本：0.5.86
运行环境：桌面端
操作系统：macOS26.6.1
系统架构：aarch64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

Fixes #6418.

"Toggle sidebar" (a global-scope shortcut, default `Mod+B`) stops working once the user rebinds it to `Mod+[`/`Mod+]` and the SQL editor has focus. It only starts working again after clicking somewhere outside the editor.

## Root cause

Global-scope shortcuts are dispatched from `App.vue`'s window-level `handleKeydown`, which bails out immediately if `event.defaultPrevented` is already `true`. CodeMirror's own `defaultKeymap` (from `@codemirror/commands`) hard-codes `Mod-[` → `indentLess` and `Mod-]` → `indentMore`. When the editor has focus and a user rebinds a global action to one of those combos, CodeMirror's own keymap intercepts the key first and calls `preventDefault()` before the event ever reaches the window listener, so the global shortcut silently never fires.

This is the same class of bug already fixed for `Mod-d` in `codemirrorSearchKeymap.ts` (`searchKeymapWithoutModD`, referenced in that file as fixing #4544) — CodeMirror's raw hard-coded keymap wins over DBX's configurable shortcut system.

## Solution

Added `defaultKeymapWithoutIndentBrackets()` (`apps/desktop/src/lib/editor/codemirrorDefaultKeymap.ts`), mirroring the existing `searchKeymapWithoutModD` pattern, and wired it into `QueryEditor.vue`'s keymap assembly.

`indentLess`/`indentMore` are already exposed as DBX's own configurable shortcuts (`shortcutRegistry.ts`, default `Shift+Tab` / unbound) and wired independently in `runKeymapExtension`, so dropping CodeMirror's raw `Mod-[`/`Mod-]` bindings does not remove any functionality — it just stops them from silently shadowing whatever the user has bound a global action to.

## Testing

- New `codemirrorDefaultKeymap.spec.ts`: real `EditorView` + `runScopeHandlers` (same technique as `codemirrorSearchKeymap.spec.ts`) proves the stock `defaultKeymap` consumes `Ctrl+[`/`Ctrl+]` (baseline reproduces the bug) and that the filtered keymap no longer does.
- Real end-to-end repro/fix in a running dev build (Playwright + Chromium): rebound "Toggle sidebar" to `Cmd+[` in Settings, focused the SQL editor, typed SQL, pressed `Cmd+[` — before the fix the sidebar didn't move; after the fix it toggled correctly with the exact same steps.
- Regression check: typing a literal `[...]` in the editor still just inserts text (closeBrackets still works), and `Shift+Tab` (the default "Indent less" binding) still outdents correctly.
- `pnpm typecheck`, `pnpm exec oxlint`, and the full `pnpm exec vitest run apps/desktop/src` suite (713 files / 6668 tests) all pass.

## Breaking changes

None. `Mod-[`/`Mod-]` were never DBX's documented/configurable bindings for indent-less/indent-more (those default to `Shift+Tab` / unbound), so no user-facing default behavior changes.

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -106,7 +106,7 @@ import { EDITOR_FONT_FAMILY_CSS_VAR, EDITOR_FONT_SIZE_CSS_VAR, editorDiagnosticC
 import { createStatementGutterMarkerDom, shouldShowStatementGutter } from "@/lib/editor/codemirrorStatementGutter";
 import { createQueryEditorSearchKeymap } from "@/lib/editor/queryEditorSearchKeymap";
 import { searchKeymapWithoutModD } from "@/lib/editor/codemirrorSearchKeymap";
-import { defaultKeymapWithoutIndentBrackets } from "@/lib/editor/codemirrorDefaultKeymap";
+import { defaultKeymapForGlobalShortcuts } from "@/lib/editor/codemirrorDefaultKeymap";
 import { appendSqlCompletionSpace } from "@/lib/editor/sqlCompletionInsertion";
 import { compareSqlCompletions, completionLabelPresentation } from "@/lib/editor/sqlCompletionPresentation";
 import { clampEditorFontSize, createEditorZoomCommitScheduler, fontSizeFromGestureScale, fontSizeFromWheelDelta } from "@/lib/editor/editorZoom";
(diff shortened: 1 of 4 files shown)


q396921921:
- Before (bug reproduced), full local path: 
before_1_fails_while_editor_focused
- After (same steps, bug is gone), full local path: 
after_1_fails_while_editor_focused

t8y2 (commented):
Maintainer patch pushed in `47a6388`.

Root cause: the PR removed CodeMirror's `Mod+[` and `Mod+]` bindings unconditionally, so the reported global-shortcut conflict was fixed at the cost of deleting the editor's existing indentation shortcuts even when no global action used those keys.

Fix: suppress only the CodeMirror binding that conflicts with the current global shortcut configuration. Without a conflict, the original indent/outdent behavior remains available, and the editor keymap reacts to later shortcut configuration changes.

Checks: 25 focused Vitest cases passed across macOS Meta, Windows/Linux Ctrl, single-line and multi-line indentation, and global-conflict precedence. `vue-tsc`, `oxfmt`, `oxlint`, and `git diff --check` passed. No dependencies were added.

t8y2:
Thanks for the contribution! Merged in f0cc5aefc, will be released in the next version.

PR #6539, opened 2026-08-18
feat(agent): add cancellation for driver installs and batch upgrades

Abeautifulsnow:
## Summary

Adds full cancellation support for agent driver installs across both entry points:

1. **New-connection flow** — the SQL Server legacy-compatibility driver download can now be cancelled from the install modal.
2. **Driver Store** — batch "Update All" can be cancelled as a whole (cancel-all aborts in-flight downloads, queued ones never start), and each individual driver inside a batch can be cancelled independently.

Cancellation is scoped per **operation id** (not per `db_type`), so cancelling one driver never leaks a stale token into another operation, and concurrent same-driver installs stay isolated.

## Motivation

Driver downloads (especially legacy-compat agents) can take a long time with no way out. Users were forced to wait for a multi-driver batch to finish or to restart the whole flow; a slow/stalled download held the driver list hostage with no Cancel affordance.

### Backend (`crates/dbx-core`, `crates/dbx-web`, `src-tauri`)
- **Cancel button / API** — new `cancelAgentInstall`, `cancelAgentUpgradeAll`, and `AgentUpdateBlockers` routes + Tauri commands; operation tokens registered before install begins.
- **Operation-scoped tokens** — `begin_install_cancellation` keys by operation id; `finish_install_cancellation` removes only its own token (`Arc::ptr_eq` guarded) so a fresh retry is never clobbered by a stale cancel.
- **Cancel-all for batches** — a batch-level token threads through every driver download, so "cancel all" interrupts a driver whose download already started.
- **Prompt cancellation of stalls** — downloads race every chunk against the token; stalled HTTP requests and `lock_or_cancel` waits on the per-driver lock abort promptly instead of waiting for timeouts.
- **Cancellable registry fetch** — the registry fetch (`fetch_registry_from_claimed`) is raced against cancellation too, so the whole pipeline honors cancel end-to-end.
- **Staged raw downloads** — raw artifacts download into `drivers/<db_type>/.staging/` and only commit to the live driver path after the final cancellation check; a cancel after a cache-hit download preserves the previously installed driver and state.
- **Harden the cancel boundary** — once the staged artifact is atomically committed, a newly arrived cancel no longer rolls back (deleting a working driver would lose the prior installation).

### Frontend (`apps/desktop`)
- **Connection modal** — Cancel button on the driver install dialog; operation-scoped guards prevent a stale cancelled promise from resetting a newer retry's dialog.
- **Driver Store** — per-driver cancel button (`cancelDriverInstall`) and a batch cancel button (`cancelUpgradeAll`) that appears while "Update All" runs.
- **`preparingUpgradeAll` state** — Cancel is only exposed after the backend operation has registered its token (avoids a silently lost cancel during the client-side blocker check); during preparation, per-driver install/update/import/uninstall/queue actions are disabled.
- **Progress routing** — per-driver progress routed by operation id; batch counter increments on each `done` step.

## Cancellation guarantees
- A cancel during download → partial artifact removed, installed driver/state untouched, error surfaces as "canceled by user".
- A cancel during lock wait / stalled request → aborts promptly (no waiting out timeouts).
- A cancel after the artifact is committed → install persists normally (no loss of a working driver).
- Cancel-all during a batch → in-flight downloads stop; queued installs never start.

## Testing
- New regression test `registry_install_cancel_after_cached_download_preserves_existing_driver` (cancel after cache-hit download preserves old driver + state).
- `cargo test --no-default-features -p dbx-core --lib agent_registry_install_tests`: **39 passed / 0 failed** (after merging latest `main`).
- `cargo check -p dbx-core --no-default-features`: passes.
- Pre-commit hooks (lint-staged + oxfmt + Rust fmt) pass on every commit.

## Commits
- `feat(agent): add cancel button to abort driver install downloads`
- `fix(agent): scope install cancellation by operation id, not db_type`
- `fix(agent): cancel-all aborts in-flight batch downloads`
- `fix(agent): abort stalled requests and lock waits promptly on cancel`
- `fix(agent): make registry fetch cancellable on install cancel`
- `fix(agent): stage raw driver downloads and harden cancel boundary`

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

e6db76e5-9051-4b6e-b273-ecd2c568cad1

c8e4ee1d-1d92-4a78-b170-e46f915d05f8

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过


t8y2 (commented):
I applied a focused maintainer follow-up in `3c703b333` before merge.

Patch: Align install cancellation with the agent commit boundary and UI progress state.

Validation: 13 focused frontend tests, 41 Agent Registry tests, and 2 commit-boundary tests passed; typecheck and formatting passed.

The PR head and contributor fork ref now both point to the patched commit. Exact-head CI still needs to complete successfully before squash merge.

t8y2:
Thanks for the contribution! Merged in c712be3e0, will be released in the next version.

PR #6552, opened 2026-08-18
feat(objects): add clear button to database and object browser search

### LINKED ISSUE #6476: [Feature] 数据库列表展示窗口表格搜索框增加一键清除按钮
### 数据库类型和版本

Mysql

### 支持信息

DBX 版本: v0.5.86
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 当前痛点

数据库列表展示窗口进行表搜索时无法一键清除搜索内容。如下图

Image

### 期望方案

最好是增加一键清除的按钮

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须做

### 补充信息

_No response_

Abeautifulsnow:
## Problem

The database browser and object browser search fields have no visible way to
clear the active filter. Users must manually select the text and delete it.
Additionally, the database browser's global Cmd/Ctrl+F focus path is a silent
no-op. (Closes #6476)

## Root cause

The search inputs only supported clearing via the keyboard. On the database
browser, the template ref attached to the `Input` component resolves to the
component's closed public instance, which has no `.focus()` method, so both
the new clear action and the existing `focusSearch()` shortcut were silently
dropped.

## Fix

- Add a clear (`X`) button inside both search inputs (`v-if="search"`,
  `type="button"`, `aria-label` from `common.clear`), plus right padding
  (`pr-6`) so typed text never runs under the button.
- Clicking the button clears the query and refocuses the input.
- In `ObjectBrowser`, extract the shared clear logic so the Escape shortcut and
  (`pr-6`) so typed text never runs under the button.
- Clicking the button clears the query and refocuses the input.
- In `ObjectBrowser`, extract the shared clear logic so the Escape shortcut and
  the new button use the same handler.
- In `DatabaseBrowser`, resolve the input ref to the actual DOM element via
  `.$el?.focus()` (the pattern already used by `RedisValueViewer`), which fixes
  both the new refocus and the previously broken `focusSearch()`.

## Compatibility

- Default behavior unchanged; the button appears only while text is entered.
- Escape-key behavior in the object browser is preserved (clear + keep focus).
- Pure UI change — no database/driver/platform impact.

## Validation

- `npx vitest run apps/desktop/src/components/objects/__tests__/SearchClearButtons.spec.ts` — passed
- `npx vitest run apps/desktop/src/components/objects/__tests__/` — passed (9 files / 27 tests)
- `npx vue-tsc --noEmit --project apps/desktop/tsconfig.json` — passed
- `npx oxlint --vue-plugin apps/desktop/src/components/objects/{DatabaseBrowser,ObjectBrowser}.vue apps/desktop/src/components/objects/__tests__/SearchClearButtons.spec.ts` — passed
- `git diff --check` — passed
- lint-staged hook (oxfmt + oxlint) on commit — passed

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

323a65a3-598e-41f2-9aa0-a4830f7266d9

## 验证

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

Related #6476

### DIFF
--- a/apps/desktop/src/components/objects/DatabaseBrowser.vue
+++ b/apps/desktop/src/components/objects/DatabaseBrowser.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { computed, onBeforeUnmount, ref, watch } from "vue";
-import { ArrowDown, ArrowUp, Database, GripVertical, LayoutGrid, List, Loader2, RefreshCw, Search } from "@lucide/vue";
+import { ArrowDown, ArrowUp, Database, GripVertical, LayoutGrid, List, Loader2, RefreshCw, Search, X } from "@lucide/vue";
 import { useI18n } from "vue-i18n";
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
@@ -32,7 +32,7 @@ const props = defineProps<{
 const { t } = useI18n();
 const queryStore = useQueryStore();
 const settingsStore = useSettingsStore();
-const searchInput = ref<HTMLInputElement>();
+const searchInput = ref<InstanceType<typeof Input>>();
 const search = ref("");
 const rows = ref<DatabaseRow[]>([]);
 const loading = ref(false);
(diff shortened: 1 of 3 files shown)


t8y2:
Thanks for the contribution! Merged in aeba7d81a, will be released in the next version.

PR #6553, opened 2026-08-18
fix(postgres): prevent duplicate columns in generated DDL

### LINKED ISSUE #6547: [Bug] 表的ddl显示了三个重复的字段
### 数据库类型和版本

PostgreSQL 18.3

### 问题描述

Image

这张表显示了三个一样的字段，导出为sql、或者比较构架的方式执行的话是会失败的

我用navicat打开表结构是正常的：

Image

### 支持信息

DBX 版本: v0.5.87
运行环境: 桌面端
操作系统: Windows 10.0.26200.8246
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

0verme:
## Summary

PostgreSQL column metadata queries could return the same `pg_attribute` column multiple times. `POSTGRES_COLUMNS_SQL`, `POSTGRES_COLUMNS_COMPAT_SQL`, and their batched `postgres_columns_for_relations_*` siblings left-join `pg_depend` on every `pg_class` dependency pointing at a column (`deptype IN ('a', 'i')`). A serial/identity column whose table also has secondary indexes collects one extra row per index, so `get_columns()` / `get_columns_for_relations()` returned the same column any number of times.

Those duplicate `ColumnInfo` rows propagated into generated table DDL, database exports, and schema comparison, producing duplicated column definitions that fail to replay.

## Root cause

The `pg_depend` join in the modern column queries (added by `a8f966dcf` for the #6405 serial-metadata fix) widened the dependency scan to `dep.deptype IN ('a', 'i')` to catch both the serial-owned sequence (`'a'`, AUTO) and the identity internal sequence (`'i'`, INTERNAL).

The problem: an ordinary secondary index also declares an AUTO (`'a'`) dependency on its table's columns. So a column can legitimately match many `pg_depend` rows — one for its owned sequence and one per index:

```
attnum | attname | dep object                 | deptype
     1 | id      | issue6547_repro_id_seq     | a   (sequence, relkind 'S')
     1 | id      | idx_issue6547_repro_id     | a   (plain index, relkind 'i')
```

A column dependency is not a sequence dependency. `pg_depend` cardinality for a column is one-to-many even when the column is a `serial`; only one of those rows refers to a sequence. The Rust layer maps every result row to an independent `ColumnInfo` (no `attnum`/identity grouping), so the fan-out became N identical `ColumnInfo` entries.

The `serial_dep` join in the compat queries had the same shape (`deptype = 'a'` without requiring the dependent object to be a sequence), and while the `serial_seq.relkind = 'S'` filter kept the *detection* correct, the row count was still amplified.

This is a regression from `a8f966dcf` ("fix(postgres): preserve serial columns in copied table DDL"): before it, the join was `deptype = 'i'`, which did not match plain-index AUTO dependencies. The original test only asserted SQL text fragments (`sql.contains("dep.deptype IN ('a', 'i')")`) and never exercised the query against a real catalog, so the cardinality break was not caught.

## Fix

Restrict the column-side `pg_depend` joins to dependency rows whose dependent object is an actual sequence — `relkind = 'S'` in `pg_class`:

```sql
LEFT JOIN pg_depend dep ON dep.classid = 'pg_catalog.pg_class'::pg_catalog.regclass
  AND dep.refclassid = 'pg_catalog.pg_class'::pg_catalog.regclass
  AND dep.refobjid = a.attrelid AND dep.refobjsubid = a.attnum AND dep.deptype IN ('a', 'i')
  AND dep.objid IN (SELECT oid FROM pg_catalog.pg_class WHERE relkind = 'S')
```

Applied to all four query paths:

| Query | Purpose |
| --- | --- |
| `POSTGRES_COLUMNS_SQL` | single-table get_columns (modern) |
| `POSTGRES_COLUMNS_COMPAT_SQL` | single-table get_columns (compat / PG10) |
| `postgres_columns_for_relations_sql()` | batched relation metadata (partition DDL, modern) |
| `postgres_columns_for_relations_compat_sql()` | batched relation metadata (partition DDL, compat) |

This fixes metadata cardinality at the source — one `pg_attribute` column now yields exactly one `ColumnInfo` — instead of deduplicating generated DDL downstream. Serial (`smallserial`, `serial`, `bigserial`) and identity detection are unaffected: the owned sequence dependency for a serial column is `deptype 'a'` with `relkind 'S'`, and an identity column's internal sequence is `deptype 'i'` with `relkind 'S'`, so both still match the narrowed join.

## Regression coverage

- Live catalog reproduction on PostgreSQL 16 and 18 (`serial PRIMARY KEY` + secondary index): before the fix the DBX metadata query returned 3 rows for a 2-column table; after the fix each column returns exactly 1 row.
- New live tests (run against real PostgreSQL through `DBX_TEST_POSTGRES_URL`):
  - `postgres_column_metadata_dependency_joins_return_each_column_once` — `get_columns()` returns each of 9 columns exactly once (serial PK + 2 secondary indexes + unique index); `serial`/`smallserial`/`bigserial`/identity extras intact; ordinary `integer`/`text`/`varchar`/`numeric`/`timestamp` metadata unchanged; `pg_ddl()` renders each column definition exactly once.
  - `postgres_column_metadata_compat_query_returns_each_column_once` — direct `POSTGRES_COLUMNS_COMPAT_SQL` check.
  - `postgres_batched_column_metadata_returns_each_column_once` — `get_columns_for_relations()` (the partition-DDL metadata path) plus its compat query stay duplicate-free.
- Existing #6405 regression `postgres_owned_serial_table_ddl_round_trips_without_external_sequences` and the partition-DDL live tests (`postgres_partition_metadata_renders_replayable_children_and_subpartitions`, `postgres_partition_ddl_uses_foreign_table_syntax_for_foreign_partitions`) pass on PostgreSQL 16 and 18.
- `cargo test -p dbx-core --lib` (db::postgres 184, schema 451, transfer 214, database_export 69, schema_diff 209), `cargo fmt --check`, and `cargo clippy -p dbx-core --lib -- -D warnings` all pass.

Fixes #6547


t8y2 (commented):
Patched in 236ae8b72b.

The sequence-only filter still allowed multiple `OWNED BY` sequences to fan out one column into multiple metadata rows. The modern queries now select exactly one semantic dependency per column: identity columns use only the internal identity dependency, while serial columns require the column default to depend on the selected owned sequence. The PostgreSQL 9.x compatibility queries use an equivalent correlated scalar subquery instead of `LATERAL`.

Validation:
- focused SQL-shape Rust tests
- live PostgreSQL 18 metadata tests for modern, compatibility, and batched paths
- live PostgreSQL 9.2 metadata tests for compatibility and batched fallback paths
- `git diff --check`

The implementation keeps a single metadata request per operation and bounds the catalog dependency lookup to one row per column.

t8y2:
Thanks for the contribution! Merged in 6f5ec6c81, will be released in the next version.

PR #6582, opened 2026-08-18
feat(s3): add S3-compatible object storage connections

perundio:
## Summary
- Add native S3-compatible object storage connections for AWS S3, MinIO, and other SigV4-compatible endpoints across desktop and web.
- Finish the S3 browser with bucket management, folder/prefix creation, paginated object listing, upload/download/preview/delete, and copy/move/rename workflows.
- Harden the implementation for merge by moving temporary S3 session tokens into the existing secret/session lifecycle, fixing SigV4 path canonicalization for special-character keys, rebasing onto the latest `main`, and expanding targeted test coverage.

## What’s Included
- Connection dialog support for endpoint, region, addressing style, optional session token, and read-only mode.
- Native S3 runtime integration in `dbx-core`, `dbx-web`, and Tauri commands.
- S3 browser UI with:
  - list/create/delete bucket
  - create folder/prefix
  - paginated `ListObjectsV2` browsing with `NextContinuationToken`
  - upload, download, preview, delete object
  - copy, move, and rename object
- Updated S3 driver branding/icon.
- Secret-handling fixes so temporary S3 session tokens do not persist in plaintext when `save_password = false`.
- SigV4 fixes for already-escaped paths plus regression coverage for path-style and virtual-hosted URLs with spaces, Unicode, `%`, and `?` in keys.

## Test Plan
- [x] `cargo test -p dbx-core s3:: --lib`
- [x] `cargo test -p dbx-core session_credentials::tests --lib`
- [x] `pnpm typecheck`
- [x] `pnpm vitest run apps/desktop/src/components/s3/__tests__/S3Browser.spec.ts`
- [ ] Manual smoke test against local MinIO / AWS-compatible endpoint
  - connect using endpoint + credentials
  - create/delete bucket
  - create prefix, upload/download/preview/delete object
  - copy/move/rename object
  - browse paginated folders with 200+ entries
  - verify read-only mode blocks write operations

## Notes For Review
- The PR head has been rebased onto the current `main` and the previous merge conflict state is resolved.
- The review-requested security and signing fixes are included in the latest head commit.


t8y2 (changes requested):
I'm requesting changes for now. Three issues need fixing before this is safe to merge:

1. `ConnectionDialog.vue` stores the temporary S3 session token in `external_config`. The existing persistence, no-save-password, and cloud-sync scrubbers do not treat this field as a secret, so it can remain in plaintext configuration and shared runtime state. Please move it through the same secret-store/session-isolation lifecycle as the secret access key.

2. `sigv4.rs` passes the already escaped `Url::path()` through `percent_encode_path` again. Keys containing spaces, Unicode, `%`, `?`, or similar characters are therefore signed with `%25` escapes that differ from the transmitted path, causing `SignatureDoesNotMatch`. Canonicalize the raw key exactly once and add path-style and virtual-hosted regression tests with special-character keys.

   The `url` crate documents that `Url::set_path` percent-encodes its input, while AWS SigV4 requires the canonical URI to match the transmitted absolute path: https://docs.rs/url/2.5.8/url/struct.Url.html#method.set_path and https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html

3. `S3Browser.vue` always requests 200 entries with a null continuation token and ignores `isTruncated`/`nextContinuationToken`, so larger folders silently hide every later object and common prefix. Add load-more or bounded page accumulation with a 201+ object/prefix test. AWS documents that truncated `ListObjectsV2` results must continue with `NextContinuationToken`: https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#AmazonS3-ListObjectsV2-response-NextContinuationToken

GitHub also currently reports this head as conflicting, while the main CI workflow stopped at `action_required` without running any jobs.

perundio:
Thanks for the detailed review and for calling out the risky parts so precisely.

I’ve addressed the requested changes in the latest head:
- moved the temporary S3 session token into the existing secret-store / session-isolation flow so it no longer stays in plaintext config when `save_password = false`
- fixed SigV4 canonical URI handling so already-escaped paths are signed exactly as transmitted, and added regression tests for path-style and virtual-hosted URLs with special-character keys
- updated `S3Browser.vue` to continue truncated `ListObjectsV2` results via `NextContinuationToken`, and added a 201+ object/prefix UI test
- rebased the branch onto the latest `main` and resolved the previous conflict state

Also took the opportunity to finish the missing object-browser workflows in the same pass: bucket delete, folder/prefix creation, download, and copy/move/rename operations.

Thanks again for the careful review.

ysicing:
@perundio  Does this PR intend to include S3 deep-link support (dbx://connection/new?...)?

perundio:
@ysicing No — this PR does **not** intend to add S3 deep-link support (`dbx://connection/new?...`).

Scope here is the S3/MinIO connection type plus the object browser (list/upload/download/copy/move/rename/delete, session-token handling, etc.). Deep-link / `connectionUrl` scheme support for S3 would be a separate follow-up if we want it.

antwa:
Hi @t8y2 (and @perundio) — I noticed this PR while working on a similar native S3-compatible object storage driver on my own fork, and wanted to check in on its status before continuing further.

Is this PR still actively being reviewed, or is there a rough timeline for merging it? I want to avoid duplicating effort if this lands soon — happy to hold off on my own implementation, or coordinate if there's anything useful 😄 

here is my current implementation 
image
image
image

perundio:
@t8y2 Rebased onto the latest `main` again and pushed a CI-fix commit (`25198a96`):

- resolved merge conflicts in `AppTabBar.vue`, `ConnectionDialog.vue`, `agent_driver.rs`, `ddl_profile.rs`
- applied `cargo fmt`
- synced `agents/drivers/duckdb/Cargo.lock`
- fixed `ConnectionConfig` test fixtures for new main fields

The three review items from the earlier CHANGES REQUESTED should still be covered in this head (session-token secret lifecycle, SigV4 path canonicalization + tests, S3 browser pagination/load-more).

Could you take another look when CI finishes? Happy to address anything else.

HKMV:
很需要，快合进来了吗

t8y2:
Thanks for all the work here — the three items from the earlier review (secret-store handling for the session token, single-encoding of SigV4 paths, and NextContinuationToken pagination) were all properly addressed, and that effort is appreciated.

We're closing this PR as superseded: DBX now ships a first-party plugin system, and S3-compatible object storage is already available as an official plugin ([DBX S3](https://github.com/t8y2/dbx-plugin-s3), installable from the in-app marketplace). New non-database data sources like object storage will live in plugins rather than the core connection registry going forward, which also keeps SigV4 signing and bucket browsing out of the core.

If you'd like to continue with this, contributions to the plugin repo would be very welcome — the edge cases you already worked through (path-style vs virtual-hosted addressing, key encoding, pagination cursors) are exactly the hard parts there.

ysicing:
@t8y2  插件会支持deep link么

PR #6584, opened 2026-08-18
fix(sidebar): reconcile object-group tree nodes on re-expand

### LINKED ISSUE #6536: [Bug] 侧边栏新增表或者修改表名后，刷新、关闭连接重新打开都显示不出来
### 数据库类型和版本

Mysql8

### 问题描述

侧边栏新增表或者修改表名后，刷新、关闭连接重新打开都显示不出来

### 支持信息

DBX 版本: v0.5.87
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

Once a database object-group node (Tables, Views, ...) in the sidebar has been
expanded once, re-collapsing and re-expanding it takes a purely client-side
shortcut (`canUseLoadedTreeNodeToggle`) that just replays the in-memory
snapshot with **no server round-trip at all**. A table created or renamed
outside DBX while the node stays loaded never surfaces until an explicit
right-click "Refresh" or a disconnect/reconnect.

## Solution

When this shortcut re-expands an already-loaded node, also kick off a
background `connectionStore.refreshTreeNode(node)` call. The toggle itself
stays instant (no `await`), and the tree silently corrects itself shortly
after — matching the behavior explicit Refresh and reconnect already have.

## Testing

Reproduced against a real MySQL 8 instance (Playwright-driven browser, real
dev backend/frontend): created a table and renamed another one outside DBX,
then collapsed/re-expanded the `Tables` node in DBX — the new/renamed table
was missing before this fix and shows up correctly after it. Also confirmed
that explicit right-click "Refresh" and disconnect/reconnect already worked
correctly (so this fix only had to make plain re-expand consistent with
those two).

- `pnpm typecheck`, `pnpm lint`, `oxfmt --check` all pass
- Full `connectionStore` + `sidebar` vitest suites (70 files / 624 tests) pass

Fixes #6536

### DIFF
--- a/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
+++ b/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
@@ -631,6 +631,15 @@ async function openDirectNavigationNode(node: TreeNode, requestId: number) {
   }
 }
 
+// Re-expanding an already-loaded object-group node (Tables, Views, ...) replays the
+// in-memory list instantly with no server round-trip at all, so external DDL (a table
+// created or renamed outside DBX) never surfaces until an explicit Refresh or a
+// disconnect/reconnect. Reconcile it silently in the background instead.
+function revalidateReopenedGroupNode(node: TreeNode, wasExpanded: boolean) {
+  if (wasExpanded || !node.isExpanded) return;
+  void connectionStore.refreshTreeNode(node).catch(() => undefined);
+}
+
 async function toggle(requestId = beginNavigationRequest()) {
   const node = activeNode.value;
   const treeLoadSearchOptions = sidebarTreeContext?.getTreeLoadSearchOptions?.(node);
@@ -690,6 +699,7 @@ async function toggle(requestId = beginNavigationRequest()) {
   if (databaseObjectGroup && connectionStore.canUseLoadedTreeNodeToggle(node)) {
     node.isExpanded = !node.isExpanded;
     if (wasExpanded && shouldReleaseCollapsedTreeNodeChildren()) connectionStore.releaseCollapsedTreeNodeChildren(node.id);
+    else revalidateReopenedGroupNode(node, wasExpanded);
     emitNodeToggled(node, wasExpanded);
     return;
   }
@@ -709,6 +719,7 @@ async function toggle(requestId = beginNavigationRequest()) {
   if (node.type === "group-extensions" && connectionStore.canUseLoadedTreeNodeToggle(node)) {
     node.isExpanded = !node.isExpanded;
     if (wasExpanded && shouldReleaseCollapsedTreeNodeChildren()) connectionStore.releaseCollapsedTreeNodeChildren(node.id);
+    else revalidateReopenedGroupNode(node, wasExpanded);
     emitNodeToggled(node, wasExpanded);
     return;
   }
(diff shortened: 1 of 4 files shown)


t8y2 (changes requested):
This adds a database request to every normal re-expand of an already loaded object group and bypasses the existing in-flight refresh deduplication.

The previous shortcut was intentionally client-side. With this change, repeated collapse/expand actions always call `refreshTreeNode()`, so ordinary sidebar navigation becomes metadata I/O. Multiple quick re-expands can also start overlapping refreshes because this path does not reuse the explicit refresh lifecycle or its pending request.

Please keep the normal loaded-node toggle request-free. If automatic reconciliation is desired, it needs an explicit bounded freshness policy and the same in-flight deduplication/cancellation used by refresh, with tests proving that repeated re-expands do not add requests. An explicit user Refresh remains an acceptable one-shot request boundary.

q396921921:
Pushed a follow-up commit addressing all three points.

**What was wrong (verified against the code, not just taken at face value):**
- `revalidateReopenedGroupNode` fired `refreshTreeNode` unconditionally on every re-expand of a loaded object group, with no staleness check.
- `refreshTreeNode`'s own generation tracking (`activeTreeRefreshGenerations`) is last-write-wins result arbitration, not request dedup — and for object-group nodes specifically it never even runs (object-group nodes take an early-return branch that skips that map entirely).
- `MetadataLoadCoordinator.run()` explicitly skips its in-flight map when `force: true`, and `refreshTreeNode` always force-loads — so two overlapping re-expands really did fire two real backend metadata queries in parallel, only the dedup skip you described didn't exist for this path.

**Fix:** replaced the direct `refreshTreeNode` call with a new `reconcileLoadedObjectGroupNode(node)`:
- Reads the persisted schema-tree cache's `cachedAt` for the node (a cheap local disk read, not a backend request) and only proceeds if it's past the existing `SCHEMA_TREE_CACHE_TTL_MS` (15 min) — the same bounded-freshness window already used everywhere else in the sidebar cache.
- When (and only when) it's actually stale, it delegates to the existing `refreshStaleTreeNode` helper, which is already used for exactly this object-group case inside `loadObjectGroupChildren` (`if (cached.isStale) refreshStaleTreeNode(node)`), and already dedupes concurrent calls via `staleTreeRefreshIds`.
- No new dedup/freshness mechanism was invented — this just wires the re-expand path into the reconciliation pattern the codebase already has and already trusts for this node type.

**Tests added:**
- `connectionStore.refresh.spec.ts` — `reconcileLoadedObjectGroupNode`: (1) repeated calls while the cache is fresh fire zero backend requests, (2) two overlapping calls after the cache goes stale fire exactly one backend request (proves the dedup is actually exercised through this path), (3) no persisted cache entry also fires zero requests.
- `SidebarTreeRuntimeHost.expansion.spec.ts` — re-expanding a loaded node calls `reconcileLoadedObjectGroupNode` once; collapsing it does not call it at all.

**Verification:** `pnpm typecheck`, `pnpm lint`, `oxfmt --check` all pass; full `connectionStore` + `sidebar` + `composables` vitest suites (118 files / 1184 tests) pass.

q396921921:
Closing this out.

After the last round of feedback I looked into whether the automatic reconciliation was worth keeping at all, given your note that an explicit user Refresh is an acceptable one-shot request boundary. Decided it wasn't - reverted both commits, so this branch is now byte-identical to its base again (no code changes left to review).

Reason: I re-verified the two paths this PR assumed already worked - explicit right-click "Refresh" and disconnect/reconnect - on a real Windows 11 (ARM64) install against a real MySQL 8.4.6 instance (not just code reading or macOS testing this time). Both correctly surface an externally created/renamed table. So the only concretely reproducible gap here is the narrow one this PR targeted (collapse → re-expand of an already-loaded object-group node doing a pure client-side replay), and per your review that's an acceptable trade-off to leave as-is rather than adding request-triggering logic to the normal toggle path.

Follow-up on #6536 with the verification details.

PR #6601, opened 2026-08-18
fix(mcp): support pnpm global shim with separate node runtime

### LINKED ISSUE #6370: [Bug] 使用pnpm全局安装的mcp无法在内置的agent中使用
### 数据库类型和版本

PostgreSQL 16

### 问题描述

Image

使用pnpm全局安装的mcp无法在内置的agent中使用，但是在终端打开的agent中是可以使用的

### 支持信息

DBX 版本: v0.5.85
运行环境: 桌面端
操作系统: macOS 26.5.2
系统架构: aarch64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

ordinary-s:
Fixes #6370

Root cause:
DBX MCP discovery only searched MCP package bound to selected Node runtime and ignored valid pnpm global shims from separate PNPM_HOME.

Solution:
Resolve PATH shims safely by validating package identity and Node compatibility before launching.

Testing:
- Added regression test for split PNPM_HOME layout
- Existing MCP tests pass

### DIFF
--- a/src-tauri/src/commands/mcp.rs
+++ b/src-tauri/src/commands/mcp.rs
@@ -52,28 +52,26 @@ struct NodeRuntime {
     node_path: PathBuf,
     npm_cli_path: PathBuf,
     npm_root: PathBuf,
-    package_manager: McpPackageManager,
     node_version: String,
-    mcp_version: Option<String>,
-    mcp_script_path: Option<PathBuf>,
-    mcp_bin_path: Option<PathBuf>,
-    mcp_native_bin_path: Option<PathBuf>,
+    mcp_installation: Option<McpInstallation>,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 struct McpPackage {
-    version: Option<String>,
+    version: String,
     script_path: PathBuf,
-    minimum_node_version: Option<NodeVersion>,
+    minimum_node_version: NodeVersion,
 }
 
 #[derive(Debug, Clone)]
 enum McpPackageManager {
     Npm,
-    Pnpm { command_path: PathBuf },
+    Pnpm { command_path: PathBuf, pnpm_home: PathBuf, global_dir: PathBuf },
+    PnpmUnavailable { pnpm_home: PathBuf, global_dir: PathBuf },
+    Unmanaged { launcher_dir: PathBuf },
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 struct LocatedMcpPackage {
     package_root: PathBuf,
     package: McpPackage,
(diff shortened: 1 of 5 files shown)


t8y2 (changes requested):
The real pnpm global-shim path is still not supported by this implementation.

`node_script_from_launcher()` accepts a direct JavaScript file or a launcher containing the custom `# cmd-shim-target=` marker, but standard pnpm shims do not contain that marker. A clean pnpm 10.27.0 global install of `@dbx-app/mcp-server@0.4.66` produces a normal POSIX shell shim; pnpm's cmd-shim implementation likewise produces standard shell, `.cmd`, and PowerShell launchers. Those files are rejected by the current parser, so the original split `PNPM_HOME` case still resolves to `[dbxMcpMissing]`. The new test only constructs a synthetic launcher with the private marker and therefore does not reproduce pnpm output.

The same discovery result also needs to drive status, version reporting, update, and uninstall. Currently the PATH shim can make status appear installed while `script_path` and version remain missing, update targets the unrelated Node runtime, and uninstall refuses because that runtime does not own the package.

Please use real pnpm-generated POSIX/Windows fixtures and implement one safe, shared discovery result for launch, status, update, and uninstall. Keep the existing package-identity, path-containment, and Node-engine checks.

ordinary-s:
Addressed the requested changes in amended commit `de8752911`.

What changed:
- Replaced the synthetic launcher coverage with fixtures captured from a real pnpm 10.27.0 global install: POSIX, Windows CMD, and PowerShell. Only machine-specific absolute `NODE_PATH` prefixes are normalized.
- Added one shared `McpInstallation` discovery result containing the package root, launcher, script, bound Node runtime/version, package version, package manager metadata, and native binary.
- Launch, status, version reporting, update, and uninstall now reuse that same installation result.
- pnpm update/uninstall use the discovered pnpm executable, `PNPM_HOME`, and explicit `--global-dir`. When ownership cannot be verified, the operation fails safely.
- Preserved exact `@dbx-app/mcp-server` identity validation, canonical path/bin containment checks, and `engines.node` compatibility checks.
- Added split `PNPM_HOME` regression coverage, real shim parsing tests, status/version/launch consistency checks, pnpm update/uninstall routing tests, npm regression coverage, and negative identity/containment/engine tests.

Validation performed:
- MCP tests: 27 passed, 0 failed on Windows.
- `cargo +1.97.0 fmt --all -- --check`: passed.
- `cargo +1.97.0 check -p dbx --lib --no-default-features`: passed.
- `cargo +1.97.0 clippy -p dbx --lib --no-default-features -- -D warnings`: passed.
- Full Rust unit suite: 226 passed, 2 failed. Both failures are pre-existing Windows path-separator assertions in `external_db` and `external_sql`; neither file is changed by this PR.
- Root format, lint, and typecheck passed. The broader frontend run had 9,881 passing, 2 unrelated Windows-specific failures, and 1 skipped test.

The PR head was amended rather than adding a second commit.

t8y2 (commented):
Maintainer conflict resolution applied and pushed.

- Old head: `de8752911a543fb13b1b4875c8ebca8a9833e354`
- New head: `ae3fcae8325bf333648c31080e65a45747e72bb1`
- Rebased onto: `d1d8dc4aec275fb86e35309b71c8ce2342754dae`

The resolution preserves the PR's unified `McpInstallation` model and split pnpm launcher/runtime discovery, while retaining current `main`'s guarded mise-shim fallback when no managed installation is available. The pnpm fixtures and all five contributor regression tests were preserved; the current-main mise regression remains in place. No dependency manifest or lockfile changed.

Checks run: 7 pnpm-focused Rust tests and the mise command-resolution regression passed; `cargo fmt --check`, range-diff review, and `git diff --check` passed.

t8y2 (approved):
Approved after the maintainer patch and exact-head CI passed.

t8y2:
Thanks for the contribution! Merged in 316a1f2a7, will be released in the next version.

PR #6609, opened 2026-08-18
fix(redis): mark key-tree folder counts as partial while still scanning

### LINKED ISSUE #6392: [Bug] Data missing from Redis query with latest DBX (0.5.86)
### 数据库类型和版本

Redis 8.6.2

### 问题描述

Install the latest release of DBX (0.5.86), there are still data missing from Redis query. The old closed issues is [6022](https://github.com/t8y2/dbx/issues/6022)

The left dataset is from DBX, right dataset is from RESP.
Image

### 支持信息

DBX Version: v0.5.88
Runtime: Desktop
Operating System: Windows 10.0.26100.9106
Architecture: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Summary
- The Redis key-tree folder count (`loadedLeafCount`) only reflects the keys the bounded auto-load SCAN has fetched so far, not the folder's true total — but it rendered identically to an exact count with no visual hint it was incomplete.
- When a keyspace is large enough that auto-load stops before the SCAN cursor reaches 0, folder counts land far below the real per-prefix cardinality (e.g. showing `12` for a folder that actually has `281` keys), which reads as DBX losing/missing data when compared against raw `SCAN`/RESP output.
- While more keys remain unscanned (`hasMore`), the count now renders with a `+` suffix (e.g. `12+`) and a tooltip explaining it's partial; once the scan actually finishes, the count renders exact again with no `+`.

Fixes #6392

## Test plan
- [x] `pnpm exec vitest run apps/desktop/src/components/redis apps/desktop/src/lib/redis` — 115/115 passed, including 2 new tests covering the partial-count `+`/tooltip and the exact-count-after-full-scan case
- [x] `pnpm typecheck` — clean
- [x] `oxlint --vue-plugin` on changed files — clean
- [x] Manually reproduced against a real Redis instance with 6,050 keys under one prefix (real SCAN-verified counts vs. DBX's tree counts), confirmed the folder counts were misleadingly partial before the fix and correctly marked with `+`/tooltip after, and confirmed they return to exact numbers once "Fetch all" finishes

### DIFF
--- a/apps/desktop/src/components/redis/RedisKeyBrowser.infiniteScroll.spec.ts
+++ b/apps/desktop/src/components/redis/RedisKeyBrowser.infiniteScroll.spec.ts
@@ -574,3 +574,40 @@ describe("RedisKeyBrowser loadMore failure handling (PR #6313 review)", () => {
     expect(mocks.redisScanKeysBatch.mock.calls.length).toBeLessThanOrEqual(1 + AUTO_LOAD_MAX_CALLS);
   });
 });
+
+// A group's `loadedLeafCount` only reflects keys the SCAN happened to return
+// before auto-load stopped (see `redisKeyTree.ts`); when the cursor hasn't
+// reached 0 yet, that count is not the folder's real total. Issue #6392: a
+// user compared DBX's tree against `redis-cli --scan` and saw folder counts
+// far below the real per-prefix cardinality, with nothing in the UI hinting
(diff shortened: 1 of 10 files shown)


t8y2:
Thanks for the contribution! Merged in 7be64f45a, will be released in the next version.

PR #6616, opened 2026-08-18
feat(xugu): preserve index partition metadata in DDL

mapan0424:
## Summary

This change improves XuguDB table metadata reconstruction so index scope and index partition definitions are preserved in generated table DDL.

## Why

XuguDB distinguishes ordinary indexes, LOCAL indexes, and GLOBAL partitioned indexes. The previous metadata path listed index columns but discarded these distinctions, so viewing or exporting a table could produce executable-looking SQL that changed the index semantics or omitted partition clauses.

## What changed

- Read Xugu index scope and partition attributes from `ALL_INDEXES`.
- Read index partitions and subpartitions from `ALL_IDX_PARTIS` and `ALL_IDX_SUBPARTIS`.
- Keep the additional metadata internal to the Xugu agent so the shared index protocol remains unchanged.
- Reconstruct ordinary, LOCAL, GLOBAL, RANGE, LIST, HASH, and subpartition clauses in table DDL.
- Keep catalog reads best-effort for older Xugu versions where the newer catalog columns are unavailable.
- Preserve the existing database-scope filtering and avoid exposing internal/system indexes as user indexes.

## Validation

- Xugu Agent unit tests pass.
- Xugu Agent live tests pass with both an administrator account and a normal account.
- Live verification covers ordinary indexes, LOCAL indexes, GLOBAL LIST indexes, and GLOBAL RANGE indexes with HASH subpartitions.
- Reconstructed table DDL was executed against XuguDB and the indexes were verified after replay.
- Core database-export tests pass, including table DDL/data export and partition DDL replay.
- `go vet ./...` and `git diff --check` pass.

This PR is limited to XuguDB metadata and DDL reconstruction. No shared database-driver behavior or other database drivers are changed.

### DIFF
--- a/agents/drivers/xugu/index_partition_live_test.go
+++ b/agents/drivers/xugu/index_partition_live_test.go
@@ -1,0 +1,197 @@
+package main
+
+import (
+	"os"
+	"strings"
+	"testing"
+)
+
+// TestLiveXuguIndexPartitionDDL verifies the four index cases that are easy to
+// conflate in reconstructed DDL: an ordinary index, a LOCAL index, a GLOBAL
+// index with explicit partitions, and a GLOBAL index with subpartitions. It is
+// opt-in because CI does not provide a XuguDB service.
+func TestLiveXuguIndexPartitionDDL(t *testing.T) {
+	if os.Getenv("XUGU_LIVE_TEST") != "1" {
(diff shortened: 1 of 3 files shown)


t8y2 (approved):
Reviewed the Xugu index partition metadata changes and focused validation; no blocking correctness or compatibility issue remains.

t8y2:
Thanks for the contribution! Merged in dcbd034, will be released in the next version.

PR #6620, opened 2026-08-19
fix(grid): edit multi-cell selections inline

xzxlove:
## 变更说明

- 支持在范围或整列多选后直接输入，复用现有单元格编辑器，不打开批量修改弹窗
- 按 Enter 或失焦时将最终值批量应用到选区，按 Escape 取消
- 过滤快捷键、组合键和输入法合成事件，保持单单元格编辑行为不变

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

本次是键盘交互修复，未附带包含本地数据库信息的原始截图；行为由回归测试覆盖。

## 验证

- [ ] `make check` 通过
- [x] 相关测试通过

- `pnpm vitest run apps/desktop/src/components/grid/__tests__/DataGridInlineBulkEdit.spec.ts apps/desktop/src/composables/__tests__/useDataGridEditor.spec.ts apps/desktop/src/composables/__tests__/useDataGridSelection.spec.ts packages/app-tests/dataGridContextMenu.test.ts` (57 tests)
- `pnpm typecheck`
- `git diff --check`

## 关联 Issue

N/A

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -242,6 +242,7 @@ import { dataGridCanvasDevicePixelSize, useDataGridCanvasRuntime, type DataGridC
 import { useDataGridScrollbars, type DataGridScrollbarsRuntime } from "@/composables/useDataGridScrollbars";
 import { useDataGridSelection } from "@/composables/useDataGridSelection";
 import { dataGridNavigationOrigin, dataGridPageScrollTop, dataGridRowScrollTop, moveDataGridCell, navigateDataGridCell, type DataGridNavigationDirection, type DataGridScrollAlignment } from "@/lib/dataGrid/dataGridNavigation";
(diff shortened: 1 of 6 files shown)


t8y2 (changes requested):
Request changes: empty bulk edits currently bypass the grid's established NULL semantics.

When a selected cell is cleared, the new inline multi-cell commit path coerces the empty value by inferred type (`0`, `false`, or `""`) instead of producing the same NULL/empty decision used by the existing single-cell editor. A user clearing several cells can therefore silently write real values that were never entered.

Please route every selected cell through the canonical value parser/NULL handling used by single-cell edits, preserve per-column type semantics, and add regression tests for numeric, boolean, text, nullable, and non-nullable columns. The PR is also currently conflicting, so rebase after the behavior is corrected.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in d2e2270d8, will be released in the next version.

PR #6625, opened 2026-08-19
feat(data-grid): add conditional bulk updates for MySQL

### LINKED ISSUE #6588: [Feature] 修改某个字段值，希望能选择根据查询条件进行更新
### 数据库类型和版本

Mysql

### 支持信息

DBX 版本: v0.5.87
运行环境: 桌面端
操作系统: Windows 10.0.19044.7663
系统架构: x86_64

### 当前痛点

修改某个字段时，默认只能针对选中的数据更新，选择的数据并不是所有查询条件匹配的数据

Image

### 期望方案

希望修改某个字段值的时候，能够选择按查询条件匹配结果进行更新

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须做

### 补充信息

_No response_

eryajf:
## 变更说明

- 为 MySQL 表数据页的“批量修改所选”增加“修改条件匹配的全部行”范围，可更新未加载分页中的匹配行。
- 支持复用当前表格筛选、通过字段/运算符/值构建多条 `AND` / `OR` 条件，以及高级 `WHERE` SQL；在执行前可预览影响行数。
- 在核心层生成带标识符引用、类型化值、非空 `WHERE` 与不可写字段保护的 `UPDATE`；已接入桌面端和 Web 后端。
- 更新操作会显示完整 SQL 的危险确认，写入历史记录并刷新表格；批量编辑弹窗标题栏支持拖动。
- 更新中英文数据网格文档及全部现有界面语言文本。

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [x] 文档更新
- [ ] CI / 构建

## 涉及前端

表格选择列之后，支持条件匹配修改

image

如果已有 where 条件，默认复用过滤条件

image

## 验证

- [ ] `make check` 通过（未运行）
- [ ] `make cargo-check-fast` 通过（未运行）
- [x] 相关测试通过

已执行：

- `git diff --check`
- `pnpm exec vitest run apps/desktop/src/composables/__tests__/useDataGridFilterBuilder.spec.ts apps/desktop/src/composables/__tests__/useDataGridEditor.spec.ts packages/app-tests/dataGridEditor.test.ts`（86 项通过）
- `cargo test -p dbx-core --no-default-features mysql_conditional_update_uses_typed_value_and_requires_a_writable_column_and_where`
- `pnpm typecheck`

## 关联 Issue

Close #6588


t8y2 (changes requested):
Request changes: the feature still needs a safe execution lifecycle for potentially large conditional updates.

1. The current `executeBatch` path has no execution ID or cancel control, and a timeout/error is always recorded as a failed update even when the database outcome may be unknown. Please define explicit `completed` / `not-started` / `unknown` outcomes and retain a cancellation path until completion is known.
2. Editing remains enabled during execution, while the success reload discards edits made in that interval. Block concurrent edits or preserve and reconcile them instead of silently dropping them.
3. Value coercion depends on an arbitrary displayed row. A nullable MySQL `TINYINT(1)` can serialize `"true"` differently depending on whether the sampled old value is `NULL` or `false`. Make coercion column-metadata-driven and add focused behavior tests.

The branch is also `CONFLICTING/DIRTY` with broad semantic overlap and includes an unrelated export-test stack change, so please rebase and split that commit before the next review.

t8y2 (changes requested):
Request changes: this head resolves the edit locking and MySQL `TINYINT(1)` coercion issues, but the execution lifecycle still has two correctness gaps.

1. `apps/desktop/src/composables/useDataGridEditor.ts:1581` treats `cancelQuery() == true` as a confirmed terminal outcome. The backend return value only confirms that the cancellation token or interrupt was triggered; the asynchronous database operation may still be running or its result may remain unknown. Please keep the grid locked until a terminal execution state is confirmed, then reload/reconcile the data.

2. `crates/dbx-web/src/routes/query.rs:375` drops the execution registration when the Web request returns a timeout/error. A later cancellation by execution ID therefore cannot find the task. Please preserve the registration on timeout, align the Web behavior with the Tauri path, and add a delayed MySQL UPDATE regression covering timeout → cancel → confirmed terminal state → reload.

The unrelated `database_export_prefetch.rs` stack-size change should also be removed from this PR or moved to a separate PR.

t8y2 (commented):
Maintainer conflict resolution pushed in d1e5a37, rewriting old head 78a11ac onto current main 10d1509d9240043b25611325856e348ad6ce6c2e.

The conflict came from the cancellation registry redesign already on main: interrupt callbacks now live inside each running task and late registration after cancellation executes immediately. The resolved version preserves that race/leak fix while adding terminal-confirmation tracking only for conditional updates, so cancellation reports terminal only after the owning execution guard exits; recent terminal records expire after the existing grace period.

The patch also distinguishes terminal SQL/constraint failures from timeout or transport-unknown outcomes so the grid unlocks, records failed history, and reloads safely; restores sampled numeric/boolean coercion only when column metadata is missing while keeping metadata authoritative; and aligns the app-test numeric ID fixture with that rule. No dependency files changed.

Verified 22 query-cancel tests, the conditional-update SQL unit test, 118 focused desktop/app tests, dbx-web and Tauri route compilation, pnpm typecheck, changed-file oxfmt/rustfmt, git diff/conflict checks, and a real MySQL 5.7.44 delayed-trigger cancellation test proving reload occurs only after terminal cancellation. Full range-diff was recorded against 68d291921932986452065f4c7e94b9b424fde16b..78a11acb01977aedb1d576ead85fad4d0f2e9069.

t8y2 (approved):
Maintainer resolution verified on exact head d1e5a37: focused regressions, type/format checks, real MySQL 5.7 cancellation, and exact-head CI all pass.

t8y2:
Thanks for the contribution! Merged in a7edb8f, will be released in the next version.

PR #6633, opened 2026-08-19
fix(schema-diff): keep Compare button reachable when choosing specific tables

### LINKED ISSUE #6627: [Bug] 比较架构指定表比较，看不到比较按钮
### 数据库类型和版本

MySQL8.0

### 问题描述

操作步骤：比较架构->选择数据->指定表比较

Image

### 支持信息

DBX 版本: v0.5.88
运行环境: 桌面端
操作系统: Windows 10.0.22000.2538
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

0verme:
Fixes #6627

## Root cause
Not a missing button or a `canCompare` bug. After #6533/#6540 ("compare specific tables") the config step grew with the table multi-select and the same-name target match block. The dialog content container is a fixed-height flex column:

```
<div class="flex-1 min-h-0 overflow-hidden flex flex-col">
```

With `overflow-hidden` plus the default `flex-shrink: 1` on the config step, the taller content was clipped and could not scroll, pushing the whole bottom action bar (Save / Options / **Compare**) out of the visible area with no way to reach it.

## Fix
Scroll the content container vertically on every step except `result` (which relies on splitpanes to manage its own scroll/heights), and mark the config step `shrink-0` so its taller-than-dialog content actually overflows and becomes reachable instead of being compressed:

```
<div :class="[step === 'result' ? 'overflow-hidden' : 'overflow-y-auto', 'flex-1 min-h-0 flex flex-col']">
  <SchemaDiffConfigStep v-if="step === 'config'" class="shrink-0" ... />
```

## Tests
Added a source-level regression test to `SchemaDiffDialog.spec.ts` locking both constraints (non-result steps scroll; config step is `shrink-0`). All diff/lib tests (`12`) and `SchemaDiffDialog.spec.ts` (`5`) pass, `pnpm typecheck` and `pnpm lint` are clean.

### DIFF
--- a/apps/desktop/src/components/diff/SchemaDiffDialog.vue
+++ b/apps/desktop/src/components/diff/SchemaDiffDialog.vue
@@ -870,10 +870,16 @@ const targetConnectionInfo = computed(() => {
         </DialogTitle>
       </DialogHeader>
 
-      <div class="flex-1 min-h-0 overflow-hidden flex flex-col">
+      <!-- Result step relies on splitpanes to manage its own scroll/heights, so it keeps
+           `overflow-hidden`; the config step's tall content (e.g. the table multi-select
+           added in the "compare specific tables" feature) can overflow a fixed-height
+           dialog, so it must be allowed to scroll vertically instead of being clipped --
+           otherwise the Compare button at the bottom becomes unreachable. -->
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in e9ca697c3, will be released in the next version.

PR #6661, opened 2026-08-19
fix(editor): keep indenting a selection instead of letting a stray completion hijack Tab

### LINKED ISSUE #5419: [Bug] 编辑器里面按Tab键对齐问题

q396921921:
## Problem

Pressing Tab to indent a multi-line selection in the SQL editor works the first time, but a completion popup can appear as a side effect of that edit (the doc-changed handler re-triggers SQL completion based on cursor context, regardless of what kind of edit just happened). The next Tab press then accepts that unrelated completion instead of indenting further, replacing the selected text with a suggestion.

## Fix

`acceptCompletionOrNextSnippetField` (used by the Tab keymap) now only goes through the completion-accept/wait path when the selection is empty. A non-empty selection always goes straight to block indent (`indentMore`), so a completion popup that appears mid-indent can never hijack a subsequent Tab press. Snippet-field navigation is unaffected either way.

## Testing

- Added two regression tests in `queryEditorCompletionKeymap.spec.ts` that extract the real `handleTab`/`acceptCompletionOrNextSnippetField` function bodies from `QueryEditor.vue` (same harness this file already uses) and assert that a multi-line selection keeps calling `indentMore` instead of `acceptCompletion`, both when completion is already `active` and when it's `pending` and resolves after the Tab press. Both fail against the pre-fix code and pass after.
- Rebased onto latest `main` and re-ran `queryEditorCompletionKeymap.spec.ts` — 16/16 tests passing.

Fixes #5419

---
Resubmitted from #6184 (closed, no review activity since 2026-08-13) — branch rebased onto current `main`, same fix and tests, no functional changes.

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -1973,9 +1973,14 @@ function extendQueryEditorSelectionForView(currentView: EditorViewType): boolean
 }
 
 function acceptCompletionOrNextSnippetField(view: EditorViewType): boolean {
-  const completionStatus = codeMirrorCompletionStatus?.(view.state) ?? null;
-  if (completionStatus === "active" && (codeMirrorAcceptCompletion?.(view) ?? false)) return true;
-  if (completionStatus) return waitForCompletionTab(view);
+  // A non-empty selection means Tab is being used for block indent, not word
+  // completion. A completion popup can still appear as a side effect of the
+  // indent edit itself, so it must never hijack this (or a following) Tab press.
+  if (view.state.selection.main.empty) {
+    const completionStatus = codeMirrorCompletionStatus?.(view.state) ?? null;
+    if (completionStatus === "active" && (codeMirrorAcceptCompletion?.(view) ?? false)) return true;
+    if (completionStatus) return waitForCompletionTab(view);
+  }
   return codeMirrorNextSnippetField?.(view) ?? false;
 }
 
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch applied on the contributor branch.

- Reviewed head: `d51cb1b67ee61392e880dd7d3cfb4d95366292da`
- Patched head: `b98e3da20f64c48a1be479e9a5a8e400e6b158d9`
- Fix: treat Tab as indentation whenever any selection range is non-empty, and validate every range before accepting a delayed completion.
- Regression coverage: mixed multi-range active and pending completion paths; 18 focused tests passed.
- Validation: focused Vitest, `oxfmt --check`, `pnpm typecheck`, and `git diff --check` passed on the development server.
- Dependencies: none added.
- Performance: no extra requests, listeners, timers, or background I/O; the existing Tab path now performs one bounded scan of editor selection ranges.

t8y2:
Thanks for the contribution! Merged in 45819df9e, will be released in the next version.

PR #6667, opened 2026-08-19
fix(mysql): show scheduled EVENT objects in the sidebar tree

### LINKED ISSUE #6638: [Bug] mysql 仅仅显示被动触发TRIGGERS ，主动触发EVENTS不显示
### 数据库类型和版本

mysql8.0

### 问题描述

对比了其他工具，DataGrip各种触发器都可以显示。

参考sql语句

SHOW TRIGGERS;

SHOW EVENTS;

参考的定时任务事件
create definer = root@`%` event event_daily_middle_db_sync on schedule
    every '1' DAY
        starts '2026-08-06 01:00:00'
    enable
    do
    CALL `water_middle_db`.`sp_sync_daily_data`(TRUE);

### 支持信息

Image

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

q396921921:
## Problem

MySQL scheduled `EVENT` objects never show up in the sidebar object tree, even though `TRIGGER` objects (created the same way) show up fine. Comparing to other tools (e.g. DataGrip), Events should be listed alongside Triggers.

## Root cause

The backend (`crates/dbx-core/src/db/mysql.rs`) already queried `information_schema.EVENTS` and returned `EVENT` objects from `list_objects` — but the frontend's MySQL sidebar object-kind list (`MYSQL_OBJECTS` in `databaseObjectCapabilities.ts`) never included `"EVENT"`, so the sidebar tree never requested that object type from the backend in the first place. The group never appeared at all (not "empty" — just missing).

## Solution

Adds `EVENT` as a first-class object kind end-to-end:

- New `ObjectSourceKind::Event` Rust enum variant. The compiler's exhaustive-match requirement surfaced every dialect-specific site that needed to account for it (Postgres/Oracle/SQL Server/SQLite all explicitly treat it as unsupported, matching how they already treat `Trigger`).
- Real `SHOW CREATE EVENT` DDL support for MySQL. Verified against a live MySQL 8.0 instance that the DDL column index is `3`, not `2` like `Trigger`/`Procedure`/`Function` — `SHOW CREATE EVENT` has an extra `time_zone` column before the DDL.
- Sidebar wiring: new "Events" group (icon, i18n label added to all 8 locales), tree-group membership sets, view-source + copy-name context menu (mirroring the existing `Sequence` node — no drop action, matching that same minimal-risk precedent).

Out of scope for this PR (none of these are what's reported here, and each would be a separate change): the standalone "Object Browser" tab, SQL autocomplete, schema-diff, and a "Create Event from template" shortcut.

## Testing

- Reproduced against a real MySQL 8.0 instance with both a real `TRIGGER` and a real `EVENT`: before the fix, the sidebar showed Tables/Views/Procedures/Functions/Triggers but no Events group at all; after the fix, an Events group appears with the event listed, and double-clicking it opens a source tab with the correct `SHOW CREATE EVENT` DDL.
- Re-verified the existing Trigger flow end-to-end — no regression.
- 2 new Rust unit tests in `schema.rs` (`SHOW CREATE EVENT` SQL generation + DDL column index) — pass.
- `cargo check --workspace` (all crates incl. dbx-web/dbx-cli/dbx-mcp/tauri) — clean. `cargo clippy -p dbx-core --lib -D warnings` — clean. `cargo fmt --check` — clean.
- 2 new/updated frontend vitest specs; full suite `pnpm exec vitest run` — 1051 files / 10061 tests, all passing. `pnpm typecheck` — clean. `pnpm run lint` — clean (pre-existing unrelated warnings only).

Before/after screenshots to follow in a comment.

Fixes #6638


t8y2 (commented):
Maintainer patch pushed on top of the updated reviewed head.

- Old head: `ba1d8597825ce7441a98d00e9d41a4c243ff059b`
- New head: `1510d95642bb8a3bee847bc3bba0a7376555cd63`
- Base checked before push: `e65c3ab9149f1b3f477bdeea3fbb93f56e94ae73`
- Fix: mark MySQL Event object sources read-only so saving cannot replay `CREATE EVENT`, and show the Events group object count in the sidebar.
- Remote checks on the rebased head: focused TreeItem Vitest (13/13), frontend format/lint/typecheck, two focused dbx-core Rust tests, `cargo fmt --check`, and `git diff --check`.

t8y2 (commented):
Maintainer conflict resolution pushed after the base advanced.

- Old head: `1510d95642bb8a3bee847bc3bba0a7376555cd63`
- New head: `e3a2fb6c992fea3e36567eb796599b3cdd5726e6`
- Merged base: `7a707e77d8cc09b9379da9a6ec3bbf44e4fe361c`
- Conflict cause: both branches added adjacent programmable-object tree tests in `tableTree.spec.ts`.
- Resolution: preserved both the MySQL Event grouping regression and the new Kingbase table-scoped trigger regression. The complete PR diff still contains the original Event feature plus the read-only/count patch; range-diff maps both commits exactly.
- Remote checks: 4 focused frontend files / 70 tests, frontend format/lint/typecheck, two focused dbx-core Rust tests, `cargo fmt --check`, and `git diff --check`.

t8y2 (approved):
Maintainer patch and conflict resolution are complete. Both Event and current-main behaviors are preserved, range-diff is stable, and exact-head frontend/Rust CI is green.

t8y2:
Thanks for the contribution! Merged in 2b11af94b, will be released in the next version.

PR #6690, opened 2026-08-19
feat(xugu): expand table editor data types

mapan0424:
## Summary

Expands the XuguDB data-type catalog used by the DBX table editor, so the
types offered by the UI match the server capabilities that the Xugu agent can
create and read through the bundled Go driver.

The editor now exposes 54 selectable XuguDB types, including:

- `TINYINT`, `DOUBLE`, `DATETIME`, and time-zone-aware temporal types;
- all supported `INTERVAL` qualifiers;
- `GUID`, `ROWID`, `JSON`, `BIT`, and `VARBIT`;
- the supported `CHAR[]`, `INTEGER[]`, `DOUBLE[]`, and `CLOB[]` array forms.

Pseudo-types and grammar-only entries reported by `SHOW DATA_TYPES` are
intentionally not exposed as normal column choices. This excludes `NULL`,
bare `ARRAY`, `ROWVERSION`, and the spatial family, because the generic table
editor cannot emit a safe standalone column definition for those entries.

## Design

The Xugu agent and desktop editor now use the same curated, alphabetically
ordered catalog. This prevents a type from being selectable in one layer but
unavailable in the other.

The generic `TYPE(length)` editor needs database-specific guards for Xugu:

- array suffixes, interval qualifiers, and fixed-width types do not accept a
  generic length suffix;
- `TIME` accepts precision `0..3`;
- `TIMESTAMP` accepts precision `0..6`.

Invalid temporal precision is omitted from the generated SQL rather than
producing invalid DDL. These rules are explicitly scoped to `xugu`; existing
profiles for other databases remain unchanged.

### Automated checks

- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `pnpm exec vitest run apps/desktop/src/lib/__tests__/table/tableStructureEditorState.spec.ts`
  - 35 tests passed
- `pnpm typecheck`
- `git diff --check`

### Live XuguDB coverage

Added an opt-in integration test that uses the same `database/sql` and
go-xugu-driver path as the agent. Against the XuguDB test environment it:

1. verifies the server catalog via `SHOW DATA_TYPES`;
2. creates, inserts into, and reads back a table for each of the 27 newly
   exposed types;
3. cleans up each temporary table after the test.

### Packaged UI verification

Built an isolated ARM64 macOS package and launched it successfully. In the
packaged application, created a table through the table editor using
`TIME WITH TIME ZONE`, `TINYINT`, `JSON`, `INTERVAL DAY TO SECOND`,
`INTEGER[]`, and `CLOB[]`. The generated SQL executed successfully and the
schema tree subsequently re-read the same six types from database metadata.


t8y2 (commented):
Maintainer patch pushed after revalidating the contributor rebased head. It fixes Xugu timezone type syntax and preserves BIT/VARBIT/TIME/TIMESTAMP metadata round-trips. Go, Rust, JDBC, frontend, typecheck, lint/format, and diff checks passed; the optional live Xugu test was not run because credentials were unavailable.

t8y2 (approved):
Maintainer patch verified at exact head 55dfd3da007c14f38c868604195d1adfac5759a2; Xugu/JDBC/Rust/frontend validation and exact-head CI passed.

t8y2:
Thanks for the contribution! Merged in 82da9cb6d, will be released in the next version.

PR #6712, opened 2026-08-19
fix(transfer): defer MySQL foreign keys so table cycles don't block transfer

### LINKED ISSUE #6665: [Bug] mysql数据库传输失败
### 数据库类型和版本

mysql 5.7

### 问题描述

mysql将数据从一个数据库传输到另一个数据库时报错，使用sqlyog正常，已尝试多次

Image

Image

### 支持信息

DBX 版本: v0.5.88
运行环境: 桌面端
操作系统: macOS 26.6.1
系统架构: aarch64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

无

q396921921:
## Problem

MySQL-family data transfer creates target tables by replaying the source's `SHOW CREATE TABLE` DDL, which includes inline `FOREIGN KEY` constraints. Table creation follows a dependency-based sort, but if the selected tables' foreign keys form a cycle (or reference a table outside the selection), there is no valid `CREATE TABLE` order at all — creation fails with `ERROR 1452`/`ERROR 1215`/`ERROR 1824` and the whole transfer aborts.

## Fix

Mirror the approach already used by the Postgres transfer path: strip inline foreign keys out of the `CREATE TABLE` DDL, create bare tables first, then add every foreign key via a deferred `ALTER TABLE ... ADD CONSTRAINT` once all selected tables exist. Table creation order no longer depends on foreign key dependencies at all.

Also fixes 4 issues found in review of the initial version of this fix:
- A failed deferred `ALTER TABLE ADD CONSTRAINT` is now surfaced as a transfer failure (`TransferStatus::Error`, naming the affected table) instead of only being logged — previously the transfer could report success while the target schema silently ended up missing foreign keys.
- Foreign keys referencing a table in a different database (MySQL supports `REFERENCES other_db.table`) now keep their original schema qualifier instead of being resolved against the wrong (target) database.
- Foreign key metadata already fetched once while sorting tables by dependency is now reused per table instead of being re-queried.
- The constraint-grouping logic shared with the Postgres path is now one function instead of two near-identical copies.

## Testing

- Real repro against a live MySQL 8 instance (docker) + real browser (Playwright) driving the actual Data Transfer dialog, not a direct API call:
  - Three tables with a genuine foreign key cycle (`b_department` ↔ `c_employee`, plus `a_orders` → `c_employee`): before the fix, transfer fails with `3 table(s) failed` / `ERROR 1824: Failed to open the referenced table`; after the fix, all 3 tables transfer successfully with all 3 foreign keys (including the cycle) correctly created on the target.
  - A simple two-table, non-cyclic foreign key case (the original issue screenshot's shape) confirmed unaffected — no regression.
  - Cross-database foreign key: source table `t` in database `A` references `B.parent`; transferring only `A.t` (not `B.parent`) now correctly generates a schema-qualified `REFERENCES `B`.`parent`` constraint, verified with a real cross-database `JOIN` against the target.
  - Forced deferred-ALTER failure (transferring a table without its foreign-key target present on the target side): the transfer now correctly reports `1 table(s) failed: 1 foreign key(s) on: <table>` instead of silently reporting success.
- `cargo test -p dbx-core --lib transfer::`: 215/215 passing (209 pre-existing + 6 new, covering DDL stripping, deferred ALTER generation, same-database vs cross-database schema resolution, constraint grouping, and FK-cycle survival through the dependency sort).
- `cargo test -p dbx-core --lib database_export:: / schema::`: 68/68, 164/164 passing — confirms `sort_tables_by_fk_dependency`'s two unrelated callers (batch drop ordering, database export) are unaffected by the refactor that lets it reuse fetched foreign key metadata.
- `cargo clippy -p dbx-core -p dbx-web --tests -- -D warnings`, `cargo clippy --manifest-path src-tauri/Cargo.toml -- -D warnings`, `cargo fmt --check`: all clean.
- Rebased onto latest `upstream/main` immediately before opening this PR; re-ran the full check/test/clippy/fmt suite post-rebase — all still clean.

Fixes #6665

### DIFF
--- a/crates/dbx-core/src/transfer.rs
+++ b/crates/dbx-core/src/transfer.rs
@@ -1725,6 +1725,10 @@ fn is_mysql_family_target(target_db: &DatabaseType) -> bool {
     )
 }
 
+fn supports_deferred_mysql_foreign_keys(target_db: &DatabaseType) -> bool {
+    is_mysql_family_target(target_db) && crate::table_structure_sql::supports_foreign_keys(*target_db)
+}
+
 /// QuestDB is not included. It only uses the PGWire protocol. SQL DDL syntax is not compatible.
 fn is_postgres_family_target(target_db: &DatabaseType) -> bool {
     matches!(
@@ -7020,7 +7024,7 @@ where
             // Postgres transfers.
             let mut ddl = ddl;
             let mut deferred_fk_alters: Vec<String> = Vec::new();
-            if is_mysql_family_target(target_db_type) {
+            if supports_deferred_mysql_foreign_keys(target_db_type) {
                 // Reuse the FK metadata `sort_tables_by_fk_dependency_with_foreign_keys`
                 // already fetched for the whole batch when the caller provided it;
                 // only fall back to a live query for callers that don't pre-fetch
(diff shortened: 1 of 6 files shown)


t8y2 (commented):
Maintainer patch pushed in `9c172f903`.

Fixes: deferred foreign-key creation is now gated by actual target capability so Doris is not sent unsupported `ADD CONSTRAINT` statements; failure reporting counts failed foreign keys rather than deduplicated table names.

Validation: 215 transfer tests, `cargo check -p dbx-web`, `cargo check -p dbx`, formatting, and `git diff --check` passed. No dependency changes.

t8y2:
Thanks for the contribution! Merged in f36d854, will be released in the next version.

PR #6714, opened 2026-08-19
docs: fix broken star history chart

OctoBored:
The star history chart in the README is currently broken because the underlying service hit GitHub stargazer API limits. Point it at a working alternative so the chart renders again.


t8y2:
Thanks for the contribution! Merged in b537c4b, will be released in the next version.

PR #6783, opened 2026-08-20
fix(jdbc): keep JDBC plugin local install fully offline

### LINKED ISSUE #6689: [Bug] 完全内网环境下无法通过导入 jdbc 驱动 方式使用
### 数据库类型和版本

oracle 19c

### 问题描述

步骤
1 纯内网环境下使用 本地 jdbc 驱动连接数据库

2 提示需要一些 json 文件.

期望

可以在纯内网下不依赖外部下载方式运行程序

### 支持信息

0.5.79
MAC 桌面版

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Summary

Fixes #6689. In a fully air-gapped environment, importing a local JDBC driver JAR already worked without network access, but installing the DBX JDBC plugin package from a local file (Local Install) still hit the network — it reused the same status check as the online install/update path, which unconditionally queries GitHub/R2 for the latest plugin version. On a machine with no internet access this could block for a long time trying to reach a network that isn't there, defeating the purpose of the local/offline install path.

- Add a local-only status helper (mirroring the pattern already used by `uninstall_jdbc_plugin`) and use it after a local plugin install, skipping the network version check entirely. The online download/update path is unaffected and still checks for updates as before.
- Generalize the "missing manifest.json" error message: it used to read as if the caller had just picked the wrong file in the Local Install dialog, but the same code path is also reached from the automatic online download and from importing a full offline Agent bundle. It now points at the existing `JDBC_PLUGIN_DOWNLOAD_URL` constant instead of a hand-written sentence that could drift out of sync.
- Add a tooltip (reusing the `Tooltip` component already used elsewhere in this file) explaining the offline install flow on the Local Install button, replacing a native `title` attribute.
- Add the `driverStore.localInstallHint` string to every locale that already has the sibling `localInstall` key.
- Trim the JDBC section of `driver-management.mdx` down to what's actually JDBC-specific, instead of repeating the generic offline-import steps already covered by the parent section.

## Test plan

- [x] `cargo test -p dbx-core --lib jdbc::` — 10/10 passing
- [x] `cargo clippy -p dbx-core --lib -- -D warnings` — clean
- [x] `cargo fmt --check -p dbx-core` — clean
- [x] `pnpm exec vue-tsc --noEmit` — clean
- [x] `pnpm exec vitest run apps/desktop/src/i18n/__tests__/` — 340/340 passing
- [x] Manually verified with a real dev backend + browser (Playwright): before the fix, clicking Local Install with a valid plugin package took ~1.3s and round-tripped a real update check against GitHub/R2; after the fix the same action completes in ~0.06s with no such request, confirming the local install path no longer touches the network.

### DIFF
--- a/apps/desktop/src/components/config/DriverStoreDialog.vue
+++ b/apps/desktop/src/components/config/DriverStoreDialog.vue
@@ -1875,10 +1875,17 @@ watch(driverStoreTab, (tab) => {
                   <Button v-else type="button" variant="default" class="rounded-md" :disabled="isInstallingJdbcPlugin" @click="installJdbcPlugin">
                     {{ isInstallingJdbcPlugin ? t("common.loading") : t("settings.jdbcPluginInstall") }}
                   </Button>
-                  <Button type="button" variant="outline" class="rounded-md" :disabled="isInstallingJdbcPlugin || isUninstallingJdbcPlugin" @click="installJdbcPluginLocal">
-                    <FolderOpen class="h-3.5 w-3.5 mr-1" />
-                    {{ t("driverStore.localInstall") }}
-                  </Button>
+                  <Tooltip>
+                    <TooltipTrigger as-child>
+                      <Button type="button" variant="outline" class="rounded-md" :disabled="isInstallingJdbcPlugin || isUninstallingJdbcPlugin" @click="installJdbcPluginLocal">
+                        <FolderOpen class="h-3.5 w-3.5 mr-1" />
+                        {{ t("driverStore.localInstall") }}
+                      </Button>
+                    </TooltipTrigger>
+                    <TooltipContent side="bottom" class="max-w-100 break-all text-xs">
+                      {{ t("driverStore.localInstallHint") }}
+                    </TooltipContent>
+                  </Tooltip>
                 </div>
               </div>
             </div>
--- a/apps/desktop/src/i18n/locales/en.ts
+++ b/apps/desktop/src/i18n/locales/en.ts
@@ -6648,6 +6648,7 @@ export default {
     updating: "Updating",
     update: "Update",
     localInstall: "Local install",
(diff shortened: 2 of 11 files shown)


t8y2 (changes requested):
The local-install post-check is offline now, but an older mount-time status request can still overwrite the successful local state.

`DriverStoreDialog` starts an asynchronous online status query when the dialog mounts. The backend reads the pre-install manifest before waiting for the network. If a local plugin installation finishes while that request is still pending, the install path correctly reports the new local manifest, but the earlier request can later write its stale pre-install snapshot back into the dialog and make the newly installed plugin appear uninstalled again—especially on an isolated or slow network.

Please move the manifest read until after the network wait, or add request generation/cancellation so only the current status request may update the dialog. Add a regression test that overlaps a delayed initial status request with a successful local installation and verifies that the installed state cannot be reverted.

q396921921:
Thanks for catching that — you're right, the mount-time status check could win the race and clobber a state that had already been correctly updated.

Fixed in 2d5cd504b by moving the manifest read in `jdbc_plugin_status_from_dir` until after the network wait completes, so a delayed status request always reads whatever is on disk at the moment it resolves (post-install, if one finished in the meantime) instead of a snapshot taken before the network call started. Went with this option over request generation/cancellation since it closes the race at the source for every caller of that function, not just the dialog's mount-time check.

Added `delayed_network_status_check_does_not_revert_a_concurrent_local_install`, which spawns a status check with an injected delayed "network" future, writes the post-install manifest to disk while that check is still pending, and asserts the returned status reflects the install rather than the stale pre-install snapshot. Confirmed it fails against the old ordering and passes with the fix.

t8y2 (approved):
Reviewed the current head. The previous offline-installation and concurrency blockers are resolved, the local JDBC plugin path remains fully offline, and the focused regression suites plus exact-head CI are green.

t8y2:
Thanks for the contribution! Merged in 0e55190, will be released in the next version.

PR #6787, opened 2026-08-20
fix(sidebar): address review follow-ups on the cancellable table ops

q396921921:
## Summary

Follow-up to #6675, which made Drop/Empty/Truncate Table cancellable instead of silently failing on large tables (#4990). A second review round on that PR found a few regressions in how it was wired up, which had already merged before this follow-up was ready. This PR contains just the fixes for those findings, rebased on current `main`.

- `executeTreeNodeSqlWithProductionGuard` generated a backend execution id (and the timeout/cancel tracking that goes with it) for every caller, not just the three danger-dialog operations that actually surface a Cancel button. Unrelated callers such as `confirmPasteTable`'s data-copy step got an untracked execution id nobody could act on — a client-side timeout with no way to cancel left the query registered server-side for the full 30-minute detach grace period with no UI reachable to it. The connection's configured timeout is still applied everywhere (matching the main editor), but only callers that opt in now get an execution id.
- Cancelling after the 10s confirmation window elapsed with no answer from the database silently reset the Cancel button with zero feedback, and a retry that later succeeded past that window had its result discarded. The handler now keeps the running-execution state alive and surfaces a toast, and the losing side of the timeout race still finalizes a late confirmation instead of dropping it.
- Extracted the three near-identical `confirmDropTable`/`confirmEmptyTable`/`confirmTruncateTable` bodies into a shared `runDangerOperation` helper.
- Added a guard against opening a second danger dialog while a prior one is still mid-cancellation, `RunningQueries::cancel` now actually removes the task entry (it previously only cleared the interrupt token, leaving it "active" until the 30-minute grace period expired even after a confirmed cancel), and `RegisteredQuery::finish` centralizes the detach-on-timeout branch that was duplicated across every HTTP/Tauri query-execution entry point.

## Test plan

- [x] `cargo test -p dbx-core --lib query_cancel::` — 14/14 passing
- [x] `cargo test -p dbx-core --lib query::` — 140/140 passing
- [x] `cargo clippy -p dbx-core -p dbx-web -p dbx --lib -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `pnpm exec vue-tsc --noEmit` — clean
- [x] `pnpm exec vitest run` — full suite, 1079 files / 10341 tests passing

### DIFF
--- a/apps/desktop/src/components/sidebar/ConnectionTree.vue
+++ b/apps/desktop/src/components/sidebar/ConnectionTree.vue
@@ -1594,6 +1594,10 @@ function openSidebarContextMenu(event: MouseEvent, node: TreeNode, openContextMe
 }
 
 function openSidebarDangerDialog(request: SidebarDangerDialogRequest) {
+  if (sidebarDangerRunningExecutionId.value) {
+    toast(t("contextMenu.dangerOperationAlreadyRunning"), 4000);
+    return;
+  }
   sidebarDangerDialogRequest.value = request;
   sidebarDangerDialogConfirming.value = false;
   // Defense in depth: sidebarDangerDialogCancelling is a singleton shared
--- a/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
+++ b/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
@@ -70,7 +70,6 @@ import { useDatabaseOptions } from "@/composables/useDatabaseOptions";
 import type { ColumnInfo, DatabaseType, TreeNode, TreeNodeType } from "@/types/database";
 import * as api from "@/lib/backend/api";
 import { queryTimeoutSecsForConnection } from "@/lib/sql/queryTimeout";
-import { uuid } from "@/lib/common/utils";
 import { resolveDefaultDatabase } from "@/lib/database/defaultDatabase";
 import { connectionUsesVisibleSchemaFilter } from "@/lib/database/visibleDatabases";
 import { canTreeNodePin, canTreeNodeShowExpander } from "@/lib/sidebar/sidebarTreeItemLayout";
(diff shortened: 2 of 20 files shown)


t8y2 (changes requested):
Two cancellation races still need to be fixed:

1. Each `cancelQuery()` call is raced against a two-second timeout. If the first backend cancellation succeeds slightly later, that `true` result is permanently discarded. The backend has already removed the execution ID, so every retry returns `false`, the frontend never confirms cancellation, and the sidebar singleton can remain blocked. Please keep observing the original promises after the per-attempt timeout, or make cancellation acknowledgement idempotent and queryable.

2. `cancel()` removes the task before DuckDB/MySQL interrupt registration is synchronized. A late `register_interrupt()` can therefore insert a stale closure after the task was removed; guard cleanup then sees no task and leaves the interrupt registered. Please make task/interrupt transitions atomic and ensure late registration is immediately rejected or executed and always cleaned up.

Please add deterministic regressions for a first cancellation that returns `true` after 2.5 seconds and for `register -> cancel -> register_interrupt -> drop`, asserting that both task and interrupt registrations are zero. This PR also has the same stable patch as #6784, so only one of the sibling PRs should remain after the fixes.

q396921921:
Both remaining races are fixed in 5ebd660:

1. **`confirmCancelWithRetry`'s per-attempt race** — the retry loop now tracks every fired `cancelQuery()` attempt instead of discarding the one whose per-attempt timeout elapsed. A slow-but-successful first attempt (e.g. answering at 2.5s) still confirms the cancel even after the loop has moved on to firing a retry; the overall result only reports unconfirmed once every fired attempt has settled `false`. Added a regression test (`confirms the cancel when the first backend response arrives after its own per-attempt timeout`) that resolves the first attempt `true` 0.5s after its own 2s timeout fires.
2. **`RunningQueries::cancel`/`register_interrupt` atomicity** — the interrupt closure now lives inside `RunningTask` itself instead of a second `Mutex<HashMap<..>>`, so task presence and interrupt registration change under one lock. A `register_interrupt()` that arrives after `cancel()` already removed the task now runs the interrupt immediately instead of storing an orphaned closure that would never be cleaned up. Added `late_interrupt_registration_after_cancel_runs_immediately_and_does_not_leak`, which does `register -> cancel -> register_interrupt -> drop` and asserts `registration_counts()` is `(0, 0)`.

Also: since this PR and #6784 carried the same stable patch, I've closed #6784 as a duplicate of this one — this PR now has both the original fix and these two follow-ups.

Verified locally:
- `cargo test -p dbx-core --lib query_cancel::` — 16/16
- `cargo test -p dbx-core --lib query::` — 140/140
- `cargo clippy -p dbx-core -p dbx-web -p dbx --lib -- -D warnings` — clean
- `cargo fmt --check` — clean
- `pnpm run typecheck` — clean
- `pnpm exec vitest run` — 1079/1079 files, 10341/10342 tests (1 pre-existing skip)

t8y2 (commented):
Maintainer patch pushed in f9ce7c8: Web query timeouts now detach only when the client supplied a reachable execution ID, while sidebar destructive-operation cancellation waits for an already-running cancel confirmation before choosing its toast/cleanup path. Added deterministic backend registration and frontend cancellation-race regressions. Verified dbx-core query-cancel tests (16/16), dbx-web query-route tests (5/5), focused desktop Vitest (26/26), app sidebar tests (4/4), pnpm typecheck, oxfmt, cargo fmt, and git diff --check.

t8y2 (approved):
Reviewed the maintainer-patched exact head. Unreachable server-generated timeout registrations are no longer detached, client-supplied execution IDs retain late-cancel support, and the sidebar waits for an in-flight cancel confirmation before selecting its toast and cleanup path. Backend/frontend race regressions and exact-head CI are green.

t8y2:
Thanks for the contribution! Merged in b4015a4, will be released in the next version.

PR #6823, opened 2026-08-20
fix(ai): require exact SQL confirmation for Agent writes

Abeautifulsnow:
## Problem

  AI Agent mode could display a generic write confirmation without showing the exact SQL to be executed. If the model then attempted a direct write tool call, DBX could
  present a second confirmation prompt.

  ## Root cause

  Write confirmation detection accepted generic write-oriented questions without requiring one unambiguous SQL code block. Backend safety handling also lacked a semantic
  fallback for direct unconfirmed write attempts.

  ## Fix

  - Require normal write proposals to contain exactly one reviewable SQL code block.
  - Bind confirmed execution to that exact SQL statement and target snapshot.
  - Intercept direct unconfirmed write/DDL tool calls before database dispatch.
  - Show a localized exact-SQL confirmation for non-production targets.
  - Block production writes with manual-review guidance and no executable confirmation.
  - Add localized messages and regression coverage for frontend and Rust safety paths.

  ## Compatibility

  Read-only Agent queries and existing action confirmations remain unchanged.

  Confirmed writes remain one-shot and limited to the exact SQL reviewed by the user. Production databases remain protected from AI-initiated writes. MongoDB command
  handling remains read-only.

  ## Validation

  - `pnpm exec vitest run apps/desktop/src/lib/__tests__/ai/aiWriteSqlConfirmation.spec.ts apps/desktop/src/components/editor/__tests__/
  AiAssistant.writeConfirmationProposal.spec.ts` — passed (72 tests)
  - `pnpm typecheck` — passed
  - `git diff --check` — passed

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [x] `make check` 通过
- [x] 相关测试通过


t8y2 (commented):
Maintainer patch applied from `21a9b00e2cd82769a8f3286b9ef80072df6416a9` to `11d55348cc6741acb4530492db6ad92dc59f15a3`.

Root cause: the exact confirmed SQL was stored in `AgentSqlPermissions`, then cloned for every sequential `execute_query` call and reused on later agent turns. A confirmation described as one-shot could therefore authorize the same write more than once.

Fix: snapshot the permissions for the current `execute_query` call and atomically clear the run-local confirmation before dispatching the first matching write/DDL. Read-only calls do not consume it; duplicate calls in the same batch and matching calls in later turns must request a new confirmation.

Verification on the patched remote worktree:
- `cargo test -p dbx-core --no-default-features confirmed_write -- --nocapture` — 11 relevant tests passed, including same-batch replay, cross-turn replay, target binding, production fail-closed, and read-before-write coverage.
- `rustfmt --edition 2021 crates/dbx-core/src/agent_loop.rs crates/dbx-core/src/agent_tools.rs`

t8y2 (approved):
Maintainer patch verified with focused one-shot authorization regressions.

t8y2:
Thanks for the contribution! Merged in 12da615daa, will be released in the next version.

PR #6933, opened 2026-08-21
fix(sql): resolve MyBatis <where> tags before executing SQL

### LINKED ISSUE #6902: [Bug] 占位符好像没识别出来<where>关键字
### 数据库类型和版本

Mysql

### 问题描述

Image

如图，他能识别foreach但是没法识别到 where。

不过能识别大部分已经很厉害了，其实也足够了用，抱歉我吹毛求疵了

### 支持信息

DBX 版本: v0.5.91
运行环境: 桌面端
操作系统: macOS 26.6.1
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Problem

`sqlParameters.ts` only recognized `<foreach>` as a MyBatis dynamic-SQL tag. A `<where>` wrapper (a very common MyBatis pattern, often nesting a `<foreach>`) was left completely untouched: the "SQL Parameters" dialog substituted the bound values inside it, but the literal `<where>` / `</where>` tags stayed in the SQL text sent to the database — causing a real syntax error at execution time (e.g. MySQL `ERROR 1064`), not just a missing highlight as the title suggests.

Example from the issue:

```sql
SELECT collect.collect_uid, count(1) AS bagNum
FROM collect_info collect
JOIN collect_big_bag bag ON collect.collect_uid = bag.collect_uid
<where>
  collect.collect_uid IN
  <foreach collection="collectUidList" item="uid" open="(" separator="," close=")">
    #{uid}
  </foreach>
</where>
GROUP BY ...
```

`<foreach>` correctly expanded, but the SQL preview kept `<where>` / `</where>` as literal text, which MySQL then rejected with a syntax error.

## Solution

- Generalized the existing foreach-closing-tag matcher (`findMatchingForeachClose` / `readForeachTagAt`) into tag-name-agnostic `findMatchingXmlTagClose` / `readXmlTagAt`, and reused it to recognize `<where>...</where>` the same way `<foreach>` is recognized (still comment/string/dollar-quote aware, so a fake `<where>` inside a comment or string literal doesn't confuse the matcher).
- On substitution, the `<where>` body is resolved recursively first (so nested placeholders and a nested `<foreach>` still work), then trimmed, a leading `AND`/`OR` is stripped (case-insensitive), and the result is prefixed with `WHERE ` when non-empty — or dropped entirely when the body is empty. This matches real MyBatis `WhereSqlNode` semantics.
- `<where>` itself is not surfaced as a bindable SQL parameter in the "SQL Parameters" dialog (it has no value of its own), but placeholders nested inside it (including a nested `<foreach>`) still are.

## Testing

- Added 8 new unit tests in `sqlParameters.spec.ts` covering: basic `<where>` stripping, leading `AND`/`OR` removal, empty-body handling (no `WHERE` at all), that `<where>` itself isn't listed as a parameter while nested placeholders are, a `<foreach>` nested inside `<where>`, ignoring where-like tags inside strings/comments, and the `enabledSyntaxes` opt-out path.
- Full existing suite still green: `sqlParameters.spec.ts` 99/99, plus `useSqlExecution` / `sqlVariableSyntax` / `executableStatementRangeCache` / `sqlFormatter` 132/132, plus the whole `apps/desktop/src/lib/__tests__/sql/` directory 1032/1032. `vue-tsc --noEmit` is clean.
- Manually reproduced against a real MySQL 8.4.6 instance through the actual UI (dev build): before the fix, executing a `<where><foreach>...</foreach></where>` template SQL through the "SQL Parameters" dialog produced a real `ERROR 1064` syntax error from MySQL. After the fix, the same steps produce a valid `WHERE ... IN (...)` clause that executes successfully.

Fixes #6902

### DIFF
--- a/apps/desktop/src/lib/__tests__/sql/sqlParameters.spec.ts
+++ b/apps/desktop/src/lib/__tests__/sql/sqlParameters.spec.ts
@@ -666,6 +666,56 @@ describe("substituteSqlParameters", () => {
     expect(substituteSqlParameters(sql, {}, { enabledSyntaxes: ["shell"] })).toBe(sql);
   });
 
+  it("strips a MyBatis <where> wrapper and prefixes its body with WHERE", () => {
+    const sql = "select * from tasks <where> status = #{status} </where>";
+
+    expect(substituteSqlParameters(sql, { status: { kind: "string", value: "open" } }, { enabledSyntaxes: ["mybatis"] })).toBe("select * from tasks WHERE status = 'open'");
+  });
+
+  it("strips a leading AND/OR from a MyBatis <where> body, case-insensitively", () => {
+    const andSql = "select * from tasks <where> AND status = #{status} </where>";
+    const orSql = "select * from tasks <where> or status = #{status} </where>";
+
+    expect(substituteSqlParameters(andSql, { status: { kind: "string", value: "open" } }, { enabledSyntaxes: ["mybatis"] })).toBe("select * from tasks WHERE status = 'open'");
+    expect(substituteSqlParameters(orSql, { status: { kind: "string", value: "open" } }, { enabledSyntaxes: ["mybatis"] })).toBe("select * from tasks WHERE status = 'open'");
+  });
+
+  it("renders an empty MyBatis <where> body as no WHERE clause at all", () => {
+    const sql = "select * from tasks <where>   </where> order by id";
+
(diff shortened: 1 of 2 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in a2c9317, will be released in the next version.

PR #6957, opened 2026-08-22
fix(sidebar): prevent table search from blanking on Enter

### LINKED ISSUE #6951: [Bug] 数据库内搜索表回车后空白
### 数据库类型和版本

MYSQL 8.0

### 问题描述

mysql在数据库内搜索表时出现空白情况，啥都没了

### 支持信息

DBX 版本: v0.5.91
运行环境: 桌面端
操作系统: macOS 26.6.1
系统架构: aarch64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

Image

0verme:
## Summary

- refresh recycled sidebar rows when table-search results replace rows without changing the flattened row count
- keep Enter scoped to the table-search input
- add regression coverage for the virtual-row transition and keyboard ownership

## Root cause

Enter was not invoking a tree action: the table-search input is not inside a form, its containing control stops `keydown` propagation, and the sidebar row handler has no Enter action. The blank state was caused by the following tree projection/rendering transition.

`ConnectionTree` already reconciled `RecycleScroller` after a row-count change, but returned early when a table-search projection replaced rows with the same flattened count. The recycled pool could therefore retain a stale viewport/row binding; after the next repaint (including the Enter interaction on the focused input), the database area could appear blank until another scroll/update.

The existing `34a158765` fix covered the count-change case. This change also compares stable row render identities (`renderKey`, `poolType`, and depth) so same-length replacements refresh the scroller.

## Verification

- `pnpm vitest run apps/desktop/src/composables/__tests__/useFlatTree.spec.ts packages/app-tests/useFlatTree.test.ts packages/app-tests/sidebarTableSearchControl.test.ts packages/app-tests/sidebarTableSearchIndex.test.ts packages/app-tests/sidebarTreeAffordances.test.ts apps/desktop/src/components/sidebar/__tests__` ✅ (21 files, 103 tests)
- `pnpm typecheck` ✅
- `pnpm exec oxfmt --check <changed files>` ✅
- `pnpm exec oxlint --vue-plugin <changed files>` ✅
- pre-fix regression assertion failed; post-fix it passes ✅
- Cargo/Rust and full Tauri build: not run (frontend-only change)
- macOS ARM manual verification: not run

Fixes #6951

### DIFF
--- a/apps/desktop/src/components/sidebar/ConnectionTree.vue
+++ b/apps/desktop/src/components/sidebar/ConnectionTree.vue
@@ -46,7 +46,7 @@ import {
   type SidebarNodeScrollAlign,
 } from "@/lib/sidebar/sidebarActiveTabTarget";
 import { findLoadedTableTargetForCandidate, queryContextTargetFromCandidate, queryCursorTableCandidate, type QueryCursorTableCandidate } from "@/lib/sql/queryCursorTableTarget";
-import { createFlatTreeIndex, SIDEBAR_TREE_ROW_HEIGHT, SIDEBAR_TREE_PRERENDER_COUNT, SIDEBAR_TREE_SCROLL_BUFFER, flattenTree, shouldVirtualizeFlatTree, type FlatTreeNode } from "@/composables/useFlatTree";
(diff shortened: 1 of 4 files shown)


t8y2 (approved):
Reviewed at exact head ba87c3115fe0e0b5065a25e75482f0a36291e10e.

t8y2:
Thanks for the contribution! Merged in 0908a944e, will be released in the next version.

PR #6991, opened 2026-08-23
feat(xugu): expose spatial query values as WKT

mapan0424:
## Summary

This change improves Xugu spatial query results by decoding supported geometry and geography values into readable WKT text before they reach the result grid. It keeps ordinary binary values and explicit textual expressions unchanged.

## Why

Xugu spatial columns and spatial expressions are returned through the Go driver as binary values. In the desktop client, users could see an opaque BLOB-like value instead of the geometry content, which also prevented the existing spatial preview from recognizing the result.

## Implementation

- Added Xugu-specific spatial value normalization in the Xugu agent.
- Detects supported Xugu geometry/geography metadata and converts valid WKB payloads to WKT.
- Preserves NULL values as NULL.
- Leaves ordinary BLOB values untouched to avoid false spatial detection.
- Keeps explicit `ST_AsText(...)` results as text/CLOB values.
- Preserves SRID information so the UI can identify supported coordinate systems.
- Reuses the existing result-grid and spatial-preview flow; no changes are made to other database agents.

## Validation

The following checks passed after rebasing onto the latest upstream `main`:

- `go test . -count=1`
- `go test -race . -count=1`
- `go vet .`
- `go build .`
- `git diff --check`
- Rust WKB unit tests: 6 passed

Desktop DBX validation also covered:

- Geometry values including point, line, polygon, and collection results.
- Geography values with EPSG:4326 and map preview.
- Geometry with EPSG:3857 and map preview.
- NULL spatial values.
- Multiple spatial columns in one result.
- Ordinary BLOB/WKB results, confirming they are not misclassified as spatial values.
- Explicit WKT text results.
- A 512-row spatial result paged through the final page, with WKT still rendered correctly.

The Xugu Agent used for the desktop verification was temporary and the original local Agent was restored afterward.

### DIFF
--- a/agents/drivers/xugu/main.go
+++ b/agents/drivers/xugu/main.go
@@ -238,6 +238,12 @@ var xuguDataTypes = []string{
 	"DOUBLE[]",
 	"FLOAT",
 	"GUID",
+	"GEOMETRY",
+	"GEOGRAPHY",
+	"BOX2D",
+	"BOX3D",
+	"SPHEROID",
+	"RASTER",
 	"INT",
 	"INTEGER",
 	"INTEGER[]",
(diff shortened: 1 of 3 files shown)


t8y2:
Thanks for the contribution! Merged in d7adff54b77d, will be released in the next version.

PR #7090, opened 2026-08-24
fix(kingbase): 显示表结构字符字段长度

zipg:
## 问题

KingbaseES V9 的列元数据可能使用内部类型名 `bpchar`，并通过 `character_maximum_length` 单独返回长度。表结构编辑器的字符长度类型集合遗漏了 `bpchar`，导致 DDL 中可见 `character(n)`，但“字段”页长度为空。

## 修改

- 将 `bpchar` 纳入表结构编辑器的字符长度元数据类型
- 增加 `bpchar + character_maximum_length` 前端回归覆盖
- 扩展 Kingbase 集成测试，同时校验 `varchar(64)` 和 `character(32)` 的类型及长度元数据

## 实际验证

在真实 `KingbaseES V009R001C010` 实例上创建临时测试表并通过当前 Kingbase Go Agent 读取，确认：

- `varchar(64)` 返回 `character varying(64)`，长度为 `64`
- `character(32)` 返回 `character(32)`，长度为 `32`
- 测试结束后临时对象由集成测试自动清理

通过：

- `pnpm vitest run apps/desktop/src/lib/__tests__/table/tableStructureEditorState.spec.ts`：36/36
- `go test ./...`（`agents/drivers/kingbase-go`）
- `go test -run TestKingbaseIntegration -v`（真实 KingbaseES V9）
- `pnpm typecheck`

### DIFF
--- a/agents/drivers/kingbase-go/integration_test.go
+++ b/agents/drivers/kingbase-go/integration_test.go
@@ -58,7 +58,7 @@ func TestKingbaseIntegration(t *testing.T) {
 		}
 	})
 
-	mustExecute(t, server, "CREATE TABLE "+qualifiedSchema+"."+quoteIdentifier(parent)+" (id integer PRIMARY KEY, name varchar(64) NOT NULL)")
+	mustExecute(t, server, "CREATE TABLE "+qualifiedSchema+"."+quoteIdentifier(parent)+" (id integer PRIMARY KEY, name varchar(64) NOT NULL, code character(32))")
 	mustExecute(t, server, "COMMENT ON TABLE "+qualifiedSchema+"."+quoteIdentifier(parent)+" IS '订单父表'")
 	mustExecute(t, server, "COMMENT ON COLUMN "+qualifiedSchema+"."+quoteIdentifier(parent)+".id IS '主键编号'")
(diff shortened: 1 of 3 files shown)


t8y2 (approved):
Thanks for the PR! Root cause confirmed: KingbaseES V9 returns the internal type name `bpchar` (without a length suffix) in some modes, with the length carried separately by `character_maximum_length` (`atttypmod - 4`), so adding `bpchar` to `CHARACTER_LENGTH_METADATA_TYPES` is the right fix. The length semantics match DBeaver's `PostgreStringTypeHandler` and the native postgres driver, and native PG-family drivers aren't affected since `format_type()` renders `character(n)` there. The GaussDB M-mode `bpchar` columns get fixed by the same change — a bonus, not a regression. Integration test coverage for `character(32)` metadata is good.

Two non-blocking notes: `columnDDLDataType` in `agents/drivers/kingbase-go/kingbase_metadata.go` (~line 2137) doesn't list `"bpchar"`, so if a mode ever returns bare `bpchar` with an empty `FullDataType`, the generated DDL would drop the length — worth a follow-up. And the earlier CI `frontend` failure was a flaky `queryStore.hiddenPrimaryKey.spec.ts` (passes repeatedly on main locally), unrelated to this change.

t8y2:
Thanks for the contribution! Merged in 79042315, will be released in the next version.

PR #7101, opened 2026-08-24
fix(xugu): preserve spatial SRIDs in export and add regressions

mapan0424:
## Summary

This PR completes the XuguDB spatial-data compatibility pass for map previews, spatial indexes, and data export/replay.

The change is intentionally scoped to XuguDB export handling and regression coverage. Existing PostgreSQL, MySQL, and other database export paths are not changed.

## What changed

- Confirmed and regression-tested the shared map-preview path for `GEOGRAPHY` values, including EWKT input and per-cell SRID metadata.
- Added opt-in live XuguDB coverage for WKB, EWKB, EWKT, SRID handling, and multiple geometry types: `POINT`, `MULTIPOINT`, `MULTILINESTRING`, `MULTIPOLYGON`, `POLYGON`, and `GEOMETRYCOLLECTION`.
- Added live coverage for Xugu RTREE spatial indexes, including index listing, indexed columns, and reconstructed table DDL.
- Updated Xugu database export selection to use `ST_AsEWKT` for `GEOMETRY` and `GEOGRAPHY` columns so non-zero SRIDs are not lost.
- Replays Xugu EWKT values with `ST_GeomFromEWKT(...)`; NULL and legacy plain WKT values remain supported.
- Added unit coverage proving the Xugu-only export branch does not affect PostgreSQL behavior.

## Compatibility notes

- Xugu spatial indexes retain `INDEXTYPE IS RTREE`; PostgreSQL `USING GIST` syntax is never emitted for Xugu.
- The live test confirms that `GEOGRAPHY` uses a longitude/latitude SRID (4326), while projected SRIDs such as 3857 remain valid for `GEOMETRY` only.
- No changes were made to the generic frontend geometry parser; the new test verifies that the existing shared parser correctly handles Xugu EWKT output.

## Validation

- Opt-in live Xugu regression suite: `TestLiveXuguSpatialQueryRegression`, `TestLiveXuguSpatialIndexDDL`, and `TestLiveXuguSpatialReplay`
- `cargo test -p dbx-core database_export::tests::xugu_spatial -- --nocapture`
- `pnpm exec vitest run apps/desktop/src/lib/__tests__/dataGrid/geometryMapPreview.spec.ts apps/desktop/src/lib/__tests__/dataGrid/geometryPreview.spec.ts` (13 tests passed)
- `pnpm typecheck`
- `pnpm exec oxfmt --check apps/desktop/src/lib/__tests__/dataGrid/geometryMapPreview.spec.ts`
- `git diff --check`

All local and live Xugu validations passed.

### DIFF
--- a/agents/drivers/xugu/main_test.go
+++ b/agents/drivers/xugu/main_test.go
@@ -1011,6 +1011,11 @@ func TestXuguIndexScopeDDL(t *testing.T) {
 			want:  " INDEXTYPE IS BTREE",
 		},
 		{
+			name:  "spatial index preserves Xugu RTREE type",
+			index: indexInfo{IndexType: indexTypePtr("RTREE")},
+			want:  " INDEXTYPE IS RTREE",
+		},
+		{
 			name:  "local partition index",
 			index: indexInfo{IndexType: &indexType, IsLocal: true},
 			want:  " INDEXTYPE IS BTREE LOCAL",
@@ -1077,6 +1082,26 @@ func TestXuguIndexScopeDDL(t *testing.T) {
 			}
 			if tc.name == "incomplete subpartition keeps valid first level" && strings.Contains(got, "SUBPARTITION") {
 				t.Fatalf("incomplete subpartition metadata must not produce a partial clause: %q", got)
+			}
+		})
+	}
+}
+
+func TestXuguIndexTypeName(t *testing.T) {
+	for _, tc := range []struct {
+		value any
+		want  string
+	}{
+		{value: int64(0), want: "BTREE"},
+		{value: int64(1), want: "RTREE"},
+		{value: int64(2), want: "FULLTEXT"},
+		{value: int64(3), want: "BITMAP"},
+		{value: "RTREE", want: "RTREE"},
+		{value: "vendor-specific", want: "vendor-specific"},
+	} {
+		t.Run(fmt.Sprint(tc.value), func(t *testing.T) {
+			if got := indexTypeName(tc.value); got != tc.want {
+				t.Fatalf("indexTypeName(%v) = %q, want %q", tc.value, got, tc.want)
 			}
 		})
 	}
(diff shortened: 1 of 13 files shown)


t8y2 (commented):
Maintainer patch applied and pushed.

- Old head: `0987c1d15d3c69643933cf03482bdbf5129df604`
- New head: `8874cfaf0df28e369fb35217230b3d1299154a84`
- Rebased onto: `d1d8dc4aec275fb86e35309b71c8ce2342754dae`

Xugu SQL export now carries spatial-column and per-cell SRID metadata through local, query-result, and table-export paths. When `column_types` has a gap, spatial metadata recovers the geometry type and writes WKT plus SRID `3857` as `ST_GeomFromEWKT('SRID=3857;...')` instead of losing the SRID.

Checks run: the focused Rust regression passed; 64 focused frontend Vitest tests passed; changed Rust/TypeScript files were formatted and `git diff --check` passed.

t8y2 (commented):
Maintainer CI follow-up applied and pushed.

- Old head: `8874cfaf0df28e369fb35217230b3d1299154a84`
- New head: `b857679fb03acab44ced374c7aaadabae1e7ee9f`
- Rebased onto: `7dfd0fdd5ac3be34896baf01272ff3b1f5313fff`

The follow-up adds the missing spatial fields to the local SQL-formatting options, safely remaps spatial column indexes without an invalid type predicate, and updates the Rust benchmark initializer. The refreshed base also includes the explicit app-close prompt test correction.

Checks run: frontend typecheck, changed export tests, the full query-store test file, the Rust example build, the Xugu spatial regression, range-diff review, and `git diff --check` passed.

t8y2 (approved):
Approved after the maintainer CI follow-up, exact-head CI, and latest-main merge validation passed.

t8y2:
Thanks for the contribution! Merged in e9fea304b, will be released in the next version.

PR #7123, opened 2026-08-25
feat(import): 添加 SQL 脚本导入功能支持

### LINKED ISSUE #6965: [Bug] mysql导入数据不支持.sql扩展名
### 数据库类型和版本

MYSQL8.0

### 问题描述

Image

无法导入.sql 的数据库文件格式

### 支持信息

DBX版本：0.5.92
Mysql：8.0

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

mr-dadong:
- 在导入对话框中添加 SQL 格式选项和文件图标
- 实现 SQL 脚本中 INSERT 语句的数据解析功能
- 支持多种 SQL 标识符格式和字符串转义
- 添加 SQL 文件的文本编码检测和处理
- 扩展文件选择器支持 .sql 文件扩展名
- 更新多语言资源文件中的 SQL 相关翻译
- 添加 SQL 导入功能的单元测试用例
- 重构代码以统一文本源的编码处理逻辑

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

- [ ] `make check` 通过
- [x] 相关测试通过

## 关联 Issue

Close #6965


t8y2 (changes requested):
request changes: the SQL-script import path can silently change valid SQL data and statement semantics.

1. `crates/dbx-core/src/table_import.rs:1309` applies MySQL-style backslash escaping to ordinary strings, while the `E'...'` handling around line 1472 does the opposite of PostgreSQL semantics. The same parser converts expressions such as `NOW()` into the string `'NOW()'`, and accepts MySQL `REPLACE` but later regenerates a normal `INSERT`.

   PostgreSQL only gives backslashes special meaning in escape string constants: https://www.postgresql.org/docs/current/sql-syntax-lexical.html
   MySQL `REPLACE` has delete-and-insert conflict semantics, not ordinary INSERT semantics: https://dev.mysql.com/doc/refman/8.0/en/replace.html

   Please make the source dialect explicit and preserve strings, expressions, special literals, and statement type. Unsupported constructs should be rejected rather than silently converted.

2. `register_sql_insert_columns()` compares already-unquoted table names with `eq_ignore_ascii_case()`, so PostgreSQL `"Foo"` and `"foo"` are merged into one import target. Preserve quote state and apply the source dialect's identifier rules.

3. The implementation reads, decodes, splits, and materializes the complete file before its first cancellation check. A permitted 100 MiB script therefore has no parsing progress or cancellation and creates several full in-memory copies. Please use a bounded streaming statement/row producer with cancellation and byte progress during parsing.

DBX already has a dialect-aware splitter in `crates/dbx-core/src/sql.rs`; MySQL client `DELIMITER` semantics are documented here: https://dev.mysql.com/doc/refman/8.0/en/mysql-commands.html

PR #7177, opened 2026-08-25
fix(duckdb): work around CURRENT_DATE literal-arithmetic binder bug

### LINKED ISSUE #7153: [Bug] DuckDB driver agent v0.1.11: binder error for 'CURRENT_DATE - integer' ((DATE, INTEGER_LITERAL)); view queries time out / worker exits
## 环境 Environment

- DBX: 0.5.93 (Windows, installed via scoop)
- DuckDB driver (Driver Manager): **0.1.11** (currently the latest in the agent registry)
- Embedded engine version reported by the driver: `SELECT version()` → `v1.5.5`

## 问题描述 Problem

Any query containing `CURRENT_DATE - <integer>` fails or hangs when executed by the DuckDB driver agent:

- In the DBX UI it shows as **"Query timed out after 30 seconds"** (stage: execute), or
- The **DuckDB worker process exits** (`duckdb_worker_exited`), or
- The worker hangs until killed by the 30s query timeout.

The same query runs in **milliseconds** on official DuckDB 1.5.5 (CLI / DBeaver `duckdb_jdbc-1.5.5.0.jar`).

## 最小复现 Minimal reproduction

```sql
SELECT CURRENT_DATE - 4;
```

**Actual (agent v0.1.11):**

```
Binder Error: No function matches the given name and argument types '(DATE, INTEGER_LITERAL)'.
You might need to add explicit type casts.
	Candidate functions:
```

**Expected (official DuckDB 1.5.5):**

```
2026-08-21
```

## 已排查的变体 More variants tested

| Expression | agent 0.1.11 | official DuckDB 1.5.5 |
|---|---|---|
| `CURRENT_DATE - 4` | ❌ Binder Error `(DATE, INTEGER_LITERAL)` | ✅ |
| `CURRENT_DATE - CAST(4 AS INTEGER)` | ❌ Binder Error `(DATE, INTEGER)` | ✅ |
| `CURRENT_DATE::DATE - 4` | ❌ Binder Error | ✅ |
| `CAST(CURRENT_DATE AS DATE) - 4` | ❌ Binder Error | ✅ |
| `current_date() - 4` (function form) | ✅ OK | ✅ |
| `DATE '2026-08-17' - 1` | ✅ OK | ✅ |
| `date_trunc('week', CURRENT_DATE)` | ⚠️ worker process crashes / hangs | ✅ |

Only the **`CURRENT_DATE` keyword node** fails to bind with arithmetic operators; the function form `current_date()` works fine. This looks like a defect in the operator resolution for the keyword node in the driver's embedded DuckDB build (same reported version `v1.5.5` as the official build, but different behavior).

## 影响 Impact

- Any view/macro query that internally uses `CURRENT_DATE` arithmetic times out or exits the worker.
- In my case: a view containing 4 occurrences of `CURRENT_DATE - 4 / - 1` made **all 7 dependent view/macro queries time out** (30s query_timeout kills them).
- `date_trunc('week', CURRENT_DATE)` can crash the worker process.

## 临时规避 Workaround (verified)

Replace `CURRENT_DATE` with `current_date()` in the view/query. The same queries then complete in **0.3–1.7 s** in DBX instead of timing out at 30 s, and results are byte-identical to official DuckDB.

Happy to provide more details or test a patched driver build if needed.

q396921921:
## Summary

DuckDB's bundled embedded engine (reports `v1.5.5` via `SELECT version()`) fails to bind `CURRENT_DATE - <int literal>` — e.g. `SELECT CURRENT_DATE - 4` — with:

```
Binder Error: No function matches the given name and argument types '(DATE, INTEGER_LITERAL)'.
```

even though the functionally identical `current_date() - 4` binds and executes fine. This is a genuine bug in the bundled DuckDB engine itself (confirmed with a minimal repro directly against the `duckdb` crate, independent of any dbx code), not something introduced by dbx. Since views/macros commonly use `CURRENT_DATE` arithmetic, this also caused the view-timeout / worker-exit symptoms described in the issue.

Confirmed the failure is narrowly scoped to a raw (uncast) numeric literal directly against the `CURRENT_DATE` keyword node via `+`/`-`:
- `CURRENT_DATE - 4` → fails
- `current_date() - 4` → works (function-call form)
- `CURRENT_DATE - CAST(4 AS INTEGER)` / `CURRENT_DATE - 4::INTEGER` → works (already-cast literal)
- `CURRENT_DATE + 4` → works (only `-` with a raw literal is affected in some cases; rewriting `+` too is harmless)

## Fix

`agents/drivers/duckdb/src/sql.rs`: added `rewrite_duckdb_current_date_literal_arithmetic`, a narrowly-scoped SQL text rewrite that converts `CURRENT_DATE` to `current_date()` only when:
- not inside a string literal or comment
- not immediately preceded by `.` (so `t.current_date`, a real qualified column, is left alone)
- not immediately preceded/followed by identifier characters (so aliases like `AS current_date` and DDL column definitions are left alone)
- immediately followed (optionally after whitespace) by `+`/`-` and a raw numeric literal

Wired into the single query-execution entry point in `agents/drivers/duckdb/src/query.rs` (`duckdb_execute_with_max_rows`), so it covers both plain queries and `CREATE VIEW ... AS SELECT ... CURRENT_DATE - N` (the view scenario from the issue).

## Testing

- New unit tests in `sql.rs` covering the rewrite (including that it stays clean for `CURRENT_DATE - 4::INTEGER`, `t.current_date - 4`, `SELECT 1 AS current_date`, `CREATE TABLE (current_date INTEGER)`, and comments/strings containing the pattern).
- New integration test in `query.rs` (`duckdb_execute_resolves_current_date_arithmetic`) that runs `SELECT CURRENT_DATE - 4` and `CREATE VIEW ... CURRENT_DATE - 4 ...` against a real in-memory DuckDB connection. Verified it genuinely fails without the fix (reproduces the exact issue error) and passes with it.
- `cargo test --release` (agents/drivers/duckdb): 48/48 passing.
- `cargo clippy --release --all-targets` / `cargo fmt --check`: clean.
- Manually verified end-to-end through the real `dbx-web` app + browser: before the fix, running `SELECT CURRENT_DATE - 4;` in the query editor against a DuckDB connection shows the exact binder error from the issue; after the fix, it returns the correct date and the result column is labeled `(current_date() - 4)`, confirming the rewrite path executed.

Fixes #7153

### DIFF
--- a/agents/drivers/duckdb/src/query.rs
+++ b/agents/drivers/duckdb/src/query.rs
@@ -145,6 +145,21 @@ mod tests {
     }
 
     #[test]
+    fn duckdb_execute_resolves_current_date_arithmetic() {
+        let con = duckdb::Connection::open_in_memory().expect("connect in-memory DuckDB");
+
+        let result = duckdb_execute(&con, "SELECT CURRENT_DATE - 4 AS d").expect("execute CURRENT_DATE - literal");
+        let expected = (Utc::now().date_naive() - ChronoDuration::days(4)).to_string();
+        assert_eq!(result.rows[0][0], serde_json::Value::String(expected));
+
+        con.execute_batch("CREATE VIEW recent AS SELECT CURRENT_DATE - 4 AS start_date, CURRENT_DATE - 1 AS end_date")
+            .expect("create view using CURRENT_DATE arithmetic");
+        let view_result =
+            duckdb_execute(&con, "SELECT * FROM recent").expect("query view built on CURRENT_DATE arithmetic");
+        assert_eq!(view_result.rows.len(), 1);
+    }
+
+    #[test]
     fn duckdb_execute_formats_temporal_values_by_column_type() {
         let con = duckdb::Connection::open_in_memory().expect("connect in-memory DuckDB");
         let result = duckdb_execute(
@@ -378,6 +393,8 @@ pub fn duckdb_execute_with_max_rows(
 ) -> Result<db::QueryResult, String> {
     let start = std::time::Instant::now();
     let row_limit = query_result_row_limit(max_rows);
+    let sql = crate::sql::rewrite_duckdb_current_date_literal_arithmetic(sql);
+    let sql = sql.as_ref();
 
     if crate::sql::starts_with_duckdb_result_sql_keyword(sql) {
         let mut stmt = con.prepare(sql).map_err(|e| e.to_string())?;
--- a/agents/drivers/duckdb/src/sql.rs
+++ b/agents/drivers/duckdb/src/sql.rs
@@ -193,6 +193,120 @@ fn parse_identifier_after_as(input: &str) -> Option<String> {
     (!name.is_empty()).then(|| name.to_string())
 }
 
+/// Works around a DuckDB binder bug: `CURRENT_DATE - 4` (a bare numeric literal directly
+/// against the `CURRENT_DATE` keyword node) fails with "No function matches ... (DATE,
+/// INTEGER_LITERAL)", while the functionally identical `current_date() - 4` binds fine.
+/// Rewrites only that narrow pattern (`CURRENT_DATE` immediately followed by `+`/`-` and a raw
+/// numeric literal) to the function-call form, so plain `CURRENT_DATE`, qualified references
+/// (`t.current_date`), aliases (`AS current_date`), and already-cast literals are left untouched.
+pub fn rewrite_duckdb_current_date_literal_arithmetic(sql: &str) -> std::borrow::Cow<'_, str> {
+    const KEYWORD: &str = "CURRENT_DATE";
+    let bytes = sql.as_bytes();
+    let mut rewritten: Option<String> = None;
+    let mut copied_to = 0usize;
+    let mut index = 0usize;
+    let mut in_single = false;
+    let mut in_double = false;
+
+    while index < bytes.len() {
+        match bytes[index] {
+            b'\'' if !in_double => {
+                if in_single && bytes.get(index + 1) == Some(&b'\'') {
+                    index += 2;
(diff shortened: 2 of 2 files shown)


t8y2 (commented):
Maintainer patch applied in `e343f473a`.

The `CURRENT_DATE` literal-arithmetic workaround scanned dollar-quoted string bodies as executable SQL, so text such as `$$CURRENT_DATE - 4$$` was silently changed. The scanner now reuses the existing dollar-quote delimiter parser and skips complete untagged and tagged dollar-quoted strings, including unmatched delimiters through end of input, while preserving the intended rewrite for real `CURRENT_DATE +/- <numeric literal>` expressions.

Regression coverage now includes untagged strings, tagged strings, different nested-looking tags, unmatched dollar quotes, and real arithmetic following a quoted value.

Checks run:

- `cargo test --manifest-path agents/drivers/duckdb/Cargo.toml --locked sql::tests -- --nocapture` — 6 passed
- `cargo test --manifest-path agents/drivers/duckdb/Cargo.toml --locked query::tests::duckdb_execute_resolves_current_date_arithmetic -- --exact --nocapture` — 1 passed
- `cargo test --manifest-path agents/drivers/duckdb/Cargo.toml --locked` — 39 unit tests and 10 worker-process tests passed
- Sequential worker stdio probes preserved both `$$CURRENT_DATE - 4$$` and `$tag$CURRENT_DATE + 2$tag$` as string values while real arithmetic still returned dates
- `cargo fmt --manifest-path agents/drivers/duckdb/Cargo.toml -- --check`
- `git diff --check`

t8y2:
Thanks for the contribution! Merged in aa91a0aaa, will be released in the next version.

PR #7192, opened 2026-08-25
feat(mcp): add governed table imports and Milvus semantic tools

xingranya:
## Summary

This PR adds governed local table-import and Milvus semantic-search capabilities to the DBX MCP server while reusing `dbx-core` parsing and database drivers.

- Add `dbx_preview_import_file`, `dbx_prepare_table_import`, `dbx_start_table_import`, `dbx_get_import_status`, and `dbx_cancel_import`.
- Add dedicated `dbx_vector_search` and `dbx_vector_upsert_file`; keep v1 batch deletion registered but fail-closed as `VECTOR_DELETE_DISABLED_V1` until DBX has an authoritative publication-state source.
- Classify vector REST entity operations so `dbx_execute_query` cannot bypass the dedicated tools.
- Stream governed XLSX/XLSM imports into a server-generated, unique staging table whose source and lineage columns are all `TEXT`.
- Preserve duplicate headers by 1-based source position and stable canonical names.
- Require exact `semantic_version`, `approval_status=approved`, and a `YYYY-MM-DD` activity window for every semantic search.
- Correct Milvus single-query response flattening, including `data: [[]]` as zero rows.
- Make connection management opt-in for an unscoped maintenance process, and correctly resolve every named connection in a multi-ID runtime scope.

## Safety model

- Local import tools fail closed unless `DBX_MCP_IMPORT_ROOTS` is configured; web mode returns `IMPORT_UNSUPPORTED_IN_WEB_MODE_V1`.
- Absolute paths, canonical allowed roots, regular files, `O_NOFOLLOW`, file-descriptor identity checks, SHA-256 revalidation, single-use 30-minute plans, and task-private snapshots protect the preview/prepare/start boundary.
- Inspection/import concurrency, timeouts, ZIP entry and decompression budgets, row/cell/string/output limits, disk-space checks, bounded XLSX channels, cancellation, and RAII cleanup bound resource use.
- Imports finish their governed snapshot before opening the database and can write only a new server-generated staging relation.
- Semantic JSONL is fully validated against the Milvus collection contract before the first network write, including UTF-8 byte limits, checksums, version consistency, embedding shape, and card ownership.
- Generic vector entity search/query/upsert/delete calls are blocked; collection diagnostics remain available where already allowed.
- `dbx_add_connection`, `dbx_duplicate_connection`, and `dbx_remove_connection` are hidden by default. They require `DBX_MCP_ENABLE_CONNECTION_MANAGEMENT=1` and no runtime scope; method-level guards remain in place.

## v1 boundaries

- Governed imports support XLSX/XLSM and explicitly UTF-8 CSV/TSV.
- Legacy `.xls`, JSON table imports, GBK, UTF-16, and automatic encoding detection remain bounded preview-only and fail before database access during prepare.
- Preview row counts and used ranges expose exactness flags; a bounded preview is not reported as a full-file count.
- Semantic deletion remains disabled rather than trusting a caller-provided publication flag.

## Validation

- `cargo test -p dbx-mcp --no-default-features`: 93 passed, 4 ignored, 7 filtered out.
- `cargo test -p dbx-core --lib`: 5,447 passed, 60 ignored.
- `cargo check -p dbx-mcp`: passed.
- `cargo check -p dbx-core --example data_transfer_bench --locked --no-default-features`: passed; covers the all-target benchmark initializer added for source row lineage.
- `cargo clippy -p dbx-mcp --no-default-features`: 0 errors; two pre-existing `nonminimal_bool` warnings outside this change.
- `pnpm --filter @dbx-app/mcp-server test`: 1 test file, 2 tests passed.
- An opt-in local XLSX regression fixture verified a 145-column sheet, repeated headers, stale used-range metadata, 81 all-empty columns, absolute source row numbers, complete row hashes, and all-`TEXT` governed mappings without opening a database or logging business cell values. The fixture itself is not included.
- Protocol tests cover structured MCP output, path/scope rejection, immutable plans, staging isolation, connection-management hiding, multi-connection routing, vector permission boundaries, and semantic-version filtering.

No production database, Milvus collection, credentials, or private fixture data is included in this PR.


xingranya:
CI follow-up: the original rust-test and rust-fmt-clippy failures had the same compile cause: `ParsedImportFile` gained `source_row_numbers`, while `crates/dbx-core/examples/data_transfer_bench.rs` still used the old initializer. This is fixed, and the branch has been cleanly rebased onto the current `main` (no workflow-file changes in the PR diff). The latest compatibility commit also preserves 1-based source record numbers for the new SQL import path on current main. Local merged-base validation passed: `cargo fmt --check`; 19 SQL-import tests; `cargo check -p dbx-mcp --locked`; `cargo check -p dbx-core --example data_transfer_bench --locked --no-default-features --features sqlite-bundled`; and 93 DBX MCP tests. The new CI run 33080745104 is `action_required` with zero jobs. Could a maintainer approve this Actions run?

t8y2 (changes requested):
Request changes — this is careful work with a genuinely good safety model (fail-closed vector risk classification, TOCTOU hardening, zip budgets, single-use plans), but it needs one rebase plus a policy integration before merge.

**Blocker — allowlist bypass.** The branch is based on a main from before the per-tool MCP allowlist landed, so it currently conflicts (8 files) and none of the eight new tools call `ensure_tool_allowed`. As it stands, an operator restricting `allowed_tool_names` in MCP settings to read-only tools could still invoke `dbx_start_table_import` and `dbx_vector_upsert_file`. Please rebase onto current main and add the `ensure_tool_allowed("<tool>")` guard to each new tool, and update the tool-count assertions (main is at 17 tools, so 25 with your additions; also decide/confirm whether `list_tools` should filter them via `policy_filtered_tools`).

Two smaller items while you're in there:

- `dbx_vector_upsert_file` reads a local JSONL under `DBX_MCP_IMPORT_ROOTS` but, unlike the five import tools, has no `web_mode` guard — if file-based tools are local-mode-only by contract, it should return `IMPORT_UNSUPPORTED_IN_WEB_MODE_V1` too.
- `validate_embedding` hardcodes 1024 dimensions while every other knob (top-k, collections, output fields) has an env override; a `DBX_MCP_VECTOR_DIMENSION`-style option would keep 1024 as the default without locking out 768/1536-dim collections. Also worth gating the per-row JSON measurement in the default XLSX import on `max_batch_bytes != usize::MAX`.

Flagging for maintainers (product decisions, not code issues — we'll follow up here): connection add/duplicate/remove are now default-off behind `DBX_MCP_ENABLE_CONNECTION_MANAGEMENT=1`, Postgres-only staging, and Asia/Shanghai business dates baked into the semantic-card contract.

xingranya:
Implemented the requested review changes in a1b0423 and rebased the branch onto upstream main 715261f. The PR remains mergeable after later main activity.

- Added ensure_tool_allowed at the entry of all eight import/vector tools, before file, plan, connection, or write work.
- Confirmed list_tools uses policy_filtered_tools. A protocol regression verifies all eight tools are hidden by allowed_tool_names and that direct stale calls return TOOL_OUT_OF_SCOPE.
- Added the local-mode guard to dbx_vector_upsert_file. Web mode returns IMPORT_UNSUPPORTED_IN_WEB_MODE_V1 before connection resolution or JSONL access.
- Added DBX_MCP_VECTOR_DIMENSION as a positive integer setting with a 1024 default. Search vectors and semantic JSONL validation share it; 768 and 1536 are covered.
- Skipped per-row XLSX JSON byte measurement when max_batch_bytes is usize::MAX, while preserving finite-budget enforcement.
- Updated tool counts for current main: 25 defined and 22 in default daily mode; without mq-admin, 24 and 21.
- Preserved current product boundaries: connection management remains default-off, staging remains PostgreSQL-only, and active_at remains an Asia/Shanghai business date.

Local validation:
- cargo test -p dbx-mcp --locked: 109 passed, 5 ignored
- dbx-core table_import tests: 154 passed
- no-mq tool-count unit and protocol tests passed with sqlite-bundled
- the exact GitHub Actions fast-feature cargo clippy workspace command passed with all targets and -D warnings
- cargo fmt --check, changed-file rustfmt checks, and git diff --check passed

Both force updates used exact force-with-lease protection. Please re-review when the new workflow is available.

PR #7256, opened 2026-08-26
perf(kafka): batch consumer group subscriptions into one agent round trip

### LINKED ISSUE #7163: [Bug] Kafka加载主题、订阅、消费组速度慢
### 来源
QQ 群反馈 · **反馈人**：悟空

### AI 摘要
Kafka 加载主题、订阅、消费组时整体速度过慢，点击后至少需要十几秒才能响应，请优化性能。

### 原始描述
kafka加载主题，订阅，消费组整体都显示速度好慢啊，优化一下啊，点一下之少十几秒

0verme:
Fixes #7163

## Problem

Loading Kafka **Topics / Subscriptions / Consumer Groups** could take tens of seconds on remote clusters. The worst offender was the Subscriptions list path.

## Root Cause

`KafkaAdmin::list_subscriptions` issued **`1 + 2N` serial JSON-RPC round trips** through the `AgentDriverClient` mutex (single-request dispatch on the agent's stdin/stdout):

```text
mq_list_consumer_groups                (1 RPC)
  → for each of N consumer groups:
      mq_describe_consumer_group        (N RPC)
      mq_get_consumer_lag               (N RPC)
```

Measured against the shared Kafka 4.3.1 test instance with 40 consumer groups (all EMPTY, with committed offsets):

| Path | JSON-RPC calls | Elapsed |
| --- | --- | --- |
| Subscriptions (before): list + describe×40 + lag×40 | **81** | lag probes alone **27.7 s** (avg 691 ms/RPC) |
| Consumer-group snapshot (batch) | 1 | 1.56 s |

The Topics path was already batched (`listTopics` + one `describeTopics` for all names), and the Consumer-Groups page already used the batched `mq_get_consumer_group_snapshot`; the subscriptions list was the hot spot.

## Solution

Move the per-group work into the Kafka agent so the whole listing happens in **one JSON-RPC call** with a handful of Kafka Admin batch requests:

- `KafkaAgent.listConsumerGroups` (a.k.a. `mq_list_consumer_groups`) now accepts an optional `topic`:
  1. `listConsumerGroups` (unchanged)
  2. `describeConsumerGroups(groupIds)` — batch description of every group (members + assignments)
  3. `listConsumerGroupOffsets(specs)` — batch committed offsets for every group
  4. when `topic` is set: deduplicate that topic's partitions and resolve end offsets with one `listOffsets(latest)`
  5. each group row now carries `members[]`, `committedOffsets[]`, `endOffsets[]`
- `KafkaAdmin::list_subscriptions` sends one `mq_list_consumer_groups {topic}` call and filters/aggregates in memory (new pure helper `kafka_subscription_from_group_row`), matching the previous semantics:
  - visible if it has active assignments on the topic **or** committed offsets on the topic
  - lag = Σ(max(0, endOffset − committedOffset)) over the topic's partitions
  - a partition without a resolved end offset contributes nothing (unknown ≠ zero lag)

**Partial failure handling:** a group that fails describe or offsets degrades to empty members/offsets instead of failing the batch or the page; a missing end offset for one partition keeps the rest usable.

**Compatibility:** without `topic`, the RPC returns the original shape (plus the new fields); batch DescribeGroups/OffsetFetch/ListOffsets downgrade automatically on older brokers via kafka-clients; the snapshot RPC and the single-group describe/lag RPCs are untouched.

## Validation

- **Rust (targeted):** `cargo test -p dbx-core --lib --no-default-features --features mq-admin -- mq::` → **151 passed** (kafka adapter: 22, incl. 4 new cases: offline group with committed offsets, active assignment without offsets, missing end offset, cross-topic offset filtering)
- **Java agent (targeted):** `:kafka:test` → **88 passed** (4 new cases for the offset-row helpers)
- **Real Kafka 4.3.1 shared test instance, 40 consumer groups seeded with committed offsets:**

| | JSON-RPC calls | Elapsed |
| --- | --- | --- |
| Before (per-group describe + lag) | 81 | 27.7 s+ |
| After (`mq_list_consumer_groups` + `topic`) | **1** | **0.32–0.42 s** |

(with `topic` filtering, one RPC returns all groups with members, committed offsets and end offsets; the same data the old path needed 81 round trips for)

Seeded test groups were deleted afterwards.

### DIFF
--- a/agents/drivers/kafka/src/main/java/com/dbx/agent/kafka/KafkaAgent.java
+++ b/agents/drivers/kafka/src/main/java/com/dbx/agent/kafka/KafkaAgent.java
@@ -956,20 +956,182 @@ public final class KafkaAgent {
     private static Object listConsumerGroups(JsonObject params) throws Exception {
         AdminClient admin = requireAdmin();
         int timeout = requestTimeout(params);
+        String filterTopic = stringOrEmpty(params, "topic");
+
         Collection<ConsumerGroupListing> groups = admin.listConsumerGroups(
                 new ListConsumerGroupsOptions().timeoutMs(timeout))
             .all().get(timeout, TimeUnit.MILLISECONDS);
 
+        List<String> groupIds = groups.stream()
+            .map(ConsumerGroupListing::groupId)
+            .sorted()
+            .toList();
+        if (groupIds.isEmpty()) {
+            return Collections.singletonMap("groups", Collections.emptyList());
+        }
+
+        // Batch-describe every group in one Kafka Admin request. Per-group
+        // failures degrade to an empty member list instead of failing the batch.
+        Map<String, ConsumerGroupDescription> descriptions = new HashMap<>();
+        DescribeConsumerGroupsResult described = admin.describeConsumerGroups(
+            groupIds,
+            new DescribeConsumerGroupsOptions().timeoutMs(timeout)
+        );
+        for (String groupId : groupIds) {
+            try {
+                descriptions.put(
+                    groupId,
+                    described.describedGroups().get(groupId).get(timeout, TimeUnit.MILLISECONDS)
+                );
+            } catch (Exception error) {
+                logger().debug("Consumer group details unavailable for {}: {}", groupId, normalizeErrorMessage(error));
+            }
+        }
+
+        // Batch-resolve committed offsets for every group in one Admin request.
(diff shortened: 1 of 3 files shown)


t8y2:
Thanks for the contribution! Merged in ef22b84, will be released in the next version.

PR #7315, opened 2026-08-26
fix(http-tunnel): reduce round-trip latency under active traffic

### LINKED ISSUE #7295: [Bug] http隧道过慢
### 数据库类型和版本

5.5.64-MariaDB

### 问题描述

同样一张表，内有数据26条，同样通过http隧道连接后打开表，navicat耗时约1秒，dbx这边需要30多秒（试了两次），dbx_tunnel.php & ntunnel_mysql.php放同一个目录代理的，php是7.4，系统是centos7.9

Image

### 支持信息

dbx桌面端0.5.96
win11_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

ordinary-s:
## Problem

HTTP tunnel connections could accumulate substantial latency during
sequential database protocol exchanges. In #7295, opening a 26-row
MariaDB table took about 1 second through Navicat but more than 30
seconds through DBX.

## Root cause

The PHP worker drained the inbound file queue once and then waited up
to 200 ms in `stream_select()`. That select call only watched the
database socket, so a newly queued HTTP write could not wake it.
Sequential MySQL/MariaDB exchanges therefore accumulated roughly one
200 ms delay per protocol turn.

The Rust-side 1000 ms read request is a long-poll lifetime rather than
a forced response delay: the PHP endpoint returns as soon as outbound
data is available. The existing queue encoding and locking were also
secondary in the reproduction.

## Fix

Use bounded adaptive polling in the PHP worker:

- 10 ms while protocol traffic is active
- 50 ms after brief inactivity
- 200 ms when steadily idle
- reset to the active interval on inbound or outbound activity

This keeps sequential exchanges responsive without creating a
permanent high-frequency idle loop. The HTTP request cadence, queue
format, session lifecycle, and generic TCP forwarding behavior remain
unchanged.

## Tests

- expand the Rust bridge test from one ping to 100 sequential round trips
- verify ordered variable payload forwarding
- verify the close-session request
- test active, warm, idle, and activity-reset polling transitions
- run 100 sequential exchanges through the real PHP worker and file queues
- verify clean worker shutdown

Validation commands:

```text
php -l deploy/dbx_tunnel.php
php -l deploy/tests/http_tunnel_worker_test.php
php deploy/tests/http_tunnel_worker_test.php
cargo fmt --manifest-path crates/dbx-core/Cargo.toml -- --check
cargo test -p dbx-core --no-default-features --lib http_tunnel
cargo check -p dbx-core --no-default-features --lib
git diff --check
```

All focused checks pass. The Rust test result was 6 passed, 0 failed.

## Before / after

Real PHP worker and queue implementation under PHP 7.4/Linux, using a
local deterministic TCP echo target:

| 100 sequential round trips | Total | Per turn |
| --- | ---: | ---: |
| Before | 20.100 s | 201.0 ms |
| After | 1.073 s | 10.7 ms |

This is approximately 18.7x faster. Steady idle polling returns to the
original 200 ms cadence (about five polls per second).

The reporter's exact MariaDB 5.5.64 / CentOS 7.9 / PHP-FPM environment
was not available; validation used PHP 7.4 on Linux with the real
worker, queues, locks, and TCP forwarding path.

Fixes #7295


t8y2:
Thanks for the contribution! Merged in 4b454b72, will be released in the next version.

PR #7367, opened 2026-08-27
fix(dialog): clamp grid track and wrap footer so dialogs stop clipping

### LINKED ISSUE #7325: [Bug] Update dialog clips release notes and buttons on wide locales
# [Bug] Update dialog clips release notes and buttons on wide locales

## Screenshot
Image

## Description

The update-available dialog (`UpdateDialog.vue`) has a layout overflow issue. `DialogContent` is a CSS grid container, and its single implicit track sizes to the **max-content** of its children. Long release notes (e.g., CJK text) and a 4-button footer in wide locales (Spanish, German, etc.) exceed the dialog's 520px width. The dialog's `overflow-hidden` then clips the right side, making content unreadable.

## Symptoms

- **Release notes**: Each bullet renders as a single long line that overflows the dialog boundary. The text is clipped at the right edge, making it unreadable.
- **Footer buttons**: In locales with longer button labels (e.g., Spanish "Descargar e instalar"), the last button is cut off at the dialog edge.

## Root cause

`DialogContent.vue` sets `grid` display on the dialog box. The implicit grid track sizes to `max-content` of its children, which can exceed the dialog's `max-width`. The dialog's `overflow-hidden` then clips the overflowing content without any visual indication (no scrollbar, no ellipsis).

## Suggested fix

1. Add `grid-cols-[minmax(0,1fr)]` to the `UpdateDialog.vue` DialogContent class prop to pin the grid track to the dialog width.
2. Increase the dialog width so all 4 buttons fit in a single row.

## Environment

- DBX 0.5.94 / 0.5.96
- Windows 10/11
- Spanish locale

## Additional note

For non-Chinese locales, `latest-en.json` on the R2 CDN returns 403, so the updater falls back to curated Chinese release notes. This is a separate upstream data issue — the dialog layout bug is independent of the notes language.

Abeautifulsnow:
## Summary

The update-available dialog (`UpdateDialog.vue`) — and any dialog reusing the shared primitives below — could clip its right edge on wide locales.

`DialogContent` renders a single-column CSS grid with **no explicit `grid-template-columns`**. A grid's implicit track sizes to the **max-content** of its children, so a `DialogFooter` holding several `whitespace-nowrap` buttons on one row (e.g. Spanish `"Descargar e instalar"`) forces the track wider than the dialog's `max-w-*`. The dialog's `overflow-hidden` then clips the overflowing edge — long CJK release notes overflow the same way (they should have scrolled but couldn't, because the column they live in was widened past the dialog).

This is a shared-component fix (not an `UpdateDialog`-only override), so every dialog that reuses `DialogContent` is protected from the same hidden overflow.

## Root cause

- **`DialogContent.vue:45`** — implicit grid track sizes to `max-content` of children → can exceed `max-w-*` → `overflow-hidden` clips.
- **`DialogFooter.vue:19`** — footer is a no-wrap `flex-row` at the `sm` breakpoint, so its `min-content` is the full row of buttons, which is what inflates the grid track.

## Changes

- **`DialogContent.vue`**: pin the implicit track to the dialog width with `grid-cols-[minmax(0,1fr)]`, letting grid items shrink (`min-width: 0`).
- **`DialogFooter.vue`**: add `sm:flex-wrap` so a long button row wraps instead of inflating the footer's min-content beyond the dialog.
- **`dialogLayoutGuard.spec.ts`** (new): source-contract test asserting both class strings stay on the shared primitives, guarding against regression of #7325.

## Test plan

- [x] `npx vitest run apps/desktop/src/components/ui/dialog apps/desktop/src/components/layout/__tests__/AppTabBar.spec.ts` — passed
- [x] Full `apps/desktop/src/components` — 181 files / 1126 tests passed
- [x] `oxlint --vue-plugin` (modified files) — no warnings
- [x] `oxfmt --check` (modified files) — passed
- [x] Manual QA: screenshot the four-button update dialog under Spanish/German locale as a post-merge acceptance check (I could not run a GUI session in this environment).

afac3d8a-867a-415a-81d4-c2300ceeaeb5

## Type

- [x] Bug fix
- [x] Tests

## Related issue

Close #7325


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 40549c9573, will be released in the next version.

Issue #5010, opened 2026-07-31
[Bug] 星环数据库JDBC连接问题


zhoujun681:
### 数据库类型和版本

星环数据库

### 问题描述

1.复现步骤：
新建连接使用星环JDBC驱动，测试连接成功。展开数据库无法展开。
2.期望结果：
能正常展开数据库查看内容。
3.实际结果：
展开数据库无法展开。

<img width="2174" height="1772" alt="Image" src="https://github.com/user-attachments/assets/4fb21bd1-0f9a-4a47-9014-77e2b876dcb6" />
日志：
DBX debug log
Exported: 2026-07-31T10:52:52.755+08:00
User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36 Edg/150.0.0.0
Platform: Win32
Timezone: Asia/Shanghai

[2026-07-31T10:52:31.193+08:00] [DEBUG] [ui:click] {"tag":"button","label":"关闭"}
[2026-07-31T10:52:33.308+08:00] [DEBUG] [DBX][metadata-load:coordinator] {"event":"start","key":"[[\"connectionId\",\"6522289e-ec6e-4ca1-b905-952cb2fabec4\"],[\"driverProfile\",\"jdbc\"],[\"kind\",\"connection-databases\"]]","kind":"connection-databases","active":1,"force":false}
[2026-07-31T10:52:33.308+08:00] [DEBUG] [DBX][api:start] checkConnectionHealth
[2026-07-31T10:52:33.317+08:00] [DEBUG] [DBX][api:success] {"operation":"checkConnectionHealth","elapsedMs":9}
[2026-07-31T10:52:33.317+08:00] [DEBUG] [DBX][api:start] deleteSchemaCachePrefix
[2026-07-31T10:52:33.317+08:00] [DEBUG] [DBX][api:start] loadSchemaCache
[2026-07-31T10:52:33.320+08:00] [DEBUG] [DBX][api:success] {"operation":"deleteSchemaCachePrefix","elapsedMs":3}
[2026-07-31T10:52:33.321+08:00] [DEBUG] [DBX][api:success] {"operation":"loadSchemaCache","elapsedMs":3}
[...]

PR #7370, opened 2026-08-27
fix(jdbc): connect Hive2 JDBC through registered driver

Mo-Moore:
## Summary
- Prefer `registeredDriver.connect()` over `DriverManager.getConnection()` so Hive/Inceptor `UnsupportedOperationException` from other drivers does not abort connect.
- Skip injecting `loginTimeout` / `connectTimeout` for `jdbc:hive2:` and Hive-like driver classes.
- Fall back to `SHOW DATABASES` when catalog metadata throws or returns empty.

This is intentionally scoped to generic Hive2 JDBC connect/metadata behavior, not Inceptor-only UI features.

## Related issues
- #912
- #1019
- #5010

## Test plan
- [ ] `./gradlew :plugins:jdbc:test --tests app.dbx.jdbc.DbxJdbcPluginTest.connectUsesRegisteredDriverWhenOtherDriversThrowUnsupportedOperationException`
- [ ] `./gradlew :plugins:jdbc:test --tests app.dbx.jdbc.DbxJdbcPluginTest.listDatabasesFallsBackToShowDatabasesWhenGetCatalogsUnsupported`
- [ ] `./gradlew :plugins:jdbc:test --tests app.dbx.jdbc.DbxJdbcPluginTest.listDatabasesFallsBackToSchemasForHiveJdbc`
- [ ] Manual: Inceptor JDBC test connection + expand databases

Made with [Cursor](https://cursor.com)

### DIFF
--- a/plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
+++ b/plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
@@ -146,6 +146,7 @@ public final class DbxJdbcPlugin {
         new JdbcDriverQuirkRule("jdbc:taos-rs:", TAOS_QUIRKS)
     );
     private static String registeredDriverKey = "";
+    private static Driver registeredDriver;
     private static String sharedConnectionKey = "";
     private static Connection sharedConnection;
     private static boolean manualTransactionActive;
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Patch pushed in a2c33f6e — three fixes on top of the PR:

1. `addDatabasesFromShowDatabases` now null-guards `conn.createStatement()`. The old try-with-resources could dereference a null statement and NPE outside the catch — exactly the path the pre-existing `listDatabasesFallsBackToSchemasForHiveJdbc` test exercises; it now falls through to the schemas fallback as before.
2. The `getCatalogs()` catch no longer silently swallows plain `SQLException` for every `useCatalogFallbackSql` driver (mysql family included). The original exception is rethrown unless the SHOW DATABASES fallback actually recovered databases; `AbstractMethodError` / `UnsupportedOperationException` handling is unchanged.
3. `registerDrivers` now calls `setAccessible(true)` on the declared constructor. Non-public driver classes failed with `IllegalAccessException` — the PR's own `connectUsesRegisteredDriverWhenOtherDriversThrowUnsupportedOperationException` test was failing deterministically before this (PR CI never ran the test job, so it was invisible).

Verified with `./gradlew test` in plugins/jdbc: 129/129 passing, including the three tests above.

t8y2:
Thanks for the contribution! Merged in 9e4fe8ce56, will be released in the next version.

PR #7371, opened 2026-08-27
feat(jdbc,desktop): browse Hive/Inceptor routines from system views

Mo-Moore:
## Summary
- List/read Hive/Inceptor routines from `system.procedures_v` / `system.functions_v`.
- Show PROCEDURE/FUNCTION sidebar groups for `hive` connections.
- Retry stream queries with an adhoc hint on Inceptor when required.
- Keep partial system-view routine results when one of the views fails (avoids duplicate procedures via JDBC metadata fallback).

**Stacked on:** #7370 (`fix/hive2-jdbc-connect`). This PR currently includes PR-A commits because the fork cannot target that branch on `t8y2/dbx`. After #7370 merges, GitHub will drop the shared commits (or rebase onto `main`). Please review the Inceptor/routines commits first.

## Related issues
- #5598
- #5010

## Test plan
- [ ] `./gradlew :plugins:jdbc:test --tests app.dbx.jdbc.DbxJdbcPluginTest.connectUsesRegisteredDriverWhenOtherDriversThrowUnsupportedOperationException`
- [ ] `./gradlew :plugins:jdbc:test --tests app.dbx.jdbc.DbxJdbcPluginTest.listDatabasesFallsBackToShowDatabasesWhenGetCatalogsUnsupported`
- [ ] `./gradlew :plugins:jdbc:test --tests app.dbx.jdbc.DbxJdbcPluginTest.listDatabasesFallsBackToSchemasForHiveJdbc`
- [ ] Manual (Inceptor): test connection succeeds (no UOE during connect)
- [ ] Manual (Inceptor): expand database list (SHOW DATABASES / schema fallback)
- [ ] Manual (Inceptor): sidebar shows PROCEDURE and FUNCTION groups
- [ ] Manual (Inceptor): list procedures/functions under a database
- [ ] Manual (Inceptor): open routine source from sidebar (`full_text` from system views)
- [ ] Manual (Inceptor): run a query that previously required adhoc hint (stream query / error 10750)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/desktop/src/lib/database/databaseObjectCapabilities.ts
+++ b/apps/desktop/src/lib/database/databaseObjectCapabilities.ts
@@ -69,7 +69,8 @@ const DATABASE_TYPE_OBJECTS = new Map<DatabaseType, SidebarObjectKind[]>([
   // MV support that the backend cannot route.
   ["doris", TABLE_VIEW_OBJECTS],
   ["starrocks", TABLE_VIEW_MV_OBJECTS],
-  ["hive", TABLE_VIEW_OBJECTS],
+  // Inceptor/Hive routines can be listed via JDBC plugin fallbacks (system.procedures_v/functions_v).
+  ["hive", ROUTINE_OBJECTS],
   ["kyuubi", TABLE_VIEW_OBJECTS],
   ["impala", TABLE_VIEW_OBJECTS],
   ["spark", TABLE_VIEW_OBJECTS],
--- a/plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
+++ b/plugins/jdbc/src/main/java/app/dbx/jdbc/DbxJdbcPlugin.java
@@ -2322,7 +2322,7 @@ public final class DbxJdbcPlugin {
                 }
             }
         } catch (SQLException | AbstractMethodError | UnsupportedOperationException ignored) {
-            return false;
+            return added > 0;
         }
 
         return added > 0;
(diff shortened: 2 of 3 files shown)


t8y2 (commented):
Patch pushed in bbdd1cc1 — two fixes on top of the PR:

1. The adhoc-hint retry is now scoped: it only applies to plain SELECT statements (leading comments handled; `WITH ... SELECT` excluded, restoring pre-PR no-retry behavior for them), and the 10750 match uses a word-boundary pattern so incidental substrings like `107500` or port numbers no longer trigger a silent rewrite-and-retry of failed statements.
2. The routines pre-filter is now case-insensitive (`lower(col) LIKE lower(?)`) for both `system.procedures_v` and `system.functions_v`, consistent with the sidebar's in-memory filter.

Also included two test repairs that were needed to get the plugin suite green (both verified failing on the pristine PR head): the pre-existing `listDatabasesFallsBackToSchemasForHiveJdbc` double now serves SHOW DATABASES with an empty result set (schema-fallback coverage and assertions preserved), and `Hive2ConnectGoodDriver` was made public so reflective instantiation in `registerDrivers` works.

Added `hiveAdhocRetryOnlyRewritesSelectStatements` covering the new gating (INSERT failing with a real Hive 10750 message is executed exactly once; SELECT is retried with the hint). `./gradlew test` in plugins/jdbc: 130/130 passing.

Note: this branch still carries the #7370 commits — I'll rebase it onto main once #7370 merges so the final diff is just the routines work.

t8y2 (commented):
Rebased onto current main (dfd4d3b7) now that #7370 has merged, so the PR diff is just the routines work.

Old head: bbdd1cc1 → New head: 2daeb0c789. The two stacked #7370 commits were dropped; `git range-diff` confirms the three remaining commits (routines feature, partial-results fix, adhoc-retry scoping patch) replayed with identical content. Net diff vs main: 3 files, +370/−8.

Verified: `./gradlew test` in plugins/jdbc on the rebased head — 130/130 passing (includes the null-statement and reflective-instantiation fixes that landed with #7370 plus this PR's adhoc gating and LIKE case-insensitivity tests).

t8y2:
Thanks for the contribution! Merged in 7933b7656f, will be released in the next version.

PR #7379, opened 2026-08-27
feat(safety): temporary 1/5 minute write-unlock window for read-only connections

fanchenggang:
## 变更说明

只读连接执行 INSERT/UPDATE/DELETE/DDL 等写操作时，弹出确认，可选 1 分钟或 5 分钟临时开写；到期或点「立刻锁回」后自动恢复只读。持久只读设置不会被关掉；按连接生效。MCP/CLI 无确认框，仍走原只读拦截。

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

- [x] 本 PR 涉及前端改动（新增只读临时开写确认弹窗与倒计时状态提示），截图待补充

## 验证

手动验证步骤：
1. 对某个连接开启只读模式；
2. 在 SQL 编辑器执行 INSERT，出现确认弹窗；
3. 选择「1 分钟」确认，写操作可执行；
4. 等待到期（或点「立刻锁回」），再次执行 INSERT，写操作被拦截；
5. 确认持久只读设置仍为开启，且仅对当前连接生效；
6. MCP/CLI 场景无确认框，仍按原只读拦截。

- [ ] `make check` 通过
- [ ] 相关测试通过

## 关联 Issue

无


t8y2 (commented):
Patch pushed in edc7c892 — three small fixes on top of the PR:

1. `readOnlyUnlockStore.confirm()` is no longer re-entrant: an in-flight guard prevents a double-click from issuing a second `unlockConnectionWrites` call that would restart the full TTL.
2. `remainingMs()` / `isUnlocked()` now compare against `Date.now()` at check time instead of the ticker-fed `nowMs`, which throttled background tabs let go stale — an expired window can no longer look active to the frontend (the backend gate was already authoritative; this fixes the missing re-prompt).
3. The five hardcoded `source: "Data editor"` literals in `useDataGridEditor.ts` now use the new `readOnlyUnlock.sourceDataEditor` key, added to all 8 locales.

Verified: `readOnlyWriteAccess.spec.ts` 7/7, `useMqMutationGuard.spec.ts` 4/4, `useSqlExecution` / `useDataGridEditor` / `useSqlExecutionMultiDbFocus` specs 90/90, `vue-tsc` clean, oxfmt clean. Also wrote a temporary closed-loop spec (not committed) proving both the double-call and the stale-window behavior fail against the pre-patch code.

fanchenggang:
感谢 review，也感谢直接补了这几处。确认过 double-click 和后台 tab 过期的问题，这样更稳，没问题。

t8y2 (commented):
Patch pushed in fb050dd8 — fixes the frontend failure introduced by the feature commit.

Root cause: `startExecution` began with an unconditional `await ensureReadOnlyWriteAccess(...)`. Even for a writable connection (where the promise resolves immediately), the `await` deferred `running.value = true` by one microtask, so the dialog's footer render flipped after the test's `nextTick()` had already queried the DOM — the cancel button wasn't there yet. This is why `SqlFileExecutionDialog.retry.spec.ts` ("does not restore a completed run's file summary after an early retry cancellation") failed on this PR's heads while passing on main (the PR's own CI runs had been stuck awaiting approval, so it was never caught).

Fix: the unlock guard is now only awaited when `connectionIsEffectivelyReadOnly(selectedConnection.value)` — the only case where it can prompt or block. Writable and already-unlocked connections return to the pre-PR synchronous click-to-running transition. Read-only enforcement is unchanged: a locked read-only connection still gets the prompt and can be blocked.

Verified: retry spec 3/3 (was 1 failed / 2 passed), `readOnlyWriteAccess.spec.ts` 7/7, `useMqMutationGuard` + `useSqlExecution` + `useDataGridEditor` 89/89, `vue-tsc` clean. The spec itself was not modified.

t8y2:
Thanks for the contribution! Merged in 955558c81f, will be released in the next version.

PR #7425, opened 2026-08-27
fix(backup): match db.schema.table exclude patterns when database equals schema

### LINKED ISSUE #7314: PostgreSQL scheduled backup exports an excluded pgvector table and times out during stream cleanup
### Summary

DBX Desktop v0.5.96 still executes a data export query for a PostgreSQL table that is explicitly excluded by a saved scheduled-backup rule.

The exported table contains a pgvector column. The row stream then makes no visible progress for 30 seconds. DBX sends a PostgreSQL CancelRequest, but stream cleanup exceeds the fixed 3-second budget, and transaction rollback may also time out.

This is reproducible when running the backup directly from the saved schedule. The schedule uses **Exclude matches**, not Include or All.

### Environment

- DBX Desktop: v0.5.96
- Client OS: Windows 10/11 x64
- PostgreSQL: 17, provided by a ParadeDB image
- pgvector: 0.8.1
- Connection paths reproduced: PgBouncer with `pool_mode=session`, and an SSH tunnel to the raw PostgreSQL listener that bypasses PgBouncer
- Target database: PostgreSQL database containing a schema with pgvector-backed tables

### Related issue

This appears related to #7087 and was reproduced on a release containing commit `9e135f76095a6bd970a827df06e0c28e7040b134` (`fix(export): bound stalled PostgreSQL backup streams`).

The bounded timeout prevents the original indefinite hang, but the backup still fails.

### Scheduled-backup configuration

The table filter mode is **Exclude matches**.

The saved patterns include the fully qualified target:

```text
agentsview.agentsview.vector_chunks_g1
```

Other excluded pgvector-related tables show the same naming shape. The backup is started using **Run now** from the saved schedule.

One unrelated sibling pattern initially contained an accidental whitespace character inside an identifier. That typo was corrected and the schedule was saved again before the reproduced run described here. The failing target pattern shown above contained no internal whitespace.

DBX trims leading and trailing whitespace around each pattern, but embedded whitespace changes the identifier and would prevent only that malformed pattern from matching. It does not explain the reproduced query against `agentsview.agentsview.vector_chunks_g1` after the corrected schedule was saved.

### Steps to reproduce

1. Create a PostgreSQL scheduled backup.
2. Select a database containing a table with a pgvector column.
3. Set the table filter mode to **Exclude matches**.
4. Add a fully qualified pattern matching that table, for example:

   ```text
   agentsview.agentsview.vector_chunks_g1
   ```

5. Save the schedule.
6. Run the backup directly from the saved schedule.
7. Observe the current object and PostgreSQL server log.

### Expected behavior

The excluded table is removed from the data-export set before DBX creates or executes its row query.

No data query should be sent for the excluded relation, and the backup should continue with the remaining selected tables.

### Actual behavior

PostgreSQL records DBX executing the following data query for the excluded table:

```sql
SELECT "doc_key", "chunk_index", "embedding"
FROM "agentsview"."vector_chunks_g1"
```

DBX then reports:

```text
Query timed out after 30 seconds; PostgreSQL stream cleanup timed out after 3 seconds. Transaction cleanup failed: ROLLBACK failed: PostgreSQL manual_txn.rollback timed out after 3 seconds
```

A separate run produced the same query and stream-cleanup timeout but ended with an automatic rollback instead of a rollback timeout.

### Raw PostgreSQL control

The same backup was also run through an SSH tunnel to the raw PostgreSQL listener, bypassing PgBouncer entirely. DBX reproduced the same client-side query timeout, stream-cleanup timeout, and rollback-cleanup failure.

This control makes PgBouncer an unlikely primary cause of the excluded-table query or the RowStream stall. The approximately 3.30-second cancellation timing below is evidence from the PgBouncer-backed run only; I have not yet correlated PostgreSQL server timestamps for the raw control run.

### Correlated PostgreSQL and PgBouncer evidence

The timestamps below are relative and sanitized.

1. PgBouncer accepts the DBX connection for the target database using session pooling.
2. After the DBX timeout, PgBouncer accepts the separate cancellation connection and logs:

   ```text
   successfully sent cancel request
   ```

3. Approximately 3.30 seconds later, PostgreSQL processes the cancellation:

   ```text
   ERROR: 57014: canceling statement due to user request
   STATEMENT: SELECT "doc_key", "chunk_index", "embedding"
              FROM "agentsview"."vector_chunks_g1"
   ```

4. The next type-resolution query receives:

   ```text
   ERROR: 25P02: current transaction is aborted,
          commands ignored until end of transaction block
   ```

   The query is the standard `pg_catalog.pg_type` lookup for a type OID.

5. No successful `ROLLBACK` is present in the PostgreSQL log for the failing run. PgBouncer subsequently records the DBX client closing with an unexpected EOF.

There was no PgBouncer pool exhaustion or client waiting in the same window. PgBouncer appears to forward CancelRequest successfully, so I do not currently have evidence that PgBouncer is the primary cause.

### Possible interaction with pgvector type resolution

This part is an inference from the log order, not a confirmed root cause.

The exported query includes the `embedding` pgvector column. DBX prepares a `pg_catalog.pg_type` lookup while the row stream is stalled, but PostgreSQL executes that lookup only after the data query is cancelled. This may indicate that row-stream consumption and extension-type resolution are waiting on the same protocol sequence.

Even if this type-resolution behavior is a separate problem, the query should not have been generated because the table was excluded.

### Questions

1. Is `tableFilterMode=exclude` and the saved `tablePatterns` configuration propagated to `excludedTables` when **Run now** is invoked from a saved schedule?
2. Can excluded PostgreSQL tables be filtered before row-query generation rather than after schema discovery?
3. After CancelRequest, should cleanup wait for PostgreSQL to return to a usable protocol state instead of using a fixed 3-second stream and rollback budget?
4. Would a minimal pgvector fixture or an additional sanitized native log help isolate the RowStream/type-resolution behavior?

### Workaround

Removing the entire affected database from the scheduled backup avoids DBX discovery, snapshot creation, and data export for that database. Formal backups remain covered separately by pgBackRest and managed `pg_dump`.

I can provide a smaller sanitized debug-log bundle if needed. The original exported log contains private connection metadata and is not attached here.

q396921921:
## Summary

`databaseBackupTableMatchesPattern` (used by scheduled database backups' table include/exclude filters) skips generating the fully-qualified `database.schema.table` candidate whenever `database === schema`. This is a common case in PostgreSQL, where a schema frequently shares its name with the database (as in the reporter's example, `agentsview.agentsview.vector_chunks_g1`).

As a result, an **Exclude matches** rule written in the fully-qualified `db.schema.table` form silently never matches, and the table meant to be excluded is still exported in full — including all of its data.

## Root cause

```ts
// before
if (database && schema && database !== schema) candidates.push(`${database}.${schema}.${table}`);
```

The extra `database !== schema` check suppresses the three-part candidate exactly when database and schema names coincide — the opposite of what's needed, since that's precisely when a user is most likely to write the fully-qualified pattern to disambiguate.

## Fix

Drop the redundant `database !== schema` condition; only require both to be non-empty:

```ts
// after
if (database && schema) candidates.push(`${database}.${schema}.${table}`);
```

## Evidence (non-visual — see PR comment for real repro details)

This bug can't be judged by a screenshot: a saved schedule's UI looks identical whether the exclude rule works or not. The real evidence is the exported SQL content itself.

Reproduced end-to-end against a real PostgreSQL 16 instance with database `issue7314` and schema `issue7314` (same name), one excluded table `vector_data` (50 rows) and one kept table `keep_rows`:

- **Before fix**: exported SQL contained the full `CREATE TABLE`/`INSERT` for `vector_data` despite the exclude rule `issue7314.issue7314.vector_data` being configured.
- **After fix**: exported SQL contains only `keep_rows`; `vector_data` does not appear at all.

Fixes #7314

## Test plan

- [x] New unit tests in `scheduledDatabaseBackup.spec.ts` covering the same-name and different-name cases, verified to fail on the pre-fix code and pass after
- [x] Full frontend suite: 883 files / 8465 tests passing
- [x] `vue-tsc --noEmit` clean
- [x] `oxlint` clean (no new warnings on touched files)
- [x] Rebased onto latest `upstream/main`, no conflicts

### DIFF
--- a/apps/desktop/src/lib/__tests__/backup/scheduledDatabaseBackup.spec.ts
+++ b/apps/desktop/src/lib/__tests__/backup/scheduledDatabaseBackup.spec.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { databaseBackupRunsToPrune, normalizeDatabaseBackupRun, type DatabaseBackupRun } from "../../backup/scheduledDatabaseBackup";
+import { databaseBackupRunsToPrune, databaseBackupTableMatchesPattern, normalizeDatabaseBackupRun, resolveScheduledDatabaseBackupTableScope, type DatabaseBackupRun } from "../../backup/scheduledDatabaseBackup";
 
 const startedAt = "2026-08-12T00:00:00.000Z";
 
@@ -44,5 +44,25 @@ describe("database backup run persistence", () => {
 
     const normalizedRuns = [...scheduledRuns, oneShotRun].map((value) => normalizeDatabaseBackupRun(value)!).filter((value): value is DatabaseBackupRun => !!value);
     expect(databaseBackupRunsToPrune(normalizedRuns, "schedule-1", 1).map((value) => value.id)).toEqual(["scheduled-old"]);
+  });
+});
+
+describe("database backup table pattern matching", () => {
+  it("matches a fully-qualified db.schema.table pattern when the database and schema share the same name", () => {
+    // Regression for #7314: PostgreSQL commonly has a schema with the same
+    // name as its database, and the fully-qualified candidate was previously
+    // suppressed whenever database === schema, silently breaking exclude
+    // rules written as "db.schema.table".
+    expect(databaseBackupTableMatchesPattern("vector_data", ["issue7314.issue7314.vector_data"], "issue7314", "issue7314")).toBe(true);
+  });
+
+  it("still matches a fully-qualified db.schema.table pattern when the database and schema differ", () => {
+    expect(databaseBackupTableMatchesPattern("vector_data", ["mydb.myschema.vector_data"], "mydb", "myschema")).toBe(true);
+  });
+
+  it("excludes the matched table from the backup scope when database and schema share the same name", () => {
+    const scope = resolveScheduledDatabaseBackupTableScope("exclude", ["issue7314.issue7314.vector_data"], ["vector_data", "keep_rows"], "issue7314", "issue7314");
+    expect(scope.includedTables).toEqual(["keep_rows"]);
+    expect(scope.excludedTables).toEqual(["vector_data"]);
   });
 });
--- a/apps/desktop/src/lib/backup/scheduledDatabaseBackup.ts
+++ b/apps/desktop/src/lib/backup/scheduledDatabaseBackup.ts
@@ -79,7 +79,7 @@ function tablePatternRegex(pattern: string, caseSensitive: boolean): RegExp {
 export function databaseBackupTableMatchesPattern(table: string, patterns: readonly string[], database = "", schema = "", caseSensitive = true): boolean {
   const candidates = [table];
   if (schema) candidates.push(`${schema}.${table}`);
-  if (database && schema && database !== schema) candidates.push(`${database}.${schema}.${table}`);
+  if (database && schema) candidates.push(`${database}.${schema}.${table}`);
   return patterns.some((pattern) => {
     const matcher = tablePatternRegex(pattern, caseSensitive);
     return candidates.some((candidate) => matcher.test(candidate));


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 1016ccc, will be released in the next version.

PR #7435, opened 2026-08-27
fix(grid): allow rerunning exact total row count from status bar

difagume:
## What changed
- Shows a rerun action in the grid status bar when an exact total row count is already displayed and recalculation is possible.
- The action reuses the existing count logic and displays a spinner while the count is in flight.

Before: once an exact total was shown, there was no way to recalculate it without re-running the query.
After: an inline rerun button appears next to the exact total.

## How to test
1. Run a query that returns a total row count.
2. Verify the exact total is shown in the status bar.
3. Click the rerun button next to the total and verify it recalculates.

Checks run:
- `pnpm test` (relevant package tests)

## Screenshots
image

## Notes
- Scope is limited to the grid total row count UI.
- Branch: `fix/grid-rerun-total-row-count` against `t8y2/dbx` `main`.

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -38,6 +38,7 @@ import {
   PanelBottom,
   PanelRight,
   RefreshCw,
+  RefreshCcw,
   TableProperties,
   UserRound,
   Database,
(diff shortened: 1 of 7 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in c0abc26, will be released in the next version.

PR #7509, opened 2026-08-28
fix(editor): align wrapped line numbers

eryajf:
## Summary

- Keep line numbers vertically centered for single visual rows.
- Detect wrapped logical lines from their rendered gutter height.
- Anchor wrapped line numbers to the first visual row instead of centering them across the full wrapped block.
- Add regression coverage for the alignment decision and theme rules.

## Validation

- `pnpm exec vitest run apps/desktop/src/lib/editor/__tests__/queryEditorLineNumbers.spec.ts`
- `pnpm exec vitest run apps/desktop/src/lib/__tests__/editorThemes.spec.ts`
- `pnpm typecheck`

Related to #7492


t8y2:
Thanks for the contribution! Merged in 038ff4e, will be released in the next version.

PR #7518, opened 2026-08-28
feat(mcp): add global query timeout override with integer validation

Abeautifulsnow:
## Summary

This PR adds a **global "query timeout (seconds)" override** to the MCP access policy. Users can now set a policy-level timeout that applies to every MCP query, independent of the connection setting:

- **Empty** → inherit the connection's query timeout.
- **0** → no limit.
- **Positive value** → override the connection setting for every MCP query.

The value is re-read from the persisted policy on **every tool call**, so a change on the settings page takes effect on the next MCP query with **no client-config regeneration**.

## Change Type

- [x] New feature
- [ ] Bug fix
- [ ] Code refactor
- [ ] Documentation update
- [ ] CI / Build

## Frontend Changes

The MCP settings section in **Settings → MCP Server** gains a "Query timeout (seconds)" input with a hint line ("Leave empty to inherit the connection's query timeout. 0 = no limit. A positive value overrides the connection setting for every MCP query."). The input rejects non-integer values (e.g. `1.5`) with the existing "non-negative whole number" toast instead of silently rounding them to a different number.

## Verification

Rust (vendored OpenSSL builds need `export PATH="/e/software/Scoop/apps/perl/5.42.3.1/perl/bin:$PATH"` on this Windows machine — see note below):

- [x] `cargo test -p dbx-mcp --lib` → **55 passed**
- [x] `cargo test -p dbx-mcp --test protocol` → **8 passed, 0 failed** (incl. new `execute_query_injects_and_omits_timeout_secs_from_policy`)
- [x] `cargo test -p dbx-mcp --test local` → **6 passed, 3 ignored** (ignored tests require an external MongoDB, unrelated to this change)
- [x] `cargo fmt -- --check crates/dbx-mcp/tests/protocol.rs` → clean

Frontend:

- [x] `npx vitest run src/components/editor/__tests__/EditorSettingsDialog` → **5 files / 35 tests passed**
- [x] `npx vitest run src/stores/__tests__/settingsStore.spec.ts` → **109 tests passed**
- [x] `npx vue-tsc --noEmit` → clean
- [x] `oxfmt` (lint-staged) → clean

Related #7473


t8y2 (commented):
Thanks for the clean implementation — the precedence chain (policy > connection effective timeout, 0 = no limit with the Spanner floor preserved), the per-call policy reload, and building the injected arguments from the typed request so clients can't self-inject `timeout_secs` are all correct. I traced all three backends (desktop bridge, web, CLI `--timeout`) and the resolved value reaches query execution in each.

I pushed a maintainer patch 2dfd8bc with the two minor items from review:

1. `agent_tools.rs` — EXPLAIN statement execution now resolves its timeout through `agent_query_timeout_secs` instead of the hardcoded `Some(30)`, so the global override truly applies to every MCP query.
2. `EditorSettingsDialog.vue` — integers beyond the u64 range now hit the invalid-value toast instead of failing backend deserialization with a generic save error.

`cargo check -p dbx-core` clean on the patched head. (The `rust-test` CI failure is the pre-existing `table_export` stack overflow that reproduces identically on `main` — unrelated.)

t8y2:
Thanks for the contribution! Merged in e6f8b21, will be released in the next version.

PR #7536, opened 2026-08-28
fix(cli): Set default rustls CryptoProvider to prevent panic

mmorella-dev:
## Description

Adds this line to the main.rs for dbx-cli:

```rs
// Set aws_lc_rs as the process-level default CryptoProvider to prevent a rustls panic
let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
```

This prevents a panic that occurs when querying MySQL servers:

```text
thread 'main' (4361874) panicked at .../rustls-0.23.43/src/crypto/mod.rs:249:14:

Could not automatically determine the process-level CryptoProvider from Rustls crate features.
Call CryptoProvider::install_default() before this point to select a provider manually, or make sure exactly one of the 'aws-lc-rs' and 'ring' features is enabled.
See the documentation of the CryptoProvider type for more information.
```

Similar call occurs in the GUI and MCP builds and the Postgres driver, but not the CLI.

image

## Change Type

- [ ] New feature
- [x] Bug fix
- [ ] Code refactoring
- [ ] Documentation update
- [ ] CI / Build

## Verification

- [x] `make check` passed
- [x] `make cargo-check-fast` passed
- [x] Relevant tests passed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2063,6 +2063,7 @@ dependencies = [
  "async-trait",
  "dbx-core",
  "dbx-mcp",
+ "rustls 0.23.43",
  "serde",
  "serde_json",
  "tempfile",
--- a/crates/dbx-cli/Cargo.toml
+++ b/crates/dbx-cli/Cargo.toml
@@ -11,6 +11,7 @@ path = "src/main.rs"
 [dependencies]
 dbx-core = { path = "../dbx-core", default-features = false, features = ["sqlite-bundled"] }
 dbx-mcp = { path = "../dbx-mcp" }
+rustls = { version = "0.23", features = ["aws-lc-rs"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 tokio = { version = "1", features = ["fs", "macros", "rt-multi-thread"] }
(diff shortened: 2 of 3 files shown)


t8y2 (commented):
Thanks for the fix — the root-cause analysis is correct. Because `mysql_async` is built with `default-rustls` (ring) while `dbx-core` enables `rustls/aws-lc-rs`, both providers end up in the CLI binary, and rustls 0.23 can't auto-pick one. Installing `aws_lc_rs` explicitly matches what the GUI, web, MCP, and Postgres code paths already do, and ignoring the double-install error is the right handling.

I pushed a maintainer patch 2dc309f: the new `dbx-cli` → `rustls` dependency edge was missing from `Cargo.lock`, which made both `rust-fmt-clippy` and `rust-test` fail under `--locked`. The patch is the one-line lockfile entry generated by cargo (no other changes).

Optional follow-up (not blocking): moving the `install_default()` call to the top of `run()` would also cover the early-return commands, though none of them use TLS today.

t8y2:
Thanks for the contribution! Merged in 694c344, will be released in the next version.

PR #7537, opened 2026-08-28
fix(postgres): query foreign keys via pg_catalog for read-only users

difagume:
### Summary

Migrates PostgreSQL foreign key metadata queries from `information_schema` views (`table_constraints`, `key_column_usage`, `referential_constraints`) to system catalogs (`pg_catalog.pg_constraint`, `pg_class`, `pg_namespace`, and `pg_attribute`).

### Context & Motivation

Standard `information_schema` views restrict returned rows based on the current user's table privileges. Consequently, read-only roles (e.g. users granted only `SELECT`) observe zero foreign keys even when constraints exist. Querying `pg_catalog.pg_constraint` ensures foreign key metadata is consistently visible regardless of role privilege limitations, aligning with existing catalog patterns used in `postgres_check_constraints_sql` and `postgres_constraints_sql`.

### Screenshots
(current situation)
image

### Key Changes

- **Queries Migrated to `pg_catalog`**:
  - `postgres_foreign_keys_sql` (single relation lookup).
  - `postgres_foreign_keys_for_relations_sql` (batched relation lookup via array `unnest`).
  - `postgres_foreign_keys_for_relations_compat_sql` (batched relation lookup via `generate_subscripts`).
- **Composite Key Handling**: Joined `LATERAL unnest(con.conkey) WITH ORDINALITY` and `LATERAL unnest(con.confkey) WITH ORDINALITY` on `pk.ord = fk.ord` to guarantee declared column pairing and ordering.
- **Action Label Mapping**: Mapped `confupdtype` and `confdeltype` action codes (`a`, `r`, `c`, `n`, `d`) through `postgres_fk_action_label` to preserve standard referential action strings (`CASCADE`, `SET NULL`, `NO ACTION`, etc.).
- **Tests**: Updated unit test assertions in `crates/dbx-core/src/db/postgres.rs` to validate catalog query construction and clauses (`JOIN LATERAL`, `ORDINALITY`, `pg_catalog.pg_constraint`).

### Verification

- Fast PostgreSQL test suite passed:
  ```bash
  cargo test -p dbx-core --lib -- db::postgres::tests
  ```
  *(219 unit tests passing)*

### DIFF
--- a/crates/dbx-core/src/db/postgres.rs
+++ b/crates/dbx-core/src/db/postgres.rs
@@ -3854,6 +3854,9 @@ fn postgres_foreign_keys_for_relations_sql() -> &'static str {
 // PostgreSQL 9.3 and older only accept one array argument to unnest(). Pair
 // the schema/table arrays through their shared subscript so the fallback stays
 // one bounded query and preserves each relation tuple's position.
+// WITH ORDINALITY is equally unavailable before 9.4, so pair conkey/confkey
+// positions with generate_series over the array length plus plain subscripts —
+// the same pre-9.4 technique the index compat SQL relies on.
 fn postgres_foreign_keys_for_relations_compat_sql() -> &'static str {
     "SELECT rel.rel_schema, rel.rel_table, \
      con.conname AS constraint_name, \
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Thanks for the clean migration — the approach matches how DBeaver reads foreign keys (`pg_catalog.pg_constraint` with `confupdtype`/`confdeltype` letter mapping), the action labels preserve the previous `information_schema` strings, and the ORDINALITY pairing is equivalent to the old `position_in_unique_constraint` join.

I pushed a maintainer patch 5ecdb8f closing the two compat gaps (plus a third one I found while patching):

1. **Pre-9.4 compat tiers**: both `postgres_foreign_keys_for_relations_compat_sql` and the single-table `list_foreign_keys` path (which had no compat tier at all) now pair conkey/confkey positions via `generate_series(1, array_length(con.conkey, 1))` + plain array subscripts instead of `unnest(...) WITH ORDINALITY` — the same pre-9.4 technique the index compat SQL uses. A mirror invariant test asserts no `WITH ORDINALITY` in compat tiers. Note: `conkey` is `smallint[]` (comma-braced text form), so the index compat's `string_to_array(..., ' ')` trick doesn't apply; `array_length` + subscripts are used instead.
2. **OpenGauss routing**: added `list_opengauss_foreign_keys` (reads `conkey::text`/`confkey::text` and resolves attribute numbers in Rust via the existing `opengauss_relation_attributes` helper, no array expansion in SQL), and `list_foreign_keys_core` now routes `is_opengauss_constraint_config` pools to it, mirroring the constraints dispatch.

Checks run: `cargo check --no-default-features -p dbx-core` clean. Local `cargo test` is blocked by a pre-existing machine-level rusqlite `load_extension` linker failure (reproduces on `main`); the new tests are plain string assertions and will run in CI `rust-test` on this head.

One minor pre-existing note kept as-is: unmapped action letters now render blank instead of raw text — theoretical only, since PG only emits a/r/c/n/d for `contype='f'`.

t8y2:
Thanks for the contribution! Merged in 4ac3624, will be released in the next version.

cpfo:
厉害，刚好解决了一个我遇到的ddl查询速度特别慢的问题，也是这个外键导致的

PR #7562, opened 2026-08-29
fix(editor): stabilize batch column selection

t8y2:
Follow-up to merged PR #7543.

- Clear batch-column selection state when Escape or completion closure ends.
- Apply the batch action only while completion is active and guard a missing snippet factory.
- Make INSERT replacement consume optional whitespace/closing parentheses and avoid duplicating an existing `VALUES` clause.

Validation:
- helper runtime checks passed
- focused static/regression checks added
- `git diff --check` passed

The remote environment has no `node_modules`, so Vitest was not available.


t8y2:
CI 阻塞已代修（71b7cd058）：
1. `queryEditorCompletionKeymap.spec.ts` 的 harness 参数表补上了 `isBatchColumnSelectionCompletionActive`（默认 `(status) => status === "active"`），消除 6 个用例的 ReferenceError；
2. `QueryEditorContextMenu.spec.ts` 的两处静态断言同步为新守卫式调用。
两个 spec 修后本地 26/26 通过。

t8y2:
Thanks for the contribution! Merged in 143be57, will be released in the next version.

PR #7563, opened 2026-08-29
fix(objects): preserve MySQL event create editor

t8y2:
Follow-up to merged PR #7556.

- Keep the MySQL CREATE EVENT editor mounted even when no EVENT row exists yet.
- Scope the editor instance by create/open request and row identity so repeated CREATE and ALTER-to-CREATE flows cannot reuse stale draft state.

Validation:
- pure Node/static regression checks passed
- `git diff --check` passed

The remote environment has no `node_modules`, so Vitest, `vue-tsc`, `oxfmt`, and `oxlint` were not available.


t8y2:
Thanks for the contribution! Merged in de2c4b3, will be released in the next version.

PR #7564, opened 2026-08-29
fix(schema-diff): guard deploy script execution

t8y2:
Follow-up to merged PR #7552.

- Remove execution from the focused-only script preview because the execution path uses selected SQL.
- Clear selected SQL before regeneration and block execution when generation fails or returns no SQL, preventing stale SQL execution.
- Normalize the Rust dependency graph shape and include the full recursive dependency closure for focused plans.

Validation:
- focused static/regression checks added
- `git diff --check` passed

The remote environment has no `node_modules`, so Vitest was not available.

### DIFF
--- a/apps/desktop/src/components/diff/SchemaDiffDdlPanel.vue
+++ b/apps/desktop/src/components/diff/SchemaDiffDdlPanel.vue
@@ -465,10 +465,6 @@ function copyDeploySqlAll() {
             <Copy class="w-3 h-3" />
             {{ t("diff.copy") }}
           </Button>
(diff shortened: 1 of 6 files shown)


t8y2:
CI 阻塞已代修（5047f57a）：`SchemaDiffDialog.spec.ts` 中一处断言的引号风格不符合 oxfmt（单引号内无嵌套引号时应使用双引号），已运行 oxfmt 修复，仅此一行变更。lint/typecheck/test 在上一 head 已全绿，本 head 触发重新运行。

t8y2:
Thanks for the contribution! Merged in f664918, will be released in the next version.

PR #7565, opened 2026-08-29
feat(diff): show schema comparison progress and next step

t8y2:
Replaces #7504 (its head branch was deleted, so that PR could no longer be updated). Rebase of @0verme's original two commits onto current main with the SchemaDiffDialog.vue conflicts resolved by the maintainer: the progress reset and generation counters unioned with the selected/focused deploy-SQL split, and the generating-phase progress write wrapped around both regenerate calls. Also moves role=progressbar onto the track element so the indeterminate variant is covered, and clears the progress state instead of writing a dead complete phase.


t8y2:
Thanks for the contribution! Merged in 1d71a5f, will be released in the next version.

PR #7581, opened 2026-08-30
Feature: 添加 SQL Server 2000 兼容支持

zipg:
## 变更说明
- 完善 SQL Server 2000 的连接、TLS fallback、查询、分页、筛选、搜索、补全和导出兼容路径
- legacy Agent 使用 jTDS 1.3.1 作为 SQL Server 2000 fallback
- 修复 SQL Server 2000 空 varchar 被 jTDS 返回为全 NUL 字符时的显示问题，保留真实 NULL 和混合内容
- 修复存储过程-查看源码报错

## 验证
- legacy Agent 单测和 shadow JAR 构建通过
- Windows x64 测试包构建通过
- 已提供 Windows 测试包由用户进行 SQL Server 2000 实际验证

fix：#5730

### DIFF
--- a/.github/workflows/sqlserver2000-windows-user-test.yml
+++ b/.github/workflows/sqlserver2000-windows-user-test.yml
@@ -48,7 +48,8 @@ jobs:
           cache-on-failure: true
 
       - name: Build SQL Server legacy agent
-        run: .\\agents\\gradlew.bat :sqlserver-legacy:shadowJar --no-daemon
+        run: .\\gradlew.bat :sqlserver-legacy:shadowJar --no-daemon
+        working-directory: agents
 
       - name: Stage SQL Server legacy agent
         shell: pwsh
--- a/.github/workflows/sqlserver2000-windows-user-test.yml
+++ b/.github/workflows/sqlserver2000-windows-user-test.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - codex/sqlserver2000-support
+  workflow_dispatch:
 
 permissions:
   contents: read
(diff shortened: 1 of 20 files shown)


t8y2 (commented):
Maintainer patch pushed in 3f980d42 (on top of the reviewed 2754880).

**Root cause confirmed:** `isSqlServer2000Unsupported` required the substring `sql server 8`, but mssql-jdbc 13.2 rejects SQL Server 2000 during prelogin with the English-only `"SQL Server version 8 is not supported by this driver."` — the word "version" breaks the substring match, so the jTDS fallback would never have triggered. The test strings were invented (including a localized message that driver doesn't ship).

**Patch:** the trigger now accepts both `"sql server 8"` and `"sql server version 8"` shapes (plus the existing 不支持 variant), and adds a secondary trigger for the `R_notSQLServer` wording that names the supported floor (`"sql server 2005 or later"`). Tests rewritten to assert the real message strings, keeping the negative cases. Also trimmed the dead `push` trigger from `sqlserver2000-windows-user-test.yml` (kept `workflow_dispatch`).

**Maintainer note:** `net.sourceforge.jtds:jtds:1.3.1` (LGPL-2.1, bundled into the downloadable sqlserver-legacy agent JAR) is accepted per the repo's existing per-driver JDBC bundling pattern — same approach DBeaver takes for its legacy SQL Server driver.

**Checks run:** `./gradlew :sqlserver-legacy:test` (19 tests, 0 failures, including the rewritten trigger cases). Real SQL Server 2000 end-to-end verification still rests on the issue reporter's test build as planned.

t8y2:
Thanks for the contribution! Merged in afe3b3f, will be released in the next version.

PR #7593, opened 2026-08-30
feat(xugu): add scheduler job discovery and management

mapan0424:
## Summary

Xugu exposes scheduled jobs through `DBMS_SCHEDULER` and the `ALL_JOBS` catalog, but DBX previously did not surface these database-scoped objects in the sidebar. This change adds a dedicated Scheduler Jobs experience for Xugu without treating jobs as ordinary schema-owned objects.

- Add a synthetic `Scheduled jobs` root after real schemas and Public synonyms. The reserved internal scope remains only a routing detail and is resolved to a localized label at render time, including for persisted tree state.
- Discover jobs from `ALL_JOBS`, constrained by `CURRENT_DB_ID`, so a connection only lists jobs belonging to its current database.
- Add source reconstruction for `DBMS_SCHEDULER.CREATE_JOB`, plus existing lifecycle actions: enable, disable, run now, and drop.
- Add dedicated scheduler-root and job icons, localized labels, tree grouping, and context-menu support.
- Normalize nullable scheduler timestamps before source generation. In particular, a null end time returned by the Go driver must remain SQL `NULL` rather than becoming a sentinel date.
- Keep scheduler nodes out of default-schema selection and schema-level destructive/configuration operations.

## Design

Scheduler jobs are database-level operational objects. Keeping them in a separate synthetic root makes their ownership and scope explicit, avoids mixing them into user schemas, and still reuses DBX's normal object-source and lifecycle-action conventions. The root is placed after real schemas and Public synonyms so it cannot be confused with a real owner/schema.

The implementation is Xugu-specific: discovery, source generation, object type handling, and actions are gated to the Xugu driver. No other database dialect changes behavior.

## Validation

- `go test -count=1 ./...` in `agents/drivers/xugu`
- Focused Vitest suite: 5 files / 26 tests passed, covering scope ordering/display, scheduler SQL quoting, tree routing, context-menu actions, and default-schema handling.
- `pnpm typecheck`
- Frontend formatting check and `git diff --check`
- `pnpm lint` (no lint errors; repository baseline warnings remain)
- `pnpm build` and a fresh debug Tauri bundle
- Packaged-app UI verification using an isolated build, not the installed DBX application: the system database shows the localized `Scheduled jobs` root, it expands successfully, and an empty catalog renders as `Scheduled jobs 0` without exposing the internal reserved scope.

## Deployment note

This change discovers and manages jobs only when the connected Xugu database provides the scheduler package and catalog access. It does not install or emulate `DBMS_SCHEDULER` in databases where that package is unavailable.


t8y2 (approved):
Thanks for the thorough PR — the synthetic-scope design mirrors the public-synonyms precedent nicely, and the safety wiring checks out (read-only gate via EXEC write classification, production confirmation, drop confirm, synthetic scope excluded from schema-level destructive actions). i18n coverage across all locales is complete and the gated live test is the right call.

Two minor, non-blocking notes for a possible follow-up:
1. `renderXuguSchedulerJobDDL` hardcodes `'default_class'`. If `ALL_JOBS` exposes the job class, replay fidelity would improve by selecting it; otherwise a comment documenting the assumption would help future readers.
2. "Run now" executes `RUN_JOB` immediately with no confirmation (only drop confirms). Production-flagged connections still get the guard dialog, but consider a lightweight confirm for run as well.

t8y2:
Thanks for the contribution! Merged in a849979, will be released in the next version.

PR #7595, opened 2026-08-30
feat(xugu): distinguish public synonym scope in sidebar

mapan0424:
## Summary

Xugu public synonyms are database-global aliases, not ordinary user schemas. The sidebar already exposes them through a dedicated synthetic `Public synonyms` scope, but the root still used the same folder icon as a real schema.

This focused UI change gives only that Xugu synthetic scope the existing blue link icon used by synonym objects. Real schemas, including a real schema named `GUEST`, keep their existing folder icon and behavior.

## Design

The icon is selected only when all three conditions match: the connection is Xugu, the node is a schema node, and its schema key is the reserved public-synonym scope. The reserved key is retained for routing; no naming, ordering, metadata query, permission, or context-menu behavior changes.

This makes the database-global namespace visually distinguishable from ordinary owners while avoiding any behavior change for other database types.

## Validation

- Focused Vitest checks: 2 files / 8 tests passed, including the public-synonym identity guard and the dedicated icon mapping.
- `pnpm typecheck`
- Frontend formatting check and `git diff --check`
- Fresh debug Tauri bundle completed successfully.
- The isolated packaged build was launched separately from the installed DBX application; the Xugu system tree displayed `Public synonyms` with the distinct link icon.

## Scope

- Xugu only.
- Two frontend files only: the tree-item icon mapping and its focused regression test.
- No database driver, metadata, SQL, or object-action behavior is changed.

### DIFF
--- a/apps/desktop/src/components/sidebar/TreeItem.vue
+++ b/apps/desktop/src/components/sidebar/TreeItem.vue
@@ -78,6 +78,7 @@ import { focusSidebarRenameInput } from "@/lib/sidebar/sidebarRenameFocus";
 import { ensureSqlExtension, stripSqlExtension } from "@/lib/savedSql/savedSqlFileName";
 import { savedSqlErrorMessage } from "@/lib/savedSql/savedSqlErrors";
 import { useSavedSqlStore } from "@/stores/savedSqlStore";
+import { isXuguPublicSynonymTreeNode } from "@/lib/sidebar/xuguPublicSynonyms";
 // --- Drag and Drop ---
 import { useDragSort } from "@/composables/useDragSort";
 import { sidebarTreeRuntimeKey } from "@/lib/sidebar/sidebarTreeRuntime";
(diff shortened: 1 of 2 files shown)


t8y2 (approved):
Thanks! Reviewed at bd7c3b2. The guard is correctly scoped (xugu + schema node + the NUL-prefixed reserved scope key), so real schemas — including one actually named GUEST — keep the folder icon, and the link icon matches the existing synonym/group-synonyms styling nicely. Focused tests and typecheck pass on my side too.

One tiny nit, feel free to ignore: in `xuguPublicSynonymActions.spec.ts`, the assertion on `return { icon: Link2, colorClass: "text-sky-500" }` already passes on main via the pre-existing `synonym` case, so only the first assertion guards the new branch. Scoping it to the `case "schema"` block would make it a true regression guard.

t8y2:
Thanks for the contribution! Merged in 0a239bd, will be released in the next version.

PR #7629, opened 2026-08-31
finish issue 4510, support more file format

andystenhe:
## 变更说明

SQL文件功能中增加一个设置按钮, 显示使用正则过滤文件, 默认*.sql. 
允许用户修改, 以显示制定文件.

## 变更类型
新功能

## 涉及前端

本 PR 涉及前端改动，已附截图/录屏（见下方）

## 验证

相关测试通过

## 关联 Issue

Related #4510


andystenhe:
@t8y2 Hi, the workflow is awaiting approval, please review~

t8y2 (commented):
Maintainer patch pushed in 8bcd0eaa (on top of your 47972e3f).

**Root causes confirmed and fixed:**

1. **glob/regex heuristic** (`is_glob_filter`): `.` was not treated as regex context, so a regex like `.*sql` was silently converted to the glob `^(?i)\.*sql$` (zero-or-more literal dots); a `?`-only glob like `file?.sql` was parsed as a regex and matched nothing. The rule now treats a wildcard directly after `.` as a regex quantifier (`.*`/`.?` stay regex, `*.sql` stays glob) and recognizes `?`-containing globs. Covered by two new Rust tests.
2. **invalid filter persisted before validation** (`saveFileFilter`): a broken pattern was saved to localStorage first and then failed every refresh (`loadFailed` toast per folder, Quick Open silently dropping external files) across restarts. `loadFolderEntries` now surfaces the error message, and `saveFileFilter` restores the previous filter and re-runs the refresh when the backend rejects the pattern with "Invalid file filter", keeping the dialog open with a new `sqlFileTree.filterInvalid` toast (key added to all 8 locales).
3. **dead i18n key**: the placeholder now renders `t('sqlFileTree.fileFilterPlaceholder')` (translated in all locales already); the hardcoded const is gone. The regex example stays in `fileFilterHint` as before.

**Checks run:** `cargo test --no-default-features --features sqlite-bundled -p dbx --lib list_sql_files` (8 passed, incl. the 2 new tests); `vitest run` on `SqlFilePanel.spec.ts` (8 passed, 2 new source assertions); `oxfmt --check` / `oxlint` clean; `cargo fmt` applied.

t8y2:
Thanks for the contribution! Merged in f725c50, will be released in the next version.

PR #7631, opened 2026-08-31
fix(grid): copy NULL cells as empty fields

### LINKED ISSUE #7610: [Feature] 查询出来数据为null的字段，粘贴到Excel里变成了字符串null
### 数据库类型和版本

MySQL5.7

### 支持信息

DBX 版本：v0.5.97
运行环境：桌面端
操作系统：Windows 10.0.26200.9168
系统架构：x86_64

### 当前痛点

sql查询出来的数据，有的数据为null(空)，复制以后粘贴到本地Excel里null(空)变成了字符串null

### 期望方案

null(空)值，粘贴出来应该是空格最好

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

_No response_

xiaou61:
## Summary

- copy NULL cells as empty fields for cell, row, TSV, and detail clipboard paths
- keep DBX internal clipboard metadata so NULL remains recoverable when pasting back into the grid
- allow empty raw/DSV output while preserving empty-result validation for SQL and document extractors

Fixes #7610

## Tests

- `F:\dbx-7592\node_modules\.bin\vitest.cmd run apps/desktop/src/composables/__tests__/useDataGridExport.spec.ts apps/desktop/src/lib/__tests__/dataGrid/dataGridCopyExtractor.spec.ts apps/desktop/src/lib/dataGrid/__tests__/dataGridDetail.spec.ts apps/desktop/src/lib/__tests__/dataGrid/dataGridClipboard.spec.ts` (96 passed)
- `oxfmt --check` on changed TypeScript files (passed)
- `rustfmt --check` on changed Rust files (passed)
- Rust cargo test was not completed because dependency fetch stalled on GitHub network access

### DIFF
--- a/apps/desktop/src/composables/__tests__/useDataGridExport.spec.ts
+++ b/apps/desktop/src/composables/__tests__/useDataGridExport.spec.ts
@@ -730,6 +730,66 @@ describe("useDataGridExport prepared row statements", () => {
     expect(copyToClipboard).toHaveBeenCalledWith(text);
   });
 
+  it("copies a NULL cell as an empty clipboard value", async () => {
+    const state = createExportState(editableTable, ["id", "name"], undefined, [1, null], undefined, undefined, [], DEFAULT_DATA_GRID_EXTRACTOR_OPTIONS, false, undefined, false, 1, 1);
+
+    await state.copyCell();
+
+    expect(copyToClipboard).toHaveBeenCalledWith("");
+  });
+
+  it("copies all rows with empty fields for NULL cells", async () => {
+    const text = "id\tname\n1\t\n2\tAda";
+    const state = createExportState(editableTable, ["id", "name"], undefined, undefined, undefined, [
+      [1, null],
+      [2, "Ada"],
+    ]);
+
+    await state.copyAll();
+
+    expect(copyToClipboard).toHaveBeenCalledWith(text);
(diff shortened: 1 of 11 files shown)


t8y2 (approved):
Thanks for the fix — verified against issue #7610 and happy to merge as-is.

The approach is right on all layers: the new `clipboardCellValue` helper keeps `displayCellValue` untouched for rendering, `copyAll` now reuses the canonical `formatTsv` (which already emits empty for NULL and adds minimal quoting — a small bonus fix for values with embedded tabs), and the Rust raw extractor skips `value_text` for NULL cells. Paste-back fidelity is preserved because `rememberDataGridClipboardCopy` keeps the logical matrix with real `null`s, and the Rust DSV writer force-quotes empty strings when `nullText` is `""`, so NULL vs `""` stays distinguishable.

One thing to be aware of (no change needed): flipping `DEFAULT_DATA_GRID_EXTRACTOR_OPTIONS.dsv.nullText` to `""` also affects csv / pipe-separated / one-row clipboard copies, not just TSV. That's consistent with the file-export path and the issue's intent, and users who explicitly configured nullText keep their setting — just worth calling out in the release note.

t8y2:
Thanks for the contribution! Merged in 0e9f92c, will be released in the next version.

PR #7655, opened 2026-08-31
feat(xugu): expose tablespaces and datafiles in schema tree

mapan0424:
## Summary

Add read-only tablespace and datafile metadata to the XuguDB connection tree.

XuguDB exposes storage layout metadata (tablespaces and their datafiles), but this information was not available in DBX. Users therefore had no way to inspect the database storage structure from the schema tree and had to switch to another administrative tool.

## What changed

- Added an XuguDB agent operation that lists tablespaces and their datafiles.
- Added primary queries against XuguDB system metadata with a compatible ALL_* fallback.
- Added permission-aware degradation: metadata access failures do not prevent the connection tree from loading.
- Added the operation through the Rust core/web and Tauri bridge layers.
- Added an Xugu-only `Storage` tree group at connection level.
- Added the hierarchy `Storage -> tablespace -> Files -> datafile`.
- Displayed the datafile path as secondary metadata in the tree.
- Kept the feature read-only; no create, alter, resize, drop, or other storage mutations are exposed.
- Kept non-Xugu behavior unchanged: other database drivers return no Xugu storage group.
- Added localized labels and icons for the new nodes.

The datafile size is intentionally not mapped into the generic byte-size field yet. Xugu reports the value in a database-specific unit, while the generic tree contract expects bytes; avoiding an implicit conversion prevents misleading values.

## Design rationale

The feature is deliberately scoped to XuguDB because tablespace/datafile metadata and its system views are Xugu-specific in the current driver contract. Capability gating keeps the shared tree model reusable without changing behavior for PostgreSQL, MySQL, Oracle, DM, or other drivers.

A read-only tree representation provides useful operational visibility while avoiding unsafe storage operations. The metadata query is isolated in the Xugu agent and degrades gracefully for users without sufficient metadata privileges.

### Automated tests

- Xugu live integration coverage for tablespace/datafile discovery against the configured test environments
- `cargo check -p dbx-core -p dbx-web --lib`
- `cargo check -p dbx --lib`
- Rust protocol unit tests
- Targeted frontend Vitest coverage for Xugu tablespace tree behavior
- `vue-tsc` type checking
- `git diff --check`

### Packaged UI verification

Built the local debug macOS application and verified the actual UI flow:

1. Open an XuguDB connection.
2. Expand the connection-level storage group.
3. Expand multiple tablespaces.
4. Expand each `Files` child and confirm datafile names and paths.
5. Select a datafile node.
6. Collapse and re-expand both the storage group and a tablespace.
7. Confirm the datafile children remain available after re-expansion.
8. Confirm a non-Xugu connection does not show the storage group.

No database writes were performed during validation.

The local debug app was generated successfully. Release updater signing was not attempted because the local environment does not contain the repository's private signing key.


t8y2 (commented):
Maintainer patch pushed in 7f9a3036 (on top of the reviewed 1d53d3a).

**Root causes confirmed and fixed:**
1. **Tablespace/datafile nodes could not expand.** `toggle()` only routed `group-tablespaces`, and `tablespace`/`group-datafiles` nodes are built collapsed with eagerly-materialized children, so nothing ever flipped their `isExpanded` (the generic fallthrough no-ops on live nodes). Added a local-toggle branch for both types, mirroring `group-partitions`, with an expansion spec case asserting expansion works without re-invoking `loadXuguTablespaces`.
2. **ALL_DATAFILES fallback used invented column names** (`NODE_ID`, `FILE_PATH`, `FILE_SIZE`). Per the official docs (docs.xugudb.com), ALL_DATAFILES has the same columns as SYS_DATAFILES — the fallback now selects `NODEID, SPACE_ID, PATH, FILE_NO, MAX_SIZE, STEP_SIZE, CURR_SIZE, RESERVED1`, matching the positional 8-column scanner.

**Checks run:** `go test ./...` in `agents/drivers/xugu` (ok), `vitest run` on `SidebarTreeRuntimeHost.expansion.spec.ts` (9 passed), `oxfmt`/`oxlint` clean, `gofmt -l` empty.

**Noted, not changed:** `loadXuguTablespaces` still swallows all errors into an empty marked-loaded group; worth a follow-up to narrow that to metadata-unavailable errors.

t8y2:
Thanks for the contribution! Merged in 351a8b6, will be released in the next version.

PR #7680, opened 2026-08-31
feat(argo): add ArgoDB connection type with routine browsing

wangfengxiang:
## Summary

ArgoDB (Transwarp Inceptor fork of Hive) ships its own procedure and function catalog views (`system.procedures_v` / `system.functions_v`) that vanilla Apache Hive does not expose. This PR introduces a dedicated `argo` connection type that opts in to routine browsing only for servers known to support it, leaving vanilla Hive unchanged.

## Background

DBX 0.5.98 switched Hive-family connections from the bundled Java JDBC plugin to a native Go agent (`hive-go`). The native agent treated routine objects the same way it treats tables (always falling back to `SHOW CREATE TABLE`), so ArgoDB procedures were invisible. Rather than force Hive users into an empty Procedures node, this PR adds a dedicated `argo` connection type.

### New connection type
- New `plugins/connection-types/argo.yaml` + catalog profile. Reuses the `hive-go` agent (`agentKey: hive`) but flips `objectSource` on for argo only.
- `hive.yaml` `objectSource` stays `false` — vanilla Hive users see no routine nodes.

### hive-go agent (`agents/drivers/hive-go/`)
- `listObjects` / `getObjectSource` gate routine listing and source fetching behind `supportsRoutines()`, true only for `argo` / `inceptor` / `inceptor2` / `transwarp`. Vanilla Hive connections never fire the catalog-view queries.
- `listRoutines` / `getRoutineSource` quote `database_name` and routine name filters as **single-quoted string literals** (not backtick identifiers). ArgoDB rejects `lower(\`ods\`)` against string columns with `ERROR_STATUS`; `lower('ods')` works.

### Rust core (`crates/dbx-core/`)
- Thread `DatabaseType::Argo` through every exhaustive match on `DatabaseType` where Hive already appears (transfer, db_admin_sql, connection URL rendering, ddl profile, table select, agent dispatch, capabilities tests).
- `agent_driver.rs`: `list_tables_constrained` / `list_objects_constrained` now emit `object_types` under **both** `object_types` (snake_case) and `objectTypes` (camelCase) keys, so native Go agents that read the camelCase field receive the object-type filter instead of silently dropping it.

### Frontend (`apps/desktop/src/`)
- Extend the Hive-family Sets and the kerberos/timeout/icon helpers to cover argo for consistent connection dialog, icons, and SQL completion behavior.

## Verification

| Check | Result |
|---|---|
| `pnpm generate:connection-types` | 79 descriptors + 101 profiles |
| `cargo check -p dbx-core` | ✅ |
| `cargo test -p dbx-core --test database_capabilities` | 10 passed |
| `cargo test -p dbx-core --lib` (skipping 2 pre-existing flaky) | 5610 passed |
| hive-go `go test` | ✅ (5 routine tests added) |
| `pnpm typecheck` | 0 error |
| `pnpm tauri build` | DBX.app built (54.9 MB) |
| **End-to-end** | ArgoDB `ods` schema shows **18 procedures** ✅ |

### End-to-end trace (DBX log)
```
listRoutines type=PROCEDURE schema_in="ods" target="ods"
sql=SELECT procedure_name FROM system.procedures_v
     WHERE lower(database_name) = lower('ods')
     AND lower(procedure_name) LIKE '%' ORDER BY procedure_name
[query:done] row_counts=[18]
```

## ArgoDB server-side catalog (no changes required)

ArgoDB ships these views natively:
- `system.procedures_v`: `procedure_name, parameters, full_text, owner_name, owner_type, create_time, database_name`
- `system.functions_v`: `function_name, parameters, full_text, owner_name, owner_type, create_time, database_name`

## Design decisions

1. **Dedicated argo type, not hive reuse**: vanilla Hive has no `system.procedures_v` view; reusing hive would show Hive users an empty Procedures node. The dedicated type + `supportsRoutines()` gate keeps Hive clean.
2. **camelCase dual-emit**: Rust emits both `object_types` + `objectTypes` so existing agents don't need a coordinated rollout.
3. **Single-quoted string literals**: `database_name` is a string column; backtick identifier quoting is semantically wrong and ArgoDB rejects it.

## Checklist

- [x] Code follows the existing style (cargo fmt + vue-tsc clean)
- [x] Tests added for the new behavior
- [x] Verified against a real ArgoDB instance (18 procedures visible in the schema tree)


t8y2 (commented):
Maintainer follow-up pushed: merge of current main in 2389101a, plus fixes in 03bcc9b0.

**Conflict resolution (2389101a):** the single conflicting arm in `build_select_columns` (table_select.rs) — main added `InfluxDb`, this PR added `Argo` to the same Hive-family match. Resolution keeps both.

**Fixes (03bcc9b0):**
1. **Rust identifier/schema registration** — `Argo` was missing from the backtick arms of `quote_table_identifier` and `quote_transfer_identifier` (identifiers.rs) and from `is_schema_aware` (capabilities.rs), so dbx-core-generated SQL would have used ANSI double quotes (a string literal in the Hive dialect) and dropped schema qualifiers, contradicting the frontend sets and `argo.yaml`. All three arms now include `Argo`, mirroring kyuubi/impala, with a unit test in `sql_dialect/tests.rs`.
2. **Deleted `.github/workflows/build-windows-argodb.yml`** — branch-local build helper (default ref `feat/hive-routines-object-source`, draft release `argo-windows-local`) that should not land on main.

**Checks run:** `cargo test --no-default-features --features sqlite-bundled -p dbx-core --lib sql_dialect` (113 passed), `cargo fmt` clean, merge-tree clean after resolution.

t8y2:
Thanks for the contribution! Merged in 2f71a1a, will be released in the next version.

PR #7687, opened 2026-08-31
feat(tabs): add "close tabs to the left" context menu action

guoziyangnb:
## What

Adds a **"Close Tabs to the Left"** action to the tab-bar context menu, mirroring the existing **"Close Tabs to the Right"** action.

Right-clicking a tab now offers an item that closes every tab to the left of the target **within the same pinned group** (pinned tabs only close pinned tabs, regular tabs only close regular tabs). If the currently active tab is among the closed ones, the target tab becomes active instead. The special surfaces (Settings / Driver Store) sit at the rightmost end of the tab bar, so they are never affected.

## Why

"Close Tabs to the Right" has been around for a long time, but there was no symmetric way to clear the tabs on the left. Users who pin frequent connections on the left often accumulate a run of tabs they want to clear while keeping the reference tab — the only workarounds were "Close Other Tabs" (which also kills the reference tab) or closing tabs one by one. This makes the tab-bar menu symmetric and gives a precise one-shot way to declutter a tab group.

## Changes

- `apps/desktop/src/stores/queryStore.ts` — new `closeLeftTabs(id, onComplete?)` mirroring `closeRightTabs`: groups by pinned state, closes the tabs before the target, and keeps the target tab active when the active tab would otherwise be closed.
- `apps/desktop/src/components/layout/AppTabBar.vue` — new `tabsToLeftInGroup` / `hasTabsToLeft` / `closeTabsToLeftFromTab` helpers and the menu item placed between "Close Other Tabs" and "Close Tabs to the Right". The special-surface menu (Settings / Driver Store) intentionally has no left-side item, since those surfaces are always at the right end of the tab bar.
- i18n — `contextMenu.closeLeftTabs` added to all 8 locale files (en, zh-CN, zh-TW, es, it, ja, ko, pt-BR). Also fills the previously missing Korean `closeRightTabs` string, which had been falling back to English.
- Tests — source-assertion coverage in `apps/desktop/src/components/layout/__tests__/AppTabBar.spec.ts` for the new item's label, action, disabled state, and ordering.

## Testing

- `pnpm exec vitest run apps/desktop/src/components/layout/__tests__/AppTabBar.spec.ts` — all tests pass, existing "close right" assertions unaffected.
- `pnpm check` — format, lint, and typecheck all pass.
- Manual — right-clicked regular and pinned tabs: the new item appears, is disabled when nothing is to the left, and correctly closes the left group and re-activates the target tab.

### DIFF
--- a/apps/desktop/src/components/layout/AppTabBar.vue
+++ b/apps/desktop/src/components/layout/AppTabBar.vue
@@ -209,6 +209,20 @@ function closeTabsToRightFromTab(tab: QueryTab) {
   });
 }
 
+function tabsToLeftInGroup(tab: QueryTab) {
+  const groupedTabs = tab.pinned ? fixedTabs.value : regularTabs.value;
+  const targetIndex = groupedTabs.findIndex((item) => item.id === tab.id);
+  return targetIndex < 0 ? [] : groupedTabs.slice(0, targetIndex);
+}
+
+function hasTabsToLeft(tab: QueryTab) {
+  return tabsToLeftInGroup(tab).length > 0;
+}
+
+function closeTabsToLeftFromTab(tab: QueryTab) {
+  queryStore.closeLeftTabs(tab.id);
+}
+
 function hasSpecialRegularSurfaceToRight(surface: SpecialRegularSurface) {
   return surface === "settings" && !!props.driverStoreOpen;
 }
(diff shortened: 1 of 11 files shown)


t8y2 (approved):
Thanks! Reviewed and this looks good — approved.

Verified the key behaviors: `closeLeftTabs` mirrors `closeRightTabs` exactly (pinned-group isolation, empty-ids early return, same final-active-tab logic), and since it routes through `beginBatchClose`, dirty tabs still get the shared save/discard confirmation rather than being silently closed. Omitting the left action from the Settings / Driver Store menu is the right call — those surfaces sit rightmost, so a "close left" there would just duplicate "close other regular tabs". i18n coverage across all 8 locales is complete, and thanks for also filling in the missing Korean `closeRightTabs` string along the way.

t8y2:
Thanks for the contribution! Merged in 6620372, will be released in the next version.

PR #7716, opened 2026-09-01
feat(editor): preview DML changes as read-only SELECT with new-value column

doersa:
## Summary

Adds a **preview** action for DML statements in the query editor: instead of executing an `INSERT` / `UPDATE` / `DELETE` directly, the editor rewrites it into a read-only `SELECT` that shows the rows that *would* change, including a **new-value** column with the post-change values.

## What changed

- **`dbx-core`** — new `dml_preview_sql.rs` modules `dmlChangePreview` / `previewDmlChanges` that parse a DML statement and generate the equivalent read-only `SELECT` with a `new_value` column (INSERT/UPDATE/DELETE).
- **Desktop editor (Vue)** — `QueryEditor.vue` runs the preview against the execution candidate; `EditorToolbar.vue` exposes the preview button, hidden when the current statement is not DML; `App.vue` / `ContentArea.vue` surface the result panel for the preview query.
- **Backend** — routes/commands in `dbx-web` and `src-tauri` to serve the preview rewrite.
- **i18n** — `en` and `zh-CN` labels for the preview action.
- **Docs** — query-editor docs (EN + CN) updated.

## Files

19 files changed, 822 insertions(+), 4 deletions(-)

### Status
- [x] Preview SQL generation covered by unit tests (`dmlChangePreview.spec.ts`)
- [x] i18n for en + zh-CN

_Cross-repo PR from fork [`doersa/dbx`](https://github.com/doersa/dbx) → `t8y2/dbx`._


t8y2 (commented):
Maintainer patch pushed in 85e7cc8a (on top of the reviewed 1066c2a1).

**Three review findings folded in:**

1. **Docs no longer promise a side-effect-free dry run.** The EN/CN `query-editor` pages now state that the preview re-evaluates the written expressions, that side-effecting expressions (advancing a sequence, volatile functions) still take effect during preview, and that non-deterministic functions (`uuid()`, `rand()`, `now()`) may show values differing from the real execution. The examples were updated to the new label and error wording, and EN lists `UPDATE ... FROM` among unsupported forms.
2. **`UPDATE … FROM` is now conservatively rejected** (matching the existing tuple-assignment precedent) instead of generating a cross-join preview that could show a target row once per matching source row while the real UPDATE touches it once. New test asserts the exact unsupported error.
3. **Locale-neutral strings:** the new-value alias suffix is now ` (new)` (ASCII, quoted per dialect) and all backend errors are short English matching dbx-core's convention ("Preview is not supported: …", "Failed to parse the statement: …") — English users no longer see `price（新值）` headers or Chinese toasts. Frontend unchanged (it renders aliases verbatim and toasts backend errors); 12 test assertions updated.

**Checks run:** `dml_preview` suite 20 passed (incl. the new rejection test); `cargo check -p dbx-core` clean; fmt clean; clippy no new warnings; `dmlChangePreview.spec.ts` 3 passed; `git diff --check` clean.

**Known scope boundaries (deliberate):** MySQL `UPDATE a JOIN b` / `DELETE … USING` still render as cross joins (review scoped the fix to `update.from`); side-effecting expressions still execute during preview — documented rather than blocked.

t8y2:
Thanks for the contribution! Merged in 333c3a3, will be released in the next version.

PR #7744, opened 2026-09-01
feat(influxdb3): add InfluxDB 3.x engine (HTTP transport)

Clarence404:
Follow-up to the closed InfluxDB 3.x PR — this one drops Flight SQL
entirely in response to the previous review's dep/architecture concern.

## 变更说明

Adds InfluxDB 3.x as a native engine. Users pick it from the **InfluxDB
card**'s version dropdown (`1.x` / `2.x` / `3.x`), mirroring how v2 was
added to v1; there is no standalone "InfluxDB 3.x" card. Save-side
translates `db_type=influxdb + version=3` into `db_type=influxdb3` for
backend routing; load-side reverses the swap so the same card round-trips.

**Wire path**: `POST /api/v3/query_sql` for queries + `GET
/api/v3/configure/database?format=json` for the database list. Uses
`reqwest` — no new heavy dependencies (tonic / arrow-flight and the
arrow-* stack are gone versus the earlier PR).

**Explicit answers to the previous review's blocking issues:**

1. **Database creation is not advertised**. The driver manifest sets
   `databaseCreate: false` (see `plugins/connection-types/influxdb3.yaml`),
   and `databaseNamespaceCreation.ts` marks `influxdb3` as `deferred`
   with a comment pointing to the admin endpoint that would be needed.
   The UI / API no longer offer an action that cannot succeed.

2. **`max_rows` is pushed into the SQL, not applied post-materialization**.
   `execute_query(client, database, sql, max_rows)` calls `inject_limit`,
   which rewrites eligible top-level `SELECT` statements to end in
   `LIMIT N + 1` (the `+1` lets the caller detect truncation). Statements
   that already carry a `LIMIT`, and non-SELECT statements, run verbatim.
   Because HTTP responses are atomic, the win is that the server never
   serializes more than `N + 1` rows into JSON to begin with — bounded
   network and memory before parsing starts.

3. **Connection test runs the same code path as real queries**. Every
   metadata / query / test call funnels through a single
   `post_query_json(client, database, sql)` primitive. `test_connection`
   issues `SELECT 1` through that function against the configured database
   (or `_internal`, which is present on every Core install). A
   configuration cannot pass the test but fail the first query — the wire
   is either good for both or bad for both.

**Test coverage** (inline `#[cfg(test)]`, 10 tests):
- Bearer token + default database extraction from `ConnectionConfig`
- `--without-auth` case: empty password produces no `Authorization` header
- URL params concatenation
- Five `inject_limit` cases: bare SELECT, existing LIMIT, non-SELECT,
  leading comments, trailing semicolon
- `build_query_result` preserves first-seen column order across rows
- SQL literal escaping

## 变更类型

- [x] 新功能

## 涉及前端

- [x] 本 PR 涉及前端改动，已附截图

InfluxDB card gets a 3rd version option (1.x / 2.x / 3.x). Selecting 3.x
switches the port default to 8181 and shows database + token fields
(no user/password, no org/bucket).

image

image

## 验证

- `cargo check -p dbx-core --lib` — full features, clean
- `cargo check -p dbx --lib` — Tauri layer, clean
- `cargo test -p dbx-core --lib db::influxdb3_driver` — 10/10
- `pnpm typecheck` — clean
- End-to-end against a local `influxdb:3.10-core` container running
  `--without-auth`: connection test succeeds, `list_databases` returns
  the admin-configured DBs, `list_tables` filters `iox`-schema tables
  only, `SELECT * FROM cpu` round-trips.

- [x] `cargo check -p dbx-core --lib` 通过
- [x] `pnpm typecheck` 通过
- [x] 相关测试通过


t8y2 (commented):
Maintainer patch pushed: fixes in 45c5ee8, plus a merge of current main in 12251d7 (clean) so CI finally has a compiled base — the original head only ever ran label/notify.

**Fixes (45c5ee8), all three review findings:**
1. **Tokenless InfluxDB 3 Core can connect again** — the hard `throw` in `buildInfluxDbExternalConfig` is removed for version 3 (the driver already treats an empty password as "no Authorization header", matching `--without-auth`); Test and Save no longer demand a token. InfluxDB 2 keeps its token/org/bucket requirements untouched.
2. **Backend/manifest agreement on database creation** — `InfluxDb3` removed from `supports_create_database_target`; the manifest correctly advertises `databaseCreate: false` and the v3 SQL endpoint would have rejected the generated `CREATE DATABASE` anyway.
3. **Enterprise-safe connection test** — `test_connection` with no default database now probes the first database the token can actually see (via the already-implemented `GET /api/v3/configure/database`), falling back to `_internal` only as a last resort (it is admin-only on Enterprise).

**Verification run locally (given no compile CI had ever run on this PR):** `cargo check --no-default-features -p dbx-core` clean on both the patched tree and the post-merge tree; `influxdb3` test suite 10/10 passed; `db_admin_sql` suite 35/35 passed; `pnpm typecheck` 0 errors; rustfmt clean; oxfmt/oxlint clean. The pushed head should now trigger the full CI matrix — will merge once green.

t8y2:
Thanks for the contribution! Merged in 264f5c4, will be released in the next version.

PR #7764, opened 2026-09-01
fix(xugu): accept scheduler job object sources

mapan0424:
## Summary

Fixes scheduler job source loading for XuguDB connections.

## Why this is needed

The desktop tree already exposes scheduler jobs and maps the selected node to the `JOB` object type. The XuguDB agent also provides the corresponding source reconstruction for scheduler jobs. However, the dbx-core object-source model did not accept `JOB` when deserializing the request. As a result, selecting **View Source** failed before the request reached the agent with an error similar to:

`invalid args objectType for command get_object_source: unknown variant JOB`

This made an object that was visible and otherwise supported by the XuguDB integration appear unusable in the UI.

## Implementation

- Added `Job` to the shared `ObjectSourceKind` enum.
- Added `JOB` keyword parsing and serde round-trip coverage.
- Added exhaustive handling in schema/source-generation and transfer paths.
- Kept non-Xugu providers unchanged: their job branches are no-op/unsupported, and Oracle does not attempt to call an invalid metadata API for jobs.
- Preserved the existing XuguDB agent source-generation path without changing its SQL behavior.

## Validation

- `cargo check -p dbx-core`
- `cargo check -p dbx-web -q`
- `cargo test -p dbx-core --lib object_source_sql::tests::` (70 passed)
- `cargo test -p dbx-core --lib types::tests::object_source_kind_accepts_job_wire_value -- --exact` (passed)
- `go test ./...` in `agents/drivers/xugu` (passed)
- Built the debug macOS application with `pnpm tauri build --debug --bundles app --no-sign`.
- Performed a packaged UI verification: expanded the scheduler-jobs node, selected a job, and opened its source successfully. The generated `DBMS_SCHEDULER.CREATE_JOB(...)` source was displayed and the unknown-variant error did not recur.

The full dbx-core library suite was also attempted; it reaches an unrelated existing stack-overflow failure in `table_export::tests::external_driver_table_export_closes_cursor_at_row_limit`. No changed-code test failed.


t8y2 (commented):
Maintainer conflict resolution pushed in ada7b548a (old head 550016970, merged main@7e24f4007 — merge commit with both sides preserved). The single conflict was the tail of `mod tests` in `crates/dbx-core/src/types.rs`: this PR's `object_source_kind_accepts_job_wire_value` vs main's `completion_candidate_kind_accepts_uppercase_agent_wire_values`. Resolution: keep BOTH tests. All other files auto-merged or were untouched by main; `git range-diff` maps the PR commit identically, and every PR-added line (the `Job` variant, all no-op match arms in schema/transfer/object_source_sql, both wire-value tests) is present in the resolved tree.

Checks: `git diff --check` clean, no conflict markers; Rust CI on this head is the merge gate (targeted types tests were interrupted locally by a disk-full environment issue, not by the code).

t8y2:
Thanks for the contribution! Merged in 1818387da, will be released in the next version.

PR #7770, opened 2026-09-01
fix(sql-file): preserve mysql statement boundaries

### LINKED ISSUE #7738: [Bug] 执行SQL文件时语句顺序错乱导致报错
### 来源
QQ 群反馈 · **反馈人**：☆名洋＠

### AI 摘要
在DBX 0.5.99中，通过执行SQL文件功能时，SQL语句未按文件顺序执行，导致依赖顺序的语句报错；将相同内容粘贴到SQL窗口执行则正常。建议检查文件执行时的语句调度逻辑。

### 原始描述
DBX版本：0.5.99
执行sql文件时，会出现没按顺序执行，导致执行sql报错，但是sql文件中内容粘贴在sql窗口执行没问题

Eve-z1:
## Summary

- preserve source statement boundaries for MySQL-compatible SQL file imports instead of merging separate single-row `INSERT` statements
- keep PostgreSQL insert batching and existing source multi-row `INSERT` statements unchanged
- add unit coverage for MySQL, MariaDB-profile JDBC, Doris, dump controls, and multi-row passthrough
- add a real MySQL regression test for `LAST_INSERT_ID()` dependent statements

## Root cause

The SQL file optimizer combined adjacent compatible MySQL `INSERT` statements into one multi-row statement. Execution remained sequential, but the merge changed session semantics: after two auto-increment inserts, `LAST_INSERT_ID()` referred to the first row of the merged statement instead of the second source statement. A later dependent insert could then fail even though the same SQL executed statement-by-statement in the editor succeeded.

## Verification

- reproduced on MySQL 8 before the fix: the regression script failed with foreign-key `ERROR 1452`
- verified the same script after the fix: import succeeded and the child referenced the second parent
- `cargo test -p dbx-core --no-default-features --features sqlite-bundled --lib` focused SQL import tests
- ignored live MySQL regression and existing create/use/import live test
- `make cargo-check-fast`
- scoped `rustfmt --check` and `git diff --check`

The full fast suite was also attempted. Its SQL import tests passed, while unrelated existing MCP pnpm-fixture tests failed; a separate `query.rs` external-driver timeout was reproduced unchanged on the exact upstream main base.

Closes #7738

### DIFF
--- a/crates/dbx-core/src/sql.rs
+++ b/crates/dbx-core/src/sql.rs
@@ -1671,11 +1671,15 @@ pub fn optimize_sql_file_import_statements(
                 });
             }
             SqlFileStatementAction::Execute(sql) => {
-                let options = db_type.map(SqlParsingOptions::for_database_type).unwrap_or_default();
                 // Combining separate MySQL INSERT statements changes session
                 // semantics such as LAST_INSERT_ID(), so execute them with the
                 // same statement boundaries as the source file.
-                let mergeable_insert = merge_adjacent_inserts.then(|| parse_mergeable_insert(&sql, options)).flatten();
+                let mergeable_insert = merge_adjacent_inserts
+                    .then(|| {
+                        let options = db_type.map(SqlParsingOptions::for_database_type).unwrap_or_default();
+                        parse_mergeable_insert(&sql, options)
+                    })
+                    .flatten();
                 if let Some(insert) = mergeable_insert {
                     match pending_insert.as_mut() {
                         Some(batch) if batch.can_accept(&insert) => batch.push(insert),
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch pushed in b76ff1999 (old head a2ee2b281c): small polish from review — `SqlParsingOptions` is now constructed inside the `merge_adjacent_inserts.then(...)` closure instead of for every `Execute` statement, so MySQL-compatible targets (where merging is disabled) skip the per-statement profile lookup on large dumps. Behavior is identical (`bool::then` never calls the closure when disabled; `DatabaseType`/`SqlParsingOptions` are `Copy`).

Checks: rustfmt clean, `git diff --check` clean; full Rust CI will re-verify on this head.

t8y2:
Thanks for the contribution! Merged in 616c0daf2, will be released in the next version.

PR #7771, opened 2026-09-01
feat(xugu): show tablespace details in sidebar

mapan0424:
## What changed

Xugu tablespace and datafile nodes are now inspectable from the sidebar using the same detail popover interaction already used by other database objects.

The tablespace popover includes:

- tablespace name, node ID, and tablespace ID;
- tablespace type and datafile count;
- total, free, and derived used chunks;
- derived percentage usage;
- media-error state with localized yes/no labels.

The datafile popover includes:

- display name and full physical path;
- node ID, tablespace ID, and file number;
- current size, maximum size, and growth step with explicit MB units;
- the Xugu unlimited-size sentinel rendered as a localized “Unlimited” value;
- optional reserved metadata when the server returns it.

## Why this is needed

The storage-space tree previously exposed only names. Users had to issue system-table queries to understand where a tablespace or datafile resides, how much capacity it uses, and whether the server reports a media error. Showing this metadata in the existing sidebar popover makes the feature discoverable without introducing a new panel or changing the established navigation model.

## Scope and compatibility

- The detail branches are guarded by the effective database type and only activate for Xugu connections.
- Common metadata loading and non-Xugu tree behavior are unchanged.
- Nullable server fields are omitted instead of displaying misleading placeholder values.
- All new labels are localized across the supported desktop locales.

## Validation

- Targeted Xugu tablespace tests: 5/5 passed.
- Desktop TypeScript typecheck: passed.
- Oxlint for `apps/desktop/src`: passed with only pre-existing warnings.
- `git diff --check`: passed.
- Debug macOS app bundle built successfully with `pnpm tauri build --debug --bundles app --no-sign`.
- Packaged DBX UI verification completed against an Xugu instance: storage nodes loaded, tablespace details opened, datafile details opened, MB/unlimited formatting verified, and the existing popover visual style was preserved.


t8y2:
Thanks for the contribution! Merged in 7e24f4007, will be released in the next version.

PR #7773, opened 2026-09-01
fix: positional order by with group by only for group by cols

Tssshhhh:
## Change Description

Positional ORDER BY 1,2,3... N (https://github.com/t8y2/dbx/pull/1844) breaks clickhouse aggregate function
As a solution - if sql contains group by -- order by works only for group by columns, not all 
## Change Type

- [ ] New Feature

- [x] Bug Fix

- [ ] Performance Optimization

- [ ] Code Refactoring

- [ ] Documentation Update

- [ ] CI / Build

## Front-end Involvement

- [ ] This PR involves front-end changes, screenshots/screen recordings are attached (see below)

## Verification

- [x] `make check` passed

- [x] `make cargo-check-fast` passed

- [x] Related tests passed

## Related Issue

```sql
SELECT city, dept, argMaxState(name, hired_at) AS latest_hire, sumState(salary) AS payroll
FROM employees
GROUP BY city, dept
```
dbx sends ... ORDER BY 1, 2, 3, 4 LIMIT 1000

error from CH:
```
Code: 44. DB::Exception: Data type AggregateFunction(argMax, String, DateTime64(3, 'UTC'))
is not allowed in ORDER BY keys, because its values are not comparable. (ILLEGAL_COLUMN)
```
A GROUP BY query now sorts on its grouping keys alone

it is not same as https://github.com/t8y2/dbx/pull/6788 - i tried not to break pagination and not hardcode specific engines

### DIFF
--- a/crates/dbx-core/src/query_result_sql.rs
+++ b/crates/dbx-core/src/query_result_sql.rs
@@ -274,8 +274,8 @@ pub fn build_paginated_query_sql(options: PaginatedQuerySqlOptions) -> QuerySqlB
             if options.database_type == Some(DatabaseType::Kingbase) && has_top_level_top(&statement) {
                 return err("unsupported");
             }
-            let dedup_count = dedup_projection_count_without_order_by(&options.original_sql);
-            ok(add_standard_limit(&statement, options.database_type, safe_limit, safe_offset, dedup_count))
+            let dedup_order_by = dedup_projection_count_without_order_by(&options.original_sql);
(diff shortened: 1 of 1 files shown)


Tssshhhh:
@onceMisery maybe you will also take a look, is it ok for starrocks & doris ? now if you have group by 3 cols - order by have only exact 3 cols

t8y2 (approved):
Reviewed the exact head and current CI. The positional GROUP BY / ORDER BY fix is scoped and ready to merge.

t8y2:
Thanks for the contribution! Merged in e79db23, will be released in the next version.

PR #7796, opened 2026-09-01
fix(storage): restrict database file permissions to the owner

omerfarukalkilic:
## Problem

`connection_secrets` stores connection passwords in plaintext, and `dbx.db` is created with the process umask — 0644 on a default setup. On platforms whose per-user data directory is world-traversable this leaves the credential store readable by every account on the machine. Most Linux desktops create `~/.local/share` as 0755; macOS is not affected in practice because `~/Library` is 0700, but the file mode is the only thing carrying that guarantee, and it also follows the database into backups.

## Change

`Storage::open` clears the group and other bits from `dbx.db` and its `-wal` / `-shm` sidecars. Owner bits are preserved, so 0644 becomes 0600 and an already-restrictive mode is left alone.

Three deliberate choices:

- **Files only, not the directory.** `Storage::open` and `enable_wal_mode` both document that a portable install can be shared by several users on one machine. Tightening the directory to 0700 would break that; tightening the database files is enough to protect the credential store.
- **Best-effort, never fatal.** A portable data directory can sit on a filesystem without POSIX modes, and a file owned by another user fails `chmod` with `EPERM` rather than being re-permissioned behind that user's back. Neither should stop the app from starting, so failures are logged at debug level.
- **After `init_schema`.** The `-wal` and `-shm` sidecars only exist once the schema pass has written to the database.

It runs on every open, so existing installations are corrected as well as new ones.

## Tests

`open_clears_group_and_other_permissions_on_db_files` opens a database, sets it back to 0644 to stand in for a pre-existing installation, reopens it, and asserts the group/other bits are gone while owner access survives. Verified that it fails when the new call is removed.

`cargo test -p dbx-core --no-default-features --features sqlite-bundled --lib storage::` — 106 passed. `cargo fmt --check` clean; `clippy` reports no new warnings.

## Note for maintainers

If a shared portable data directory is a supported configuration where several accounts are expected to *read* the same `dbx.db`, this change ends that. Writing was already impossible under 0644, so it only affects shared reads of a plaintext credential store — but it is a behavior change and I'd rather flag it than hide it.

### DIFF
--- a/crates/dbx-core/src/storage.rs
+++ b/crates/dbx-core/src/storage.rs
@@ -740,6 +740,9 @@ impl Storage {
         // the app from starting.
         storage.enable_wal_mode().await;
         storage.init_schema().await?;
+        // After the schema pass, so the `-wal` and `-shm` sidecars exist and
+        // are covered too.
+        restrict_db_file_permissions(&storage.path);
         Ok(storage)
     }
 
@@ -916,6 +919,39 @@ fn remove_sqlite_db_files(db_path: &Path) -> Result<(), String> {
     }
     Ok(())
 }
+
+/// Drop group and other permissions from the SQLite database files.
+///
+/// `connection_secrets` stores connection passwords in plaintext, so the file
+/// mode is what keeps other local accounts out of the credential store on
+/// platforms whose per-user data directory is world-traversable: most Linux
+/// desktops create `~/.local/share` as 0755, unlike `~/Library` on macOS.
+///
+/// Runs on every open so existing installations are corrected as well, and is
+/// deliberately best-effort. A portable data directory can live on a
+/// filesystem without POSIX modes, and a file owned by another user fails
+/// `chmod` with `EPERM` instead of being re-permissioned behind that user's
+/// back. Neither case should stop the app from starting.
+#[cfg(unix)]
+fn restrict_db_file_permissions(db_path: &Path) {
+    use std::os::unix::fs::PermissionsExt;
+
+    for path in [db_path.to_path_buf(), db_path.with_extension("db-wal"), db_path.with_extension("db-shm")] {
+        let Ok(metadata) = std::fs::metadata(&path) else { continue };
+        let mode = metadata.permissions().mode() & 0o777;
+        // Owner bits are preserved; only group and other are cleared.
+        let owner_only = mode & 0o700;
+        if mode == owner_only {
+            continue;
+        }
+        if let Err(err) = std::fs::set_permissions(&path, std::fs::Permissions::from_mode(owner_only)) {
+            log::debug!("Could not restrict permissions on {}: {err}", path.display());
+        }
+    }
+}
+
+#[cfg(not(unix))]
+fn restrict_db_file_permissions(_db_path: &Path) {}
 
 fn ensure_history_columns_sync(conn: &Connection) -> Result<(), String> {
     const COLUMNS: &[(&str, &str)] = &[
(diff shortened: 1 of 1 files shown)


t8y2 (changes requested):
Request changes.

The permission hardening needs to cover the complete SQLite journal set and preserve the documented shared-directory behavior:

- `crates/dbx-core/src/storage.rs:939` only checks the main database, `-wal`, and `-shm`. A rollback journal (`<database>-journal`) can still retain database contents and should be included in the same hardening pass.
- The helper is called only after WAL setup and schema initialization at `storage.rs:745`. If either step fails, an existing database can remain at its old mode; please make the security guarantee independent of successful schema initialization.
- Clearing all group/other bits makes a portable data directory unusable for the documented case where multiple local users share it. Please define the supported sharing model or use a permission policy that protects `connection_secrets` without silently breaking shared portable directories.

Please add regression coverage for the rollback-journal path and the supported shared-directory behavior before merging.

omerfarukalkilic:
Thanks — all four are fair. Pushed ec727fc.

**1. Rollback journal.** Added `<database>-journal` to the same pass. While doing that I also stopped deriving the sidecar names with `Path::with_extension`, which silently assumes the database is named `*.db`; they are now built by appending to the filename the way SQLite names them. Master journals (`-mjXXXXXXXX`) are only written for a transaction spanning several attached databases, which the storage handle never performs — happy to include them anyway if you'd rather be belt-and-braces.

**2. Independence from schema initialization.** The pass now runs as soon as the database file exists, and again after `init_schema` on both the success and failure paths:

```rust
restrict_db_file_permissions(&storage.path);
storage.enable_wal_mode().await;
let schema = storage.init_schema().await;
restrict_db_file_permissions(&storage.path);
schema?;
```

The first call covers the file that actually holds the secrets; the second picks up journals created by the schema pass, including when it fails.

**3. Sharing model.** You're right that flagging it in the description while the code silently broke it was not good enough. The data directory's own mode now selects the model:

- **Group-writable directory** — several local accounts have to be able to write to a shared data directory for it to work at all, so this is read as the operator opting into group access. World bits are dropped, group bits are preserved.
- **Any other directory** — treated as single-user; files become owner-only.

World-readable is not treated as a sharing model, because it cannot be narrowed to a set of accounts, and under the previous 0644 it never granted write access anyway — a genuinely shared portable directory already needed a common group.

If you'd rather express this as an explicit setting than infer it from the directory mode, say so and I'll rework it — the inference keeps the change self-contained, but it is a heuristic and it is your call.

**One case the heuristic reads conservatively.** On distributions that pair per-user groups with umask 002 (the Red Hat family), `~/.local/share` is created group-writable, so it is read as the shared model and the files end up 0660 rather than 0600. The group there is the user's own private group, so they stay effectively private, and the world bit — the actual exposure — is dropped either way. Flagging it because it means the heuristic is conservative by default rather than maximally strict; an explicit setting would not have that property.

**4. Regression coverage.** Three tests, each verified to fail when its own behavior is reverted:

- `open_restricts_the_database_and_its_journals_to_the_owner`
- `open_preserves_group_access_in_a_shared_data_directory`
- `open_restricts_the_database_even_when_schema_initialization_fails` (a read-only directory keeps the database file writable while denying the journal SQLite needs, so the schema pass fails deterministically)

The tests create their own data directory with an explicit mode — the process temp directory is group-writable on Linux (`/tmp` is 1777) and would otherwise be read as the shared model.

**Left alone deliberately.** `remove_sqlite_db_files` still derives its sidecars with `Path::with_extension`. It carries the same latent assumption, but it is not on this change's path and I would rather not widen the diff — say the word and I'll align it here or in a follow-up.

`cargo test -p dbx-core --no-default-features --features sqlite-bundled --lib storage::` — 108 passed, `fmt` clean, no new `clippy` warnings. That was on macOS; the workflow runs on this PR are still awaiting approval, so the Linux job has not reported yet.

t8y2 (approved):
Re-reviewed at ec727fc — all four points from my previous review are addressed, and I verified the new tests locally (`cargo test -p dbx-core --no-default-features --features sqlite-bundled --lib storage::`, 108 passed):

- `-journal` is now included in the hardening pass, and deriving sidecar names by suffix append (instead of `with_extension`) is the right fix; master journals are correctly out of scope since the storage connection never ATTACHes.
- The two-pass placement (immediately after connect, and again after schema init on both success and failure) makes the guarantee independent of initialization.
- The directory-mode heuristic works as the sharing model: group-shared portable layouts stay usable while the world bits are always dropped.
- Regression tests cover the rollback-journal path (including a stale `-journal`), the group-shared directory, and schema-init failure.

Two non-blocking notes: `remove_sqlite_db_files` still misses `-journal` (pre-existing, fine as the follow-up you proposed), and the sub-second first-creation window before the first chmod is inherent to post-hoc chmod. The Linux CI job still needs a workflow approval before merge — approving it on my side.

t8y2:
Thanks for the contribution! Merged in d3a6a69796, will be released in the next version.

PR #7800, opened 2026-09-01
fix(structure): preserve DDL after tab switch

### LINKED ISSUE #7778: [Bug] 编辑表时候从DDL切换到其他例如“字段”，再切换DDL，DDL空白
### 数据库类型和版本

MySQL 8.0

### 问题描述

CTRL+D 编辑表，打开之后DDL显示正常，切换到字段，再切换到DDL，DDL显示空白

Image

### 支持信息

DBX 已是最新版本 (0.5.99)。
运行环境桌面端
操作系统windows

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

Eve-z1:
## Summary

- keep only the DDL tab panel mounted while inactive so Reka does not remove the CodeMirror container before the existing reinitialization runs
- hide the inactive DDL panel while preserving the existing editor destroy/recreate lifecycle and cached DDL
- add a real-tabs component regression covering DDL → Columns → DDL without a second DDL request

## Verification

- `node node_modules/vitest/vitest.mjs run --maxWorkers 4` (1244 files, 12241 tests passed)
- `node node_modules/vue-tsc/bin/vue-tsc.js --noEmit --project apps/desktop/tsconfig.json`
- focused DDL lifecycle/search/concurrent-transition tests (4 files, 7 tests)
- reproduced before the fix and verified after the fix in the real macOS DBX desktop app against MySQL 8.0; DDL remained visible after repeated tab switches

Closes #7778

### DIFF
--- a/apps/desktop/src/components/structure/TableStructureEditor.vue
+++ b/apps/desktop/src/components/structure/TableStructureEditor.vue
@@ -4759,7 +4759,7 @@ watch([activeTab, loading, ddlLoading, ddlContent], ([tab, structureIsLoading, d
             </div>
           </TabsContent>
 
-          <TabsContent ref="ddlScrollerRef" v-if="tableMetadataCapabilities.ddl" value="ddl" class="relative m-0 min-h-0 flex-1 overflow-auto p-[var(--structure-cell-px)]" @scroll.passive="onStructureContentScroll('ddl', $event)">
+          <TabsContent ref="ddlScrollerRef" v-if="tableMetadataCapabilities.ddl" value="ddl" force-mount class="relative m-0 min-h-0 flex-1 overflow-auto p-[var(--structure-cell-px)] data-[state=inactive]:hidden" @scroll.passive="onStructureContentScroll('ddl', $event)">
             <div v-if="ddlLoading" class="flex items-center justify-center gap-2 py-10 text-muted-foreground">
               <Loader2 class="h-4 w-4 animate-spin" />
               {{ t("common.loading") }}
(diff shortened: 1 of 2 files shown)


t8y2 (approved):
Thanks! Verified the approach and it looks correct.

Root cause matches: the DDL CodeMirror container lived inside reka-ui's `TabsContent`, which unmounts its children while inactive; on returning to the tab, `initDdlEditor` could bail on its container/state guards after the remount, leaving the panel blank. `force-mount` + `data-[state=inactive]:hidden` makes the container's existence independent of tab state (and `display:none` keeps the inactive panel out of the tab order despite the `tabindex="0"` reka sets), while the existing destroy-on-leave / init-on-return watches keep the editor lifecycle and DDL cache semantics untouched — confirmed by the test asserting `.cm-content` is removed while inactive and exactly one `loadObjectDdl` call across the switch cycle.

Also checked the surrounding surface: the scroll helpers only act on the tab being activated, the hide class survives `cn()`/tailwind-merge, and the arbitrary `data-[state=...]` variant pattern already has precedent in `TooltipContent`/`DropdownMenuContent`. Keeping it scoped to the DDL panel is right — the other tabs have no imperative editors.

One minor nit (non-blocking): the test asserts the literal class string `data-[state=inactive]:hidden`, which couples it to the hiding mechanism — fine as-is given happy-dom's computed-style limitations, just something to update if the mechanism ever changes.

t8y2:
Thanks for the contribution! Merged in a0d868b637, will be released in the next version.

PR #7809, opened 2026-09-01
fix(influxdb): use SELECT * and inject a rolling time window for table quick-open

Clarence404:
## 变更说明

**Problem**. Today's sidebar quick-open on an InfluxDB table emits SQL
with no time predicate:

    SELECT "time", "device", "-_acc_param_-EP_EXP_kWh", ... 55 more names ...
    FROM "device_data" ORDER BY time DESC LIMIT 100;

Without a `WHERE time > ...` clause, InfluxDB has no shard-pruning
signal:

- **v1 / v2 (TSM)**: the engine iterates **every shard for the
  measurement** — with default 1w / 1d shard duration and a year of
  retained data that is hundreds of shards — then merges them
  in-memory to satisfy `ORDER BY time DESC`. `LIMIT 100` only clips
  the tail; the full scan has already happened.
- **v3 (Parquet + DataFusion)**: without a `time` predicate the
  planner can't prune Parquet files by min/max statistics and reads
  every row group into memory before applying `ORDER BY … LIMIT`.

On a production `device_data` table this turns a "peek at the latest
rows" click into a multi-second, gigabyte-scale scan that also
degrades every concurrent query on the same server. It's the exact
pattern InfluxData's own docs list under "queries to avoid," and why
Grafana / Chronograf / InfluxDB Studio all require or default to a
time range.

**Fix**. Two orthogonal changes, both scoped to `DatabaseType::InfluxDb`
and `DatabaseType::InfluxDb3`:

1. **Fall back to `SELECT *` like the other single-table dialects.**
   `build_select_columns` treats most engines as `"*"` and only
   branches into per-column enumeration for the SQL-on-Hadoop family
   (Hive / Kyuubi / Impala / Argo) where identifier aliasing matters.
   InfluxDb sat inside that same branch, which explains the wall of
   50+ quoted column names above. This PR moves InfluxDb back to the
   `*` default — InfluxQL supports it natively. InfluxDb3 already
   falls through to `*` today, no change there.

2. **Inject a rolling five-minute default WHERE.** When the caller
   has not supplied their own predicate, `default_time_series_predicate`
   emits a `time > now() - 5m` clause so the server prunes shards
   (v1/v2) or Parquet files (v3) up front. Five minutes matches
   InfluxDB Studio's default; users can widen it by editing the
   generated SQL or the WHERE input.

The default is scoped: any user-supplied `where_input` bypasses it
entirely, so widening the range still works exactly as before.
`ORDER BY time DESC LIMIT N` is preserved intentionally — it drives
dbx's grid pagination (right-hand "100 rows/page" dropdown) and
gives DataFusion a deterministic order it does not provide
otherwise.

Per-engine syntax lives in `default_time_series_predicate`:

- `InfluxDb` (v1.x / 2.x InfluxQL) → `time > now() - 5m`
- `InfluxDb3` (DataFusion SQL) → `time > now() - INTERVAL '5 minutes'`

Extending to other time-series engines is a one-line change.

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [x] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

- [ ] 本 PR 涉及前端改动，已附截图/录屏

Backend SQL generation only.

## 验证

- `cargo test -p dbx-core --lib sql_dialect::table_select` — 15/15
  (four focused new/updated tests: InfluxDb emits `SELECT *` with
   the 5-minute window, InfluxDb3 uses the DataFusion INTERVAL
   literal, a user-supplied WHERE bypasses the default, and
   non-time-series engines are untouched).
- Manually against a local InfluxDB 1.11 container holding roughly
  1M points in `device_data`: sidebar quick-open completes in tens
  of milliseconds instead of the multi-second cross-shard scan it
  took before, and the generated SQL fits on one line.

- [x] 相关测试通过


t8y2 (commented):
Maintainer patch pushed in 887bdc6 on top of 5ffc32b9.

Root cause: the rolling 5-minute window was injected inside the shared `build_table_data_select_sql_with_database`, so every caller passing an empty `where_input` silently inherited the filter. Two internal best-effort callers never opted into quick-open semantics:

- `agent_tools.rs` `build_sample_data_sql` (AI `sample_table_data`): measurements with no points in the last 5 minutes would sample 0 rows and the LLM would conclude the table is empty.
- `csv_export.rs` `export_table_data_csv_core`: whole-table CSV export would be silently truncated to a rolling 5-minute window, with `now()` sliding per page.

Patch:

- Added `inject_default_time_series_where` to `TableDataSelectSqlOptions` (serde default `false`); the window is now injected only when the flag is set and no user WHERE exists.
- The frontend passes `injectDefaultTimeSeriesWhere: true` only from the data-tab surfaces (sidebar quick-open, `buildTableSql` in `useDataGridActions`, the `useNavigationTargets` builds), so page 1 and pagination stay consistent; agent sampling and CSV export keep the historical unbounded semantics.
- Removed the redundant `InfluxDb` early return in `build_select_columns` (the `!matches!` fallback already returns `*` after the list removal) and dropped the unverifiable "matches InfluxDB Studio's default" comment claim.

Checks run: `cargo test -p dbx-core --no-default-features --features sqlite-bundled --lib sql_dialect::table_select` (16 passed, including the new `influxdb_table_select_without_opt_in_stays_unfiltered` regression), `cargo check -p dbx-core --no-default-features --features sqlite-bundled --tests` clean, `vue-tsc --noEmit` clean, and vitest on the tableSelectSql / useSidebarDataOpenRuntime / useNavigationTargets (×2) / useDataGridActions specs: 119 passed.

Two product-level notes remain as follow-ups (not blockers): hourly/daily rollup tables quick-open as an empty grid within the 5-minute window (the generated SQL is visible in the editor tab), and the total-row count is computed without the window.

t8y2 (approved):
Approved. The quick-open behavior and both dialect syntaxes are correct (verified against the InfluxDB docs), the test coverage is solid, and the maintainer patch above (887bdc6) scopes the rolling window to data-tab callers so agent sampling and CSV export keep their historical semantics while page 1 and pagination stay consistent.

The two product-level notes (hourly/daily rollups quick-opening as an empty grid within the 5-minute window, and the total-row count ignoring the window) are recorded as follow-ups, not blockers.

t8y2:
Thanks for the contribution! Merged in 84669b0d1a, will be released in the next version.

PR #7817, opened 2026-09-02
feat(mcp): execute SQL batches with opt-in rollbackable transactions

Abeautifulsnow:
## Summary

Adds `dbx_execute_batch` to the MCP server (issue #7548) and hardens its
`use_transaction` path to **fail closed** instead of silently degrading to
auto-commit when the active backend cannot guarantee a rollbackable batch.

A batch requested with `use_transaction: true` must be atomic. Previously,
backends without a verifiable transactional path (unsupported drivers, and
`TxPath::None`) ran the statements one-by-one in auto-commit — so a later
statement failing left the earlier ones already committed. This PR rejects
those requests up front, and also closes two gaps found during review:
Oracle DDL (which implicitly commits like MySQL-family DDL) and a
single-statement script being misrejected as an invalid option combination.

### `dbx_execute_batch` tool (core + MCP)
- New MCP tool that splits a script with a dialect-aware parser and returns
  one result per statement, or a single merged transaction outcome when
  `use_transaction` is set on a multi-statement script.
- Returns a structured array (`structuredContent`) so callers no longer parse
  concatenated text (issue #7548).
- Rejects the tool for Redis / MongoDB connections.

### Fail-closed `use_transaction`
- **core**: removed the `exec_tx_none` auto-commit fallback; unsupported
  backends now map to `BatchTransactionPath::Unsupported` and are rejected
  before execution. The transaction gate now runs **before** the SQL Server
  and HTTP-SQLite batch fast paths, so those backends actually enter the
  transaction wrapper instead of silently ignoring the option.
- **agents** (JDBC `TransactionExecutor`, Hive, IoTDB): a transaction request
  the driver cannot roll back throws / returns an error instead of degrading
  to one-by-one auto-commit, so a failed batch can never leave partial writes.
- **Oracle DDL**: the DDL-rejection check now uses the shared
  `database_supports_transactional_ddl` capability predicate (the same one the
  schema-diff deploy path uses), instead of a hard-coded MySQL-family list —
  so Oracle (which implicitly commits around DDL like MySQL) is also rejected
  for DDL batches. `Goldendb` was added to the MySQL-family set.
- **Single-statement scripts** now ignore `use_transaction` entirely (plain
  auto-commit), and the `session_id` / `continue_on_error` / DDL option-validity
  rejects fire **only** when the script actually enters transaction mode —
  a single statement combined with `session_id` or `continue_on_error` is no
  longer misrejected.

### Query timeout on transactional batches
- Applies the MCP global query timeout to the SQLite and Agent transaction
  paths, and interrupts SQLite transaction statements that overrun the budget.

## Tests
- Core unit tests for Oracle/MySQL-family DDL rejection, SQLite rollback of a
  failed batch, and unsupported-backend rejection before execution.
- MCP tests: MySQL-family + Oracle DDL rejection, SQL Server merged outcome,
  single-statement option-combination allowance, structured content, and live
  regression / protocol tests.
- Go driver tests (Hive, IoTDB) for transaction rejection before statements run.

## Notes
- `use_transaction` is rejected (not silently ignored) for backends that
  cannot provide a rollbackable batch — including SQL Server-compatible /
  HTTP-SQLite backends that were previously allowed to silently ignore it.
  If you prefer these to keep running in atomic-batch mode instead of being
  rejected, that should be decided against the issue requirements.
- `use_transaction` cannot be combined with `session_id` (runs on a pooled
  connection) or `continue_on_error` (rolls back at the first failure).


t8y2 (changes requested):
Thanks for the thorough work — the fail-closed transaction model, the SQLite watchdog design, and the test quality here are excellent. One blocking issue before merge:

**Schema Diff deploy regression**: The new DDL rejection at the top of `execute_statements_in_transaction_on_pool_typed` (query.rs) is not scoped to the MCP `use_transaction` option — it also fires for `execute_schema_diff_deploy` (used by the desktop Schema Diff deploy feature via the Web route). On `main`, MySQL-family/Oracle DDL deploy scripts execute through this path with documented "mixed outcome on failure" semantics; after this change they are rejected up front with an error referencing `use_transaction`, an option that caller never sets, so schema diff deploy stops working for those engines. Could you gate the rejection so it only applies to opt-in batch transactions, or let the schema-diff path keep its run-and-report-mixed behavior?

Two smaller notes: (1) the MCP pre-check splits with `sql_execution_plan_for_database` while the core gate uses `query_execution_plan(..., is_sqlserver_agent)` (`split_sql_batches` for SQL Server agents) — if the splitters disagree, the MCP-level rejects can diverge from the core's actual transaction entry; (2) `tracing = "0.1"` is added to dbx-mcp for a single `warn!` — the rest of the codebase uses the `log` facade, which would avoid a second logging dependency.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 543bf6690, will be released in the next version.

PR #7937, opened 2026-09-02
feat(mcp): list stored routines and fetch routine source via MCP

wangfengxiang:
## Summary

Adds two read-only MCP tools so clients can browse stored procedures/functions without dropping to raw SQL:

- **`dbx_list_routines`** — lists procedures and functions for a schema as a markdown table (name, type, schema, signature, comment). Optional `routine_type` filter (`PROCEDURE` / `FUNCTION`); omit for both.
- **`dbx_get_routine_source`** — returns a routine's full source text, with optional `signature` for databases that overload routine names (e.g. Oracle).

## Design: generic, no per-database code

The tools are thin wrappers over machinery that already exists — every database family's routine support is already implemented for the desktop sidebar:

- **LocalBackend** → `dbx-core` `list_objects_core(object_types=[...])` / `get_object_source_core`, which dispatch per connection type (JDBC `DatabaseMetaData.getProcedures`, Hive-family `system.procedures_v`/`functions_v`, MySQL `information_schema.ROUTINES`, Postgres `pg_proc`, ...).
- **WebBackend** → reuses the existing `/api/schema/objects` and `/api/schema/object-source` routes the desktop app already calls.
- Routines are re-filtered locally after listing, so agents that ignore the `objectTypes` hint cannot leak tables into the results.
- Unsupported databases (sqlite, duckdb, ...) surface their normal "not supported" errors through the standard tool error path.

Both tools honor the existing `ensure_tool_allowed` fine-grained authorization and are registered in the settings-page MCP tool allowlist with localized labels (en / zh-CN / zh-TW / ja / ko / es / pt-BR / it).

## Testing

- `cargo test -p dbx-mcp` green (tool registration counts and JSON-schema optionality checks updated)
- `cargo fmt --check` clean
- `pnpm typecheck` green

### DIFF
--- a/apps/desktop/src/components/editor/EditorSettingsDialog.vue
+++ b/apps/desktop/src/components/editor/EditorSettingsDialog.vue
@@ -2286,6 +2286,8 @@ const mcpToolOptions = [
   { name: "dbx_list_databases", labelKey: "settings.mcpToolListDatabases" },
   { name: "dbx_list_tables", labelKey: "settings.mcpToolListTables" },
   { name: "dbx_describe_table", labelKey: "settings.mcpToolDescribeTable" },
+  { name: "dbx_list_routines", labelKey: "settings.mcpToolListRoutines" },
+  { name: "dbx_get_routine_source", labelKey: "settings.mcpToolGetRoutineSource" },
   { name: "dbx_get_schema_context", labelKey: "settings.mcpToolGetSchemaContext" },
   { name: "dbx_execute_query", labelKey: "settings.mcpToolExecuteQuery" },
   { name: "dbx_open_session", labelKey: "settings.mcpToolOpenSession" },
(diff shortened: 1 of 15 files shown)


t8y2 (commented):
Maintainer patch pushed in e30a78db (old head dafb657e). Three items from review:

1. Fixed the WebBackend query-param mismatch: `WebBackend::list_routines` now sends `object_types=` instead of `objectTypes=` — `SchemaQuery` binds snake_case only, so the filter was silently dropped (every web-mode call fetched the full object list, and a `routine_type=PROCEDURE` request also returned functions). `get_routine_source` params were audited and are correct as-is.
2. `LocalBackend::list_routines` now defaults to `["PROCEDURE", "FUNCTION"]` when `routine_type` is omitted, hitting the targeted routine group-load path instead of enumerating every object kind (mirrors the WebBackend fallback).
3. Documented the two new tools in mcp.mdx / mcp.cn.mdx / mcp-server README, and updated the stated tool counts (15 → 17).

Checks: `cargo check -p dbx-mcp --no-default-features` clean, rustfmt clean.

t8y2:
Thanks for the contribution! Merged in aac1d7853, will be released in the next version.

PR #7975, opened 2026-09-03
fix(desktop): let data header name chips flex instead of hard truncating

### LINKED ISSUE #7880: [Bug] UI显示Bug表名显示不完整
### 数据库类型和版本

starrrocks3,5,17

### 问题描述

Image如图

### 支持信息

如上图

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

_No response_

Abeautifulsnow:
## Problem

In the data-grid header, the table-name chip was hard-capped at `max-w-48`
(192px). Long table names were truncated even when the header row had plenty
of free space on the right — the workaround was manually squeezing the panel
(#7880). The connection and schema chips in the same row had the same fixed
caps (`max-w-48` / `max-w-56`).

## Root cause

The chips rely on truncation, but the fixed `max-w-*` utilities cap them at a
static width that never reacts to available space. The bug is generic UI
behavior, not StarRocks-specific — any database with a table name longer than
~28 characters at 12px triggers it.

## Change

- Remove the fixed `max-w-48` / `max-w-56` caps from the three name chips in
  the data-mode header. They now flex with `min-w-0` + `truncate`, so names
  are shown in full whenever space permits and only clip (with ellipsis) when
  the row itself runs out of space. Right-side toolbar controls were already
  `shrink-0` and are unaffected.
- Add `:title` tooltips to the table and schema chips (the connection chip
  already had one) so the full value stays reachable on hover.
- Add `ContentArea.dataHeader.spec.ts` to guard against reintroducing fixed
  width caps and dropping the tooltips.

## Validation

- New spec fails on the unfixed source (2/2 assertions) and passes after the
  fix; targeted suites pass: `vitest run` → 15/15 (ContentArea output-view,
  data header, column-layout popover).
- `pnpm typecheck` (vue-tsc), `oxlint`, `oxfmt --check` — all clean on the
  touched files.
- Rendered behavior verified against a real Tailwind v4 compile in a browser
  with the exact chip classes and the issue's data: at 1600px the full name
  `warehouse_management_order_detail_wide` is displayed; at 620px chips
  truncate proportionally with ellipsis and the toolbar buttons remain intact.

Closes #7880

### DIFF
--- a/apps/desktop/src/components/layout/ContentArea.vue
+++ b/apps/desktop/src/components/layout/ContentArea.vue
@@ -1902,13 +1902,15 @@ defineExpose({
     <template v-else-if="activeTab.mode === 'data'">
       <div class="flex-1 min-h-0 flex flex-col">
         <div class="h-9 shrink-0 border-b bg-background/80 px-3 flex items-center gap-2 text-xs">
-          <span v-if="activeConnection?.name?.trim()" data-data-header-connection class="inline-flex max-w-48 min-w-0 items-center truncate rounded border border-border bg-muted/30 px-2 py-0.5 text-muted-foreground" :title="activeConnection.name">
+          <!-- No fixed max-w cap on these chips: they must flex (min-w-0 + truncate)
+               so long names only clip when the header row itself runs out of space (#7880). -->
+          <span v-if="activeConnection?.name?.trim()" data-data-header-connection class="inline-flex min-w-0 items-center truncate rounded border border-border bg-muted/30 px-2 py-0.5 text-muted-foreground" :title="activeConnection.name">
             {{ activeConnection.name }}
           </span>
-          <span class="inline-flex max-w-48 min-w-0 items-center truncate rounded border border-border bg-muted/50 px-2 py-0.5 font-medium">
+          <span class="inline-flex min-w-0 items-center truncate rounded border border-border bg-muted/50 px-2 py-0.5 font-medium" :title="activeTab.tableMeta?.tableName || activeTab.title">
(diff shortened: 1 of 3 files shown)


Weiki886:
我检查了当前 frontend CI 失败原因：`packages/app-tests/dataTabHeaderConnection.test.ts:30` 仍断言 `data-data-header-connection` 必须包含 `max-w-48`，这与本 PR 移除固定宽度的目标冲突。建议将该断言改为验证该 chip 不包含 `max-w-48`/`max-w-56`，并包含 `min-w-0` 与 `truncate`，例如：

```ts
assert.doesNotMatch(template, /data-data-header-connection class="[^"]*max-w-48/);
assert.match(template, /data-data-header-connection class="[^"]*min-w-0[^"]*truncate/);
```

我用 PR 当前源码确认失败点就是这个旧断言，不涉及 Rust 或数据库。

t8y2 (commented):
Maintainer patch pushed in a24c5839 (old head 97ce5204f): aligned the stale assertion in `packages/app-tests/dataTabHeaderConnection.test.ts` with the PR's new chip contract — the test still required the removed `max-w-48` cap, which was the only cause of the red `frontend` check. Now requires `min-w-0`+`truncate` and forbids the removed caps scoped to the connection chip's class attribute (the remaining `max-w-48` occurrences in ContentArea.vue are result-run buttons, not header chips). Both spec files pass locally (3/3).

t8y2:
Thanks for the contribution! Merged in 3c4d61339, will be released in the next version.

PR #8004, opened 2026-09-03
fix(editor): anchor wrapped line numbers via inline style

### LINKED ISSUE #7920: [Bug] sql编辑器换行选中后，行号变成了居中
### 数据库类型和版本

通用

### 问题描述

sql编辑器换行选中后，行号变成了居中。

### 支持信息

DBX 版本: v0.6.0
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

Image

Image

Abeautifulsnow:
## Problem

Fixes #7920.

In the query editor, the line number of a soft-wrapped line drifts between
two alignments: anchored to the first visual row for a while, then reverting
to vertically centered over the whole wrapped block, then top-anchored again.
The reversion is triggered by selection changes / cursor movement /
active-line refreshes; scrolling, editing or resizing restores the correct
alignment. Single-line numbers must stay vertically centered.

## Root cause

Line-number gutter elements are styled `align-items: center`, and the query
editor's measuring ViewPlugin temporarily added a `cm-db-wrapped-line-number`
class (`align-items: flex-start`) to wrapped ones. CodeMirror rebuilds the
gutter element's `className` from its markers on every marker update
(`GutterElement#setMarkers` in @codemirror/view), which wipes that toggled
class on active-line and selection updates. The plugin only re-measured on
doc/geometry/viewport changes — not on selection updates — so the class was
not re-added immediately, leaving the number centered until the next
scroll/edit/resize.

## Fix

Anchor only wrapped lines by writing an inline `align-items: flex-start`
from the measuring plugin instead of toggling a class
(`setQueryEditorLineNumberAlignment`). `setMarkers` rewrites `className` but
never touches `element.style`, so the wrapped-line alignment survives
selection/active-line updates and viewport virtualization without widening
the re-measure triggers. Single lines keep the theme's centered rule — the
inline style is cleared for them — so their appearance is unchanged.

Removed the now-unused `cm-db-wrapped-line-number` theme rule and added a
regression test proving the alignment survives a gutter className rebuild.
`buildQueryEditorLineNumbersExtension` (show/hide toggle + line-selection
`mousedown`) is unchanged.

## Compatibility

- Single lines: vertically centered, visually unchanged.
- Wrapped lines: anchored to the first visual row, stable under interaction.
- Other gutters (statement run, fold) untouched.
- No database / persistence / public API impact.

## Validation

- `pnpm exec vitest run .../queryEditorLineNumbers.spec.ts .../editorThemes.spec.ts` — passed (18/18)
- `pnpm exec oxlint --vue-plugin` (4 changed files) — passed
- `pnpm exec vue-tsc --noEmit --project apps/desktop/tsconfig.json` — passed
- `git diff --check` — passed
- Visual check in the running editor (wrapped line number stays top-anchored
  while moving the cursor / changing selection; single lines centered) — passed

## Known limitations

None.

### DIFF
--- a/apps/desktop/src/lib/__tests__/editorThemes.spec.ts
+++ b/apps/desktop/src/lib/__tests__/editorThemes.spec.ts
@@ -175,7 +175,7 @@ describe("SQL completion theme", () => {
 });
 
 describe("editor gutters", () => {
-  it("anchors line numbers to the first visual row of wrapped lines", () => {
+  it("keeps single line numbers vertically centered in the base rule", () => {
     const rules = buildEditorFontThemeRules();
 
     expect(rules[".cm-lineNumbers .cm-gutterElement"]).toMatchObject({
@@ -183,6 +183,5 @@ describe("editor gutters", () => {
       display: "flex",
       justifyContent: "flex-end",
     });
-    expect(rules[".cm-lineNumbers .cm-gutterElement.cm-db-wrapped-line-number"]).toMatchObject({ alignItems: "flex-start" });
   });
 });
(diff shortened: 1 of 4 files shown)


t8y2 (approved):
Thanks for the contribution! Reviewed and looks good — merging soon.

t8y2:
Thanks for the contribution! Merged in dd11e4274, will be released in the next version.

PR #8039, opened 2026-09-03
fix(postgres): prevent custom type stream deadlock

### LINKED ISSUE #7986: [Bug] 导出数据库->选定指定表（其中一张表是向量数据），会提示失败：
### 数据库类型和版本

PostgreSQL16

### 问题描述

导出数据库->选定指定表（其中一张表是向量数据），会提示失败：
Image

### 支持信息

DBX版本：v0.6.1

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

_No response_

CN-Scars:
## 变更说明

Prevent PostgreSQL database exports from deadlocking while streaming large results that contain user-defined types such as pgvector.

`query_typed_raw` sends Describe and Execute together for unnamed queries. When an uncached custom result OID is encountered, tokio-postgres performs a type lookup on the same connection before returning the row stream. A sufficiently large result can fill the connection buffers while that lookup waits behind it, so DBX never receives the first row and eventually reaches both the query and stream-cleanup timeouts.

This change describes the query once without executing it, allowing custom result metadata to be resolved before the actual unnamed stream starts. The data query remains unnamed, preserving compatibility with `DEALLOCATE ALL` and environments that lose server-side prepared statements.

It also adds a live regression covering 5,000 rows of `vector(1536)` and verifies that the metadata statement is not retained server-side.

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

This PR does not modify frontend code.

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

Validated with:

- The PostgreSQL snapshot live regression matrix: wide pgvector result, `DEALLOCATE ALL`, wide JSONB, stalled-stream timeout/rollback, and pending-row cancellation.
- Existing pgvector export-literal and binary-precision unit tests.
- Manual database export of 5,000 rows of `vector(1536)`: completed in 3.7 seconds and produced a 35 MB SQL file containing all 5,000 unique rows and valid 1,536-component square-bracket vector literals.

## 关联 Issue

Closes #7986

### DIFF
--- a/crates/dbx-core/src/db/postgres.rs
+++ b/crates/dbx-core/src/db/postgres.rs
@@ -1413,12 +1413,34 @@ fn postgres_select_stream_outcome(stream: tokio_postgres::RowStream) -> Postgres
     PostgresSelectStreamOutcome::Binary { stream, metadata }
 }
 
+async fn prepare_unnamed_select_metadata(
+    client: &deadpool_postgres::Client,
+    sql: &str,
+) -> Result<PreparedSelectMetadata, tokio_postgres::Error> {
+    // query_typed_raw sends Describe and Execute together. If its result has an
+    // unknown user-defined type, tokio-postgres then resolves that type with a
+    // second query on the same connection before returning the RowStream. A
+    // large result can fill the connection buffers and block that type lookup.
+    // Describe once without executing so custom types are cached before the
+    // actual unnamed stream starts. The stream remains unnamed to tolerate
+    // server-side prepared statement loss in snapshot/proxy environments.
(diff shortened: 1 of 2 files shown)


t8y2 (approved):
Thanks for the contribution! Reviewed and looks good — merging soon.

t8y2:
Thanks for the contribution! Merged in 2324d84c4, will be released in the next version.

PR #8147, opened 2026-09-04
fix(generate): stream bulk data generation with progress and configurable timeout

gggaiitx:
## Problem

Data Tools → Generate Data becomes unresponsive at large row counts and fails with timeouts (reported against TDSQL, which uses the MySQL driver):

- 300k rows: the dialog freezes after clicking generate (inserts still complete in the background); 1M–10M rows: the UI hangs entirely.
- Large volumes time out.

Root causes:

1. **Synchronous full-dataset generation on the UI thread.** All rows were generated in one loop and buffered three times over (`rows[]`, formatted `valueRows[]`, and one giant INSERT string), so 10M rows × N columns saturates the JS heap.
2. **Single giant INSERT bypassed packet-size protection.** For MySQL/TDSQL every row was merged into one multi-row INSERT, skipping the existing max_allowed_packet-aware batch route.
3. **No timeout parameter on the execution path.** Generation used `executeInTransaction`, which has no timeout argument and falls back to the hard-coded 30s default.
4. **No progress feedback or cancellation.**

## Solution

Keep the dialog responsive and make inserts streaming:

- **Chunked generation** (`dataGenerate.ts`): new exported `createTableGenerateState` / `generateTableRowsChunk` / `formatGeneratedRowValues` / `buildGenerateInsertStatements` / `splitValueRowsByByteBudget`. Cross-chunk state (unique-value sets, TDengine tag values, global row index, generated child-table name) lives in the state object so uniqueness and sequence semantics are preserved across chunks. `generateTableData` is unchanged behaviourally and now built on the same primitives.
- **Sample preview**: the preview step generates a 50-row sample instead of the full dataset.
- **Streaming inserts** (`DataGenerateDialog.vue`): rows are generated chunk by chunk (`batchRows`, default 1000) and grouped into statements under a 1 MB byte budget, then executed via `executeMultiWithProgress` with:
  - a live progress bar (inserted/total, percent, elapsed time),
  - cancellation (`cancelQuery` on the execution id),
  - configurable timeout (`timeoutSecs`, 0 = unbounded, capped at 24h),
  - existing options preserved: truncate, use transaction, extended insert, continue on error.
- **i18n**: new strings added for zh-CN, zh-TW, en, ja, ko, es, pt-BR, it, tr.

## Testing

- New `packages/app-tests/dataGenerateChunk.test.ts`: 12 tests covering chunked/one-shot equivalence, cross-chunk uniqueness, sequence continuity across chunk boundaries, chunk edge cases, TDengine stable-table tag/tbname coherence, statement building (multi-row vs per-row vs Oracle INSERT ALL vs forced single-row), and byte-budget splitting.
- All 17 existing `dataGenerate.test.ts` tests pass (29/29 total).
- `vue-tsc --noEmit` clean for `apps/desktop` (remaining errors are pre-existing nacos module-resolution issues unrelated to this change).
- Production build (`vite build`) verified.


t8y2 (changes requested):
Requesting changes for now. The streaming design itself is right — chunked generation, the byte budget, and the timeout propagation (0 → unbounded on both the Rust and JDBC agent sides) all check out — but the error accounting has two real holes:

1. `run()` in `streamInsertTable` only awaits `executeMultiWithProgress` and never inspects the returned results. On several backend paths a failed statement does not throw: with `continueOnError` the sequential executor records the error in the result array and still returns `Ok` (`crates/dbx-core/src/query.rs:3127`), and the MySQL route returns `Ok` with an embedded error result even on connection checkout failure (`crates/dbx-core/src/query.rs:3513`). In those cases `ok += group.length` (`DataGenerateDialog.vue:663`) counts failed batches as inserted, `lastError` stays empty, and with `useTransaction: false` + single-row statements the loop doesn't even stop — the result page reports full success while rows are silently missing. On main each statement was executed and caught individually, so this regresses the continue-on-error option this PR intentionally keeps. Please check the returned `QueryResult[]` for `error` (or use the `onProgress` callback already wired up) and only count genuinely successful rows.

2. In `startInsert`'s outer `catch`, `outcome = { ok: 0, ... }` (`DataGenerateDialog.vue:719`) drops the progress made before the throw. With streaming, `UniqueValueGenerationError` now surfaces mid-insert (unique space exhausted at row 500k, say); the table already contains those rows but the result page reports 0 inserted. Carry the last known count through instead.

Smaller notes: a cancel that interrupts a transaction/single-statement round trip comes back as an error, so `cancelled = insertCancelled && !outcome.error` (`DataGenerateDialog.vue:721`) mislabels user cancels as failures; and "Copy All SQL" on the preview step now copies only the 50-row sample script, which no longer matches the button label — either rename it or copy the streaming-equivalent full script.

t8y2 (approved):
Thanks — all four points from the last review are addressed, verified at 5f7cd4f5:

1. `executeBatch` now folds the returned `QueryResult[]` through `summarizeBatchResults()` and only counts rows from statements that carry neither `execution_error` nor a structured `error`, with `rowsPerStatement` mapped by `generateInsertBatches()`. This correctly covers the sequential-executor and MySQL-checkout paths that resolve `Ok` with an embedded error.
2. Both the mid-stream generation-failure path and the `startInsert` safety-net catch carry the last known `ok` count through, so a late failure reports the rows that were actually inserted.
3. Cancels are recognized via the `insertCancelled` flag plus `isQueryCanceledError()` and reported as cancelled rather than failures.
4. "Copy All SQL" is now "Copy sample SQL" with matching translations in all nine locales.

Two non-blocking nits, feel free to ignore: the production guard now evaluates the 50-row sample script (same tables and statement shapes, so equivalent coverage), and the cancellation regex could in principle match an SQL error message containing the word "cancel" — the flag keeps that exposure minimal. Thanks also for the accounting unit tests.

t8y2:
Thanks for the contribution! Merged in a3772fb442, will be released in the next version.

PR #8153, opened 2026-09-04
test(xugu): add focused live integration baseline

mapan0424:
## Summary

Add a focused, opt-in live integration baseline for the Xugu driver. The suite is intentionally scoped to common database behavior so it can serve as a reusable regression gate without duplicating the existing specialized tests for packages, triggers, scheduler jobs, spatial features, or other Xugu-specific objects.

## Why

The existing broad baseline is useful for exhaustive qualification but is too large for a fast, repeatable driver-level regression check. Common connection, metadata, DDL/source, query, transaction, paging, and error-recovery behavior needs a compact test that can run consistently during local validation and in CI when a live Xugu service is available.

## What changed

- Added `common_integration_live_test.go` with an environment-gated `TestLiveXuguCommonIntegration`.
- Covers session reuse, table/view discovery and filtering, columns, primary keys, foreign keys, indexes, comments, reconstructed table DDL, view source, DML, transactions, cursor paging, and recovery after an expected SQL error.
- Creates randomized synthetic fixtures under a test-only naming prefix and removes them with cleanup, avoiding real business object names and sensitive data.
- Keeps credentials and endpoint configuration outside the repository through the existing `XUGU_LIVE_*` environment variables.
- Skips cleanly when `XUGU_LIVE_TEST` is not enabled, so normal builds and contributors without a live database are unaffected.
- Does not change production driver or agent behavior.

## Validation

- `go test ./...` passes in `agents/drivers/xugu`.
- The live suite was run successfully against the configured Xugu environment for: an administrative system-database session, an administrative business-database session, ordinary business-database access with and without DBA, and ordinary system-database access with minimal object-creation grants and with DBA.
- Fixtures were cleaned after each run; no permanent test users or test objects remain.
- `git diff --check` passes.

## Scope and compatibility

This PR adds tests only. It does not alter SQL templates, metadata behavior, object support, or frontend code for any database. The live test remains opt-in and has no effect when the Xugu environment variables are absent.

### DIFF
--- a/agents/drivers/xugu/common_integration_live_test.go
+++ b/agents/drivers/xugu/common_integration_live_test.go
@@ -1,0 +1,377 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+	"time"
+)
+
+// TestLiveXuguCommonIntegration is the compact, opt-in XuguDB integration
+// baseline. It intentionally follows the size and scope of the Kingbase and
+// Vastbase live tests instead of duplicating the much larger TEST-01
+// certification matrix. The fixture contains only synthetic names and values
+// and is removed on every exit path.
+//
+// Coverage is deliberately limited to the common DBX contract:
+//   - connection and session reuse;
+//   - table, view, index, key and column metadata;
+//   - comments, source and reconstructed table DDL;
+//   - query execution, DML transactions and cursor paging; and
+//   - structured SQL-error recovery in the same session.
+//
+// Run it with XUGU_LIVE_TEST=1 and the XUGU_LIVE_HOST, XUGU_LIVE_PORT,
+// XUGU_LIVE_DATABASE, XUGU_LIVE_USERNAME and XUGU_LIVE_PASSWORD variables.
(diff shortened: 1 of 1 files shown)


t8y2 (approved):
Thanks for the contribution! Reviewed and approved — a clean, well-scoped addition. The gating matches the existing `XUGU_LIVE_TEST` convention used by the other Xugu live suites, CI never sets `XUGU_LIVE_*` so the test always skips there, fixture naming/cleanup hygiene is solid, and the coverage baseline mirrors the Kingbase/Vastbase integration tests without duplicating the specialized Xugu suites.

Two optional nits, not blocking:
- `common_integration_live_test.go:238`: the trailing `+ ""` in the `sql_error_recovery` subtest's SQL string is a no-op and can be dropped.
- Skipping on `openDB` error follows `tablespaces_live_test.go`, but `sql.Open` is lazy so that branch is effectively unreachable (real connect failures surface at `validateConnection`). If you touch the file again, `t.Fatal` there would match the trigger/scheduler suites.

t8y2:
Thanks for the contribution! Merged in 1c972b80c, will be released in the next version.

PR #8158, opened 2026-09-04
feat(influxdb): grid filter support for time syntax of influxdb

ptma:
## 变更说明

为表格字段过滤器增加对 InfluxDB(包括V3) timestamp 类型的时间语法支持。

支持的时间语法：
1. 绝对时间
```
'2006-01-02T15:04:05.00Z' -- RFC3339 date-time string
'2006-01-02 15:04:05.00'  -- RFC3339-like date-time string
1136189045000000000       -- Unix nanosecond epoch time
1136189045s               -- Unix epoch time
```
2. 相对时间
```
'2023-01-01T00:00:00Z' + 2h -- Resolves to 2023-01-01T02:00:00Z
'2023-01-01 00:00:00' - 20h -- Resolves to 2022-12-31T04:00:00Z
1672531200000000000 + 1d    -- Resolves to 2023-01-02T00:00:00Z
now() - 1d2h
```
另外如果用户输入的时间包含为 RFC3339 时间格式, 会自动将 RFC3339 时间用单引号括起来， 比如：
输入 `2023-01-01T00:00:00Z + 2h` ，实际值会转换为 `'2023-01-01T00:00:00Z' + 2h`

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

## 验证

增加了测试用例 `formats_influxdb_timestamp_copy_literals_with_supported_syntax` (Line: 5500)

- [x] `make check` 通过
- [x] 相关测试通过

### DIFF
--- a/crates/dbx-core/src/data_grid_sql.rs
+++ b/crates/dbx-core/src/data_grid_sql.rs
@@ -1,7 +1,9 @@
 use chrono::{Local, NaiveDateTime};
+use regex::Regex;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::collections::HashSet;
+use std::sync::OnceLock;
 
 #[path = "data_grid_neo4j_sql.rs"]
 mod data_grid_neo4j_sql;
@@ -2158,6 +2160,11 @@ pub(crate) fn format_grid_sql_literal_with_identifier_quote(
             return timestamp.to_string();
         }
     }
+    if (database_type == Some(DatabaseType::InfluxDb) || database_type == Some(DatabaseType::InfluxDb3))
+        && column_info.is_some_and(|column| column.data_type.eq_ignore_ascii_case("timestamp"))
+    {
+        return normalize_influxdb_time_expr(&text);
+    }
     if is_mysql_binary_literal_column(database_type, column_info) {
         if let Some(literal) = format_mysql_binary_literal_text(&text) {
             // DBX result values expose binary columns as prefixed hex; keep them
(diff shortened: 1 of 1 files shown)


t8y2 (changes requested):
Thanks for the follow-up — the clippy `useless_format` issue is confirmed fixed in 3cff6b3b, and the implemented InfluxQL time syntax set matches the official v1/v2 docs exactly (quoted RFC3339, epoch with unit suffix, `now() ± duration`, absolute-time arithmetic). Two things still need to be addressed before this can merge:

1. **The `InfluxDb3` arm never triggers on a real server.** The v3 driver passes `information_schema.columns.data_type` through verbatim (`influxdb3_driver.rs:295-303`), and InfluxDB 3 reports the `time` column as `Timestamp(Nanosecond, None)` (see the [information schema docs](https://docs.influxdata.com/influxdb3/core/reference/sql/information-schema/)), so the `eq_ignore_ascii_case("timestamp")` check never matches — the v3 test cases pass only because they fabricate `"timestamp"`. Also note that even when it does match, the emitted syntax is InfluxQL-specific: InfluxDB 3's SQL dialect requires `now() - INTERVAL '1 hour'` style arithmetic (see [v3 SQL docs](https://docs.influxdata.com/influxdb3/enterprise/query-data/sql/parameterized-queries/)), so `now() - 1d`, `'...' + 2h`, and `1136189045s` would be rejected. Could you either scope this to `InfluxDb` only for now (one-line change), or match the v3 type string and generate v3 SQL syntax?

2. **The fallback path loses quoting/escaping.** `normalize_influxdb_time_expr` returns the raw input for anything that doesn't match (line 3505). Before this change, non-matching text values on the time column went through the default path, which escapes `\` and `'` and wraps the value in single quotes. Now an ordinary value like `foo` is injected into `WHERE time = foo` unquoted, which at best produces a parse error and at worst widens the injection surface (the v1 `/query` endpoint accepts `;`-separated batch statements). Suggestion: only bypass the default escaping when the full input matches the strict legal-expression regex, and fall through to the default quoted/escaped path otherwise. Minor nit while you're in there: the `prev`/`next` check indexes `chars().nth()` with a byte offset (`m.start()`), which misbehaves when multi-byte characters precede the match.

Once these are addressed this will be a nice quality-of-life improvement for InfluxDB grid filtering.

PR #8191, opened 2026-09-04
feat(elasticsearch): add visual search profiler

Abeautifulsnow:
## Problem

  Elasticsearch `_search` responses with `"profile": true` contain deeply nested query timing data that is difficult to inspect as raw JSON.

  ## Root cause

  DBX already preserves REST-DSL response bodies, but it had no result view that parsed and rendered `profile.shards[].searches[].query[]`.

  ## Fix

  - Add an Elasticsearch/Easysearch-only Profile result view.
  - Render shard-specific collapsible query trees with timing, breakdown values, self-time cost shares, and heat indicators.
  - Highlight the critical path using cumulative `time_in_nanos`.
  - Preserve existing result, summary, explain, chart, and message views.
  - Add localized Profile UI strings and focused parser/component tests.

  ## Compatibility

  The Profile entry appears only for Elasticsearch-compatible responses containing a valid query profile tree. Non-profile responses and existing output views are
  unchanged.

  ## Validation

  - `pnpm exec vitest run apps/desktop/src/lib/__tests__/elasticsearchProfile.spec.ts apps/desktop/src/components/common/__tests__/ElasticsearchProfilePanel.spec.ts
  packages/app-tests/queryResultToolbar.test.ts` — passed (40 tests)
  - `pnpm lint` — passed with existing unrelated warnings
  - `git diff --check` — passed

## Related Issues
Related #7919 

## Screenshot

8d3bfd68-a713-4ec2-badc-f5df692bda15


t8y2 (commented):
Pushed a small maintainer patch in 145b719 on top of 46b38bf8 (three changes, all in-scope):

1. **Blocker — `ContentArea.vue` profile toggle emit was missing the tab id.** `emit('update:activeOutputView', ...)` expects `[tabId, view]`, but the toolbar button passed only the view. That was both the `frontend` CI typecheck failure (TS2769) and a live no-op: App.vue's handler compared `'profile'` against the active tab id, so the button never switched views. Now passes `activeTab.id` first, matching the adjacent `@select-view` call sites.
2. Removed the redundant `activeOutputView === "profile" ||` clause in `showStandaloneResultToolbar` (subsumed by the following `!== "result"` check).
3. Removed the unused `profile.tag` / `profile.emptySearch` keys from all 9 locale files — they were defined but never referenced.

Checks run locally at the patched head: `vue-tsc --noEmit` clean, `ElasticsearchProfilePanel.spec.ts` + `elasticsearchProfile.spec.ts` (22 tests) pass, oxfmt clean on changed files, `git diff --check` clean. Will merge once CI on 145b719 is green.

t8y2:
Thanks for the contribution! Merged in 3ee0640fc2, will be released in the next version.

PR #8260, opened 2026-09-05
fix(xugu): exclude synthetic scopes from schema export

mapan0424:
## Summary

Xugu exposes Public synonyms and Scheduled jobs through reserved synthetic scope names so the sidebar can display them as database-level nodes. The database export dialog was treating those protocol-only scopes as ordinary schemas and listing them in the schema selector.

This change keeps the synthetic scopes available to the sidebar while excluding them from schema export selection and all-database export planning.

## Root cause

`listSchemas` intentionally returns Xugu synthetic scopes for tree navigation. The export UI consumed that result without distinguishing real schemas from protocol-only namespaces, so both synthetic entries appeared in the export schema dropdown.

## Changes

- Added a shared Xugu-aware schema filter for export boundaries.
- Applied the filter to single-database schema selection.
- Applied the filter to all-database export plan generation.
- Kept the filter scoped to `xugu`; other database types are unchanged.
- Preserved real schema names, including names that may resemble system or guest schemas.
- Added regression coverage for Xugu filtering, non-Xugu compatibility, and all-database export planning.

## Validation

- Export schema regression tests passed.
- Xugu synthetic-scope tests passed.
- Type checking passed.
- Frontend production build passed.
- Debug macOS desktop bundle built successfully with updater artifact generation disabled for local testing.
- Manual packaged UI verification confirmed that the export schema selector shows only real schemas while the sidebar still shows Public synonyms and Scheduled jobs.

## Compatibility

This is a desktop export-boundary change only. No Agent behavior, database metadata SQL, or non-Xugu export behavior was changed.

### DIFF
--- a/apps/desktop/src/components/export/DatabaseExportDialog.vue
+++ b/apps/desktop/src/components/export/DatabaseExportDialog.vue
@@ -12,7 +12,7 @@ import * as api from "@/lib/backend/api";
 import type { ExportProgress } from "@/lib/backend/api";
 import { isSchemaAware, isSingleDatabase } from "@/lib/database/databaseFeatureSupport";
 import { databaseOptionsForConnection, fetchNamespaceOptionsForConnection } from "@/composables/useDatabaseOptions";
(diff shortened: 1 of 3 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 5d65be03a, will be released in the next version.

PR #8266, opened 2026-09-05
feat(xugu): add read-only server monitoring dashboard

mapan0424:
## Summary

This PR adds a read-only XuguDB server monitoring dashboard for operational visibility without changing database state. It is intentionally scoped to XuguDB and does not alter monitoring behavior for other database engines.

## What changed

- Added an XuguDB server dashboard with version, node, session, transaction, lock, memory, storage, I/O, and configuration summaries.
- Added native Xugu system-view queries and column-name based result parsing.
- Added refresh controls and a clear read-only presentation.
- Added permission-aware degradation: unavailable system views are reported as unavailable instead of being rendered as misleading zero values.
- Kept the dashboard gated to Xugu connections so MySQL, PostgreSQL, Oracle, and other database implementations are unchanged.
- Added focused unit coverage for catalog SQL, result parsing, role labels, and restricted-view behavior.

## Permission scope

The fully populated dashboard is validated for a connection to the SYSTEM database using SYSDBA. A business-database DBA account is not claimed as a full instance-monitoring account in this PR: Xugu distinguishes SYS_* system virtual tables from DBA_* views, and the current dashboard queries instance-level SYS_* data. When those views are unavailable, the UI shows an explicit permission/data-unavailable state instead of fabricating values. Support for a DBA connection backed by equivalent DBA_*/explicit SYS_* grants is intentionally left as a follow-up compatibility task.

## Validation

- `pnpm exec vitest run apps/desktop/src/lib/__tests__/database/xuguServerStatus.spec.ts`
- `pnpm typecheck`
- `pnpm exec oxfmt --check` on all changed files
- Packaged desktop UI validation was completed against XuguDB SYSTEM/SYSDBA and a restricted account; the restricted case degraded with an explicit warning.

## Risk

The implementation is Xugu-specific and read-only. No shared query behavior or metadata implementation for other database engines was changed.


t8y2 (commented):
Applied one small maintainer patch (2f61fc89): when the active-session query fails for a restricted user, the merged session rows now show "—" instead of a hard "0", matching the unavailable markers used elsewhere on the dashboard (the header already guards via availability flags).

One question left open, not blocking: `XUGU_TABLESPACE_SUMMARY_SQL` reads `C.CHUNK_SIZE` as a column from `SYS_CTL_VARS`, which implies a single-row wide table — if it can ever return multiple rows, the cross join would inflate the tablespace totals. If you ever see inflated counts on a cluster, constraining that join would be the fix. Also FYI, the earlier `autofill` failure was the i18n translation bot (DeepSeek API), unrelated to your change.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 42a17be7c, will be released in the next version.

PR #8286, opened 2026-09-06
feat(app): 添加全局背景图片

onceMisery:
## 变更说明

为桌面端添加全局背景图片支持：选择一张本地图片后，壁纸铺满整个应用窗口底层，工具栏、侧边栏、主内容区、SQL 编辑器等表面变为半透明从底层透出（类 JetBrains "Background Image" 效果）；卡片、对话框、下拉弹层等浮层保持实底，保证可读性。

## 变更类型

- [x] 新功能
- [ ] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

image

image

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

## 关联 Issue
8222


t8y2 (commented):
Thanks for the great work on this feature — the default-off scoping, guards and test coverage are all solid. I pushed a maintainer update to the branch (3f9acc73e3 → e523c42cd, base 772ce36de) with four changes:

1. **Conflict resolution** (`App.vue`, `EditorSettingsDialog.vue`): merged upstream's settings import/export (7961e688f) and theme hover preview, keeping both sides — the wallpaper composable call and the new `useUiFontFamilyPreview` in `App.vue`; both import sets in the dialog. Your `backgroundImage` draft key now flows into settings import/export.

2. **Fix: custom palette colors were lost while the wallpaper is active** (`useBackgroundImage.ts`): the wallpaper effect clears all surface variables before re-emitting them with alpha, but with a custom palette those variables are inline styles written by `applyCustomUiColors`, so clearing them fell back to the palette class colors — custom background/sidebar colors silently reverted (even with the wallpaper turned off). The effect now resolves custom-palette base values from the custom color settings and re-emits them (tinted when active, exact when inactive). Covered by two new component tests.

3. **Fix: missing exhaustiveness case** (`EditorSettingsDialog.vue`): `backgroundImage` was added to `EDITOR_SETTINGS_DRAFT_KEYS` without a case in the draft-to-refs switch, so `pnpm run typecheck` (vue-tsc) failed on the branch; added `case "backgroundImage"`.

4. **Fix: transfer category** (`settingsTransfer.ts`): added `backgroundImage` to the `appearance` category so import/export covers it and the import dialog lists it; round-trip test added (the "every whitelisted key maps to a category" test was otherwise red).

Checks run locally on e523c42cd: full desktop vitest suite (1348 files / 13341 tests green) and `pnpm run typecheck` clean.

Two non-blocking follow-up ideas (not addressed here, happy to discuss): returning the image bytes via `tauri::ipc::Response` instead of base64 IPC would avoid a main-thread `atob` decode on startup, and storing the file under the resolved data dir (extension only in settings) would survive portable-mode/data-dir moves.

t8y2:
Thanks for the contribution! Merged in 39aa663ebf, will be released in the next version.

PR #8308, opened 2026-09-07
feat(ai): render AI chart-json blocks as ECharts charts

Abeautifulsnow:
## Summary

Lets the AI assistant render charts directly in the conversation. When the assistant emits a fenced ```chart-json block, DBX parses it against a **controlled, versioned schema** and renders it as a **line**, **bar**, or **pie** chart with ECharts — instead of showing raw JSON text. No new dependency is added (`echarts` / `vue-echarts` are already in the frontend).

Implements the chart-rendering capability of #6467.

## What it does

- **Thin Rich Content protocol** — a small registry keyed on the raw fenced code language routes ```chart-json blocks to a chart handler. Because it matches the **raw** language (lowercased/trimmed) *before* the SQL/SHELL normalization, the existing SQL → proposal / confirm / execute path is untouched. The registry is typed generically (`AiRichBlockHandler<AiMessageRichSegment>`), so future rich block types only extend the segment union — they never change the renderer contract.
- **Controlled, versioned schema (not an arbitrary ECharts option)** — a discriminable union:
  - `version` is a **required protocol field** (V1 accepts only `1`; missing → rejected as malformed)
  - `line` / `bar`: non-empty categorical `xAxis.values` + `series` (each `data.length === xAxis.values.length`)
  - `pie`: `data` only (no axis, no `series`); a single zero is allowed, negative / all-zero datasets are rejected
  - category axis values (`"Jan"`, `"华东"`, product names) are kept as-is — never coerced; numeric columns accept `number` or numeric strings and reject `"NaN"` / `"Infinity"` / non-numeric
  - budget limits: ≤256KB JSON, ≤8 series, ≤5000 points per array, ≤10000 total
  - any invalid block is **deterministically rejected** and falls back to the raw code block — never auto-downsampled or rewritten
- **Rendering** — `AiChartRenderer.vue` renders with `vue-echarts`, follows the existing light/dark theme, and uses a fixed minimum height (≥320px) with `autoresize`.
- **Lazy-loaded** — the renderer is mounted via `defineAsyncComponent`, so the ECharts bundle (~370KB min / 128KB gzip) is fetched only when a chart actually renders, keeping plain conversations light. The chunk is shared with the existing query chart, and a render-function placeholder reserves the chart slot height so the stream does not jitter while it loads.
- **Grounded-data prompt** — the assistant prompt (injected into both normal and vector-DB system prompts) shows the exact `chart-json` schema with `version: 1` examples, asks for at most one chart per reply and only when a visual comparison / trend / distribution / proportion materially improves the answer, and requires chart data to be grounded in available data (query results, attachments, provided values) — never invented.

## Safety

- The chart block is pure data (validated finite numbers only), so there is no XSS surface.
- SQL / write-confirmation semantics are unchanged: `chart-json` is non-SQL (`isSql: false`), so it never gets the run/apply buttons or the write-confirmation flow, and `extractSql` only matches the SQL family.
- Streaming uses the existing `pending`/closed gate: an unclosed fence stays a plain code block; the chart only appears once the fence closes.

## Tests

- Unit — schema parse/validate (required version + `version: 2` reject, category-axis pass-through, numeric coercion + `"NaN"`/`"Infinity"` reject, pie zero/negative/all-zero, length & budget limits), option builder (line/bar/pie, theme colors, pie mapping), registry closed-gate (unclosed → not parsed, non-rich languages excluded, SQL/SHELL not routed).
- Integration — AI assistant template splits `chart` vs `code`; a closed fence switches code → chart; SQL buttons / `extractSql` / proposal detection confirm non-SQL behavior for `chart-json`.
- Prompt — normal + vector DB system prompts both include the `version: 1` schema examples, the single-chart limit, and the grounded-data rule (en + zh).
- typecheck / lint green; the affected suites pass (584 tests across the AI + editor surfaces).

## Notes

- No new dependencies.
- The model prompt also tells the assistant not to emit ```html blocks unless explicitly asked; no HTML rendering is included in this change.

## Related issue

This PR addresses **one part of #6467** only. That issue asks for three capabilities in the AI conversation:

1. **AI chart rendering** — the assistant can emit chart data that DBX renders visually ← **covered by this PR**
2. AI-generated HTML preview — *not* part of this PR
3. Conversation / analysis report export — *not* part of this PR

Items 2 and 3 are left for follow-up work; this change is self-contained and does not block or depend on them.


t8y2 (commented):
Maintainer patch applied in `a41bbf738`.

Root cause: after the user selected a PNG destination, filesystem or browser-download errors were caught and silently ignored. The existing `chartDownloadFailed` message was only used for canvas conversion failures, so a failed write looked like a successful download.

Fix: keep an empty save-dialog result as a normal cancellation, but show the localized failure toast for actual save or download exceptions. The regression assertion now specifically covers the catch path after `writeFile`.

Checks run:

- 7 focused AI/chart test files, 103 tests passed
- `oxfmt --check`
- targeted `oxlint`
- `pnpm typecheck`
- `pnpm build`
- `git diff --check`

t8y2:
Thanks for the contribution! Merged in 29c6d1fbe, will be released in the next version.

PR #8330, opened 2026-09-07
feat(database): introduce dependency provider abstraction

mapan0424:
## Summary

This change introduces the first provider-based foundation for database object dependency inspection.

The sidebar previously had XuguDB-specific dependency behavior embedded directly in the shared tree menu runtime. That approach would make every additional database integration expand the shared UI conditionals and increase the risk of changing existing database behavior. This PR separates the shared interaction flow from database-specific catalog logic.

## Design

- Add a small `DatabaseDependencyProvider` contract with `supports()` and `buildQuery()` operations.
- Resolve providers by `DatabaseType` through a central registry.
- Keep the sidebar responsible only for capability checks, query-tab creation, and result execution.
- Keep object type mappings, catalog SQL, permission boundaries, and database scoping inside each provider.
- Register XuguDB as the first provider using its `ALL_OBJECTS`, `ALL_DEPENDS`, and `ALL_SCHEMAS` dictionaries.
- Preserve the existing behavior of all other databases until they explicitly register a provider.
- Exclude XuguDB package members and package bodies because the database does not expose independent dependency records for those nodes.

This gives future PostgreSQL, Oracle, Dameng, and other integrations a stable extension point: each database can implement its own dictionary query without duplicating or modifying the shared sidebar workflow.

## Scope and compatibility

The provider registry currently contains only XuguDB. As a result, no dependency menu or dependency SQL is added for other database types. No shared metadata query, driver implementation, or existing database-specific behavior is changed.

## Validation

- Focused dependency/provider tests: 16 passed.
- Formatting check: passed.
- Oxlint: passed; only pre-existing repository warnings remain.
- TypeScript typecheck: passed.
- Frontend production build: passed.
- Full repository check: connection-types, formatting, lint, and typecheck passed. The unrelated docs export smoke test timed out while compiling its Rust fixture; no failure originated from the changed files.

### DIFF
--- a/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
+++ b/apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue
@@ -85,6 +85,7 @@ import { loadSidebarObjectGroup } from "@/lib/sidebar/sidebarObjectGroupRouting"
 import { isXuguTypeMemberContainer } from "@/lib/sidebar/xuguTypeMembers";
 import { isXuguSyntheticTreeNode } from "@/lib/sidebar/xuguPublicSynonyms";
 import { buildXuguSchedulerJobSql, type XuguSchedulerJobAction } from "@/lib/database/xuguSchedulerJobSql";
+import { canViewDatabaseObjectDependencies, databaseDependencyProviderFor } from "@/lib/database/databaseObjectDependencies";
 import { elasticsearchClearIndexPreview, isElasticsearchClearConfirmed, isElasticsearchIndexPattern } from "@/lib/sidebar/elasticsearchIndexActions";
 import { mysqlObjectTemplateForGroup } from "@/lib/sidebar/mysqlObjectTemplates";
 import { buildTableDeleteTemplate, buildTableInsertTemplate, buildTableSelectTemplate, buildTableUpdateTemplate } from "@/lib/table/tableSqlTemplates";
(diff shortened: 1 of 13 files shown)


t8y2 (commented):
Maintainer patch applied in `59d660135`.

Root cause: the Xugu dependency query joined `ALL_DEPENDS.OWNER_ID1/2` to `ALL_OBJECTS.USER_ID`, although those fields identify schemas, and the two dependency directions were labeled in reverse.

The Xugu catalog documentation defines `OWNER_ID1` as “被依赖对象模式ID” and `OWNER_ID2` as “依赖对象模式ID”. `ALL_OBJECTS` separately defines `USER_ID` as “拥有者的用户ID” and `SCHEMA_ID` as “模式ID”:

- https://docs.xugudb.com/content/reference/system-view/all/all_depends
- https://docs.xugudb.com/content/reference/system-view/all/all_objects

Fix: join owner IDs through `SCHEMA_ID`, match the target on side 2 for `DEPENDS_ON` and side 1 for `REFERENCED_BY`, and include object-type predicates in both the target and returned-object joins.

Checks run:

- 2 focused dependency-provider test files, 16 tests passed
- `oxfmt --check`
- targeted `oxlint`
- `pnpm typecheck`
- `pnpm build`
- `git diff --check`

t8y2:
Thanks for the contribution! Merged in c0c97ef0e, will be released in the next version.

PR #8345, opened 2026-09-07
fix(sqlite): reject ambiguous keyless updates on duplicate rows

### LINKED ISSUE #8321: [Bug] 没有主键时，日期重复时会自动修改数据
### 数据库类型和版本

sqlite

### 问题描述

1、表格字段：日期、时段+很多列数据指标。每天会统计早上和中午两个时段的数据。
在末尾增加两行，首列填写为相同日期2026-09-07，其他字段为空，点击保存后。之后再在任一行输入数据（譬如，时段列分别输入早上、中午），或者从他处复制两行完整数据粘贴到新增的行中（两行数据不是完全一致的），都会无法有效存储数据，无论如何两行数据自动绑定一致，修改一行，另一行也变动，两行数据变动完全一致。
如果新增两行时，先输入两行，日期为2026-09-08，时段分别设置早上、中午。 
而且回查之前历史数据，譬如2026-08-06就出现两行完全一致的纪录，现在要想修改回正常的早上、中午两个时段的数据，也完法正常改过来。
2、对比sqlite expert, dbeaver都不会出现这种情况。就是先只写日期的数据，保存后，其他字段继续输入数据，都能各自正确的保存。

### 支持信息

DBX 版本: v0.6.5
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

q396921921:
## Summary

A table with no primary key identifies a row for UPDATE/DELETE by matching every column's current value (the "keyless row predicate"). When another loaded row happens to have identical values in all of those columns — e.g. after inserting several rows and only filling in one column so far — that WHERE clause matches every duplicate row instead of just the one being edited, and the change is silently applied to all of them.

This affects every database that uses this keyless matching strategy (SQLite, MySQL, PostgreSQL, Oracle without a usable ROWID, etc.), since they all build the WHERE clause the same way in `build_row_where`.

This PR adds a validation check that refuses the save with a clear error instead of applying an ambiguous update/delete:

> Cannot safely update or delete this row: the table has no primary key and another row currently has identical values in every column, so the change can't be targeted at just one row. Add a primary key or unique index, or make the rows distinguishable, before editing.

Fixes #8321

## Repro

On upstream/main, in a SQLite table with no primary key:
1. Add two rows, only filling in `stat_date` (same value on both), leave everything else empty, save.
2. Edit only one row's `period` column and save.
3. **Before this fix**: both rows end up with the same `period` value — the edit silently applied to both.
4. **After this fix**: the save is rejected with the error above, and neither row is modified.

## Test plan

- [x] Added `rejects_sqlite_keyless_update_when_another_row_is_identical` — reproduces the exact issue scenario and asserts the save is rejected.
- [x] Added `allows_sqlite_keyless_update_when_rows_are_distinguishable` — asserts normal (non-ambiguous) keyless edits still work and produce the same SQL as before.
- [x] `cargo test -p dbx-core --lib -- data_grid` — 210 passed, 0 failed (all pre-existing data-grid tests plus the two new ones).
- [x] `cargo clippy -p dbx-core --lib` — clean.
- [x] `cargo fmt --check -p dbx-core` — clean.
- [x] Manually reproduced the bug and the fix in a running `dbx-web` + SQLite instance: before the fix, editing one of two duplicate rows corrupted both; after the fix, the save is safely rejected. A distinguishable-rows edit (not ambiguous) still saves normally and only touches the intended row.

### DIFF
--- a/apps/desktop/src/composables/__tests__/useDataGridEditor.spec.ts
+++ b/apps/desktop/src/composables/__tests__/useDataGridEditor.spec.ts
@@ -12,6 +12,7 @@ const mocks = vi.hoisted(() => ({
   cancelConditionalUpdate: vi.fn(),
   executeInTransaction: vi.fn(),
   executeInManualTransaction: vi.fn(),
+  executeQuery: vi.fn(),
   addHistory: vi.fn(),
 }));
 
@@ -22,6 +23,7 @@ vi.mock("@/lib/backend/api", () => ({
   cancelConditionalUpdate: mocks.cancelConditionalUpdate,
   executeInTransaction: mocks.executeInTransaction,
   executeInManualTransaction: mocks.executeInManualTransaction,
+  executeQuery: mocks.executeQuery,
   unlockConnectionWrites: vi.fn(),
   lockConnectionWrites: vi.fn(),
   connectionWriteUnlockState: vi.fn().mockResolvedValue(0),
@@ -621,6 +623,7 @@ describe("useDataGridEditor saveChanges reload", () => {
     mocks.cancelConditionalUpdate.mockReset();
     mocks.executeInTransaction.mockReset();
     mocks.executeInManualTransaction.mockReset();
+    mocks.executeQuery.mockReset();
     mocks.addHistory.mockReset();
     mocks.getConfig.mockReset();
   });
@@ -635,6 +638,9 @@ describe("useDataGridEditor saveChanges reload", () => {
       manualTransactionSessionId?: string;
       refreshSavedRows?: ReturnType<typeof vi.fn>;
       onManualTransactionMutation?: ReturnType<typeof vi.fn>;
+      connectionId?: string;
+      primaryKeys?: string[];
+      onExecuteSql?: (sql: string) => Promise<void>;
     } = {},
   ) {
     const emit = vi.fn();
@@ -651,7 +657,7 @@ describe("useDataGridEditor saveChanges reload", () => {
       result: computed(() => result.value),
       editable: computed(() => true),
       databaseType: computed(() => "mysql"),
-      connectionId: computed(() => "connection-1"),
+      connectionId: computed(() => ("connectionId" in options ? options.connectionId : "connection-1")),
       database: computed(() => "app"),
       tableMeta: computed(() => ({
         tableName: "orders_test",
@@ -659,11 +665,11 @@ describe("useDataGridEditor saveChanges reload", () => {
           { name: "id", data_type: "int" },
           { name: "status", data_type: "varchar" },
         ],
-        primaryKeys: ["id"],
+        primaryKeys: options.primaryKeys ?? ["id"],
       })),
       sourceColumns: computed(() => undefined),
       joinedWriteTargets: computed(() => options.joinedWriteTargets),
-      onExecuteSql: computed(() => undefined),
+      onExecuteSql: computed(() => options.onExecuteSql),
       customSaveHandler: computed(() => options.customSaveHandler),
       manualTransactionSessionId: computed(() => options.manualTransactionSessionId),
       onManualTransactionMutation: options.onManualTransactionMutation,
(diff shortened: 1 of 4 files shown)


q396921921:
## Before / after evidence

**Before (bug reproduced on upstream/main)** — two rows with only `stat_date` filled in, edited only row 1's `period`, both rows ended up changed:

before

**Root cause** — the "Preview SQL" panel shows the generated `UPDATE` matching on every unedited column's original value, which is ambiguous when another row shares them:

sql

**After (this fix)** — the same steps now show a clear error and neither row is touched:

after

**Zero regression** — once rows are distinguishable, a normal keyless edit still saves correctly and only affects the intended row:

regression

t8y2 (changes requested):
Request changes: the current guard still cannot guarantee that a keyless edit targets one physical row.

1. `crates/dbx-core/src/data_grid_sql.rs:1262` searches only `options.rows`, which is the currently loaded/filtered page. If an identical row exists outside that page, validation succeeds but the generated all-column `WHERE` can still match and update/delete multiple rows.

2. `compared_indexes` at `crates/dbx-core/src/data_grid_sql.rs:1238` includes result columns whose `source_columns` entry is `None`, while `build_row_where` omits those columns. The validation can therefore decide a row is unique using a value that is not present in the actual mutation predicate.

Please base the safety decision on the exact predicate sent to the database and use a reliable row identifier or server-side uniqueness check. If neither is available, keyless writes should remain disabled rather than relying on the visible page.

q396921921:
Thanks — both points were correct. The page-based check is gone; here is what replaced it.

### The safety decision is now the exact predicate

The guard is no longer computed separately. `build_data_grid_save_statements` derives it from the same `where_clause` value the `UPDATE`/`DELETE` carries, at the point that value is produced, so it is literally the same string. The guard cannot use a different column set or different values than the mutation — this closes point 2 structurally instead of keeping two computations in sync. A unit test asserts the correspondence in both directions: every guard's predicate appears in a statement, and every statement predicate is covered by a guard.

### Uniqueness is decided by the server, not by the visible page

`DataGridSavePreparation` gained `keylessGuards`, one entry per distinct keyless predicate:

```sql
SELECT COUNT(*) AS dbx_keyless_row_matches FROM "main"."daily_stats" WHERE ("stat_date" = '2026-09-08' AND "period" = 'morning' AND "visits" = 10)
UPDATE "main"."daily_stats" SET "period" = '上午'                     WHERE  "stat_date" = '2026-09-08' AND "period" = 'morning' AND "visits" = 10;
```

The editor runs the guards before executing any statement and refuses the save when a count exceeds 1. The count is taken against the whole table, so a duplicate outside the loaded or filtered page is now caught — point 1.

### If neither is available, keyless writes stay disabled

- The predicate would be empty (every result column is expression-only, so `build_row_where` drops all of them): refused during validation. There is no row identifier and nothing a server-side check could count.
- The guard cannot be executed (no backend save target, or the count query itself fails): refused. An unverified keyless write is exactly what the guard exists to prevent.

I also removed the loaded-page comparison entirely rather than keeping it as a cheap pre-check. In a joined result the same physical row can appear more than once, so "two visible rows share a predicate" does not prove two physical rows exist, and keeping it would have rejected safe edits. The server count is the single source of truth; the cost is one `COUNT` per distinct predicate per save.

### Verified against #8321 on SQLite

**1. Two rows identical in every column, editing one** — refused, no `UPDATE` issued, table unchanged.

duplicate refused

**2. Same table with page size 1, so the duplicate is off-page** — still refused. This is the case the previous guard let through.

off-page duplicate refused

**3. A distinguishable row** — saves normally, and only that row changes (rows 1, 2 and 4 untouched).

distinguishable row saved

Rebased on latest main. `cargo test -p dbx-core --lib -- data_grid` (218), `cargo clippy`, `cargo fmt --check` and the full `vitest` suite (13198) all pass.

t8y2 (approved):
Thanks — both blocking points from the previous review are addressed: the guard is derived from the exact `where_clause` its statement carries, counted on the server via `SELECT COUNT(*)` (out-of-page duplicates now caught, test included), and the save fails closed on ambiguous, uncountable, and empty predicates. Tables with a primary key keep PK predicates with zero extra round trips, and the merge with the recently landed Xugu synthetic-ROWID work is clean (synthetic keys make `primary_keys` non-empty, so guards never activate there). Verifying now; three non-blocking notes for possible follow-up:

1. The guard COUNT and the mutation run in separate statements, so a concurrent writer inserting a duplicate between the two can still slip through — inherent to non-transactional keyless edits; a comment acknowledging the window would be welcome.
2. `verifyKeylessGuards` awaits one COUNT per distinct predicate serially; bulk edits on keyless tables pay N sequential round trips.
3. Pre-existing mismatch surfaced by this work: `spanner` and `argo` are `keylessRowPredicate: true` in the frontend capabilities but absent from `uses_keyless_row_predicate` in `data_grid_sql.rs` — worth a follow-up aligning the two lists.

t8y2:
Thanks for the contribution! Merged in 5814ff008, will be released in the next version.

PR #8377, opened 2026-09-07
fix(xugu): enable safe row editing without primary keys

### LINKED ISSUE #7102: [Bug] xugu 无法修改数据
### 数据库类型和版本

xugu server 11

### 问题描述

点开查询的表，双击单元格，无法编辑，在dbeaver可编辑
另外一个连接xugu12可以
猜测是否是权限问题？

### 支持信息

dbx:0.5.92
桌面版
manjaro
linux 

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

_No response_

mapan0424:
## Summary

Fixes #7102.

XuguDB 11 can return editable table rows without exposing a declared primary key. DBX previously treated those results as non-editable because the generated result set did not contain a stable key recognized by the desktop editor. This made double-click editing fail even though the same table was editable in DBeaver and the database itself supports native `ROWID` addressing.

## Design

- Use XuguDB's native `ROWID` as an internal synthetic key only for Xugu tables.
- Project it as `ROWID AS "__DBX_ROWID"` and keep it hidden from the user-facing grid.
- Generate Xugu update/delete predicates with bare `ROWID`; Xugu rejects qualified `table.ROWID` and `ROWIDTOCHAR(ROWID)`.
- Include the hidden key during cold metadata loads, including partitioned tables, so the first open is editable instead of requiring a second refresh.
- Keep views read-only because Xugu does not support `ROWID` on views.
- Exclude the synthetic key from exports and user-visible query columns.
- Preserve existing behavior for Oracle, PostgreSQL, MySQL, and all other database dialects.

## Verification

- Xugu table without a primary key: load, edit, submit, and refresh verified.
- First-level range partition table: cold-load UI path, edit, submit, and persistence verified.
- Two-level range/hash partition table: load and ROWID update verified.
- Session temporary table: ROWID projection, insert, and update verified.
- View: ROWID rejection confirmed and read-only behavior preserved.
- Frontend tests: 24/24 passed.
- Xugu Rust regression tests: 9/9 passed.
- `pnpm typecheck` passed.
- Frontend and Rust formatting checks passed.
- Packaged Dev application UI test passed with the generated app bundle.

The change is intentionally scoped to Xugu capabilities and does not alter other database dialects.


mapan0424:
Superseded by #8379. This replacement is based directly on the latest upstream main and contains only the Xugu row-editing fix.

PR #8386, opened 2026-09-07
fix(grid): keep new top row visible after a toolbar refresh

### LINKED ISSUE #8339: [Bug] 新增数据之后刷新需要向上滚动才能看到新数据
### 数据库类型和版本

MySQL 8.0

### 问题描述

1、查看数据表数据，倒序排列
2、接口新增数据
3、点击刷新按钮，数据行定位还是原来的位置，查看新数据需要向上滚动鼠标才能看到新数据

### 支持信息

DBX版本：v0.6.5
运行环境：桌面端
操作系统：windows10

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

_No response_

q396921921:
## Summary
- Toolbar refresh anchors the row sitting at `scrollTop 0` back to the same on-screen offset after re-fetching. When a row is inserted externally while the grid is sorted descending, the new row now sorts above that anchored row and gets pushed above the visible viewport instead of being shown — matching the reported "need to scroll up to see the new row" behavior.
- Skip capturing a viewport anchor when the grid is already scrolled to the top, so a manual refresh lands on the freshest first row instead of re-pinning the old top row in place.

Fixes #8339

## Test plan
- [x] Reproduced on a real MySQL 8.4 instance: sorted a table DESC, inserted a row via an external client (not through the DBX UI), clicked the toolbar Refresh button — the new row was pushed above the visible viewport, requiring a manual scroll up (matches the issue).
- [x] Applied the fix and repeated the exact same steps — the new row now appears at the top immediately after refresh, no scrolling needed.
- [x] Regression-checked the "already scrolled into the middle of the grid" refresh case against the pre-fix baseline with the same script — behavior is identical before/after this change (not affected by this fix).
- [x] `apps/desktop/src/components/grid` test suite: 30 files / 240 tests passing, including a new regression test for this fix.
- [x] `oxlint` clean on the changed files.

### DIFF
--- a/apps/desktop/src/components/grid/DataGrid.vue
+++ b/apps/desktop/src/components/grid/DataGrid.vue
@@ -5997,6 +5997,11 @@ function captureViewportAnchorForRefresh(): { row?: PersistedDataGridSelection; 
   if (showTranspose.value || displayItems.value.length === 0) return null;
   const scroller = useCanvasGridRows.value ? canvasScrollerElement() : gridScrollerElement();
   if (!scroller) return null;
+  // Already viewing the top of the grid: don't anchor to the row currently
+  // there. Anchoring would re-pin that row to the same on-screen offset even
+  // after a refresh prepends new rows above it (e.g. newly inserted rows
+  // under a DESC sort), pushing the new rows above the visible area (#8339).
+  if (scroller.scrollTop <= 0) return null;
   const rowHeight = useCanvasGridRows.value ? CANVAS_DATA_GRID_ROW_HEIGHT : DOM_DATA_GRID_ROW_HEIGHT;
   const fallbackDisplayIndex = Math.max(0, Math.min(displayItems.value.length - 1, Math.floor(scroller.scrollTop / rowHeight)));
   const item = displayItems.value[fallbackDisplayIndex];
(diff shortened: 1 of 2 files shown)


q396921921:
## Repro & fix evidence

Steps: connect to a MySQL table, sort by `id` DESC, insert a row externally (e.g. via another client — not through the DBX UI), click the toolbar **Refresh** button.

### Before fix — bug reproduced
Before refresh (sorted DESC, 60 rows, top row is `id=60`):

before refresh

After inserting a new row and clicking Refresh (footer shows 61 rows, but the visible top row is still `id=60` at row #2 — the new row is scrolled out of view above, matching the reported behavior):

after refresh, bug present

### After fix — same steps, bug is gone
Before refresh:

fix before refresh

After inserting a new row and clicking Refresh, the new row now shows immediately at the top (row #1) — no scrolling needed:

fix after refresh

t8y2 (approved):
Thanks for the clean, well-scoped fix — the `scrollTop == 0` guard composes correctly with the `inPlaceRefreshPending` watcher, and I traced the shared `prepareFullReload` callers (auto-refresh, rollback, row-save reload): all benefit and none regress. Two non-blocking notes for a possible follow-up: the exact-zero check could use a small tolerance (`< 1`) for subpixel scroll values, and the new spec is a source-text assertion following this file's existing convention — a behavioral capture/restore test would be a nice addition someday.

t8y2:
Thanks for the contribution! Merged in e3b9bae67, will be released in the next version.

PR #8432, opened 2026-09-08
feat(object-browser): preserve object list on return via stale-cache scaffold refresh

Abeautifulsnow:
## Summary

Returning to the Object Browser list after opening a table (i.e. switching tabs) reloaded the list from empty, causing a visible flash even though the search query, filter, and viewport were already restored by #8236 / #8305. This PR removes that flash by showing the last-known object rows instantly on return and revalidating them in the background.

## Root cause

The Object Browser component is **unmounted on every tab switch** — `ContentArea` renders only the active tab and nothing in the repo wraps tabs in `<KeepAlive>`. So on return it remounts with `rows.value = []`, and the immediate watch re-runs `loadObjects({ allowCached: true })`.

The rows cache is a bounded `MetadataResultCache` with a 30s TTL (`maxEntries: 24`, LRU). `getCachedObjectBrowserRows()` read it **without** `allowStale`, so:

- return within 30s → cache hit → rows shown instantly (no flash)
- return after 30s (the common case, since you were browsing the table) → `get()` refuses the stale entry (`if (stale && !options?.allowStale) return undefined`) → `rows.value = []` + full-area spinner → **flash**

## Design (three-state read)

The fix reuses the existing bounded cache and adds an `allowStale` scaffold read — no `<KeepAlive>`, no TTL change, no cache expansion (capacity stays `maxEntries: 24`):

- **fresh** (< 30s) → apply rows → return, no DB request (unchanged semantics)
- **stale** (> 30s) → apply rows as scaffold → non-blocking background revalidate → in-place replace
- **miss** (first load / DDL-invalidated) → full loading (unchanged)

A `refreshingObjects` flag drives a non-blocking spinner instead of blanking the list. Both `loadingObjects`/`refreshingObjects` are reset at `loadObjects()` entry so a superseded request's `finally()` can't leave the icon spinning forever. On a failed background revalidate, the visible rows are kept and a lightweight `scaffoldRefreshError` banner is shown instead of a full-area error.

**Behavior note (per scope)**: the first-ever load of a given connection/database/schema still shows a normal loading state (no prior rows exist) — that is expected, not the reported "return" flash. Every return after that — fresh or stale — shows the list instantly, no blank flash. The scaffold is bounded by `maxEntries: 24` (LRU), so heavy churn across 24+ scopes in one session may evict an entry and fall back to a load.

## Changes

- `apps/desktop/src/lib/table/objectBrowserRowsCache.ts` — add `getCachedObjectBrowserRowsForScaffold()` (`allowStale: true`, returns `{ rows, stale, ageMs, cachedAt }` hit) + `ObjectBrowserRowsCacheHit`.
- `apps/desktop/src/components/objects/ObjectBrowser.vue` — three-state `loadObjects`, `refreshingObjects` + entry reset, `scaffoldRefresh`/`scaffoldRefreshError` on revalidate failure, `preserveExistingRows` threading for same-instance refresh.
- Tests: `objectBrowserRowsCache.spec.ts` (fresh/stale/miss + invalidate), `ObjectBrowserScaffoldRefresh.spec.ts` (supersede race + scaffold-preserving error), updated contract tests in `ObjectBrowserMetadataRefresh.spec.ts` / `contentAreaObjectBrowserCatalog.spec.ts`.

## Testing

- Affected suites pass (`apps/desktop/src` object browser area: 56 files / 467 tests); `packages/app-tests` (3292 tests) pass.
- `pnpm typecheck` exit 0; `oxlint` clean.
- Full `vitest` suite: 3 existing failures (`legacyWebViewRegexCompat` lookbehind assertion, `TableImportDialog` zh-CN locale text, `exportSmoke` `sqlite3.lib` link error) — confirmed present on the pre-change baseline, local-environment artifacts unrelated to this PR.

## Out of scope

- No `<KeepAlive>` state preservation, no TTL increase, no long-term cache expansion.
- Full loading retained for first load and for return after a DDL mutation (cache invalidated) — correct behavior.

## Related Issue
Related #8390

### DIFF
--- a/apps/desktop/src/components/objects/ObjectBrowser.vue
+++ b/apps/desktop/src/components/objects/ObjectBrowser.vue
@@ -154,7 +154,7 @@ import { createSidePanelRequestGuard } from "@/lib/table/sidePanelRequestGuard";
 import { runBatchTableTruncate } from "@/lib/table/batchTableTruncate";
 import { tableColumnDefaultDisplayValue } from "@/lib/table/tableColumnDefaultPresentation";
 import { gaussdbMTypeDisplayName } from "@/lib/table/postgresDataTypeHelp";
-import { cacheObjectBrowserRows, createObjectBrowserRowsCacheWriteToken, getCachedObjectBrowserRows, type ObjectBrowserRowsCacheScope, type ObjectBrowserRowsCacheWriteToken } from "@/lib/table/objectBrowserRowsCache";
+import { cacheObjectBrowserRows, createObjectBrowserRowsCacheWriteToken, getCachedObjectBrowserRowsForScaffold, type ObjectBrowserRowsCacheScope, type ObjectBrowserRowsCacheWriteToken } from "@/lib/table/objectBrowserRowsCache";
 import { createObjectBrowserRowsLoadGuard, type ObjectBrowserRowsLoadHandle } from "@/lib/table/objectBrowserRowsLoadGuard";
 import { loadObjectDdl, type ObjectDdlRequest } from "@/lib/metadata/objectDdlCache";
 import { loadObjectMetadataFacet } from "@/lib/metadata/objectMetadataCache";
@@ -210,6 +210,7 @@ const sortKey = ref<ObjectBrowserSortKey>("name");
 const sortDirection = ref<ObjectBrowserSortDirection>("asc");
 const loadingSchemas = ref(false);
 const loadingObjects = ref(false);
+const refreshingObjects = ref(false);
 const sourceLoading = ref(false);
 const sourceContent = ref("");
 const sourceError = ref("");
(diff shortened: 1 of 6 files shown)


t8y2 (approved):
Approved — this lands exactly the shape #8390 asked for: stale-cache scaffold read on the existing bounded LRU (no TTL/capacity change, no KeepAlive), three-state load, non-blocking revalidate, and in-place replace. The entry-point flag reset plus the `isCurrent`-gated `finally` handle the supersede race correctly, and DDL-invalidated scopes properly fall back to a full load. Non-blocking nits noted for later: the revalidate-failure banner could use `role="alert"` plus a dismiss affordance, and the toolbar refresh button stays enabled during background revalidates (safe today via the epoch guard). Thanks!

t8y2:
Thanks for the contribution! Merged in 8642e1a59, will be released in the next version.

PR #8391, opened 2026-09-07
feat(ai): add sandboxed HTML preview for AI rich content

Abeautifulsnow:
## Problem

  AI-generated fenced `html` blocks were rendered only as code, so users could not preview generated HTML safely inside DBX.

  Related issue: #6467.

> This PR implements one independently usable stage of the broader issue and can be adopted and reviewed on its own; the remaining stages are intentionally out of scope.

  ## Root cause

  The AI message renderer supported plain text, code blocks, and `chart-json`, but had no HTML rich-content handler or isolated preview surface.

  ## Fix

  - Added an `html` rich-content segment and handler.
  - Added validation for closed, non-empty HTML blocks up to 512 KB.
  - Wrapped preview content with a restrictive Content Security Policy:
    - `default-src 'none'`
    - `base-uri 'none'`
    - `form-action 'none'`
    - `img-src data: blob:`
  - Rendered HTML inside an empty `sandbox` iframe with no script, form, popup, or same-origin permissions.
  - Added safe actions:
    - Require confirmation before copying raw AI-generated HTML.
    - Save only the CSP-wrapped safe HTML document.
    - Provide an expanded in-app preview dialog.
  - Made the expanded dialog responsive with a maximum width of `min(1100px, 100vw - 32px)`.
  - Lazy-loaded the HTML preview component.
  - Added translations for all supported locales.
  - Added a repeatable Edge runtime canary covering both iframe and standalone-file contexts.

  ## Compatibility

  - SQL, Shell, and regular code blocks retain their existing rendering, proposal, confirmation, and execution flows.
  - Unclosed HTML fences remain regular code blocks during streaming.
  - Empty or oversized HTML blocks fall back to regular code blocks.
  - Copying still copies the raw HTML source, but the first copy requires explicit confirmation.
  - No external-browser opening path is introduced in this PR.

  ## Validation

  - `pnpm exec vitest run apps/desktop/src/components/ai/rich/__tests__/AiHtmlPreview.spec.ts apps/desktop/src/lib/ai/richContent/__tests__/aiHtmlPreviewCanary.spec.ts` —
  passed, 10 tests
  - `pnpm exec vitest run apps/desktop/src/components/ai/rich/__tests__/AiHtmlPreview.spec.ts apps/desktop/src/lib/ai/richContent/__tests__/aiHtmlPreviewCanary.spec.ts
  apps/desktop/src/i18n/__tests__/aiRichContentParity.spec.ts` — passed, 19 tests
[...]

PR #8459, opened 2026-09-08
feat(ai): export the conversation as markdown or html

Abeautifulsnow:
## Summary

Adds conversation-level export to the AI assistant: the whole visible transcript can be saved as a **Markdown** or **HTML** report from a single entry point in the conversation header. This is the third and final capability of #6467, alongside the chart rendering (#8308) and the sandboxed HTML preview (#8391). No new dependency is added.

## Related issue

This PR addresses **one part of #6467** only. That issue asks for three capabilities in the AI conversation:

1. ~~AI chart rendering~~ — merged (#8308)
2. ~~AI-generated HTML preview~~ — merged (#8391)
3. Conversation / analysis report export ← **covered by this PR**

The single-message "Export as Markdown" action is kept unchanged; the two exports coexist.

## What it does

- **Two formats, one boundary** — a header dropdown offers Markdown / HTML. The export source is the **visible transcript** (`visibleMessages`), the same messages the UI renders: hidden `contextSummary` entries are excluded (filtered by the caller and again defensively in the builder); reasoning, tool arguments/results, and attachment payloads are never reachable by construction.
- **Markdown report** — title/date header, one `##` section per message with role labels; message content is kept **verbatim**, so ```chart-json / ```html fences survive byte-identical (zero data loss, round-tripable).
- **HTML report** — each message is rendered through the same `formatAiInlineMarkdown` escape path the chat uses, so raw HTML (`<script>`, ``, full documents, `</title>` breakouts) lands **escaped**; fences come out as escaped code blocks. The document shell reuses `AI_HTML_PREVIEW_CSP` (`default-src 'none'` + `base-uri`/`form-action 'none'`), which blocks the one remaining remote-reference path (markdown images/links) in the standalone file. Title and connection name are HTML-escaped.
- **Failure markers that survive reloads and locale switches** — failed foreground turns persist a `failed` flag on the assistant message (set at the same point the localized `requestFailed` prefix is written). The flag rides the full persistence chain: `buildConversationSnapshot`, `persistPendingInputRecovery`, `chatMessagesFromConversation`, and a new serde `failed` field on dbx-core's `AiChatMessage` (`Option<bool>`, `default` + `skip_serializing_if`, backward compatible with existing stored conversations). Background turns failed/interrupted by the run registry are also marked — attributed only to an assistant reply **after the latest user turn**, so a restart that interrupts a turn before its first delta never marks the *preceding successful* reply.
- **Entry point** — a ghost icon button in the conversation header opens the format dropdown ("Export as Markdown" / "Export as HTML", localized, matching the single-message export copy so the items read standalone and never wrap). Exporting an empty conversation shows a toast; failures during save reuse the standard export-failure toast.

## Safety

- The report body is only reachable through escaping paths; the CSP is defense in depth for the standalone file, exactly as with the safe-HTML preview shell.
- The export writes only what the user already sees on screen — no hidden context, no tool internals, no attachments.
- SQL proposal / confirm / execute semantics are untouched: export is a pure string assembly with no SQL path involvement.

## Tests

- Unit (`aiConversationExport.spec.ts`) — empty/whitespace and all-`contextSummary` conversations return null; Markdown verbatim fences + role headings + failure marker; HTML escapes `<script>` / `` / full HTML documents / `</title>` injection in the document head; CSP meta present; sanitized file names; `ai` fallback without a connection name.
- Source contracts (`AiAssistant.conversationExport.spec.ts`) — native save-dialog short extensions (`md`/`html`, never the bare format key); failure recorded on message state at the prefix write site; `failed` persisted through **both** snapshot paths; marker derived from persisted state + run status, **not** locale text; turn attribution restricted to after the latest user turn (restart-before-first-delta regression).
- Existing contracts updated in lockstep (`AiAssistant.clearAndCancel.spec.ts`, `aiMessageLayout.test.ts`) — the guarded failure write now also asserts `msg.failed = true`.
- i18n — all 9 locales (en / es / it / ja / ko / pt-BR / tr / zh-CN / zh-TW) expose the 7 new `ai.*` keys; key-parity suite extended.
- Focused suites 182/182 green; full suite 1343/1346 (3 failures pre-existing/flaky, zero file overlap with this diff); `vue-tsc`, `oxlint`, `oxfmt` clean; `cargo check -p dbx-core` exit 0 (test fixture updated for the new serde field).

## Notes

- No new dependencies.
- V1 scope decision (per the task PRD): chart-json / html fences stay escaped code blocks inside the HTML report. Embedding real chart snapshots (offscreen re-render, size budgets, fence round-trip semantics) is deliberate V2 follow-up material.


t8y2 (approved):
Reviewed — this looks solid and ready to merge. Notes from triage:

- Verified the HTML escaping story end to end: raw HTML is escaped by the shared renderer, `javascript:` links are dropped by the protocol whitelist, marked 18 percent-encodes quote/space in image srcs (no attribute injection), and the report CSP blocks inline handlers and remote fetches in the standalone file. The one survivor, ``, is inert in modern browsers and blocked by `img-src data: blob:` — and it's identical to the existing chat rendering path, so no regression here.
- The persistence chain for `failed` is complete (both snapshot whitelists + restore + serde field), and the run-status attribution correctly avoids marking the preceding successful reply on restart-before-first-delta. Old conversations pre-dating the flag simply show the baked-in failure text without the new marker — fine.
- Cosmetic only: newly failed replies will show both the persisted `ai.requestFailed` prefix text and the `**failed to generate**` marker. If that reads redundant, the marker could be skipped when the prefix is present, but I don't consider it blocking.

Thanks for the clean extraction (`lib/export` builder, maximal reuse of `saveTextFile`/CSP/escape helpers) and the thorough test coverage (unit + source contracts + 9-locale parity). No blockers from me.

t8y2:
Thanks for the contribution! Merged in 2b2d1f0d9, will be released in the next version.

PR #8536, opened 2026-09-09
feat(i18n): add Azerbaijani locale

jamalkamaladdin:
This PR adds an Azerbaijani (az) translation.

New file `apps/desktop/src/i18n/locales/az.ts`: 9304 strings, all filled in, the same keys in the same order as `en.ts`, wrapped in `withEnglishFallback` like every other non-English locale.

New file `apps/desktop/src/i18n/locales/docs/az.ts`: 55 strings, the same keys as `docs/en.ts`.

`apps/desktop/src/i18n/locales/meilisearchManagement.ts`: added the `az` block and the `meilisearchManagementAz` export. It spreads the English base and overrides the same 25 keys that the `es`, `it`, `ko`, `pt-BR` and `tr` blocks override.

`apps/desktop/src/i18n/index.ts`: added `az` to the `Locale` union, to `supportedLocales`, to `localeLoaders`, and to `localeFromLanguageTag` for `az` and `az-*`.

`apps/desktop/src/lib/app/localeOptions.ts`: added the `Azərbaycan` entry after English, so `LOCALE_OPTIONS[0]` is still the English entry that `EditorSettingsDialog` falls back to.

`apps/desktop/src/components/meilisearch/MeilisearchIndexView.vue`: added the dayjs `az` locale import and its `DAYJS_LOCALES` entry.

`src-tauri/src/lib.rs`: added `LocaleFamily::Azerbaijani` with the tray labels, the app menu quit label and the Copy Support Info label, plus the matching unit test assertions.

Added `az` to the 20 locale enumerating test files: agent offline export, AI rich content, backend errors, cached result, Consul UI, custom types, docs namespace, execution namespace, SQLite rebuild notice, tab context menu, table info pin, user admin host, AI assistant write confirmation, editor settings preview position, settings search, ZooKeeper connection dialog, and the four `packages/app-tests` locale maps.

`crates/dbx-core/assets/docs-export.js` and `crates/dbx-core/assets/docs-export.manifest.json`: refreshed for the `localeOptions.ts` change. `EXPORT_LOCALES` is unchanged, so `az` does not appear in the documentation export language picker and exported documents stay on the existing 8 languages.

Product names, SQL keywords, protocol names, file extensions, CLI flags and identifiers are left in English, matching how the other locales handle them. `consulUi.ts` and `sqlServerTraceMessages.ts` stay in English for `az` as they do for every other locale.


t8y2:
Thanks for the contribution! Merged in 3408ddb, will be released in the next version.

PR #8629, opened 2026-09-09
feat(grid): preserve data-grid viewport and selection across tab switches

Abeautifulsnow:
## Problem

  Switching away from a data-result tab unmounts its `DataGrid`. Returning to the tab previously reset the viewport, selected rows/cells, and column selection.

  Conditional queries had an additional issue: asynchronous filter hydration caused the snapshot probe to mismatch, so the saved state was not restored.

  ## Root cause

  - Inactive tabs unmount their `DataGrid` instance.
  - View state was not retained independently of the component instance.
  - Restoring before structured filter hydration completed produced an invalid probe comparison.
  - Holding row arrays in snapshots could prevent inactive result payloads from being evicted.

  ## Fix

  - Added bounded, primitive-only data-grid view snapshots.
  - Preserved scroll position, row selection, column selection, cell/range selection, and focus state.
  - Added result-generation and probe checks to prevent restoring state onto a different result.
  - Added LRU and byte limits to avoid retaining heavyweight result data.
  - Added compact serialization for all-row, contiguous-row, column, range, and sparse selections.
  - Bounded oversized sparse selections while retaining viewport state.
  - Added asynchronous structured-filter hydration gating and retry logic.
  - Excluded transpose mode from snapshot restoration.
  - Added restore notifications for selections dropped due to snapshot limits.

  ## Compatibility

  - Fresh query results receive a new view generation and do not restore stale viewport state.
  - Append and disk-restore paths preserve the existing generation where appropriate.
  - Snapshot data does not retain result-row references, so inactive result eviction remains effective.
  - Existing grid rendering modes and default behavior remain unchanged.

  ## Validation

  - `pnpm exec vitest run apps/desktop/src/components/grid/__tests__/DataGridViewSnapshotRestore.spec.ts apps/desktop/src/lib/__tests__/dataGrid/
  dataGridViewStateCache.spec.ts apps/desktop/src/lib/__tests__/tabs/tabResultCache.viewGeneration.spec.ts apps/desktop/src/stores/__tests__/
  queryStore.resultViewGeneration.spec.ts apps/desktop/src/stores/__tests__/queryStore.resultViewGenerationSource.spec.ts`
    - Passed: 5 test files, 38 tests
  - `pnpm typecheck`
    - Passed
  - `pnpm lint`
    - Passed with existing repository warnings
  - `git diff --check`
    - Passed
    
 ## Related Issue
Related #8533


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 41b5e51dc, will be released in the next version.

PR #8249, opened 2026-09-05
fix(redis): bound implicit key and value scanning

ordinary-s:
## Summary

Fixes #7779

Keep Redis key browsing and value searches bounded on large databases without changing the backend protocol.

Previously, value/all searches could continue to cursor 0 automatically, while automatic main-key loading could resume pending subtrees with additional scan budgets. Lifecycle transitions and stale asynchronous subtree completions could also restart or interfere with loading.

## Changes

- Load one cursor page for an initial value/all search. Load more or a user scroll advances one additional page from the saved cursor; resize and short viewports do not auto-load these searches.
- Cap every value/all scan request at `COUNT 100`, including each interruptible batch of explicit Fetch all.
- Preserve the existing 50-iteration automatic key SCAN budget. Automatic main-key filling does not resume pending subtrees; deliberate Load more/scroll continuation may resume at most one expanded pending subtree.
- Preserve collapsed subtree cursors without background continuation, and use generation/operation ownership to prevent stale catch/finally handlers from affecting newer operations.
- Invalidate requests on KeepAlive deactivation without replenishing the retained operation's automatic budget.
- Block continuation with an old cursor while an Enter-only key pattern has unsubmitted edits, and disable repeated refreshes during loading.

Virtualization, deduplication, cursor continuation, maximum-row limits, failure-stop behavior, and explicit Fetch all remain intact. The diff is limited to the browser component and three regression-test files.

## Validation

Validated commit: `bfef725d01f12f10763182b991edeb965d30ac05`.

### Frontend checks

- Focused Redis browser suite rerun: **114 tests passed across 4 files**.
- TypeScript typecheck, connection-type validation, and Oxfmt/Oxlint for all changed files passed during implementation.
- The earlier full frontend run was **not fully green**: 1,302 test files and 12,930 tests passed, with 1 test skipped; two unrelated host/baseline failures remained:
  - docs-export: native Cargo unavailable on Windows (`spawnSync cargo ENOENT`);
  - legacy WebView compatibility: Windows path separators are not normalized in the test allowlist.

### Live Redis and one-million-key integration

Completed against **Redis 7.4.11** in isolated Docker containers, using the exact committed source and **Rust/Cargo 1.98.0**. The database contained 1,000,000 string keys and one uniquely matching value; fixtures were streamed in bounded batches.

| Check | Result |
| --- | --- |
| DBX HTTP → Rust → Redis, full key scan | 1,000,000 unique keys, cursor 0, 100 requests; **8.84 s** |
[...]

PR #8641, opened 2026-09-09
fix(redis): honor configured value scan page size

ordinary-s:
Follow-up regression fix for #8249, related to #7779.

Value/All searches currently send `COUNT 100` even when the connection's `redis_scan_page_size` is set to 5000. Restore the configured COUNT for initial pages, Load More, scroll continuation, and explicit Fetch All. Keep the `COUNT 50000` Fetch All override exclusive to Key mode.

The existing single-page behavior, empty-page cursor continuation, layout/KeepAlive guards, request ownership, and Fetch All stop/rollback remain intact. No backend, API, or schema changes.

Validation:

- Regression tests cover COUNT 200/1000/2000/5000/10000 in both Value and All, plus idle layout behavior, opaque cursor continuation, KeepAlive, and stopping/resuming an in-flight Fetch All page.
- Final review: 265 Redis component/helper tests across 25 files passed, along with application typecheck, generated connection-type consistency, scoped lint, formatting, and whitespace checks.
- Real Redis 7.4.11: standalone and three-master Cluster million-key fixtures passed full key scans and sparse value traversal through the Rust backend. Nine standalone and seven Cluster component integration checks passed.
- Twelve Chromium GUI checks against Cluster and twelve native Windows Tauri GUI checks against standalone Redis passed, covering all five COUNT settings in both modes, resize inactivity, continuation, dense matches, and Fetch All/Stop.

Larger configured COUNT means more value/metadata reads per page and potentially longer waits. Stop prevents subsequent pages; it does not abort an in-flight backend request. COUNT remains a work hint, not an exact result limit. Native GUI and Cluster were tested separately; the GUI runs are functional checks with short development-build observations, not release performance benchmarks. Cluster failover/resharding was not tested.

### DIFF
--- a/apps/desktop/src/components/redis/RedisKeyBrowser.infiniteScroll.spec.ts
+++ b/apps/desktop/src/components/redis/RedisKeyBrowser.infiniteScroll.spec.ts
@@ -5,6 +5,7 @@ import { createI18n } from "vue-i18n";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { RedisKeyInfo } from "@/lib/backend/api";
 import { defaultRedisKeyGrouping, type RedisKeyGrouping } from "@/lib/redis/redisKeyGrouping";
+import { REDIS_SCAN_PAGE_SIZE_OPTIONS } from "@/lib/redis/redisKeyPattern";
 
 const grouping = ref<RedisKeyGrouping>();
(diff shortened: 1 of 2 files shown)


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 619cd8f89, will be released in the next version.

PR #8694, opened 2026-09-10
fix(mongodb): support zero-argument and numeric shell value constructors

thailoc-dev:
## Problem

MongoDB write commands containing `new Date()` are rejected with the generic unsupported-command hint, which hides the real cause:

```js
db.reports.updateOne(
  { phone_number: "+84905421172", code: "VN" },
  { $set: {
      type: ObjectId("5bbc28701f3fd80f00e0211c"),
      created_at: new Date(),
      updated_at: new Date()
  }},
  { upsert: true }
)
```

> Query Error
> Use MongoDB shell-style commands, for example: db.collection.find({}).limit(100), …

## Cause

Both shell parsers rewrite value constructors into extended JSON with a matcher that only accepts **one quoted-string or integer argument**:

- `packages/mongo-shell/src/json.ts` → `replaceMongoShellConstructors` (desktop editor)
- `crates/dbx-core/src/mongo_shell.rs` → `transform_shell_constructors` (CLI / MCP / web)

`new Date()` matches neither shape, so it survives into the JSON text, `JSON.parse` / `json5` fails, the command parses as `null`, and the caller falls back to the generic hint. The `ObjectId("…")` in the same statement was fine — the two bare `new Date()` calls sank it.

## Change

Both matchers are replaced with a scanner that resolves a call by name and argument shape rather than by a fixed pattern:

| Form | Result |
| --- | --- |
| `new Date()`, `ISODate()` | current time |
| `ObjectId()` | generated id |
| `new Date(1735689600000)` | `{"$date":{"$numberLong":"…"}}` |
| `NumberInt`, `NumberDecimal` | `$numberInt`, `$numberDecimal` |
| `ObjectId("…")`, `ISODate("…")`, `new Date("…")`, `NumberLong` | unchanged |

A bare `Date()` is deliberately left alone, since the shell returns a string rather than a date there. Zero-argument values are filled in client-side, the way `mongosh` does.

This also closes a gap between the two parsers: the Rust side accepted neither `new Date(...)` nor `NumberLong(...)`, though the TypeScript side did, so the same statement behaved differently in the editor and in the CLI/MCP.

## Testing

- 3 new frontend tests, 2 new Rust tests, covering zero-argument, epoch-millisecond, numeric and in-string forms
- `pnpm test` — 13746 passed
- `cargo test -p dbx-core --lib` — 6092 passed
- `pnpm typecheck`, `cargo clippy`, `cargo fmt`, `oxlint` clean
- Verified the emitted extended JSON through the driver's own `json_object_to_document`, which yields `ObjectId`, `DateTime`, `Int32`, `Int64` and `Decimal128`

Two `dbx-core` tests fail identically on `main` before this change and are unrelated to it: `agent_tools::tests::mongodb_agent_registers_shell_query_tool_and_routes_find_one_as_read_only` (stack overflow at the default test stack size; passes with `RUST_MIN_STACK=16777216`) and `query::tests::external_driver_preview_retry_preserves_marker_truncation`.

### DIFF
--- a/crates/dbx-core/src/mongo_shell.rs
+++ b/crates/dbx-core/src/mongo_shell.rs
@@ -1,3 +1,5 @@
+use chrono::{SecondsFormat, Utc};
+use mongodb::bson::oid::ObjectId;
 use serde::Serialize;
 use serde_json::Value;
 
@@ -1045,6 +1047,11 @@ fn is_shell_regex_value_prefix(previous_significant: Option<char>) -> bool {
     previous_significant.is_none_or(|character| matches!(character, ':' | '[' | ',' | '('))
 }
 
+/// Shell value constructors rewritten to extended JSON before json5 parsing.
+/// `Date` is only recognised after `new`, matching the shell where a bare `Date()`
+/// returns a string rather than a date.
+const SHELL_CONSTRUCTORS: [&str; 6] = ["ObjectId", "ISODate", "Date", "NumberLong", "NumberInt", "NumberDecimal"];
+
 fn transform_shell_constructors(input: &str) -> Result<String, String> {
     let mut output = String::with_capacity(input.len());
     let mut index = 0;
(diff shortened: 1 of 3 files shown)


t8y2 (commented):
Maintainer patch pushed in `7a87cece0` (on top of `89e3c21f3`): adds integer range validation to the TypeScript constructor parser for parity with the Rust side (`mongo_shell.rs` already validates `i64` via `value.parse::<i64>()`):

- `NumberLong` args are now bounded to the signed 64-bit range, `NumberInt` to the signed 32-bit range, and `new Date(<ms>)` epoch values to the signed 64-bit range. Out-of-range input is rejected at parse time (same clean rejection as other invalid forms) instead of emitting a `$numberLong`/`$numberInt` string that later fails at the driver/server with a confusing `$`-prefixed-field error.
- Implementation is a small `fitsIntegerBounds` helper using `BigInt` comparison — exact, no float precision loss.
- New regression test pins both overflow rejections (`NumberLong(9223372036854775808)`, `NumberInt(2147483648)`, `new Date(99999999999999999999)`, negative overflow) and the boundary values (max i64 / min i32 still pass through verbatim).

Local verification: `mongoShellCommand.test.ts` 94/94 pass (93 existing + 1 new), oxlint clean, oxfmt clean. CI will run on the new head (workflow approval needed as before).

t8y2 (approved):
Thanks! Verified the head commit locally: all mongoShellCommand vitest cases and mongo_shell cargo tests pass, and I probed the edge cases directly — new Date(ms) correctly maps to epoch milliseconds, ObjectId()/new Date()/ISODate() fill in client-side values, and invalid forms (NumberLong(5.5), NumberDecimal() with no argument, non-numeric strings) are rejected rather than silently passed through, matching mongosh. The maintainer patch above adds the TS-side integer range checks for parity with the Rust parser, with regression tests at both boundaries. CI green on the patched head. Approving.

t8y2:
Thanks for the contribution! Merged in 3e707d777, will be released in the next version.

PR #8711, opened 2026-09-10
fix(structure): respect identifier quote preference

### LINKED ISSUE #8680: [Bug] 关闭双引号后编辑表结构预览中依然带双引号
### 数据库类型和版本

通用

### 问题描述

1、设置-编辑-关闭《生成SQL时为标识符添加引号》；
2、选中一张表右键《编辑表结构》；
3、新增字段abc；
期望结果：sql预览中不带双引号；
实际结果：sql预览中依然带双引号。

### 支持信息

DBX 版本：v0.6.9
运行环境：桌面端
操作系统：window
系统架构：x86

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

Image

Eve-z1:
## 变更说明

- 表结构编辑器生成 DDL 时读取“生成 SQL 时为标识符添加引号”设置
- 关闭后复用现有安全去引号逻辑，保留字符串、保留字、特殊字符和大小写敏感名称所需的引号
- SQLite 普通结构变更同样生效；专用类型重建继续保持与 revision 校验执行计划完全一致

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

- [x] 本 PR 涉及前端行为改动，无布局变化；由组件回归测试覆盖

## 验证

- [ ] `make check` 通过
- [x] `pnpm exec vitest run apps/desktop/src/components/structure/TableStructureEditor.primaryKey.spec.ts apps/desktop/src/lib/sql/__tests__/ddlDisplay.spec.ts --maxWorkers=1`（53 tests）
- [x] `pnpm exec vue-tsc --noEmit --project apps/desktop/tsconfig.json`

## 关联 Issue

Close #8680

### DIFF
--- a/apps/desktop/src/components/structure/TableStructureEditor.charsetCollation.spec.ts
+++ b/apps/desktop/src/components/structure/TableStructureEditor.charsetCollation.spec.ts
@@ -211,7 +211,7 @@ vi.mock("@/stores/queryStore", () => ({ useQueryStore: () => ({ tableStructureRe
 vi.mock("@/stores/historyStore", () => ({ useHistoryStore: () => ({ add: vi.fn() }) }));
 vi.mock("@/stores/settingsStore", () => ({
   useSettingsStore: () => ({
-    editorSettings: { structureEditorDensity: "compact", sqlFormatter: {}, tableColumnTemplateFields: [] },
+    editorSettings: { structureEditorDensity: "compact", sqlFormatter: {}, tableColumnTemplateFields: [], generateSqlQuoteIdentifiers: true },
     updateEditorSettings: mocks.updateEditorSettings,
   }),
 }));
--- a/apps/desktop/src/components/structure/TableStructureEditor.primaryKey.spec.ts
+++ b/apps/desktop/src/components/structure/TableStructureEditor.primaryKey.spec.ts
@@ -14,6 +14,7 @@ const mocks = vi.hoisted(() => ({
   executeQuery: vi.fn(),
   listDataTypes: vi.fn(),
   buildTableStructureChangeSql: vi.fn(),
+  previewSqliteTableStructureChange: vi.fn(),
   buildMysqlAutoIncrementSql: vi.fn(),
   updateEditorSettings: vi.fn(),
   loadObjectDdl: vi.fn(),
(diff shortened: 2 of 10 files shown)


t8y2 (commented):
Maintainer patch pushed in `8f124bdb1` (on top of `fb843e469`), closing the two blockers from review:

1. **CI red (spec mocks)** — the failing specs mocked `editorSettings` without the new key, so the preference read as disabled and quotes were omitted. Added `generateSqlQuoteIdentifiers: true` to the existing mocks in 6 spec files (charsetCollation, concurrentTransition, ddlTab, ddlTabLifecycle, initialTab, ddlSearch). Previously-failing suites now pass (12/12).

2. **Postgres concurrent-index guard bypassed when quoting is off** — `concurrentIndexNamesInStatements` only matched double-quoted names, so after `omitDdlIdentifierQuotes` strips quotes from a safe name, the same-name INVALID pre-apply check no longer fired. Chose to extend the extractor (rather than keeping a parallel raw-statements ref — `pendingStatements` is assigned at 11 sites and the stripped form is also the executed batch, so a shadow ref would be a stale-state hazard): the schema qualifier and index name now accept either quoted or unquoted identifiers, with an optional `IF NOT EXISTS` skip. Strict superset — every previously-matched form captures the identical name (verified against the PG builder output in `table_structure_sql/indexes.rs`). Added two regression tests in `concurrent.spec.ts` pinning dequoted extraction and same-name INVALID detection with the preference off.

Local verification: previously-failing 2 files 12/12; concurrent + primaryKey + ddlDisplay 66 passed; ddlTab + ddlTabLifecycle + initialTab + ddlSearch 14 passed; full `components/structure` + `lib/table` sweep 19 files 135 passed; vue-tsc clean. CI will run on the new head.

t8y2 (approved):
Both blockers are closed by the maintainer patch above and the new head is green. The core change — necessity-based dequoting through the shared `omitDdlIdentifierQuotes` helper with the SQLite rebuild carve-out — was verified correct in review, and the concurrent-index guard now matches both quoted and unquoted names with regression tests. Approving.

t8y2:
Thanks for the contribution! Merged in e17246c0f, will be released in the next version.

PR #8736, opened 2026-09-10
fix(mcp): persist query timeout edits reliably

### LINKED ISSUE #8616: [Question] MCP运行超时配置似乎没有保存按钮
### 数据库类型和版本

无关

### 支持信息

DBX 版本: v0.6.8
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 描述

尝试修改MCP服务器的运行超时配置，但是设置后找不到保存的地方。修改后，其他设置页面的保存按钮未激活无法点击，直接关闭再次打开配置未被保存。是否遗漏该配置的保存按钮？（如下图）
Image

### 环境信息

_No response_

Abeautifulsnow:
## Problem

  Fixes #8616.

  The MCP query-timeout field had no explicit save action. Its previous persistence depended on `change`, so edits could be lost before a reliable commit event occurred.
  The field also gave no positive indication that an automatic save had completed.

  ## Root cause

  `Input.vue` uses VueUse's passive `useVModel`, which emits `update:modelValue` asynchronously. Reading the bound ref during the same native input event can therefore
  observe a stale value.

  The debounce path also consumed its pending value before checking whether another MCP policy write held the shared mutation gate. A `type=number` bad-input intermediate
  state could be mistaken for an intentionally cleared field and persisted as `null`.

  ## Fix

  - Read the current value from the native input event in capture phase while retaining controlled `v-model` synchronization.
  - Debounce query-timeout persistence and flush it on Settings close/unmount.
  - Retain pending values while another MCP policy save is active, then retry after the gate reopens.
  - Ignore `validity.badInput` intermediate states instead of persisting them as inherited timeout values.
  - Add a field-level status indicator: `Saving…`, short-lived `Saved`, or `Save failed`.
  - Add English and Simplified Chinese status translations.

  ## Compatibility

  - Empty valid input still persists `null` and inherits the connection timeout.
  - `0` still means unlimited.
  - Invalid values retain the existing validation toast.
  - Other MCP policy controls keep their existing immediate-save behavior.

  ## Validation

  - `pnpm exec vitest run apps/desktop/src/components/editor/__tests__/EditorSettingsDialog.mcpQueryTimeout.spec.ts apps/desktop/src/components/editor/__tests__/
  EditorSettingsDialog.mcpQueryTimeout.integration.spec.ts apps/desktop/src/components/ui/input/__tests__/Input.spec.ts apps/desktop/src/lib/__tests__/mcp/
  mcpPolicySelection.spec.ts` — passed (43 tests)
  - `pnpm typecheck` — passed
  - `pnpm lint` — passed with existing unrelated warnings
  - `git diff --check` — passed

## Screenshots

42625fbe-9829-41e1-a262-378511c45659


t8y2 (approved):
Thanks — verified and approved. The root cause checks out: @change on a type=number input only commits on blur/Enter, so closing the dialog dropped the edit, and the shared mutation gate had no retry. Your fix (input capture read of the native value, 300 ms debounce, flush on close/unmount, gate retention with finally-retry, badInput guard) addresses it directly. I ran your four suites plus the whole editor __tests__ directory at head: 43/43 targeted and 191/191 total pass, and confirmed the persisted value reaches MCP query execution (dbx-mcp applies policy.query_timeout_secs to query options). Two non-blocking notes: a save completing mid-typing can momentarily clobber one keystroke via the queryTimeoutSecs watcher (self-heals); the "capture phase runs before v-model" comment overstates the ordering guarantee — harmless since you read currentTarget.value. Approving.

t8y2:
Thanks for the contribution! Merged in a9b447094, will be released in the next version.

PR #8760, opened 2026-09-10
fix(mcp): restore custom connection scope after switching to all

### LINKED ISSUE #8740: [Feature] MCP-权限管理-②分组与连接：优化建议
### 数据库类型和版本

通用

### 支持信息

DBX 版本: v0.6.9
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 当前痛点

https://github.com/user-attachments/assets/b36fc6f3-ba7f-4b35-9337-bd4929115181
如视频，我本来自定义范围的，然后点了一下“所有连接”，再切换回自定义范围，之前的范围不见了，又得全部重新选择。

### 期望方案

希望可以保留原来的配置。或者干脆加个“确定”按钮才做切换

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须做

### 补充信息

_No response_

onenewcode:
## 变更说明

修复 MCP 权限管理「分组与连接」中，从自定义范围切到「所有连接」再切回后选择被清空的问题。

切到「所有连接」前会缓存当前自定义分组和连接；切回「自定义范围」时恢复该选择，而不是从空列表重选。保存中禁用模式切换，避免连点被丢弃。

## 变更类型

- [ ] 新功能
- [x] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

- [] 本 PR 涉及前端改动，已附截图/录屏（见下方）

## 验证

- [x] `make check` 通过
- [x] 相关测试通过

## 关联 Issue

Close #8740

### DIFF
--- a/apps/desktop/src/components/settings/McpResourceScopePicker.vue
+++ b/apps/desktop/src/components/settings/McpResourceScopePicker.vue
@@ -53,6 +53,7 @@ const { t } = useI18n();
 const search = ref("");
 const expandedGroupIds = ref(new Set<string>());
 const initializedExpansion = ref(false);
+const lastCustomScope = ref<{ allowedGroupIds: string[]; allowedConnectionIds: string[] } | null>(null);
 
 const scopeMode = computed<ScopeMode>(() => (props.allowedConnectionIds === null ? "all" : "custom"));
 const connectionById = computed(() => new Map(props.connections.map((connection) => [connection.id, connection])));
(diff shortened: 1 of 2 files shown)


t8y2 (commented):
Maintainer patch pushed in `${NEW:0:9}` (on top of `a462294918`): restored the pre-switch confirmation guard that this PR had removed, keeping the new snapshot/restore feature intact.

Rationale: the snapshot makes accidental switches reversible within one settings-page session, but the cleared custom rules are still persisted permanently once the page closes — the confirm (using the existing `settings.mcpResourceScopeAllConfirm` key, live again in all locales) warns about exactly that cross-session loss. Wiring: the guard is the first statement of the `mode === "all"` branch with the same non-empty-scope condition origin/main used; on cancel nothing is snapshotted or emitted, on confirm the snapshot is taken and only then the all-mode payload is emitted. The all→custom restore path never confirms.

Spec updated: `confirm` stubbed via `vi.stubGlobal` (happy-dom here has no `window.confirm`), the round-trip test now asserts confirm-then-restore, a new test pins cancel-keeps-custom-and-emits-nothing, and the empty-scope case asserts no confirm. 7/7 pass locally; oxlint/oxfmt clean. CI will run on the new head.

t8y2 (approved):
Thanks — root cause confirmed: switching to "all connections" immediately persists the cleared scope through the settings store, so the custom list was clobbered in storage. The snapshot/restore approach is the right minimal fix given `allowedConnectionIds === null` encodes "all", and the tests pin it well. The maintainer patch above restores the pre-switch confirmation (same key and non-empty-scope condition as before, taken before the snapshot so cancel emits nothing) — the snapshot covers accidental in-session switches while the confirm warns about the cross-session loss that still exists since the cleared scope persists once the page closes. CI green on the patched head; approving.

t8y2:
Thanks for the contribution! Merged in 69098bf0d, will be released in the next version.

PR #8805, opened 2026-09-11
fix(mongodb): accept db["collection"] bracket accessors

thailoc-dev:
## Problem

A collection whose name is not a valid JavaScript identifier cannot be queried with the syntax the shell itself uses:

```js
db["orders-2024"].find({})
db['audit.logs'].find({})
```

Both are rejected with the generic unsupported-command hint. Hyphenated collection names are common, and bracket access is what `mongosh` and Compass print for them, so this is a straight copy-paste failure.

`db.getCollection("orders-2024")` works today, but nothing in the error points there.

## Cause

Collection names are resolved in one place per parser, and both accept only two shapes — `db.<identifier>` and `db.getCollection("…")`:

- `packages/mongo-shell/src/json.ts` → `parseCollectionMethodTarget`
- `crates/dbx-core/src/mongo_shell.rs` → `parse_collection_prefix`

The Rust side rejects it before anything else, since it requires the source to begin with `db.`.

## Change

Accept `db["name"]` / `db['name']` alongside the existing forms, in both parsers. The extracted name flows through the existing code path, so every collection method — `find`, `updateOne`, `aggregate`, and the rest — gains it at once.

No new command fields and no IPC changes: this is name resolution only.

## Testing

- `pnpm test` — 13912 passed
- `cargo test -p dbx-core --lib` — 6108 passed
- New cases both sides: double and single quotes, a dotted name, a chained `updateOne` with options, and the malformed forms `db[]`, `db[""]`, `db["x"]find({})`, `db["x"]`
- `cargo clippy`, `cargo fmt`, `oxfmt` clean

`query::tests::external_driver_preview_retry_preserves_marker_truncation` fails identically on `main` without this change and is unrelated.

### DIFF
--- a/crates/dbx-core/src/mongo_shell.rs
+++ b/crates/dbx-core/src/mongo_shell.rs
@@ -734,6 +734,21 @@ fn trim_mongo_outer_comments(mut source: &str) -> &str {
 }
 
 fn parse_collection_prefix(source: &str) -> Result<(String, usize), String> {
+    // db["orders-2024"] reaches names that are not valid identifiers, the same way the shell does.
+    if source.get(..3).is_some_and(|prefix| prefix.eq_ignore_ascii_case("db[")) {
+        let close = matching_bracket(source, 2).ok_or("Invalid db[\"collection\"] accessor.")?;
+        let collection = parse_string_arg(source[3..close].trim())?;
+        if collection.is_empty() {
+            return Err("Invalid MongoDB collection name.".to_string());
+        }
+        let end = close + 1;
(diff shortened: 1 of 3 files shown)


t8y2 (approved):
Thanks! Verified the bracket branch on both the TS and Rust parsers — it mirrors the existing `db.getCollection()` handling, malformed forms are covered by tests on both sides, and the write-risk classification stays aligned. Nice fix.

t8y2:
Thanks for the contribution! Merged in e1e925fba, will be released in the next version.

PR #8821, opened 2026-09-11
fix(mongodb): stop suggesting extended JSON wrapper keys as fields

thailoc-dev:
## Problem

Autocomplete offers `_id.$oid` as a field. Picking it produces a query that is valid, runs without error, and matches nothing:

```js
db.reports.find({ "_id.$oid": "6743e4bfa3f6f84bc3fff6c8" })   // 0 documents
```

On the server `_id` is an ObjectId scalar, not a subdocument with an `$oid` key — that key exists only in the extended JSON used to ship BSON over the wire. Because the query succeeds, there is nothing to suggest the field is wrong; it reads as "the document isn't there".

The same applies to `created_at.$date`, `count.$numberLong`, `raw.$binary`, and every other wrapper.

## Cause

`collectFieldTypes` in `mongoCompletion.ts` walks into every object it meets, with no notion of extended JSON. Sampling one document yields:

```
_id                (object)
_id.$oid           (string)   <- phantom
created_at         (object)
created_at.$date   (string)   <- phantom
count.$numberLong  (string)   <- phantom
```

The phantom paths are then offered as completions, and the wrapper types are reported as `object`.

The app already knows better elsewhere: `mongoDocumentValues.ts` keeps the canonical wrapper list and the data grid uses it to type its columns correctly. Only the completion path was unaware.

## Change

Add `mongoExtendedJsonValueType` next to that existing list, mapping each wrapper key to the BSON scalar it stands for, and use it in the completion to stop at a wrapper instead of descending into it.

The previous `MONGO_EXTENDED_JSON_VALUE_KEYS` set is now derived from that map, so the wrapper keys stay in one place.

Field types improve as a side effect, since the scalar is now named rather than reported as `object`:

| field | before | after |
| --- | --- | --- |
| `_id` | `object`, plus `_id.$oid` | `objectId` |
| `created_at` | `object`, plus `created_at.$date` | `date` |
| `count` | `object`, plus `count.$numberLong` | `int64` |
| `raw` | `object`, plus `raw.$binary` | `binary` |

Genuine subdocuments and arrays are untouched: `profile.email` and `profile.tags.id` are still inferred as before.

## Testing

- `pnpm test` — 13912 passed
- New test asserts the four phantom paths are absent, the four scalar types are correct, and nested `profile.email` still resolves

### DIFF
--- a/apps/desktop/src/lib/mongo/mongoCompletion.ts
+++ b/apps/desktop/src/lib/mongo/mongoCompletion.ts
@@ -1,3 +1,4 @@
+import { mongoExtendedJsonValueType } from "@/lib/mongo/mongoDocumentValues";
 import { ACCUMULATORS, COMMON_OPERATORS, EXPRESSION_OPERATORS, PIPELINE_STAGES, PUSH_MODIFIERS, QUERY_OPERATORS, STAGE_OPTION_KEYS, UPDATE_OPERATORS, UPDATE_OPERATOR_LABELS, VALUE_SNIPPETS, mongoOperatorItemType, type MongoOperatorSpec } from "@/lib/mongo/mongoCompletionTables";
 
 /**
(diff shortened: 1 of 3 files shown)


t8y2 (approved):
Clean fix, verified end to end: operator completion, genuine subdocuments, arrays and the grid path are all unaffected; the new `mongoExtendedJsonValueType` helper correctly stops field inference from descending into extended-JSON wrappers. Tests cover the phantom paths. One optional nit: `formatMongoShellLiteral` still re-implements the same wrapper check inline and could call the new helper, but merging as-is is fine.

t8y2:
Thanks for the contribution! Merged in 7c2db2d14, will be released in the next version.

PR #8861, opened 2026-09-11
fix(mongodb): support UUID, BinData, Timestamp, MinKey and MaxKey shell values

thailoc-dev:
## Problem

Five shell value constructors are rejected with the generic unsupported-command hint:

```js
db.users.find({ _id: UUID("3b241101-e2bb-4255-8caf-4136c566a962") })
db.files.find({ data: BinData(0, "AQID") })
db.oplog.find({ ts: Timestamp(1735689600, 7) })
db.items.find({ score: { $gt: MinKey, $lt: MaxKey } })
```

A UUID-keyed collection (common with Java and .NET backends) is effectively unqueryable from the editor: the shell spelling fails to parse, and a plain string matches nothing because the value is stored as binary.

## Cause

Both parsers already rewrite shell constructors into extended JSON, but only `ObjectId`, `ISODate`, `new Date`, `NumberLong`, `NumberInt` and `NumberDecimal` are wired up.

The backend is not the limitation. `parse_extended_json_value` in `mongo_driver.rs` already accepts `$uuid`, `$binary`, `$timestamp`, `$minKey` and `$maxKey`; the raw spellings work today. Only the shell-to-extended-JSON translation is missing — the same gap #8694 closed for `new Date()`.

## Change

Teach both parsers the five constructors, emitting the extended JSON the driver already understands:

| shell | extended JSON |
| --- | --- |
| `UUID("…")` / `UUID()` | `{"$uuid": "…"}` (generated when empty, as in `mongosh`) |
| `BinData(sub, "base64")` | `{"$binary": {"base64": "…", "subType": "hh"}}` |
| `Timestamp(t, i)` | `{"$timestamp": {"t": t, "i": i}}` |
| `MinKey` / `MinKey()` | `{"$minKey": 1}` |
| `MaxKey` / `MaxKey()` | `{"$maxKey": 1}` |

`MinKey` and `MaxKey` are accepted both bare and called, since the shell allows either. The bare form is only rewritten in a value position: `{MinKey: 1}` stays an object key, `"MinKey"` stays a string, and identifiers such as `fooMinKey` are untouched.

Arguments are validated before emitting, with a specific error rather than the generic hint: `BinData` subtype 0–255 with a string payload, `Timestamp` two unsigned 32-bit integers, `UUID` a string, `MinKey`/`MaxKey` no arguments.

## Testing

- `pnpm test` — 13816 passed
- `cargo test -p dbx-core --lib` — 6132 passed
- Three new tests per side: the happy path for all five (including nested and array positions), the non-value positions left alone, and nine malformed forms rejected
- Verified the emitted JSON through the driver's `json_object_to_document`: `Binary{Uuid}`, `Binary{Generic}`, `Timestamp`, `MinKey`, `MaxKey`

`query::tests::external_driver_preview_retry_preserves_marker_truncation` fails identically on `main` without this change and is unrelated.

### DIFF
--- a/crates/dbx-core/src/db/mongo_driver.rs
+++ b/crates/dbx-core/src/db/mongo_driver.rs
@@ -2695,7 +2695,15 @@ fn json_filter_value_to_bson(value: &serde_json::Value, field_name: Option<&str>
         }
         serde_json::Value::Object(obj) => {
             if obj.len() == 1 {
-                if obj.contains_key("$regularExpression") {
+                // Shell constructor wrappers (UUID/BinData/Timestamp/MinKey/MaxKey)
+                // and $regularExpression decode through the shared extended JSON
+                // parser, following the same precedent as $date below.
+                if obj.keys().next().is_some_and(|key| {
+                    matches!(
+                        key.as_str(),
+                        "$regularExpression" | "$uuid" | "$binary" | "$timestamp" | "$minKey" | "$maxKey"
+                    )
+                }) {
                     if let Ok(Some(value)) = parse_extended_json_value(obj) {
                         return value;
                     }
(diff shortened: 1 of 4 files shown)


thailoc-dev:
The first CI run failed on `query::tests::sqlite_external_interrupt_without_budget_does_not_panic`, which is unrelated to this change — it looks like a timing race in that test. It fires `interrupt()` from a retry loop expecting to land mid-statement; on the loaded runner it landed during the transaction's `BEGIN` instead, so the error was `Failed to begin transaction: interrupted` rather than the expected `Statement 1 failed … interrupted`.

Re-triggered CI with an identical commit.

t8y2 (commented):
Patch pushed on top of the reviewed head: `5718970b` → `be8b89456` (thanks for enabling maintainer edits). The dual-parser literal support itself is good as reviewed; this patch fixes the headline use case and adds UUID validation.

**Root cause (filters):** both shell parsers rewrite `UUID(...)` / `BinData(...)` / `Timestamp(...)` / `MinKey` / `MaxKey` into `{"$uuid": ...}`-style extended JSON, but `json_filter_value_to_bson` in `crates/dbx-core/src/db/mongo_driver.rs` only decoded `$regularExpression`, `$oid`, `$numberLong` and `$date` in single-key value position. So `db.users.find({_id: UUID("...")})` sent a raw `{_id: {$uuid: "..."}}` document to the server, which fails with `unknown operator: $uuid` (or silently matches nothing inside range operators).

**Fix:**
- Extended the single-key wrapper branch to delegate `$uuid`, `$binary`, `$timestamp`, `$minKey` and `$maxKey` to the existing `parse_extended_json_value` (same precedent as `$regularExpression`). `parse_extended_json_value` already accepts all five keys and the bson crate maps them to `Binary(Uuid)`, `Binary`, `Timestamp`, `MinKey` and `MaxKey`.
- Operator position is preserved: `{$gt: {$uuid: ...}}` still recurses through the generic operator-document path, and the `_id` all-`$` operator branch (`{_id: {$gt: {$uuid: ...}}}`) decodes the operand the same way the `$date` case already does. `_id`/`$oid`/`$date`/`$numberLong` special handling is untouched.
- UUID validation at parse time on both sides: `UUID("...")` now requires the canonical 8-4-4-4-12 hex form (hex digits only, dashes at fixed positions, case-insensitive, matching mongosh's documented 36-char format). Rust rejects with `MongoDB UUID() requires a canonical 8-4-4-4-12 hex string.`; the TS parser rejects by leaving the constructor untouched (its existing rejection path). Note: mongosh's bson package additionally accepts a 32-hex no-dash form; that stays unsupported here for now — happy to add it as a follow-up if wanted.

**Tests (all local, no server needed):**
- `json_filter_to_document_decodes_uuid_binary_timestamp_and_key_constants` — `_id: {$uuid}` yields `Binary(Uuid)` (not a nested doc); BinData/Timestamp equality; MinKey/MaxKey.
- `json_filter_to_document_decodes_key_constants_inside_operators` — `{$gt: MinKey}`-style operands, including the `_id` all-`$` branch and `$eq`/`$gte` operands.
- `validates_uuid_strings_at_parse_time` (Rust) and `parseMongoFindCommand validates UUID strings at parse time` (TS) — reject missing dashes, wrong length, non-hex; uppercase preserved verbatim.

**Results:** `cargo test --no-default-features -p dbx-core --features dbx-core/sqlite-bundled --lib -- json_filter_to_document mongo_shell` → 48 passed, 0 failed; `vitest run packages/app-tests/mongoShellCommand.test.ts` → 99 passed. No timeouts. `git diff --check` clean.

Not approving — leaving the final call to a maintainer pass after CI runs.

t8y2:
Thanks for the contribution! Merged in 71cc734cd, will be released in the next version.

PR #8876, opened 2026-09-11
fix(mongodb): suggest extended JSON values and complete inside $in arrays

thailoc-dev:
## Problem

The editor never suggests the extended JSON spellings of BSON values, even though the parser and driver accept them. Typing `$` in a value position goes silent right as you spell what you want:

| you type | suggestions |
| --- | --- |
| `db.c.find({_id: $` | *(none)* |
| `db.c.find({_id: {$oi` | *(none)* |
| `db.c.find({_id: {$in: [` | *(none)* |
| `db.c.find({_id: {$in: [{$` | *(none)* |

`{_id: {$oid: "…"}}` is valid and runs, but there is no way to discover it. `$in` is the worst case: it is the usual place to paste a list of ids, and it gets no help at all — the classifier treats its array like `$and` / `$or`, whose elements are sub-filters, so it offers field names for what are actually values.

The shell-constructor snippets are also stale: `NumberInt`, `NumberDecimal`, `UUID`, `BinData`, `Timestamp`, `MinKey` and `MaxKey` all parse since #8694 and #8861 but are never offered.

## Change

**A table of extended JSON wrappers** (`$oid`, `$date`, `$uuid`, `$binary`, `$timestamp`, `$numberLong`, …), offered wherever a value can go:

- in a bare value position the braces are added for you: `{_id: $o` → `{ $oid: "…" }`
- inside a `{` the body is enough: `{_id: {$oi` → `$oid: "…"`
- under a field, alongside the query operators, since `{_id: {$oid: …}}` is as valid there as `{_id: {$gt: …}}`
- in update `$set` values and insert documents, which share the value position

`$oid` and `$date` get the existing common-operator boost, so they sort first at a bare `$` rather than `$binary`.

**`$in`, `$nin` and `$all` arrays hold values.** Their elements now complete like any value, and an object inside them completes as a wrapper. `$and` / `$or` / `$nor` are unchanged and still complete fields.

**The shell-constructor snippets catch up with the parser**: the seven constructors above are now offered.

## Testing

- `pnpm test` — 13899 passed
- Three new tests: braced vs. bare wrapper insertion across find/update/insert, operators and wrappers coexisting under a field, `$oid`/`$date` sorting first; `$in`/`$nin`/`$all` as values with `$or` still completing fields and nothing offered inside a string; the seven constructors present
- The new table is included in the existing placeholder-syntax sweep

Not in scope, deliberately: `QUERY_OPERATORS` is still one flat list, so `{_id: {$o` also offers `$or` (invalid there) and a top-level `{$` offers nothing. That is a separate position-awareness change and would be its own PR.

### DIFF
--- a/apps/desktop/src/lib/mongo/mongoCompletion.ts
+++ b/apps/desktop/src/lib/mongo/mongoCompletion.ts
@@ -1,5 +1,19 @@
 import { mongoExtendedJsonValueType } from "@/lib/mongo/mongoDocumentValues";
-import { ACCUMULATORS, COMMON_OPERATORS, EXPRESSION_OPERATORS, PIPELINE_STAGES, PUSH_MODIFIERS, QUERY_OPERATORS, STAGE_OPTION_KEYS, UPDATE_OPERATORS, UPDATE_OPERATOR_LABELS, VALUE_SNIPPETS, mongoOperatorItemType, type MongoOperatorSpec } from "@/lib/mongo/mongoCompletionTables";
+import {
+  ACCUMULATORS,
+  COMMON_OPERATORS,
+  EXPRESSION_OPERATORS,
+  EXTENDED_JSON_VALUES,
+  PIPELINE_STAGES,
+  PUSH_MODIFIERS,
+  QUERY_OPERATORS,
+  STAGE_OPTION_KEYS,
+  UPDATE_OPERATORS,
+  UPDATE_OPERATOR_LABELS,
+  VALUE_SNIPPETS,
+  mongoOperatorItemType,
+  type MongoOperatorSpec,
+} from "@/lib/mongo/mongoCompletionTables";
 
 /**
  * What the cursor may usefully be completed with. Each mode maps to exactly one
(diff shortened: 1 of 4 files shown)


t8y2 (commented):
Thanks for the PR! The TS completion side looks good (already approved on the reviewed head `d6c89b3`). While re-checking the head, I found one gap on the Rust side and pushed a small maintainer patch on top as `256e4322` (`fix(mongo): decode number wrappers in query filters`), following the precedent from #8861:

**Gap:** With the completion now suggesting extended JSON number wrappers in filter contexts, `find({score: {$numberInt: "5"}})` still sent the raw `{"$numberInt": "5"}` sub-document to the server (rejected as an unknown operator, or silently matching nothing), while update/insert documents already decoded these wrappers. Only `$oid`/`$numberLong`/`$date` plus the constructor wrappers were decoded in filter position.

**Fix:** `json_filter_value_to_bson` in `crates/dbx-core/src/db/mongo_driver.rs` now routes `$numberInt | $numberDouble | $numberDecimal` through the shared `parse_extended_json_value` parser (same arm as `$regularExpression`/`$uuid`/`$binary`/`$timestamp`/`$minKey`/`$maxKey`), so typed number literals decode to `Int32`/`Double`/`Decimal128` in equality, operator, and `$in`-array positions.

**Tests:** Added filter-path unit tests next to the #8861 `json_filter_to_document` tests covering `{$numberInt: "5"}` → `Int32`, `{$numberDouble: "1.5"}` → `Double`, `{$numberDecimal: "3.14"}` → `Decimal128`, and decoding inside `$gte` and `$in` arrays.

**Results:** `cargo test --no-default-features -p dbx-core --features dbx-core/sqlite-bundled --lib -- json_filter_to_document` — 10 passed, 0 failed (8 pre-existing + 2 new); `cargo check --no-default-features -p dbx-core` clean.

No TS changes were made. Nothing else pending from my side.

t8y2:
Thanks for the contribution! Merged in 074216ed0, will be released in the next version.

PR #8881, opened 2026-09-11
fix(editor): stop replaying stale format/compress requests on tab switch

### LINKED ISSUE #8827: [Bug] sql文件页签切换的时候原本格式化的内容会被莫名压缩
### 数据库类型和版本

通用

### 问题描述

1. 复现步骤：复现了，但是不确定是否稳定。这样操作，把sql点击压缩，你再格式化。然后切换到另外一个sql文件（不要去数据页，就是另外一个sql文件），再切回这个sql文件，来回试试，大概率能复现
2. 期望结果：不应该被自动压缩
3. 实际情况：格式化过的内容会被自动压缩

### 支持信息

DBX 版本: v0.6.10
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

https://github.com/user-attachments/assets/4915c7ae-c001-416a-a44e-3221edcf2377

Abeautifulsnow:
### Problem

Fixes #8827 

On a SQL editor tab, clicking **Compress** and then **Format** and then switching to
another SQL file tab and back silently reverted the format: the text collapsed back
to the compressed single line, and the tab was flipped dirty. Going back and forth
between the two tabs reproduced it on every round trip (alternating between the
formatted and compressed text, which made it look "unstable"). Switching to a data
page did *not* reproduce it.

Reported-by: QA repro (compress → format → switch to another SQL file → switch back)

### Root cause

Toolbar format/compress requests travel as global monotonic `{ id, tabId }` refs
(`formatSqlRequest` / `compressSqlRequest`) that are never cleared. `ContentArea`
delivers them to the editor gated on the active tab
(`formatSqlRequest?.tabId === activeTab.id ? formatSqlRequest.id : undefined`), so
switching away sets the prop to `undefined` and switching back re-arms the **same
stale id** on the same reused `QueryEditor` instance (query tabs share one editor;
documents are swapped per tab via `activateTabDocument`).

Both request watchers only guarded with `val !== oldVal`, so the stale re-arm
re-fired them. `compressCurrentSql` is fully synchronous and dispatched first;
`formatCurrentSql` suspends at its `await` and then bails on its stale-state guard
— so the replayed compress won and the user's format was silently undone (and
written back into the tab store, flipping the dirty marker).

This also explains the report's details: a data page unmounts the editor (fresh
instance → no watcher fire), and only the compress-then-format order leaves both
requests armed with a non-idempotent replay.

### Fix

Guard both request watchers with **module-scope monotonic last-handled cursors**:

```ts
if (val && val > lastHandledFormatRequestId) {
  lastHandledFormatRequestId = val;
  formatCurrentSql();
}
```

- Strictly monotonic `>` (not `!==`) because the ids are a single global counter
  shared by all tabs — a lower id is always an already-handled request, never a
  pending one (delivery is same-tick to the active tab's instance).
- **Module scope** (a plain `<script lang="ts">` block, matching
  `EditorGroupTabBar.vue`) rather than `<script setup>` so the cursor also survives
  editor unmount/remount: a remounted editor consumes the stale id instead of
  replaying it.

First-time compress/format via toolbar, the editor context menu (which calls these
actions directly and never went through the request refs), keyboard shortcuts, and
selection-scoped formatting are all unchanged.

### Testing

- New regression spec `QueryEditor.formatCompressRequestReplay.spec.ts` (real
  component, real prop oscillation mirroring the App → ContentArea → editor loop):
  - replay after tab switch-away-and-back must leave the formatted text intact —
    **fails on `main`**, passes with this fix (verified by reverting the fix in a
    dirty-tree run);
  - a genuinely greater id still fires (over-blocking guard);
  - a stale id preset on a fresh mount does not fire (data-page remount path).
- `editor` + `layout` suites: 53 files / 391 tests green; `vue-tsc` typecheck clean;
  `oxlint` / `oxfmt` clean.

### Notes for reviewers

- The one semantic change: a request id delivered while the editor view is not
  ready is now consumed without acting and will not be retried on tab return.
  Previously it *could* retry — that retry *was* the replay bug, so this corner is
  the intended behavior change.
- The replay-guard contract is documented in the task's spec notes so future
  toolbar actions added to this request plumbing inherit the same cursor pattern.

### DIFF
--- a/apps/desktop/src/components/editor/QueryEditor.vue
+++ b/apps/desktop/src/components/editor/QueryEditor.vue
@@ -1,3 +1,18 @@
+<script lang="ts">
+// Replay guard for the toolbar format/compress requests. The ids are global
+// monotonic counters shared by every tab, and ContentArea delivers them by
+// gating the inactive tab's prop back to `undefined`; returning to the tab
+// re-arms the same stale id on the reused editor instance (a replay, not a new
+// command). The cursors live at module scope — not in `<script setup>` — so
+// they also survive an editor unmount/remount (data-page switches), letting a
+// freshly mounted editor consume the stale id instead of replaying it. The
+// check is strictly monotonic (`>`), not `!==`: because ids are shared across
+// tabs, a lower id is always an already-handled request from an active tab
+// (delivery is same-tick), never a pending undelivered one.
+let lastHandledFormatRequestId = 0;
+let lastHandledCompressRequestId = 0;
+</script>
+
 <script setup lang="ts">
 import { ref, onMounted, onBeforeUnmount, onActivated, onDeactivated, watch, shallowRef, computed, nextTick } from "vue";
 import { AlignLeft, Camera, CaseLower, CaseSensitive, CaseUpper, ClipboardPaste, Code2, Download, Eye, FileCode, MessageSquareText, Minimize2, Pencil, PencilRuler, Play, Copy, List, Scissors, Search, Sparkles, Table2, TextSelect, Trash2 } from "@lucide/vue";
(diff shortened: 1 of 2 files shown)


t8y2 (approved):
Verified the root cause end-to-end and ran your new spec three ways: 3/3 at the PR head, the replay test fails on the pre-fix watcher (exactly the reported symptom), and 3/3 on a merged tree with current main. Great behavioral coverage, and the module-scope cursor rationale is sound. Note after today's KeepAlive refactor on main your guard is partly defense-in-depth for the literal repro path, but it still covers real residual paths and pins the editor-level contract — keeping as-is. Approving.

t8y2:
Thanks for the contribution! Merged in 32e070f38, will be released in the next version.

PR #8892, opened 2026-09-11
fix(shortcuts): record shifted punctuation by its unshifted key

### LINKED ISSUE #8670: [Bug] 设置ctrl+shift+/ 为块注释快捷键时无法正常使用
### 数据库类型和版本

通用

### 问题描述

问题1：设置块注释快捷键为ctrl+shift+/ 时无法正常使用。
问题2：鼠标左键三击某sql语句时无法选中该行。

### 支持信息

DBX 版本: v0.6.9
运行环境: 桌面端
操作系统: Windows 10.0.26200.9445
系统架构: x86_64

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

_No response_

thanadon-dev:
## Problem

Binding an editor action to a shifted punctuation or digit key produces a shortcut that never fires. Per #8670, setting `Ctrl+Shift+/` as the block-comment shortcut leaves the editor with no working binding.

## Cause

`KeyboardEvent.key` reports the *shifted* character, so pressing `Ctrl+Shift+/` on a US layout yields `key === "?"`. That value is stored verbatim and handed to CodeMirror:

```
event.key            : "?"
recorded as          : Shift+Mod+?
registered binding   : Shift-Ctrl-?
```

For that same press, CodeMirror's `runHandlers` resolves Shift bindings against the *unshifted* key from `base[event.keyCode]` (`base[191] === "/"`), so it only ever looks up:

```
Ctrl-?  ,  Shift-Ctrl-/
```

Neither matches `Shift-Ctrl-?`, so the command is unreachable.

Letters escape the bug by coincidence: `shortcutToCodeMirrorKey` lowercases single characters, and `base[65] === "a"`, so `Shift+Alt+A` → `Shift-Alt-a` matches CodeMirror's second lookup. Punctuation and digits have no equivalent fallback.

## Fix

Derive the unshifted character from `event.code` when recording a shifted press, so the registered binding is one the keymap actually resolves:

```
Shift+Mod+/  ->  Shift-Ctrl-/   (matches)
```

`matchesShortcutKey` accepts both the shifted and unshifted form, so shortcuts saved by earlier versions keep working after upgrading rather than silently going dead.

The lookup table mirrors `base[keyCode]` from `w3c-keyname`, which is the table CodeMirror itself uses. Letters are deliberately excluded — they already round-trip correctly.

## Tests

Added to `keyboardShortcuts.spec.ts`:

- shifted punctuation and digits record their unshifted key (`/`, `1`, `;`), and `shortcutToCodeMirrorKey` produces the binding CodeMirror resolves
- a press matches both the new form and the legacy stored form; a different physical key does not match
- unshifted keys and letters are unchanged

`keyboardShortcuts`, `shortcutRegistry`, and `shortcutDisplay` specs pass together (74 tests), and `oxlint --vue-plugin` is clean on both files.

## Not covered

The second, unrelated report in #8670 — triple-click failing to select a statement — is not addressed here.

Closes #8670

### DIFF
--- a/apps/desktop/src/lib/__tests__/editor/keyboardShortcuts.spec.ts
+++ b/apps/desktop/src/lib/__tests__/editor/keyboardShortcuts.spec.ts
@@ -17,6 +17,7 @@ import {
   tabSwitcherDirectionFromShortcut,
 } from "@/lib/editor/keyboardShortcuts";
 import { formatShortcutDisplay, isMacShortcutPlatform } from "@/lib/editor/shortcutDisplay";
+import { shortcutToCodeMirrorKey } from "@/lib/editor/shortcutRegistry";
 
 describe("keyboard shortcut matching", () => {
   it("records modifier-only mouse shortcut settings", () => {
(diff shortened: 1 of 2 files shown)


t8y2:
Thanks for the contribution! Merged in 35773ac, will be released in the next version.

PR #8914, opened 2026-09-11
fix(import): unwrap the CSV force-text marker written by export

### LINKED ISSUE #8803: [Bug] 导出的csv数据包含时间字段时，导出的问题异常且该文件无法导入
### 数据库类型和版本

MySQL：5.7.44，Doris：doris-2.1.8-rc01-52396fbc88

### 问题描述

导出csv格式文件时，如果导出的表中有时间字段，会自动给时间字段加东西
下面这是一条导出数据，eg:
insert_time,id
"=""2026-06-24 02:00:07""",695350
导入的时候，又解析不了这种csv文件，直接报错

### 支持信息

DBX 版本: v0.6.10
运行环境: 桌面端
操作系统: Windows 10.0.26200.9445
系统架构: x86_64

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

Image

Image

thanadon-dev:
## Problem

Per #8803, exporting a table that has a datetime column to CSV and importing that same file back fails: the timestamp arrives as the literal `="2026-06-24 02:00:07"`, which every temporal target rejects.

```
insert_time,id
"=""2026-06-24 02:00:07""",695350
```

## Cause

The `="..."` wrapper is written deliberately. `temporal_format::wrap_csv_force_text` (and `forceCsvTextForTemporalColumns` on the desktop side) adds it so WPS/Excel/OnlyOffice stop re-guessing a quoted date cell's type and silently truncating it.

The importer has no matching step. `csv_value_with_config` stores the field verbatim, so the wrapper survives into the INSERT and the column type rejects it.

## Fix

Unwrap the marker in `csv_value_with_config`, matching only its exact shape: a leading `="`, a trailing `"`, and no `"` in between.

That shape cannot be produced by RFC4180 quoting of a value that merely starts with `=`, so CSV files dbx did not write are unaffected:

| cell | result |
|---|---|
| `="2026-06-24 02:00:07"` | unwrapped |
| `=SUM(A1:A2)` | unchanged |
| `="a"&"b"` | unchanged (inner quotes) |
| `="unterminated` | unchanged (no trailing quote) |

## Tests

Two tests in `table_import.rs`:

- `csv_import_unwraps_the_force_text_wrapper_written_by_csv_export` — the exact row from the report round-trips into `INSERT INTO ... VALUES ('2026-06-24 02:00:07', 695350)`
- `csv_import_keeps_values_that_only_look_like_the_force_text_wrapper` — the four cases in the table above

```
test result: ok. 3 passed; 0 failed
```

`cargo fmt -p dbx-core -- --check` is clean.

Closes #8803

### DIFF
--- a/crates/dbx-core/src/table_import.rs
+++ b/crates/dbx-core/src/table_import.rs
@@ -492,8 +492,29 @@ pub fn effective_delimited_config(
     })
 }
 
+/// Undo the `="..."` force-text wrapper that CSV export writes around temporal
+/// cells (see `temporal_format::wrap_csv_force_text`). Without this, exporting a
+/// table with a datetime column and importing the file straight back stores the
+/// literal `="2026-06-24 02:00:07"` instead of the timestamp, which every
+/// temporal column type then rejects.
+///
+/// Only the exact wrapper shape is unwrapped: a leading `="`, a trailing `"`,
+/// and no `"` in between. A genuine value that merely starts with `=` (or
+/// contains quotes of its own) is left untouched, so this cannot corrupt CSV
+/// files that dbx did not produce.
+fn unwrap_csv_force_text(value: &str) -> &str {
+    let Some(inner) = value.strip_prefix("=\"").and_then(|rest| rest.strip_suffix('"')) else {
+        return value;
+    };
+    if inner.contains('"') {
+        return value;
+    }
+    inner
+}
+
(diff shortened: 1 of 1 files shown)


t8y2:
Thanks for the contribution! Merged in 076c39bc215, will be released in the next version.

PR #8934, opened 2026-09-12
fix(sql): treat GET DIAGNOSTICS targets as MySQL user variables

thanadon-dev:
## Problem

Running a MySQL script whose stored procedure reads the diagnostics area fails with `ERROR 1064`. The user variable on the left of each `GET DIAGNOSTICS` assignment is replaced with an empty string before the script reaches the server:

```sql
GET DIAGNOSTICS CONDITION 1 @err_msg = MESSAGE_TEXT;
```

is sent as

```sql
GET DIAGNOSTICS CONDITION 1 '' = MESSAGE_TEXT;
```

which is exactly the fragment quoted in the reporter's error message.

## Root cause

`collectNativeSqlServerParameters` in `apps/desktop/src/lib/sql/sqlParameters.ts` records the MySQL user variables a script assigns to, so the `@name` template-parameter branch skips them. It recognises the three statements that write into one:

- `DECLARE` → `collectDeclareStatementVariables`
- `SET` → `collectSetStatementVariables`
- `SELECT` / `SELECT ... INTO` → `collectSelectAssignmentVariables`

`GET DIAGNOSTICS` writes into user variables the same way but has no branch, so its targets are never declared and fall through to the template-parameter path.

## Fix

Route the statement into `collectSetStatementVariables`, which already handles this exact right-hand side (`@name = source`, comma separated). The new `readGetDiagnosticsEnd` only skips the leading keywords, accepting the optional `CURRENT`/`STACKED` form, and returns `null` for an ordinary identifier called `get` so a column of that name keeps its surrounding template parameters.

## Tests

Four cases in the existing `extractSqlParameters` suite:

- the reported statement, with two targets on one line
- `GET DIAGNOSTICS @affected = ROW_COUNT` plus the `CURRENT` and `STACKED` forms
- a genuine template parameter sitting next to a diagnostics target, which must still be collected
- a column named `get`, which must not disable parameter collection for the rest of the statement

Reverting the source change fails 3 of the 4; they pass with it.

## Verification

- `npx vitest run apps/desktop/src/lib/__tests__/sql/sqlParameters.spec.ts` → 119 passed
- `npx vue-tsc --noEmit --project apps/desktop/tsconfig.json` → exit 0
- `npx oxfmt` on both files, no further changes
- `npx oxlint --vue-plugin` on both files reports one `no-useless-escape` warning, which is present on an unmodified tree at the same source line and is not introduced here

One note in the interest of full disclosure: `apps/desktop/src/lib/__tests__/sql/externalSqlFileTarget.spec.ts` fails all 10 of its tests on my machine. It fails identically with my changes stashed, so it is pre-existing here and unrelated to this change, but I could not confirm whether it is green in CI.

### DIFF
--- a/apps/desktop/src/lib/__tests__/sql/sqlParameters.spec.ts
+++ b/apps/desktop/src/lib/__tests__/sql/sqlParameters.spec.ts
@@ -292,6 +292,41 @@ describe("extractSqlParameters", () => {
     expect(extractSqlParameters(sql, { databaseType: "mysql" })).toEqual(["input_id", "tenant_id"]);
   });
 
+  it("ignores MySQL user variables assigned by GET DIAGNOSTICS", () => {
+    const sql = `
+      get diagnostics condition 1 @err_state = returned_sqlstate, @err_msg = message_text;
+      select concat('failed: ', @err_state, ' ', @err_msg) as result;
+    `;
+
+    expect(extractSqlParameters(sql, { databaseType: "mysql" })).toEqual([]);
+  });
+
+  it("ignores GET CURRENT/STACKED DIAGNOSTICS targets and the statement-level row count", () => {
+    const sql = `
+      get diagnostics @affected = row_count;
+      get current diagnostics condition 1 @current_state = returned_sqlstate;
+      get stacked diagnostics condition 1 @stacked_state = returned_sqlstate;
+      select @affected, @current_state, @stacked_state;
+    `;
+
+    expect(extractSqlParameters(sql, { databaseType: "mysql" })).toEqual([]);
+  });
+
+  it("keeps ordinary MySQL template parameters next to GET DIAGNOSTICS targets", () => {
+    const sql = `
+      get diagnostics condition 1 @err_msg = message_text;
+      select * from audit_log where tenant_id = @tenant_id and note = @err_msg;
+    `;
+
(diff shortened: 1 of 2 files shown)


t8y2 (approved):
Thanks for the clean, well-scoped fix. Routing `GET DIAGNOSTICS` through `collectSetStatementVariables` matches the statement grammar exactly (`target = property_name`, comma-separated, optional `CURRENT|STACKED`) and follows the same pattern as the existing DECLARE / SET / SELECT ... INTO branches. Word-boundary matching and the null fall-through for an ordinary identifier named `get` check out, and the four tests cover the reported statement plus the interesting neighbors. CI green, merging as-is.

Two non-blocking notes for potential follow-ups:
1. `GET DIAGNOSTICS CONDITION @cno @errno = MYSQL_ERRNO` (condition number as a user variable, a pattern shown in the MySQL manual) still collects `@cno` as a template parameter when no earlier statement in the script assigns it. Pre-existing behavior, unchanged by this PR.
2. `readNextKeyword` doesn't skip MySQL `#` line comments between `GET` and `DIAGNOSTICS`; the existing routine-definition helper shares this limitation, so cosmetic.

t8y2:
Thanks for the contribution! Merged in 4bac6b56d, will be released in the next version.

PR #8969, opened 2026-09-13
feat(elasticsearch): support custom CA and mutual TLS (mTLS) for HTTPS connections

yimiaoxiehou:
Add TLS support to the Elasticsearch and Easysearch drivers (Rust `dbx-core`), including custom CA trust and client-certificate (mTLS) authentication.

### Changes
- Shared `apply_tls_certificates` helper in `crates/dbx-core/src/db/mod.rs` for reqwest/rustls-based drivers.
- Extended `EsClient::from_config`/`new_with_mode` and `EasysearchClient::from_config` to accept `ca_cert_path`, `client_cert_path`, and `client_key_path`.
- Wires these paths through the connection dispatcher for both ES and Easysearch.
- UI: show CA/client-cert/key inputs in `ConnectionDialog.vue` for `elasticsearch` and `easysearch`.
- Tests:
  - Unit test generating throwaway CA + client cert via openssl and asserting the client builds.
  - Env-gated integration test that connects to a real TLS-enabled Elasticsearch in Docker.

### Test environment used
- Elasticsearch 8.15.0 with `xpack.security.http.ssl.enabled=true`, custom CA + server cert mounted under `/usr/share/elasticsearch/config/certs`.
- Integration test env vars: `DBX_ELASTICSEARCH_TLS_TEST_URL`, `DBX_ELASTICSEARCH_TLS_TEST_CA_CERT_PATH`, `DBX_ELASTICSEARCH_TLS_TEST_USERNAME`/`PASSWORD`.


t8y2 (changes requested):
request changes — the Rust-side plumbing is structurally sound (shared `apply_tls_certificates` helper is the right direction), but two things prevent the feature from working end-to-end today:

1. The dialog never persists the new fields. `connectionConfigForSubmit` (`apps/desktop/src/components/connection/ConnectionDialog.vue:4135-4141`) strips cert fields for db types not on its keep-lists — `config.client_cert_path = undefined; config.client_key_path = undefined;` (4137-4138) and `config.ca_cert_path = undefined` (4140-4141) — and elasticsearch/easysearch aren't on them, so the inputs you now show are silently dropped on every save/test and the Rust side never sees them. Please extend both keep-lists and add the same trim + cert/key-pair validation as the consul branch (and the zookeeper branch from #8977 once it lands).

2. `apply_tls_certificates` (`crates/dbx-core/src/db/mod.rs:136`) doesn't pin `.use_rustls_tls()`. The desktop build unifies reqwest with `default-tls` (`src-tauri/Cargo.toml:57`), so `Client::builder()` selects native-tls there, which rejects the PEM `Identity` at `build()` with "incompatible TLS identity type" — the repo documents this exact trap in `agent_kv.rs:1228` and `consul/client.rs:26`. The current `unwrap_or_else(|_| HttpClient::new())` fallback in `EsClient::new_with_mode` (`crates/dbx-core/src/db/elasticsearch_driver.rs:115`) then silently swaps in a default client, dropping the CA, the identity, and `danger_accept_invalid_certs`. Please call `.use_rustls_tls()` like the etcd-metrics and Consul clients do.

Smaller robustness notes on the helper (same patch): prefer `Certificate::from_pem_bundle` (single-cert `from_pem` silently fails on CA chain bundles), append a newline to the cert PEM before concatenating the key (a file without a trailing newline produces an unparseable identity that is silently dropped), and consider surfacing read/parse errors instead of degrading silently — the Consul/etcd-metrics helpers return errors, which is much easier to debug.

Heads-up on ordering: #8977 (zookeeper TLS) rewrites the exact same `ConnectionDialog.vue` region from the same base and is approved — it will likely merge first, so please rebase on it afterwards and add elasticsearch/easysearch to the rewritten keep-list conditions.

yimiaoxiehou:
Thanks for the detailed review — pushed a fixup that addresses both blockers (commit `7f6554d` on `feat/elasticsearch-tls`):

**1. Dialog now persists the cert fields (frontend)**
- `connectionConfigForSubmit` in `ConnectionDialog.vue` now keeps `ca_cert_path` / `client_cert_path` / `client_key_path` for `elasticsearch` and `easysearch` (both keep-lists extended).
- Added the same trim + client-cert/key-pair validation the consul branch uses, reusing `connection.etcdClientCertPairRequired` when only one of the pair is set.

**2. Rust TLS plumbing (db/mod.rs, elasticsearch_driver.rs, easysearch_driver.rs)**
- `apply_tls_certificates` now pins `.use_rustls_tls()` so the PEM `Identity` is accepted instead of being rejected by native-tls with `incompatible TLS identity type` — mirroring `consul/client.rs` and `agent_kv.rs`.
- Switched the CA to `Certificate::from_pem_bundle` (handles CA-chain bundles, not just a single cert).
- Append a newline before concatenating the key PEM, so a cert file without a trailing newline still parses.
- Removed the silent `unwrap_or_else(|_| HttpClient::new())` fallback. `apply_tls_certificates` now returns `Result<ClientBuilder, String>` and surfaces descriptive read/parse errors; `new_with_mode` / `from_config` (both `EsClient` and `EasysearchClient`) propagate the error up to the connection dispatcher via `?`, so a bad cert path fails the connect loudly instead of degrading. The unit test was updated to assert the error path. Verified: `cargo test -p dbx-core` passes (109 ES tests + the TLS test), and the desktop typecheck is clean.

**Ordering / rebase:** #8977 isn't merged yet, so I kept this branch based off the original shared base (to avoid polluting this PR's diff with the zookeeper changes). I'll rebase `feat/elasticsearch-tls` onto `feat/zookeeper-tls` once #8977 lands and confirm the rewritten keep-list conditions include elasticsearch/easysearch — exactly as you flagged.

t8y2 (changes requested):
request changes — both original blockers were genuinely addressed (rustls pinned, cert paths persisted with pair validation, PEM-bundle + newline guard, errors surfaced), but the PR cannot merge as-is:

1. **The desktop crate no longer compiles.** `src-tauri/src/commands/connection.rs` still calls the old 7-arg `EsClient::from_config` / `EasysearchClient::from_config` at lines 1393, 1407, 1836 and 1849. Since the constructors gained three cert args and now return `Result`, `cargo check --no-default-features -p dbx` fails with 10 errors (E0061/E0308). Please update all four sites to pass `ca_cert_path`/`client_cert_path`/`client_key_path` and propagate with `?` — the "Test Connection" flow must use the same TLS config as the real connection. Running `cargo check --workspace --no-default-features` (not just `-p dbx-core`) would have caught this.

2. **Please rebase onto current main.** ZooKeeper TLS (#8977) merged and rewrote the exact `connectionConfigForSubmit` region this PR touches — GitHub reports the branch CONFLICTING with three hunks in `ConnectionDialog.vue`. The resolution is mechanical: keep zookeeper alongside your elasticsearch/easysearch entries in the client-cert wipe exclusion, the `ca_cert_path` keep-list, and the shared cert-input `v-if` (and note `clientCertHint` is now the generic key after #8977).

3. Minor: `apply_tls_certificates` (`crates/dbx-core/src/db/mod.rs:160/171/180`) reads cert paths verbatim — the Consul and etcd-metrics clients run them through `expand_tilde`, and the dialog placeholder suggests `~/...` paths. Please do the same for all three paths.

After that, `cargo check --workspace --no-default-features`, `cargo test -p dbx-core --no-default-features tls`, and the connection vitest specs (246/246 on your head today) should all pass. Happy to review again quickly.

yimiaoxiehou:
Thanks — pushed a second fixup set (branch rebased onto current `main`, which now includes #8977). Summary:

**1. Desktop crate compiles again.** In `src-tauri/src/commands/connection.rs` I updated all four `EsClient::from_config` / `EasysearchClient::from_config` call sites (the `test_connection_with_info_inner` and `connect_db` flows) to pass `ca_cert_path` / `client_cert_path` / `client_key_path` and propagate with `?`, mirroring the already-updated `dbx-core` dispatcher. The "Test Connection" path now uses the same TLS config as the real connection.

**2. Rebased onto `main` and resolved the `ConnectionDialog.vue` conflict.** `zookeeper` (from #8977) and `elasticsearch`/`easysearch` (this PR) now coexist in all three spots: the client-cert wipe exclusion, the `ca_cert_path` keep-list, and the shared cert-input `v-if`. The `clientCertHint` generic key from #8977 is preserved (this PR never introduced a separate key).

**3. `expand_tilde` for cert paths.** `apply_tls_certificates` now runs `ca_cert_path` / `client_cert_path` / `client_key_path` through `crate::path_utils::expand_tilde` (a no-op for non-`~` paths) before reading — matching the Consul and etcd-metrics clients, so a `~/...` cert path from the dialog placeholder resolves correctly.

**Verification (all green):**
- `cargo check -p dbx --no-default-features` — desktop crate compiles.
- `cargo test -p dbx-core --no-default-features tls` — TLS unit test passes.
- `pnpm test connection` — 955 connection vitest specs pass.

Commits on the branch: `60057e8` (Tauri call sites) and `1574487` (expand_tilde). Ready for another review.

t8y2 (commented):
Thanks — all three points from the last review are confirmed addressed (Tauri call sites propagate the cert paths, the rebase coexists cleanly with #8977, and expand_tilde covers all three paths). Pushed a small maintainer commit af058f609 to close the two CI failures on the new head:

1. **clippy (rust-fmt-clippy)**: `new_with_mode` now takes 12 args → `clippy::too_many_arguments` under `-D warnings` (`crates/dbx-core/src/db/elasticsearch_driver.rs`). Added `#[allow(clippy::too_many_arguments)]` (the established in-repo pattern). Verified with `cargo clippy -p dbx-core --no-default-features -- -D warnings` — clean.

2. **rust-test TLS unit test**: the test now generates the client cert as X.509 **v3** — `gen_certs` passes an extfile (`basicConstraints=critical,CA:FALSE`) to `openssl x509 -req`. Root cause: without extensions, `openssl x509 -req` emits a **v1** cert on OpenSSL 3.0.x (the ubuntu runner; also LibreSSL locally), and rustls rejects v1 client certs at builder time with `UnsupportedCertVersion` — your local Homebrew OpenSSL 3.6 emits v3, which is why it passed for you. Also switched the assertion to print the actual build error. Verified locally on a v1-emitting toolchain (LibreSSL 3.3.6): the test went red before the fix, green after.

One behavior note for the release notes (fine to keep, matches consul/zookeeper): pinning rustls switches every ES HTTPS connection from the system trust store to webpki roots, so servers whose certificate chains to a corporate CA in the OS keychain now need `ca_cert_path` set.

t8y2 (approved):
All green on af058f609 (rust, rust-test, rust-fmt-clippy, frontend). Merging — thanks for iterating on the review points.

t8y2:
Thanks for the contribution! Merged in 805d10346, will be released in the next version.

PR #8977, opened 2026-09-13
feat(zookeeper): support TLS-secured connections

yimiaoxiehou:
## Summary
Adds TLS support to the ZooKeeper connection driver so dbx can connect to ZooKeeper clusters that require a secure client port (`secureClientPort`) with optional mTLS.

- **Go agent** (`agents/drivers/zookeeper`):
  - New `tls.go` with `buildZooKeeperTLSConfig` that builds a `*tls.Config` from `ca_cert_path` / `client_cert_path` / `client_key_path` (and `ssl` flag).
  - `connection.go` now dials via `tls.DialWithDialer` when a TLS config is present, and the reachability probe uses it as well.
- **Frontend** (`apps/desktop/src/components/connection/ConnectionDialog.vue`):
  - Preserve the `ssl` (启用加密连接) toggle for ZooKeeper connections (it was previously forced off on save / on open).
  - Persist `ca_cert_path` / `client_cert_path` / `client_key_path` when saving and when switching/editing a ZooKeeper connection.
- **Tests**: `tls_test.go` (unit) and `integration_test.go` (TLS handshake against a secure port).

## How it works
The Rust `agent_connect_params` already forwards `ca_cert_path` / `client_cert_path` / `client_key_path` to the agent; this change makes the agent actually use them and the UI stop discarding them for ZooKeeper.

## Test plan
- Unit: `cd agents/drivers/zookeeper && go test ./...`
- Manual: configure a ZooKeeper connection with `ssl=true` + CA/client cert paths pointing at a `secureClientPort`, click 测试 — connection succeeds (verified end-to-end against a TLS-enabled ZooKeeper 3.9.5 container).

## Notes
- The runtime agent binary and the ZooKeeper server TLS keystore (PKCS#12) are deployed out-of-tree; this PR contains only the source changes.


t8y2 (approved):
Thanks for the clean implementation — reviewed and approved. Verify-by-default TLS 1.2, correct per-dial `ServerName` handling for multi-host connect strings, SASL digest stacking on top of the TLS conn, and non-TLS paths bit-identical to before; local `go test` / `go vet` / `gofmt` all pass. Three non-blocking nits if you feel like polishing later: the client-auth hint shown for ZooKeeper reuses the etcd-specific string (`etcdClientCertHint`), `loadCACertPool` could use `x509.CertPool.AppendCertsFromPEM` like `etcd-go` instead of the manual PEM loop, and `insecure_skip_verify` is URL-param-only with no docs mention (fine since it defaults to verify). Heads-up: your #8969 rewrites the same `ConnectionDialog.vue` region, so expect a rebase on whichever merges second.

yimiaoxiehou:
Pushed fixes for all three nits:

1. **Client-cert hint:** replaced the etcd-specific `connection.etcdClientCertHint` (used in the shared etcd/consul/zookeeper TLS block) with a generic `connection.clientCertHint`, so the hint no longer mislabels ZooKeeper/Consul connections. Added `en` (fallback locale) + `zh-CN`; the i18n autofill picked up the remaining locales.
2. **`loadCACertPool`:** switched to `x509.CertPool.AppendCertsFromPEM` (the same helper etcd-go uses), removing the manual PEM decode loop and the now-unused `encoding/pem` import. Behavior is unchanged (still parses CA-chain bundles).
3. **`insecure_skip_verify`:** documented that it is a URL-only param (`?insecure_skip_verify=true`) defaulting to off, so callers that don't set it get full certificate-chain verification.

`gofmt` / `go vet` / `go test` all pass for `agents/drivers/zookeeper`. Ready for another look whenever convenient.

t8y2 (commented):
Maintainer conflict resolution applied: `1d352113c` → `e1a5d9d01` (merge of current `main` into the branch, `83fdd78d5` at the time).

Conflict cause: only `apps/desktop/src/i18n/locales/ko.ts` — this branch's i18n autofill added `clientCertHint` while `main` gained the `pluginPlatform` Korean block at overlapping lines. All other files (ConnectionDialog.vue, remaining locales, zk agent) auto-merged cleanly.

Resolution: took main's `ko.ts` and inserted the branch's single line (`clientCertHint: "mTLS가 요구되는 경우…"`) directly after `etcdClientCertHint`, exactly matching its placement in the other seven locales. Verified: `git diff --check` clean, the resolved `ko.ts` differs from main by exactly that one line, the full branch-vs-main diff is unchanged PR content (zk agent TLS files + ConnectionDialog + 8 locale lines), and `oxfmt --check` passes.

No behavior changed on either side; the merged tree is the union. CI will run on the new head and the merge follows once it's green.

t8y2:
Thanks for the contribution! Merged in 56c0f8567, will be released in the next version.

PR #8982, opened 2026-09-13
feat(editor): copy selected SQL as rich text with highlighting

zouchao:
## What

Adds a **Copy as Rich Text** item to the query editor context menu (right below *Copy selection*, highlighter icon). It writes both a `text/html` and a `text/plain` flavor to the system clipboard, so pasting a query into a mail client, Word or a chat app keeps the syntax colors — while plain-text targets (a terminal, another SQL editor) still receive the original SQL untouched.

Motivation: copying a query into an email previously lost all highlighting. *Screenshot Selected Code* covers this as an image, but the recipient cannot select or search the text.

Single commit, 17 files, +326/-4. Rebased onto current `main`.

## How

**`lib/common/clipboard.ts`** — new `copyRichTextToClipboard(html, text)` with a three-step degradation:

1. Tauri → `clipboard-manager` `writeHtml(html, altText)`, where `altText` carries the plain-text flavor
2. Web → `navigator.clipboard.write` with a two-flavor `ClipboardItem`
3. Otherwise → plain text via `copyToClipboard`, so a non-secure web deployment still copies something instead of throwing

**`lib/sql/sqlRichText.ts`** (new) — renders from Shiki's `codeToTokensBase` rather than `codeToHtml`, so the emitted markup is fully controlled:

- **Inline styles only** — no `class`, no `<style>` block. Mail clients strip both, so nothing may depend on a stylesheet surviving the paste.
- **Light theme forced**, ignoring app appearance. Outlook's Word-based renderer can drop `background-color`; light-on-dark text would then be unreadable on a white mail body, whereas dark text on white degrades gracefully.
- Sources over 200k characters skip tokenizing and emit uncolored monospace, so copying a very large query stays responsive.
- Reuses the existing Shiki singleton from `sqlHighlighter.ts` (`getShikiSqlHighlighter` is now exported) instead of loading the SQL grammar a second time.

**`src-tauri/capabilities/default.json`** — adds `clipboard-manager:allow-write-html`. The `default` capability covers `["main", "detached-tab-*"]`, so editors in detached tab windows get the permission too.

Locale strings added for all 10 languages.

## Verification

| Check | Result |
| --- | --- |
| New unit tests (`sqlRichText`, `clipboard`) | 12 passed |
| Editor + i18n + clipboard suites | 955 passed |
| `pnpm typecheck` (vue-tsc) | pass |
| `pnpm lint` | no new warnings |
| `oxfmt --check` | pass |
| `cargo check -p dbx --locked` | exit 0 |

The Tauri capability was confirmed resolved rather than merely accepted: `src-tauri/gen/schemas/capabilities.json`, regenerated by tauri-build during that check, lists `clipboard-manager:allow-write-html` alongside the existing clipboard permissions.

Rendered output was inspected against a query containing a string literal, a `>` operator and a CJK comment — keywords `#D73A49`, identifiers `#005CC5`, strings `#032F62`, comments `#6A737D`; line breaks and indentation preserved; `>` and quotes correctly escaped.

## Notes

- On Linux/Wayland, HTML clipboard support depends on arboard and the compositor; X11 is fine and some Wayland sessions may fall back to plain text. macOS and Windows are unaffected.
- Because both flavors are written, pasting into a plain-text target yields the original SQL with no markup leakage.


t8y2 (approved):
Thanks for the clean PR — reviewed and approved. Rendering from `codeToTokensBase` into a single self-contained `<pre>` with inline styles only is the right call for mail clients, reusing the shared Shiki singleton avoids a second grammar load, `escapeHtml` covers `& < > " '` in one pass (token colors come only from the static theme, so SQL containing markup can't survive as markup), and `writeHtml(html, altText)` matches the clipboard plugin's 2.3.2 signature exactly. The default plain-text copy path is untouched and the plain flavor is the raw selection. Targeted specs 12/12 locally.

Two tiny nits, neither blocking: the zh-CN/ja/ko labels carry a "（保留高亮）"-style suffix the English label doesn't (feel free to drop for consistency), and worst-case tokenization at the 200k cap is synchronous on the main thread — fine given the editor already full-doc highlights. The frontend failure at your head is the stale merge-base baseline; the update-branch merge will clear it.

t8y2 (commented):
Maintainer conflict resolution applied: `adbd23485` → `a6d947665` (merge of current `main` into the branch, conflict resolved).

Conflict cause: `apps/desktop/src/lib/common/__tests__/clipboard.spec.ts` — both this branch and the merged Windows-CRLF fix (#8986) extended the same import line in that spec (your `copyRichTextToClipboard` vs their `clipboardLineEndings` + an `afterEach` restore block). `clipboard.ts` itself auto-merged cleanly.

Resolution: union — the import now carries `clipboardLineEndings, copyRichTextToClipboard, copyToClipboard, ClipboardEnvironment`, with the `afterEach` block kept. No production code touched.

Checks run at `a6d947665`: `clipboard.spec.ts` 11/11 ✅ (CRLF cases included) + `sqlRichText.spec.ts` 7/7 ✅, oxfmt clean. Merging once CI on the new head settles.

t8y2 (commented):
Pushed a small maintainer patch in 78acacfc4 covering the follow-up from review: `copyRichTextToClipboard` now normalizes its plain-text flavor with `clipboardLineEndings` at entry (same rule as `copyToClipboard` / #8986), so the plain flavor of a rich copy gets CRLF on Windows for both `writeHtml(html, text)` and the web `ClipboardItem`. The HTML flavor is untouched — rich targets handle LF and CRLF there could double newlines.

Added one spec case (Windows UA → plain flavor CRLF, HTML unchanged). Focused proof: `vitest run clipboard.spec.ts sqlRichText.spec.ts` — 19/19 passing.

t8y2:
Thanks for the contribution! Merged in 2249aa30c, will be released in the next version.

PR #8986, opened 2026-09-13
fix(clipboard): copy SQL with CRLF line endings on Windows

### LINKED ISSUE #8696: [[Bug] Windows 下从 SQL 编辑器复制的文本粘贴到原生编辑框(如 PowerBuilder)后换行丢失
### 数据库类型和版本

sqlserver 2008

### 问题描述

1. 复现步骤：
   - 在 dbx 的 SQL 编辑器中写一段多行 SQL（例如包含 SELECT / WHERE 的语句）。
   - 全选并 Ctrl+C 复制。
   - 打开 PowerBuilder（或任意 Win32 原生编辑控件）的编辑框。
   - Ctrl+V 粘贴。
   - 
2. 期望结果：粘贴后保留原始 SQL 的多行格式，换行正常显示。

3. 实际结果：所有换行符丢失，整段 SQL 被压缩成一行。

### 支持信息

DBX 版本: v0.6.9
运行环境: 桌面端
操作系统: Windows 10.0.26200.9278
系统架构: x86_64

### 优先级（是否紧急）

🟠 P1 重要 · 尽快处理

### 补充信息

补充诊断信息：从 dbx 复制出的文本换行符为 LF（\n），Windows 原生控件（包括 PowerBuilder 的编辑框）通常按 CRLF（\r\n）识别换行。建议 Windows 平台复制到剪贴板时使用 CRLF，或提供“剪贴板换行符格式”设置项。

thanadon-dev:
### The problem

Text is written to the clipboard with whatever line endings it already has, which for the SQL editor means bare LF. Windows native edit controls — the Win32 `EDIT` control, and applications built on it such as PowerBuilder — only treat CRLF as a line break, so a multi-line statement copied from DBX and pasted into one of them collapses into a single line.

`copyToClipboard` in `apps/desktop/src/lib/common/clipboard.ts` passes the text straight through to all three write paths (the Tauri plugin, the Web Clipboard API, and the legacy `execCommand` fallback) without touching line endings.

### The change

- Added `clipboardLineEndings(text, platform)`, which rewrites line endings to CRLF on Windows only. `\r\n` is matched ahead of bare `\r` and `\n` so text that already uses CRLF stays byte-identical, and single-line text is untouched. macOS and Linux keep LF.
- Applied it once at the top of `copyToClipboard`, so every write path is covered rather than just the native one.
- The platform check reuses the existing `getPlatform()` helper from `@/lib/backend/platform`; no new dependency.

### The non-obvious half

The data grid remembers the text it just copied so that pasting it back can restore `null` cells (plain TSV cannot distinguish a real NULL from the string `"NULL"`), and `parseDataGridClipboard` matches that memory **by exact string equality**.

Normalizing only inside `copyToClipboard` would leave the remembered text as LF while the clipboard holds CRLF, so on Windows every in-app copy→paste round trip of a multi-row selection would miss the lookup and silently turn NULL cells into the literal string `NULL`. `useDataGridExport.copyText` now remembers the normalized text, so the round trip still matches.

### Testing

Added to the existing `clipboard.spec.ts` suite:

- `copyToClipboard` writes CRLF when the platform is Windows, and leaves LF on macOS
- `clipboardLineEndings` unit cases: bare LF and CR become CRLF; text already using CRLF is unchanged; single-line text is unchanged; other platforms pass through

Verified RED → GREEN. With the source change reverted and the tests kept, the five new cases fail:

```
× writes CRLF line endings on Windows
× converts bare LF and CR to CRLF on Windows
× leaves text that already uses CRLF untouched
× leaves single-line text untouched on Windows
× returns the text unchanged on other platforms
Test Files  1 failed (1)
```

With the fix restored, the file passes again. The two pre-existing tests in that file pass both before and after.

Gates run on Windows (Node 22.22.2, pnpm 10.8.0):

- `pnpm exec oxfmt` on the three touched files — clean, `git diff --stat` unchanged at +65/−4 afterwards
- `pnpm lint` — exit 0
- `pnpm typecheck` — exit 0
- `pnpm test` — **1345 passed, 3 files failed**

Those 3 failures are pre-existing and unrelated to this change. I baselined them by stashing my work and re-running the same three files on unmodified `main`, where they fail identically:

- `packages/app-tests/sidebarRuntimeDecomposition.test.ts` — sidebar hot-path source budget
- `apps/desktop/src/lib/ui/__tests__/legacyWebViewRegexCompat.spec.ts` — lookbehind assertions in bundled sources
- `apps/desktop/src/lib/__tests__/plugins/pluginCenterIntegration.spec.ts` — 2 plugin center cases

I checked the regex one specifically since this change adds a regex: it fails on clean `main` too, and `/\r\n|\r|\n/g` contains no lookbehind.

Fixes #8696


t8y2 (commented):
Maintainer patch pushed: `2e58af568` → `ecdcf1503` (one commit `fix(nacos): accept platform-normalized text in copy verify`, on top of a merge of current `main` `933decc98`).

Root cause confirmed: your CRLF normalization in `copyToClipboard` is correct, but `NacosAdminConsole.vue`'s `copyConfigIdentity` verifies the copy by reading the clipboard back and comparing with strict equality against the *pre-normalized* `text` — on Windows the clipboard now holds CRLF while `text` still has LF, so every multi-line Nacos config copy would toast `nacos.copyVerifyFailed` despite succeeding (the same round-trip class you fixed for the data grid).

Patch: the verify now also accepts `clipboardLineEndings(text)` (import added) — behavior unchanged on macOS/Linux where both sides reduce to `text`.

Checks run at `ecdcf1503`: `clipboard.spec.ts` 8/8 ✅, full `apps/desktop/src/components/nacos/` suite 71/71 ✅, oxfmt clean. The main merge also brings in the spec fixes so frontend CI runs against a green base.

The issue-repro note from review still stands for verification: the literal #8696 repro (Ctrl+C inside the SQL editor) is a native CodeMirror copy that WebView2 already normalizes — the LF source was the app-level copies through the Tauri plugin path, which this PR fixes.

t8y2:
Thanks for the contribution! Merged in 78f974036, will be released in the next version.

PR #9003, opened 2026-09-14
feat(data-grid): hide selected result columns from the header context menu

### LINKED ISSUE #8671: [Feature] 希望可以增加多选结果列并右键可选择隐藏这些列的功能
### 数据库类型和版本

通用

### 支持信息

DBX 版本: v0.6.9
运行环境: 桌面端
操作系统: Windows 10.0.26200.9445
系统架构: x86_64

### 当前痛点

需要隐藏某些列时需要去“字段筛选”中操作

### 期望方案

希望可以增加多选结果列并右键可选择隐藏这些列的功能，参考示例：datagrip

Image

### 优先级（是否紧急）

🟡 P2 普通 · 按排期处理

### 补充信息

_No response_

Abeautifulsnow:
Closes #8671

## Background

Issue #8671 asks for a DataGrip-style workflow on the result grid: select multiple
columns in the header and hide them via right-click, instead of having to open the
字段筛选 (Columns) popover for every hide. Today the grid only offers single-column
toggle inside the popover, and there is no in-grid way back once columns are hidden.

## What's Changed

**Three new entries in the column-header context menu** (`DataGrid.vue`, built by
`createDataGridColumnContextMenuItems` in `dataGridContextMenu.ts`):

- **Hide This Column** — hides the right-clicked column; disabled when only one
  column is visible (the helper would silently refuse, so the menu says so instead).
- **Hide Selected Columns (N)** — count-bearing, shown only for a real multi-column
  selection (shift/Ctrl-click, which already existed); disabled when the selection
  covers every visible column, rather than silently hiding N-1.
- **Show All Columns** — the recovery entry, shown only while `hiddenColumnCount > 0`,
  mirroring DataGrip's `Show All Columns` (DBE-13331). Without it, recovering hidden
  columns still required the popover — half of the reported pain.

**Batch hide primitive** — new pure helper `hiddenColumnIndexesAfterHiding()`
(`dataGridColumnVisibility.ts`): one deterministic decision for the whole set,
idempotent for duplicate/already-hidden/unavailable indexes, keeps at least one
column visible. The composable exposes `hideColumns()` — a single state commit +
single persistence write, never a loop over `toggleColumnVisibility`.

**Wiring guarantees** (same contract as the existing freeze-selected-columns flow):
the hide routes through `applyColumnOrderChange()` so stale selection indexes are
reconciled away (copy-names / freeze / drag-to-editor stay correct), cell selection
is cleared, and after the layout refresh the horizontal scroll is clamped via the
existing `clampGridScrollerBounds()` so no blank gutter remains. `showAllColumns`
is now also wrapped for the same selection reconciliation.

**Persistence semantics unchanged**: hidden columns are remembered per
connection/database/table-or-SQL scope key (same SQL keeps them hidden, a different
query starts fresh). No new persisted state, no backend changes.

**Follow-up fix (empty results)**: `selectColumns()` carried a leftover
`displayItems.length <= 0` guard from commit f41601e4d that made the entire header
column selection dead on a 0-row result (e.g. `SELECT ... WHERE 1=0`), blocking
multi-select hide exactly where users inspect schemas. The guard is removed —
column selection is column-based and row-independent; zero-row selections still
never produce phantom cell selection (covered by tests).

**i18n**: three new `grid.*` keys added to all 10 locales (parity suites enforce).

## Notes for Reviewers

- Alignment with peer tools: DataGrip has no header multi-select for hiding at all
  (DBE-12973 open) and strands users who hide the last column; DBeaver disables
  hiding all columns with a dialog. Here the same guard is expressed as a disabled
  menu item, and multi-select already exceeds DataGrip.
- Known UX note (unchanged from the popover's existing behavior): with the
  "hide null columns" setting on, auto-hidden all-null columns count toward
  `hiddenColumnCount`, so "Show All Columns" may appear even if the user hid
  nothing manually.
- Manual verification item: the post-hide horizontal scroll clamp is DOM behavior
  without unit coverage — verified by wiring/timing analysis only.
- Two pre-existing local test failures are unrelated (`legacyWebViewRegexCompat`,
  flaky `TableImportDialog.existingTarget`); full suite: 14174 passed / 2 failed
  before and after this branch's changes.

## Testing

- New unit tests: batch-hide semantics (5 cases: batch, idempotency, empty input,
  keep-last-visible, keep-first-on-hide-all); composable `hideColumns` persistence
  + interop (3 cases); menu gating (5 cases: single disabled at one visible column,
  no false-disable when the count is omitted, batch disabled on full coverage,
  recovery entry visibility); zero-row header selection (4 cases: click, range,
  shift-merge, no phantom cell selection). The zero-row tests were validated to
  fail without the fix.
- `pnpm typecheck` ✅ · `pnpm lint` ✅ (no findings in touched files) · i18n parity ✅
- Merged latest `main` (37 commits) — no conflicts; full suite green relative to
  the pre-existing baseline.

## Screenshots
13e65930-8adc-4c7a-a93e-7626a702bc9e


t8y2:
Thanks for the contribution! Merged in 638aa48f5, will be released in the next version.

PR #9004, opened 2026-09-14
Fix：修复 Oracle 查询结果直接跳转末页显示首页

### LINKED ISSUE #8993: [Bug] Oracle数据查询后翻页bug

Oracle12c

Oracle数据查询后，点击翻页到最后一页，会显示第一页数据，且之后再翻页都只显示一页数据。

DBX 版本: v0.6.12
运行环境: 桌面端
操作系统: Windows 10.0.22000.2538
系统架构: x86_64
数据库类型: mysql, oracle, redis, sqlserver, sqlserver/sqlserver-legacy
本地驱动版本: oracle 0.1.61, sqlserver-legacy 0.1.30

_No response_

zipg:
Fixes #8993

## 问题

Oracle 查询结果直接点击后续页或末页时，页面会重新显示第一页数据。原因是 Oracle agent 使用前向游标，非连续页码请求会丢失已有 session，而 Oracle 通用分页 SQL 无法表达 offset。

## 修复

当用户直接跳转到后续页时，复用一个新的 Oracle result session，自动消费目标页之前的数据，再返回目标页内容，保持连续翻页和直接跳页结果一致。

## 实测

- 使用真实 Oracle 11g 数据库。
- 执行 `SELECT LEVEL AS ID FROM DUAL CONNECT BY LEVEL <= 250`。
- 连续翻页返回 `1–100`、`101–200`、`201–250`。
- 直接跳转末页返回 `201–250`，不再显示首页。
- 已同步最新 `main` 并解决冲突，`git diff --check` 通过。


t8y2 (commented):
Pushed a maintainer patch in 99bc086c3 that closes the blocker from review — your frontend loop was correct but unreachable, because the backend plan builder only set `useAgentResultSession` for `offset == 0` or an existing session, while Oracle user queries are `Unbounded` (SQL never rewritten), so a jump page never reached your new code path.

What the patch does:

1. **Backend root cause** (`query_result_sql.rs`): when `use_agent_cursor` is set and the UserQuery pagination strategy is `AgentMaxRows`/`Unbounded` (Oracle user queries, generic JDBC — i.e. the statement cannot be rewritten with dialect pagination), the plan now keeps `use_agent_result_session = true` for `offset > 0`, which activates your session-consume loop. Rewritten dialects (PostgreSQL LIMIT/OFFSET, HighGo, SQL Server TOP, …) are unchanged, covered by a counter-test.
2. **Loop hardening** (`queryStore.ts`): batch multi-statement executions are excluded from the loop (they now take the normal `executeMultiWithProgress` path instead of being hijacked); `pageLimit!` is replaced with a guarded limit; if the session reports `has_more` without a `session_id`, the loop now throws instead of silently returning a short page as the wrong offset window; when the data ends before the requested offset it returns an empty page. Manual-transaction offset jumps keep the legacy single-shot call so the new flag cannot strand a cursor session in the txn path.
3. **Tests**: the previously failing agent-variant mock now serves three sequential session pages (rows 1–100, 101–200, 201–237) and asserts the returned window (201–237), the exact total (237), three execute-multi calls, and `resultSessionId` propagation after the first page — i.e. the #8993 last-page jump scenario end to end at the store level. Plus three new Rust plan tests (Oracle jump, JDBC jump, PostgreSQL negative).

Proof: `cargo test -p dbx-core --no-default-features --lib query_result_sql::tests` — 192 passed; `vitest run packages/app-tests/queryStore.test.ts` — 207 passed; `vue-tsc` clean; fmt/oxfmt clean. The prior frontend red is gone with these changes.

t8y2:
Thanks for the contribution! Merged in 58d902c07, will be released in the next version.

PR #9023, opened 2026-09-14
fix(import): hide the data-tab import entry for unsupported engines

Abeautifulsnow:
On SAP HANA a user could still open the table-import wizard from the data tab and get
stuck: **表数据 → 工具箱 → 导入数据** opened 导入表数据, the wizard reached the 选项 step, and
下一步 stayed disabled with no visible error.

HANA declares `tableImport: false` in the driver manifest. The two other surfaces that can
open this wizard — the sidebar context menus (`SidebarTreeRuntimeHost.vue:4607`) and the
Object Browser row menu (`ObjectBrowser.vue:381`) — already honour `supportsTableImport`.
The data-tab toolbox item was the only one without the check, and it is the one a user
reaches from a data tab.

Why the leak produced a dead end: the toolbox passes `schema: tableMeta.schema`, which is
empty for HANA, and `metadataSchemaForConnection` blanks it again for tree-schema engines,
so the target-column metadata request reaches the Java agent with an empty schema →
`schemaPattern = null` → an unscoped catalog scan. The dialog's own budget is 15 s and
`existingTargetMetadataReady` never flips, so the wizard can never advance.

This PR closes the leak: the toolbox import item is now gated on the same capability
predicate as the other two surfaces (one shared computed, no new capability logic).

Deliberately out of scope (tracked separately): implementing HANA table import
(dialect/type mapping), gating the toolbox's 生成数据 item (no capability key exists for it
yet), and a backend/web-API guard that rejects imports by capability.

## 变更类型 / Change Type

- [ ] 新功能 / New feature
- [x] Bug 修复 / Bug fix
- [ ] 代码重构 / Code refactoring
- [ ] 文档更新 / Documentation
- [ ] CI / 构建 / CI or build

## 涉及前端 / Frontend Changes

- [ ] 本 PR 涉及前端改动，已附截图/录屏（见下方） / This PR includes frontend changes and screenshots or a recording are attached below.

**Not attached yet.** The visible effect is the absence of **导入数据** in 表数据 → 工具箱 for an
engine whose manifest declares `tableImport: false` (HANA, Teradata, Vertica, Firebird, …).
Capturing it requires a live connection with a data tab open; happy to add a before/after pair
if the maintainer wants one before merge.

## 验证 / Validation

- [ ] `make check` 通过 / `make check` passes

What was actually run on Windows (Node/pnpm toolchain):

- `pnpm vitest run apps/desktop/src/components/layout/__tests__ apps/desktop/src/components/objects/__tests__ apps/desktop/src/lib/__tests__/database/databaseFeatureSupport.spec.ts`
  → **34 files / 282 tests passed**, including the new `ContentArea.tableImportEntry.spec.ts`
- The new spec fails without the guard (verified by `git stash`-ing the source change)
- `npx oxlint --vue-plugin` and `npx oxfmt --check` on both changed files → clean
- `pnpm typecheck` (vue-tsc) → exit 0
- `pnpm check:connection-types` → exit 0 (no capability descriptor touched)

Not run here: `make check` (repo-wide run) and `make cargo-check-fast` — the diff touches no
Rust/cargo files, so CI is the source of truth for those. Say the word if you want the full
local `make check` run attached as well.


t8y2:
Thanks for the contribution! Merged in 8d32cf9ff, will be released in the next version.

PR #9038, opened 2026-09-14
fix(mongodb): offer query operators by position in the filter editor

thailoc-dev:
## Problem

Query-operator completion ignores where the cursor is, so it is wrong in both directions:

| you type | suggestions | should be |
| --- | --- | --- |
| `db.c.find({$` | *(none)* | `$and`, `$or`, `$nor`, `$expr`, … |
| `db.c.find({_id: {$o` | `$or` | `$oid` only — `{_id: {$or: …}}` is invalid |

The top level of a filter is the one place `$and` / `$or` are valid, and it offers nothing. Under a field, the whole flat list is offered, including operators MongoDB rejects there.

`QUERY_OPERATORS` is a single list with no notion of position, and the filter's top level is classified as a plain `field` position — the same mode as an insert document, where no operator is valid.

## Change

**Split the table by position.** `FIELD_QUERY_OPERATORS` holds the constraint operators valid inside `{ field: { … } }` (`$eq`, `$gt`, `$in`, `$exists`, `$elemMatch`, …). `TOP_LEVEL_QUERY_OPERATORS` holds the whole-filter operators (`$and`, `$or`, `$nor`, `$expr`, `$text`, `$where`, `$jsonSchema`). `QUERY_OPERATORS` stays exported as the union for callers that do not care.

**A `filterField` mode** for positions where a *filter* is being written: the filter root, the sub-filters of `$and` / `$or` / `$nor`, `$elemMatch` bodies, and the `$match` stage. It offers the fields first and the whole-filter operators after, so fields lead on an empty prefix and operators take over once `$` is typed. Plain `field` mode is unchanged, so update bodies and insert documents still get fields only.

**Under a field**, `queryOperator` mode now draws from `FIELD_QUERY_OPERATORS`, so `$or` and friends no longer appear where they would be rejected.

Two existing tests asserted the old behaviour directly (the root mode name, and a filter argument offering exactly the field list) and were updated to describe the new one.

## Testing

- `pnpm test` — 14201 passed
- Three new tests: whole-filter vs. field operators at each level (including that `{_id: {$o` now yields only `$oid`); `$or` sub-filters, `$elemMatch` bodies and `$match` behaving as filters; update and insert documents unchanged
- The existing `$match` + `na` → exactly `["name"]` and update-vs-query separation tests pass unchanged

`DocumentBrowserFieldSearch.spec.ts › shows the value type selector and preserves a sampled string _id` is a pre-existing UI-timing flake — it fails 2 of 5 runs on `main` with this change stashed, and neither it nor `DocumentBrowser.vue` touches the completion module.

Possible follow-up, not included: `$elemMatch` over an array of scalars takes field operators directly (`{tags: {$elemMatch: {$gt: 5}}}`); its body currently offers fields and whole-filter operators but not those.


t8y2 (approved):
Thanks — verified the position logic end to end and it's correct: filter root, $and/$or/$nor sub-filters, $elemMatch bodies and $match all classify as filters now, update/insert/projection stay fields-only, and `{_id: {$o` yielding only `$oid` is a nice touch. Tests cover each level and CI is green.

Two nits, neither blocking:

1. The file-header table list in mongoCompletionTables.ts (lines 14-22) still says `QUERY_OPERATORS → filter documents`; worth mentioning the new FIELD_/TOP_LEVEL_ split so the "grouped by where they may legally appear" note stays true.
2. Follow-up idea beyond the one you listed: TOP_LEVEL_QUERY_OPERATORS could plausibly also serve the `$elemMatch`-over-scalars case by pairing it with FIELD_QUERY_OPERATORS there, but the current fields+logical-ops offer is already an improvement over fields-only.

Merging as-is is fine by me.

t8y2:
Thanks for the contribution! Merged in 608633edf, will be released in the next version.

PR #9103, opened 2026-09-15
fix(mongodb): accept .toArray() and .pretty() after aggregate() and find()

thailoc-dev:
## Problem

The `.toArray()` that mongosh and Compass append to an aggregation is rejected:

```js
db.orders.aggregate([{ $match: { status: "paid" } }]).toArray()
```

> Unsupported MongoDB aggregate form. … Chaining (for example .limit()) is not supported.

Compass's "Export to language" and the shell itself both emit this form, so it is precisely what a user gets when they copy a working pipeline into the editor. The message points them at a chaining problem when the fix is deleting six characters.

The same suffix already works after `find()` in the editor, so the two commands disagree.

## Cause

`.toArray()` and `.pretty()` are cursor methods that do nothing in DBX — results are always materialised. But the aggregate parser rejects *anything* after the closing `)`, which is the right call for `.limit()` and `.sort()` (those belong in the pipeline as `$limit` / `$sort` stages) and the wrong call for a method that changes nothing.

On the Rust side (CLI, MCP, web) `find()` rejects them too, so `db.c.find({}).toArray()` fails there while succeeding in the editor.

## Change

Both parsers accept and drop `.toArray()` and `.pretty()`, with no arguments, in any order, after both `aggregate()` and `find()`. Everything else after `aggregate()` is still rejected, and the Rust error now names the offending method the way `find()` already does:

> Unsupported MongoDB aggregate() chain: limit(). Use pipeline stages such as $sort and $limit instead.

No new command fields; the parsed command is identical to the un-suffixed form.

## Testing

- `pnpm test` — 14310 passed
- `cargo test -p dbx-core --lib` — 6351 passed
- New tests both sides: the no-ops accepted alone, combined, across a newline, with a trailing `;`, and alongside `aggregate` options and a `find` chain; `.limit()`, `.sort()`, a no-op followed by `.limit()`, and `.toArray(1)` still rejected

Not changed, deliberately: the editor's `find()` parser silently ignores *every* unknown chained method, including ones with real semantics such as `.hint()`. That is a broader policy question and would be its own change.


t8y2 (approved):
Nice find and a clean fix — shipping the `.toArray()`/`.pretty()` suffix as a true no-op is the right call since DBX materializes results itself. Rust side reusing the existing chained-calls parser is exactly right. Merging.

t8y2:
Thanks for the contribution! Merged in f95650f03, will be released in the next version.

PR #9129, opened 2026-09-15
fix(editor): keep Cmd+H for macOS hide instead of replace (#9068)

### LINKED ISSUE #9068: [Bug] Mac 下 cmd+h 一般是隐藏当前 App，但这个快捷键在 DBX 里面的查询界面，会触发一个查找替换功能？是不是不太对

数据库无关，软件本身

[Bug] Mac 下 cmd+h 一般是隐藏当前 App，但这个快捷键在 DBX 里面的查询界面，会触发一个查找替换功能？是不是不太对

Image

_No response_

Abeautifulsnow:
## Summary

On macOS, `⌘H` is the system-wide **Hide App** shortcut, but inside DBX's editors it opened the find/replace panel instead, so the app could not be hidden with the standard shortcut.

**Root cause.** `apps/desktop/src/lib/editor/queryEditorSearchKeymap.ts` registered a hardcoded CodeMirror binding `{ key: "Mod-h", preventDefault: true, run: openReplace }`, spread into `QueryEditor.vue`'s `Prec.high` keymap. `Mod` resolves to `⌘` on macOS, and because the binding calls `preventDefault()`, the key event never reached AppKit — so the application menu's Hide key equivalent could not fire even though DBX already ships the item (`PredefinedMenuItem::hide` / `hide_others`, macOS-only `build_app_menu` in `src-tauri/src/lib.rs`). The binding was added for DBX's own "search in embedded source views" work; it is not upstream behaviour (`@codemirror/search`'s stock `searchKeymap` only binds `Mod-f`).

**Fix, in two layers.**

1. *Built-in fallback key.* `replaceFallbackKey(platform)` now returns `Alt-Mod-f` (`⌥⌘F`, the platform-conventional find-and-replace key) on macOS and keeps `Mod-h` everywhere else. Both hardcoded keymaps use it — the query editor and the inline one in the Nacos config console. Replace stays reachable on macOS through the `⌥⌘F` fallback, the configurable `⌘R`, the search panel's expand-replace button, and the editor context menu.
2. *Configurable layer.* The same hijack was still possible through Settings: a user-configured (or cloud-synced) `replace: "Mod+H"` matches the DOM handler and re-consumes `⌘H`. `shortcutRegistry.ts` now exports `MACOS_RESERVED_SHORTCUTS` (`Mod+H` Hide, `Alt+Mod+H` Hide Others) and `isReservedShortcut()`, which compares on an order- and alias-insensitive canonical form (`Mod`/`Meta`/`Cmd` collapse; `Ctrl`/`Control` stay distinct, so `⌃H` — CodeMirror's `deleteCharBackward`, not a menu accelerator — remains bindable). `normalizeShortcutSettings` repairs a reserved value to that action's platform default and then **clears it if the default is already occupied by another explicit same-scope binding**, so the repair cannot steal a binding the user deliberately set. Settings rejects a reserved combination at input time with a new `settings.shortcutReserved` message, and `normalizeSqlShortcuts` drops reserved values from SQL shortcuts (they are user-bindable too, and hit the identical path).

`⌘M`/`⌘Q` are deliberately **not** in the reserved set: they are outside the reported problem and adding them would silently unbind existing user configurations.

Behaviour on Windows/Linux is unchanged (`Ctrl+H` keeps opening replace).

## Change Type

- [ ] New feature
- [x] Bug fix
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation
- [ ] CI / build

## Frontend Changes

- [ ] This PR includes frontend changes and screenshots or a recording are attached below.

This PR changes shortcut handling rather than layout. The only new visible surface is a toast in Settings when a reserved combination is pressed (new i18n key `settings.shortcutReserved`, added to all ten locales). No screenshot is attached because the behaviour under test is macOS-only keyboard handling and the change was developed on Windows — see Verification for what still needs a macOS reviewer. No visual regression is expected, and none of the touched files change markup or styles.

## Verification

Automated (repo root, Windows):

- `npx vitest run apps/desktop/src/lib/__tests__/editor apps/desktop/src/lib/editor/__tests__ apps/desktop/src/lib/__tests__/sql apps/desktop/src/components/editor/__tests__ apps/desktop/src/components/nacos apps/desktop/src/stores` → **216 files / 2895 tests passed**
- `npx vue-tsc --noEmit --project apps/desktop/tsconfig.json` → exit 0
- `npx oxlint --vue-plugin` on the touched directories → 0 errors (only pre-existing warnings)
- `npx oxfmt --check` on every changed file → clean

New tests: `lib/editor/__tests__/queryEditorSearchKeymapMac.spec.ts` drives a real CodeMirror view under a stubbed `MacIntel` navigator and asserts `⌥⌘F` opens replace, `⌘F` opens search, and `⌘H` is **not** consumed (`defaultPrevented === false`); a composed regression takes `normalizeShortcutSettings({ replace: "Mod+H" }, "MacIntel").replace` through the real replace bindings/handler and asserts `⌘H` is still not consumed, while the same input on `Win32` does consume `Ctrl+H`. Registry tests cover the reserved-key verdict table and the occupancy rule. Two pre-existing specs that dispatched synthetic `Ctrl` chords while CodeMirror resolves `Mod` from the *host* platform (`QueryEditorSearchKeymap.spec.ts`, `codemirrorSearchKeymap.spec.ts`) are now parameterised over both platforms, so they no longer depend on the machine running them.

Non-vacuity was checked by reverting each guard in turn and confirming exactly the new assertions fail (reserved-key filter removed → the macOS repair test fails; occupancy pass removed → the steal test fails), then restoring them. Every assertion above was reproduced locally; the throwaway probe files used for this were deleted.

Not run: `make check` / `make cargo-check-fast` — this PR touches no Rust or backend code.

**Still needs a macOS reviewer on real hardware:** that `⌘H` actually hides the app (the code side can only prove DBX no longer consumes the key — delivery of the key equivalent to AppKit is inferred), and that `⌥⌘F` / `⌘R` open replace and `⌘F` opens search on a real keyboard. CI has no macOS frontend job (`ci.yml` runs ubuntu + windows only), so this manual pass is the gate.

## Related Issue

Closes #9068


t8y2 (approved):
Thanks for the thorough fix! Verified the two-layer approach covers both hardcoded and configurable replace paths, and CI is green. Merging.

t8y2:
Thanks for the contribution! Merged in 3f7419583, will be released in the next version.

PR #9136, opened 2026-09-15
fix(cassandra): skip table alias completions for CQL

### LINKED ISSUE #9134: [Bug] Cassandra: table completion inserts "AS alias", which CQL rejects

Cassandra 5.0.2

1. Steps:
   - Connect to a Cassandra database and open a SQL editor tab.
   - Type `SELECT * FROM ` and pick a table (e.g. `demo_ks.user_events`) from the completion dropdown.
   - DBX inserts `SELECT * FROM demo_ks.user_events AS ue`.
   - Execute the query.
2. Expected: the completion inserts only `demo_ks.user_events` and the query runs. CQL does not support table aliases at all (neither `table AS alias` nor `table alias`), so no alias should be generated for Cassandra regardless of the "Automatically add table aliases" editor setting.
3. Actual: the query fails:
   ```
   Database returned an SQL error (stage: execute).
   line 2:26 mismatched input 'AS' expecting EOF (SELECT * FROM demo_ks.user_events [AS]...)
   ```
4. Workaround: disable Settings → Editor → "Automatically add table aliases". The setting is on by default, so every new Cassandra user hits this on the first table completion.

Root cause: `formatTableAliasApply` / `buildTableItems` in `apps/desktop/src/lib/sql/sqlCompletion.ts` only special-case Oracle-like databases (drop `AS`). Cassandra is not handled, so an `AS <alias>` suffix is always appended when the setting is on.

I would like to fix this (skip alias generation when `databaseType === "cassandra"`, plus a regression test).

/claim

```
DBX 版本 / Version: v0.6.13
运行环境 / Runtime: Desktop
操作系统 / OS: macOS 26.6.2
系统架构 / Architecture: aarch64
```

SQL inserted by completion:

```sql
SELECT *
FROM demo_ks.user_events AS ue
```

thanhthai3010:
## 变更说明 / Change Description

CQL has no table alias syntax, but the "Automatically add table aliases" editor setting (on by default) made table completions insert `SELECT * FROM ks.table AS t` for Cassandra connections. Cassandra rejects this:

```
Database returned an SQL error (stage: execute).
line 2:26 mismatched input 'AS' expecting EOF (SELECT * FROM demo_ks.user_events [AS]...)
```

This PR adds a `supportsTableAliases(databaseType)` guard in `sqlCompletion.ts` and uses it to:

- skip generated aliases in `buildTableItems` / `buildForeignKeyRelatedTableItems` when the database is Cassandra, regardless of the `autoAliasTables` setting;
- skip standalone alias suggestions (`AS t`) after a table name for Cassandra.

Other databases are unchanged. The existing Oracle-like special case (drop `AS`) is left as is.

## 变更类型 / Change Type

- [ ] 新功能 / New feature
- [x] Bug 修复 / Bug fix
- [ ] 代码重构 / Code refactoring
- [ ] 文档更新 / Documentation
- [ ] CI / 构建 / CI or build

## 涉及前端 / Frontend Changes

- [ ] 本 PR 涉及前端改动，已附截图/录屏（见下方） / This PR includes frontend changes and screenshots or a recording are attached below.

No UI, style, or copy changes; only the text inserted by the SQL completion for Cassandra changes (`ks.table AS t` → `ks.table`).

## 验证 / Validation

- Added two regression tests in `sqlCompletion.context.spec.ts`:
  - Cassandra table completion with `autoAliasTables: true` applies `order_items` (no alias).
  - Cassandra does not offer an `alias for order_items` snippet after `FROM order_items `.
  - Both fail on `main` and pass with this change.
- `make check` (connection-types, format, lint, typecheck, vitest: 1361 files / 14312 tests) passes.
- `make cargo-check-fast` not run: no Rust changes.

- [x] `make check` 通过 / `make check` passes
- [ ] `make cargo-check-fast` 通过 / `make cargo-check-fast` passes (N/A, frontend only)

## 关联 Issue / Related Issue

Close #9134


t8y2:
Thanks for the contribution! Merged in 715d834c3, will be released in the next version.

PR #9151, opened 2026-09-15
fix(agents): GBase 8s tables fail to open with "illegal character in statement"

yiqiui:
## Summary

- GBase 8s connections fail to open any table: every generated table-data query (`SELECT * FROM \`table\`) returns `java.sql.SQLException: 语句中有非法的字符` (illegal character in statement).
- Root cause: the GBase 8s JDBC driver (and other Informix-family drivers) report `"\`"` from `DatabaseMetaData.getIdentifierQuoteString()`, but the server rejects backtick-quoted identifiers outside MySQL compatibility mode (`SQLMODE=mysql`).
- Fix: in `AbstractJdbcAgent.resolveIdentifierQuote`, sanitize the driver-reported backtick to an empty quote for Informix-family connections (matched via connection string / driver class / JDBC URL), so clients emit unquoted identifiers that both SQL modes accept.

## Behavior

- Informix-family driver reports `` ` `` and no MySQL compat → quote becomes empty (unquoted identifiers).
- `mysql_compat_mode` set, or connection string contains `SQLMODE=mysql` → backtick is kept.
- Non-backtick quotes (e.g. `"`) are left untouched; non-Informix drivers unaffected.

## Test plan

- [x] New unit tests in `AbstractJdbcAgentTest`: gbasedbt JDBC URL, `jdbc:informix-sqli` connection string, `SQLMODE=mysql` keeps backtick, non-backtick quote preserved.
- [x] `./gradlew :common:test :gbase8s:test :informix:test` all pass.
- [x] Manually verified against a live GBase 8s 8.8 instance (192.168.5.65, driver 3.6.5_2L1_1, `DELIMIDENT=y`):
  - Fixed agent resolves an empty identifier quote (driver still reports `` ` ``).
  - `SELECT * FROM alert_code_record` returns 20 columns / rows successfully (previously failed to open).
  - The old backticked SQL still fails with `illegal character in statement`, confirming the root cause.


t8y2:
Thanks for the contribution! Merged in dd99786c0, will be released in the next version.

PR #9155, opened 2026-09-15
fix(mongodb): say what failed instead of the generic command hint

thailoc-dev:
## Problem

Every command the editor cannot parse gets the same message:

> Use MongoDB shell-style commands, for example: db.collection.find({}).limit(100), db.collection.aggregate([]), …

It says nothing about *what* was wrong. An unsupported value constructor, an unsupported method, a wrong number of arguments and a typo all read identically, so the user is left to diff their command against the examples by eye. #8694's `new Date()` bug took a code-reading session to diagnose for exactly this reason.

## Cause

`describeMongoCommandParseFailure` in the shared package only diagnoses aggregate-shaped input; everything else falls through to the generic hint. `mongoShellCommand.ts`, which knows every method the editor supports, re-exported it unchanged.

## Change

`mongoShellCommand.ts` now wraps the package function. Its specific messages (unclosed delimiters, aggregate problems) pass through untouched; only where it would have returned the generic hint does a diagnosis run, in this order:

1. **An unsupported value constructor**, named, with the argument it sits in:
   `Unsupported value Foo(...) in the update argument of updateOne(). Supported value constructors: ObjectId, ISODate, new Date, NumberLong, NumberInt, NumberDecimal, UUID, BinData, Timestamp, MinKey, MaxKey.`
2. **An argument that is not a document**, by role: `The filter argument of find() is not a valid document.`
3. **Trailing text** after a well-formed call: `Unexpected text after find(...): ".count".`
4. **The wrong argument shape** for a known method: `insertOne() expects one document.` / `updateOne() expects a filter, an update, and optional options.`
5. **An unsupported collection method**, with the supported list: `Collection method replaceOne() is not supported. Supported collection methods: find, findOne, …`
6. **An unsupported database method**, pointing at the supported equivalent where one exists: `db.stats() is not supported; use db.runCommand({ dbStats: 1 }).` — likewise `serverStatus`, `adminCommand`, `getSiblingDB` (→ `use <database>`), `createCollection`, `getCollectionNames`.
7. `show collections` → collections are in the sidebar; only `show dbs` is supported.

Anything that is not `db.`-shaped still gets the original hint. Leading comments are stripped before the shape is read, and bracket and `getCollection("…")` targets are recognised.

The Rust parser is unchanged: it already returns specific messages, and its copy of the generic hint is only a backstop for Mongo text that falls into the SQL executor.

## Testing

- `pnpm test` — 14360 passed
- Three new tests covering each branch above, including that constructor-like text inside a string is not reported, and that a leading comment does not hide the command shape
- The existing aggregate and generic-hint diagnostics tests pass unchanged

`EditorGroupTabBar.groups.spec.ts` and `sqlShortcutActions.spec.ts` each have one failure that reproduces identically on `main` with this change stashed; both are unrelated to MongoDB.


t8y2:
Thanks for the contribution! Merged in aeafddfae, will be released in the next version.

PR #9181, opened 2026-09-15
feat(mongodb): support BSON dumps and database backup/restore

HandSonic:
## Change Description

Add native MongoDB dump/restore support on top of the existing collection import/export workflow.

- Collection import/export supports raw BSON dumps and gzip-compressed BSON, including files produced by official `mongodump`. Existing CSV, JSON and NDJSON behavior, including upstream CSV column type overrides, is preserved.
- Database context menus open backup/restore dialogs for directory dumps and MongoDB archives, with gzip support. Web exports archives and can restore either an uploaded directory or archive; desktop also supports local directory export.
- Restore separates source selection, source/target database and collection selection, confirmation, and execution. Web directory preview sends only a manifest and metadata; selected BSON files are uploaded after confirmation.
- Stream BSON in bounded batches without a preliminary full-data scan. Optional `objcheck` runs during the same restore pass and is off by default; framing, gzip integrity and archive CRC checks remain mandatory.
- Restore collection options, indexes and dependency-ordered views. Reuse task progress/cancellation, read-only and production safeguards, and the existing `DBX_MAX_UPLOAD_MB` setting. Prevent duplicate submissions while the production confirmation is pending.
- Keep dump/restore logic in `dbx-core`; Web routes and Tauri commands adapt transport and lifecycle only. Add behavioral tests, official-tool interoperability tests, and English/Chinese user documentation.

### Compatibility and Safety

This is not a complete implementation of every `mongodump` / `mongorestore` CLI option. Oplog/PIT recovery, UUID preservation, internal databases, time-series and encrypted collections are outside this PR.

Restore appends by default, without upsert. An explicit option drops selected same-name target collections before restoring. Failure or cancellation does not roll back documents already written, and there is no automatic retry. Existing IDs can therefore produce duplicate-key errors on repeated restores. Directory preview does not count all BSON documents.

## Change Type

- [x] New feature
- [x] Bug fix
- [x] Performance improvement
- [ ] Code refactoring
- [x] Documentation
- [ ] CI or build

## Frontend Changes

- [x] This PR includes frontend changes and screenshots are attached below.

Screenshots use synthetic local data. Images are stored on a separate evidence branch in the fork, outside this PR's code diff.

### Collection BSON.gz Preview

Collection import preview for an official gzip BSON dump

### Database Backup

Database backup with collection selection and gzip archive output

### Directory Restore

Directory restore with source and target databases and optional objcheck

### Confirmation Before Execution

Restore confirmation showing target collections and append behavior

## Validation

Final CI for `9d330ec503062c92e41d43c80a694f7b44538fa0` completed successfully on 2026-09-15: [CI run](https://github.com/t8y2/dbx/actions/runs/34952308344). All 12 applicable PR checks passed, including I18n Autofill; three unrelated jobs were skipped by the repository's change filters. Windows 7 binary and installer audits also passed.

- [x] `make check` passes: its equivalent `pnpm check` passed in Linux CI. Historical Windows-local limitations are recorded below.
- [x] `make cargo-check-fast` passes: ran its equivalent `cargo check --locked --no-default-features --features sqlite-bundled -j2` successfully.
- [x] Relevant tests pass.

Re-run on this branch after integrating upstream:

- `cargo fmt --check`: passed.
- `node scripts/sync-connection-types.mjs --check`: passed.
- Frontend formatting, lint and direct Vue typecheck: passed. Built `@dbx-app/mongo-shell` before the direct typecheck because the isolated worktree dependency install skipped lifecycle scripts.
- Six focused Vitest files: **52 tests passed**, covering import dialogs, database dump/restore, backend upload preparation, sidebar commands and i18n parsing.
- `cargo test -p dbx-core --locked --no-default-features --features sqlite-bundled,openapi --test mongodb_dump_restore -j2 -- --include-ignored --nocapture --test-threads=1`: **8 tests passed**, using MongoDB 8.0.17 and official Database Tools 100.18.0. Verified official -> DBX -> official BSON/BSON.gz byte fidelity (1,207 documents each), and directory/archive with and without gzip in both directions, including six collections/views, options and indexes.
- `git diff --check` passed; merge-tree check against upstream `aeafddfae` is conflict-free.
- First CI exposed the standalone DuckDB driver's separate lockfile, which also depends on `dbx-core`. Commit `13b0f5de4` synchronizes only the new CRC/SHA-1 dependency entries; `cargo metadata --manifest-path agents/drivers/duckdb/Cargo.toml --locked --offline --format-version 1` passes, and the complete agents CI job now passes.
- Linux CI with Rust 1.97.1 passed `cargo clippy --workspace --locked --all-targets --no-default-features --features "$RUST_FULL_FEATURES" -- -D warnings` and the complete Rust test job.
- Final local follow-up: 17 focused locale/parser/dialog tests passed; six default MongoDB integration tests passed, with the two external-server interoperability tests left opt-in. The test helper returns a boxed error, with narrowly scoped lint expectations only on the two callbacks whose public API requires `MongoImportIssue` by value.

Local aggregate-check limitations (Windows, Rust 1.98; repository CI uses Linux and Rust 1.97.1):

- The full Vitest run reported **14,412 passing tests and two failures**. The new i18n parser failure was fixed and its regression rerun passes. The remaining failure is unchanged upstream `legacyWebViewRegexCompat.spec.ts`, whose forward-slash allowlist does not match Windows `path.relative` output.
- The unchanged check wrapper resolves `node` to a nonexistent `node_modules/.bin/node.CMD` on Windows. Running the connection-types script directly passes.
- An earlier local Clippy run used blanket baseline-category allowances, which also hid new integration-test diagnostics. That precheck was insufficient. The final follow-up retained complete diagnostics without those blanket allowances and found none in the integration-test target. Strict Linux full-workspace CI, rather than the earlier relaxed local result, is the passing validation above.

Earlier feature validation also covered core streaming/objcheck tests, a 20,000-view dependency chain and the Web upload/progress HTTP integration test. These are not counted as fresh post-integration runs above.

Earlier I18n Autofill runs failed on empty or truncated external translation responses. Commit `9d330ec50` supplies all 59 new keys in the six remaining languages and adds key/placeholder parity tests. I18n Autofill now passes without calling the translation service.

## Related Issue

No linked issue.


HandSonic:
Fixed both Linux Clippy findings in 35eddeffd: keep symlink detection immutable with a Windows-only shadow binding, and box the gzip encoder variant. Added behavioral regressions for ordinary files/directories, Unix symlinks, and finishing plain/gzip BSON files (including empty dumps). Local MongoDB-related core tests: 86 passed, 1 opt-in benchmark ignored. cargo fmt --check passed; scoped Clippy passed with the same previously disclosed baseline allowances, without suppressing either reported lint. Unix-only tests and strict full-workspace Clippy still require the new Linux CI run. The push also triggered a fresh I18n Autofill run; manual rerun is unavailable to this fork account.

t8y2 (commented):
Maintainer rebase + safety patch, applied to keep this PR mergeable against the reworked `main`:

- **Heads:** `9d330ec` → `b941af95` (6 PR commits replayed onto `main` @ `94a8e63`, plus 1 new commit)
- **Range-diff:** PR commits 2-6 map 1:1 (`=`); commit 1 carries only the adaptations required by `main`'s rework, detailed below.

**Conflict cause.** `main` reworked `crates/dbx-core/src/mongodb_import_export.rs` (legacy-agent export streaming via find cursors) and removed the `require_bson_types` mechanism, while this PR modified that exact line to add a `MongoImportFormat::Bson` disjunct — a single-hunk conflict.

**How both behaviors are preserved.**
- `main`'s refactor stands: `insert_documents_batch` no longer takes `require_bson_types` and the LEGACY_AGENT type-preservation guard is gone, because the legacy-agent insert path now round-trips BSON types through canonical Extended JSON (`document_to_canonical_extended_json` on the Rust side, `Document.parse` on the agent side).
- This PR's requirement — BSON dump imports keep native BSON types — now holds structurally on `main`: `stream_bson_file` yields native `Document`s that never pass through `MongoImportParseOptions` (type coercion only applies to CSV/JSON/NDJSON text), and both insert paths preserve types. The disjunct itself would have been dead code (its only consumer was the removed guard), so it was dropped rather than kept as an unused binding.
- Two further adaptations were needed where auto-merged PR hunks referenced symbols `main` had removed: `export_bson` now resolves the native client from `state` (matching `export_csv`/`export_ndjson`) and returns "BSON dump export requires the native MongoDB driver" for legacy-agent connections — the same scope as the PR's branch point, where all exports required the native driver (possible follow-up: transcode canonical Extended JSON back to BSON for the agent path). The six `MongoExportRequest` test literals added on `main` gained `gzip: false`.

**Added commit: archive-overwrite guard** (`b941af95`, `fix(mongodb): refuse overwrite of existing dump archive`). The Archive arm's `temp.persist(&target)` silently replaced an existing destination file on Unix, while the directory branch refused an existing destination. The dump now checks `target.exists()` up front for both formats ("Dump destination already exists; choose a new directory" / "... choose a new archive file") plus a TOCTOU re-check immediately before `persist`, mirroring the directory branch's existing pattern. The web UUID-temp path is untouched.

**Verification.**
- `cargo check --no-default-features`: PASSED (full workspace, includes all changed crates).
- `cargo test --no-default-features -p dbx-core --lib mongo`: the test-binary build was repeatedly OOM-killed on the maintainer machine (rustc SIGKILL), so this suite was not completed locally — CI is the gate for it.
- `git diff --check` clean; no conflict markers; range-diff verified as described above.

The real-server interop tests remain `#[ignore]` opt-in exactly as authored (follow-up: an opt-in CI job to run them against a live MongoDB).

t8y2:
Thanks for the contribution! Merged in f5673ca0a1, will be released in the next version.

PR #9192, opened 2026-09-15
fix(editor): measure shortcut capture input width to fit CJK placeholders

### LINKED ISSUE #9144: [Bug] 快捷键编辑状态字体显示不完整

通用

文字显示不完整

DBX 版本: v0.6.13
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64
数据库类型: chromadb, clickhouse, dameng/dm, duckdb, elasticsearch, etcd, gaussdb, influxdb, kingbase, milvus, mongodb, mq/kafka, mq/pulsar, mysql, mysql/mariadb, nacos, neo4j, oracle, plugin, postgres, postgres/cockroachdb, qdrant, redis, saphana, sqlite, tdengine, trino, zookeeper
本地驱动版本: dameng 0.1.67, db2 0.1.59, duckdb 0.1.18, kingbase 0.1.58, neo4j 0.1.48, oracle 0.1.62, saphana 0.1.54, sqlserver-legacy 0.1.31, tdengine 0.1.49

🟢 P3 低优 · 有空再修 / Low · Fix when possible

Image

Abeautifulsnow:
Closes #9144

### Problem

In the shortcut settings tables, entering a row's capture state ("按下快捷键") could
clip the placeholder's last glyph. The reporter's screenshot (zh-CN, Windows,
v0.6.13) shows the final character cut in half mid-stroke.

### Root cause

The capture input width was derived from char-count arithmetic:
`${label.length + 2}em` — i.e. "every character advances exactly 1em, plus 2em
for padding and slack". That assumption breaks wherever the environment's CJK
fallback font renders wider than 1em. Pixel-measuring the issue screenshot: the
chip was exactly 7em @ 13px (91px), but the five zh-CN glyphs advanced at
15px/char (ink 12px + 3px gap) instead of 13px — so the content box (69px) fit
only 4.5 characters and the input (which always clips at its content box) cut
the fifth glyph in half. Latin locales never exposed the defect because their
~0.6em/char advance leaves several em of slack.

The same formula fed all three capture inputs: the shortcuts table, the
formatter table, and the SQL-shortcut dialog.

### Fix

Stop guessing; measure. A single hidden mirror span inside the dialog carries
the same box + typography classes as the capture inputs
(`font-mono text-[13px] font-semibold px-2.5 border` + `whitespace-pre`) and
renders the placeholder label. Because it uses the real font stack in the real
layout, its border-box width includes fallback-font advances, letter-spacing,
and sub-pixel rounding that char counting cannot predict.

- New `useMeasuredWidth` composable (`apps/desktop/src/composables/`):
  ResizeObserver-driven border-box measurement following the existing
  `useToolbarOverflow` idiom — `flush: "post"` target watch, observer
  availability guard, revalidate sources for environments without
  ResizeObserver, zero-width measurements ignored ("not laid out yet" must not
  clobber a usable width), observer disconnected on unmount.
- `shortcutPressShortcutInputWidth` becomes a ref: initialized to the legacy
  formula (pre-measurement / test fallback), then updated to
  `${Math.ceil(width) + 2}px` from the mirror's measured width. The +2px
  cushion absorbs sub-pixel differences between the mirror and the input.
- Language-agnostic by construction; the label acts as a revalidate source so
  locale switches re-measure. Widths above the input's own `max-w-64` cap clamp
  exactly as before, so no locale regresses.

### Testing

- `useMeasuredWidth.spec.ts` — fallback before attach, measure on attach,
  zero-width guard (exercised via revalidate source), source-triggered
  re-measure, target swap, unmount inertness.
- `EditorSettingsDialog.shortcutCaptureInputWidth.spec.ts` — source assertions:
  exactly one mirror span with the shared classes; all three capture-input
  width bindings reference the measured ref; legacy formula survives only as
  the fallback initializer.
- `pnpm vitest run` on `components/editor/__tests__/` +
  `useToolbarOverflow.spec.ts`: 26 files / 207 tests pass; `vue-tsc` typecheck
  clean.
  
  ### Screenshots

3aa5a11e-4e92-4601-9eb0-a6ada4667ea3

e8021c71-bd0d-442a-b94a-7b1c7f58ea45


t8y2:
Thanks for the contribution! Merged in 0437f952e, will be released in the next version.

PR #9194, opened 2026-09-15
fix(db2): tolerate malformed character data via charsetDecoderEncoder

bigdogcome:
## Problem
When DB2 table/column comments (or other string metadata) contain bytes that cannot be decoded in the database codepage, the IBM JCC driver throws `CharConversionException` with `ERRORCODE=-4220`. This causes the entire schema tree to fail to load — clicking the schema node returns an RPC error and the user cannot browse any tables in that schema.

This can happen when comments were created with a client-side encoding that does not match the database codepage (e.g. GBK bytes in a UTF-8 database, or any other unsupported encoding). Other tools like DBeaver handle this gracefully by showing replacement characters instead of failing entirely.

## Fix
Add `db2.jcc.charsetDecoderEncoder=3` to DB2 connection properties. This tells the JCC driver to replace undecodable byte sequences with the Unicode replacement character U+FFFD (�) instead of throwing an exception.

The property is set by overriding `buildConnectionProperties` in `Db2Agent`, so it only affects DB2 connections and does not change behavior for other databases.

## Verification
- Tested against a DB2 database where one schema previously failed to load with `CharConversionException: ERRORCODE=-4220`
- After the fix, the schema loads successfully; affected table/column comments show � characters where the encoding was invalid
- All other schemas and queries continue to work normally

## 问题说明（中文）

当 DB2 的表注释、列注释或其他字符串元数据中包含数据库代码页无法解码的字节时，IBM JCC 驱动会抛出 `CharConversionException` 异常（错误码 `ERRORCODE=-4220`）。这会导致整个模式树加载失败——点击模式节点时返回 RPC 错误，用户无法浏览该模式下的任何表。

这种情况通常发生在注释的客户端编码与数据库代码页不一致时（例如在 UTF-8 数据库中存入了 GBK 编码的字节，或其他任意不支持的编码）。DBeaver 等同类工具对此做了容错处理，用替换字符显示而不是直接失败。

## 修复方式

在 DB2 连接属性中添加 `db2.jcc.charsetDecoderEncoder=3`。该参数告诉 JCC 驱动：遇到无法解码的字节序列时，用 Unicode 替换字符 U+FFFD（�）代替，而不是抛出异常。

通过在 `Db2Agent` 中重写 `buildConnectionProperties` 方法来设置此属性，因此只影响 DB2 连接，不会改变其他数据库的行为。

## 验证结果

- 在一个 DB2 数据库上测试，其中一个模式此前因 `CharConversionException: ERRORCODE=-4220` 无法加载
- 修复后该模式可正常加载；受影响的表/列注释在编码无效处显示为 � 字符
- 其他模式和查询均正常工作

---


t8y2 (commented):
The fix itself looks right: setting `db2.jcc.charsetDecoderEncoder=3` on the connection properties is exactly the mechanism IBM documents for substituting undecodable bytes with U+FFFD instead of throwing ERRORCODE=-4220, and overriding `buildConnectionProperties` keeps it scoped to DB2 connections only. Thanks for the live-database verification.

One thing I normalized before merging (70e30c3): the original commit converted the whole file from LF to CRLF (+506/−498 for what is really an 8-line change), which would have polluted future diffs and blame. I stripped the CR characters so the combined diff is now just the import + the property override, with LF endings restored.

One product-level note to keep an eye on: if this property also applies to the write direction, unmappable characters in user edits could be silently replaced with U+FFFD on INSERT/UPDATE — IBM's docs don't state the write-direction behavior explicitly. Worth watching for reports after release.

t8y2:
Thanks for the contribution! Merged in df22dd49b, will be released in the next version.

PR #9213, opened 2026-09-15
feat(editor): group shortcuts by scope and add conflict tiers

### LINKED ISSUE #9075: [Feature] 快捷键按照scope划分模块，使得使用体验更加直观

通用

DBX 版本: v0.6.12
运行环境: 桌面端
操作系统: Windows 10.0.26200.9168
系统架构: x86_64
数据库类型: chromadb, clickhouse, dameng/dm, duckdb, elasticsearch, etcd, gaussdb, influxdb, kingbase, milvus, mongodb, mq/kafka, mq/pulsar, mysql, mysql/mariadb, nacos, neo4j, oracle, postgres, postgres/cockroachdb, qdrant, redis, saphana, sqlite, tdengine, trino, zookeeper
本地驱动版本: dameng 0.1.65, db2 0.1.58, duckdb 0.1.17, kingbase 0.1.57, neo4j 0.1.48, oracle 0.1.61, saphana 0.1.53, sqlserver-legacy 0.1.28, tdengine 0.1.49

### 当前痛点 / Current pain point

目前的快捷键底层有按照scope冲突检测机制，不同域之间相同key的配置是允许的，这符合产品设计。但是使用过程中容易出现用户对此不够理解的方面，且所有快捷键目前揉在一块。

同时需要增加冲突检测（跨域）方面的提示

### 期望方案 / Proposed solution

既然底层有按照scope区分，则可以按照scope进行范围区分，更加容易理解和使用，后端逻辑不变，前端UI体现上更加易懂。

🔴 P0 紧急 · 下个版本必须做 / Urgent · Must do next release

_No response_

Abeautifulsnow:
Fixes #9075

## What

The shortcuts tab rendered all actions as one flat table with a per-row scope badge, so two very different situations looked the same:

- **Same-scope duplicate** — a real conflict. Only one action per scope can own a chord, so one of them must change. It was signalled by inline row text plus a disabled Apply, with no explanation of *why* Apply was disabled.
- **Cross-scope overlap** — intentional design. `find` (editor) and `focusSearch` (global) both default to `Mod+F` and are routed by focus at runtime, so there is nothing to fix.

This PR groups the tab by scope and splits conflicts into two tiers: one blocking, one informational.

## Changes

**1. Two-level classification (`EditorSettingsDialog.vue`)**
- The tab is grouped into scope cards in runtime dispatch order: `global → editor → grid → search → sidebar`.
- The group header carries the scope icon, localized label, raw scope id, visible action count, unbound count and conflict counts.
- Rows no longer repeat the scope badge.
- Only groups that still have visible rows render, so a search cannot leave empty groups or group counts that disagree with the rows below them.

**2. Same-scope duplicate = blocking**
- Offending pill turns red (`aria-invalid` + `data-conflict`), the group header shows a red count, and the footer states the blocking reason (`{count} conflicts within one scope ({pairs} duplicate pairs)`).
- Apply / Apply-and-close stay disabled — and only once the draft has actually been touched, so a pre-existing conflict loaded from disk does not lock the page before the user does anything.

**3. Cross-scope overlap = informational only**
- Thin amber pill border (`data-cross-scope`), amber group count, tooltip listing every partner as `scope · action` plus a "nothing to fix" explanation.
- The cross-scope tier never reaches the apply gate.

**4. Registry helpers (`lib/editor/shortcutRegistry.ts`)**
- `findCrossScopeShortcutConflicts()` and `countShortcutConflictPairs()`, both read-only and display-only.
- `normalizeShortcutSettings` still computes chord occupancy from the **same scope only** — this PR does not change that (see `.trellis/spec/dbx/frontend/keyboard-shortcuts.md`, Contract 3).

**5. i18n**
- 18 new `settings.shortcut*` keys across all 10 locales (en, zh-CN, zh-TW, ja, ko, es, it, pt-BR, tr, az).

## Not in this change

The high-fidelity prototype used for this work also ships a demo toolbar (theme switch, scene switch, **density compact/detailed**, **only-conflicts filter**). Those are prototype review aids, not product requirements. The *default* density (compact — conflict explanation lives in the hover tooltip, rows keep only the nearest locating hint) is the design decision, and it is what this PR implements.

The non-editing pill width formula and the measured capture-input width are governed by Contract 5 and are deliberately untouched.

## Verification

Commands actually run:

```bash
npx vitest run apps/desktop/src/lib/__tests__/editor/shortcutRegistry.spec.ts \
  apps/desktop/src/components/editor/__tests__/EditorSettingsDialog.shortcutScopeGroups.spec.ts \
  apps/desktop/src/components/editor/__tests__/EditorSettingsDialog.shortcutScopeGroups.integration.spec.ts \
  apps/desktop/src/components/editor/__tests__/EditorSettingsDialog.shortcutResetEditState.spec.ts \
  apps/desktop/src/components/editor/__tests__/EditorSettingsDialog.shortcutResetEditState.integration.spec.ts
pnpm typecheck
pnpm build
npx oxlint --vue-plugin apps/desktop/src
npx vitest run
```

| Gate | Result |
|---|---|
| Shortcut specs (5 files) | **61 passed** |
| `pnpm typecheck` (`vue-tsc --noEmit`) | exit 0 |
| `pnpm build` | exit 0 |
| `oxlint --vue-plugin apps/desktop/src` | 0 errors, 35 warnings (all pre-existing; the only warning in a changed file is untouched MCP code) |
| i18n key check | 18/18 new keys present in all 10 locales |
| `npx vitest run` | 73/75 files, 634 tests passed, 2 failed (see below) |

The new `EditorSettingsDialog.shortcutScopeGroups.integration.spec.ts` mounts the real component and drives the real controls — it enters a row's capture state, captures a chord already owned by another action in the same scope, and asserts the blocking tier (red pills, footer reason, disabled Apply); then it does a clean rebind and asserts the cross-scope tier renders while Apply stays reachable.

## Pre-existing failures (unrelated to this change)

`QueryEditorContextMenu.spec.ts` and `queryEditorCompletionSelection.spec.ts` both declare `// @vitest-environment happy-dom`, where Vite externalizes `node:fs` / `node:path` and vitest fails the suite with `No such built-in module: node:`.

Both files are untouched by this PR (`git diff` empty for them), and both still fail after stashing the entire change set — verified on a clean baseline. They are an environment-level failure, not a regression from this work; a separate fix is desirable so they stop polluting the full-suite signal.

## Uncovered risk

- **Desktop runtime walkthrough pending.** The change was verified through real component mounting plus type check, lint and production build. An interactive pass on Windows WebView2 is still recommended for: group rendering, group counts after search filtering, L1 blocking, L2 amber styling, and red/amber contrast in both themes.
- The web dev shape (`pnpm dev:web`) redirects to `/login` and needs a backend plus credentials, so it was not used as the verification surface.

## Checklist

- [x] Same-scope conflicts still block Apply; cross-scope overlaps never do
- [x] `normalizeShortcutSettings` occupancy semantics unchanged (same scope only)
- [x] No regression in capture-state measured width, Esc cancel, per-row reset or restore-defaults exit (Contracts 1–5 hold)
- [x] i18n keys complete across all 10 locales
- [ ] Interactive walkthrough on desktop WebView2

## Screenshots

b9ccd2f3-8885-4b46-9562-875662d0fb48

1c68dd69-e902-46aa-bd94-1f758e470677


t8y2 (commented):
Small pre-merge patch pushed as 4d0eded on top of your branch.

**Root cause:** `EditorSettingsDialog.vue` passes `content-class="max-w-[320px]"` to `LightTooltip`, but `LightTooltip` had no `contentClass` prop and is a multi-root component (`<span class="contents">` + `<Teleport>`), so the attribute was silently dropped (Vue dev warning) and the cross-scope conflict hints rendered at the default `max-w-xs` (200px) instead of the intended 320px.

**Patch** (`apps/desktop/src/components/ui/LightTooltip.vue`):
- Added an optional `contentClass?: string` prop (default `undefined`, all existing callers unaffected).
- The teleported tooltip div now merges classes through the repo's `cn()` helper, appending `contentClass` last. This detail matters: plain string concatenation would not work here, because Tailwind v4 emits `.max-w-[320px]` **before** `.max-w-xs` in the stylesheet, so the default would still win the cascade. `tailwind-merge` drops the conflicting `max-w-xs` so the caller's width actually takes effect.

**Test** (`apps/desktop/src/components/ui/__tests__/LightTooltip.spec.ts`): two new unit tests — one asserting the teleported tooltip carries the passed `max-w-[320px]` and no longer carries `max-w-xs`, one asserting the default `max-w-xs` is kept when no `contentClass` is passed.

**Checks run locally:**
- `pnpm vitest run apps/desktop/src/components/ui/__tests__/LightTooltip.spec.ts` — 3/3 pass.
- `pnpm vitest run apps/desktop/src/components/editor/__tests__/EditorSettingsDialog.shortcutScopeGroups.integration.spec.ts` — 5/5 pass.
- `pnpm vitest run apps/desktop/src/lib/__tests__/editor/shortcutRegistry.spec.ts` — 44/45 pass. The one failure (`findCrossScopeShortcutConflicts > reports the intentional cross-scope overlaps in the defaults`, `conflicts.viewTableDdl` undefined) is pre-existing on your head `2a35ef0` (verified via stash) and environment-dependent: the spec relies on the host `navigator.platform`, and on a macOS host `shortcutDisplayOrder` skips modifier re-ordering, so `Shift+Mod+D` vs `Mod+Shift+D` don't normalize to the same key. CI (Linux) is green. Worth a follow-up to pin the platform explicitly in that spec, not addressed here.
- `pnpm typecheck` (`vue-tsc --noEmit`) — pass.

No merge performed; leaving CI on the new head to run.

t8y2:
Thanks for the contribution! Merged in ea1de14517, will be released in the next version.

PR #9233, opened 2026-09-15
feat(mongodb): support estimatedDocumentCount, db.stats and db.serverStatus

thailoc-dev:
## What

Three shell commands that DBX rejected, each of which maps exactly onto a path DBX already executes:

```js
db.orders.estimatedDocumentCount()
db.stats()
db.serverStatus()
```

## How

No driver, IPC or agent changes — these are parser-level, because the work already exists downstream.

**`estimatedDocumentCount()`** becomes a filterless legacy count. `count_documents(..., accurate: false)` in `mongo_driver.rs` already takes the metadata fast path for an empty filter and calls `estimated_document_count()`, which is precisely what this command asks for. It takes no filter, matching the shell.

**`db.stats()`** and **`db.serverStatus()`** become `runCommand({ dbStats: 1 })` and `runCommand({ serverStatus: 1 })` — the commands the shell helpers are shorthand for — and so run through the existing, fully supported `runCommand` path. Both take no arguments.

`db.collection.stats()` is unaffected and still parses as collection stats.

Deliberately not included: `db.adminCommand()` looks like the same shape but targets the `admin` database rather than the current one, so mapping it to `runCommand` would silently run against the wrong database.

## Also updated

- Autocomplete offers `estimatedDocumentCount` after a collection, and `stats` / `serverStatus` after `db.`, plus root snippets for the latter two.
- The parse-failure tables from #9155 are kept truthful: the three are added to the supported lists and given argument shapes, and the now-obsolete "use db.runCommand({ dbStats: 1 })" hints are removed.
- `suggests only helpers the shell parser accepts` asserted `estimatedDocumentCount` was *not* offered because DBX could not run it. It now asserts the opposite, and pins `renameCollection` in its place. That guard tying completion to parser support is a good one and still holds.

## Testing

- `pnpm test` — 14467 passed
- `cargo test -p dbx-core --lib` — 6384 passed
- New tests both sides: each command across plain, bracket and `getCollection` targets, with a trailing `;` and odd spacing; arguments and chained calls rejected; `count()` / `countDocuments()` modes and `db.collection.stats()` unchanged
- `cargo clippy`, `cargo fmt`, `oxfmt`, `oxlint`, `pnpm typecheck` clean

Four frontend tests fail identically on `main` with this change stashed (`sqlShortcutActions`, and three VARBINARY copy tests), and `DocumentBrowserFieldSearch › shows the value type selector…` is flaky there — it fails roughly 2 runs in 5 on `main` as well. None touch MongoDB.


t8y2 (approved):
Thanks for the clean PR — verified end-to-end against the existing paths: `estimatedDocumentCount()` maps to the driver's metadata-backed count exactly as documented, `db.stats()`/`db.serverStatus()` wrap `dbStats`/`serverStatus` runCommands with graceful degradation for version-dependent fields, and the db-level matcher can't false-positive on `db.stats.find({})` or `db.collection.stats()`. Case handling matches on both parsers, and pinning the completion↔parser guard to `renameCollection` is the right move.

Two non-blocking notes for possible follow-ups:

1. The real shell helpers accept optional arguments this rejects (`db.stats(scale)`, `db.serverStatus({repl: 0})`, `estimatedDocumentCount({maxTimeMS})`). The "takes no arguments" error is honest and `db.runCommand(...)` remains the escape hatch — `db.stats(1024)` is the one most likely to be hit later. (Minor: the Rust error for `estimatedDocumentCount` says "takes no filter" where "takes no arguments" would be more accurate.)
2. Since these map onto `RunCommand`, they inherit its blanket mutating/dangerous classification, so MCP still blocks them — same as `db.runCommand({dbStats: 1})` today, no regression. A future whitelist for known read-only commands (`dbStats`, `serverStatus`, `ping`, `buildInfo`) would make them usable from agents too.

t8y2:
Thanks for the contribution! Merged in 69b138dc2, will be released in the next version.

PR #9257, opened 2026-09-16
feat(mongodb): support replaceOne

thailoc-dev:
## What

`replaceOne` — the one core CRUD verb DBX did not have. `insertOne`, `updateOne`, `deleteOne` and `findOneAndReplace` all existed; this fills the gap.

```js
db.orders.replaceOne({ _id: ObjectId("…") }, { name: "new", tags: [] })
db.orders.replaceOne({ sku: "A1" }, { sku: "A1", stock: 0 }, { upsert: true })
```

It runs on both the native driver and the MongoDB Legacy Agent, so it works on MongoDB 3.6 as well as current servers.

## How

A new `Replace` command threaded through every layer, modelled on `updateOne` since it has the same shape (filter, document, `{ upsert }`) and returns counts:

| layer | change |
| --- | --- |
| parser (Rust + TS) | `replaceOne(filter, replacement[, options])`; an empty filter is guarded like an update |
| native driver | `replace_one(filter, doc).upsert(…)` |
| agent protocol | capability `mongo_replace_document` and method `replace_document`, in both contract versions and both copies |
| Java agent | `replaceDocument` handler with `ReplaceOptions` |
| Tauri / web / MCP | command, `/api/mongo/replace-document`, remote mapping |
| frontend | IPC, `queryStore` dispatch, safety, completion, parse diagnostics |

**Operator documents are rejected before reaching the server**, on every path. `replaceOne({…}, { $set: {…} })` is the obvious mistake, and the message says so:

> replaceOne() replaces the whole document, so it must not contain update operators such as $set; use updateOne() to modify fields.

**Older agents fail cleanly.** The capability is advertised in the handshake and required by the client, so a user on a previous agent jar gets *"MongoDB Legacy Agent does not support replaceOne; upgrade or reinstall the MongoDB Legacy driver"* rather than an RPC error — the same mechanism as `runCommand` and `insertMany`.

The result reports `modified_count`, matching `updateOne`. As with `upsert` there, a replace that *inserts* reports 0.

Not changed: the desktop `mcp_bridge` `/data/mongo/*` surface, which has no in-repo callers and already lacks `findOneAndReplace` and `runCommand`.

## Testing

- `cargo test -p dbx-core --lib` — 6392 passed; `dbx-web` + `dbx-mcp` — 235 passed
- `./gradlew :common:test :mongodb:test` — 206 + 91 passed, including both protocol-contract tests (v1 and v2)
- `pnpm test` — 14520 passed
- New tests at each layer: Rust parser and safety (7 cases), Java options/validation/RPC recognition (3), frontend parser/safety/normalization (3), completion guard updated

**Verified live against MongoDB 3.6** (a throwaway `mongo:3.6` container) through the agent's JSON-RPC interface with the freshly built jar:

| case | result |
| --- | --- |
| replace an existing document | `modified_count: 1`; document is `{_id: 1, name: "new"}` — the old `tags` and `count` fields are gone, as replace requires |
| upsert onto a missing document | document created |
| no upsert onto a missing document | nothing written |
| replacement containing `$set` | rejected: `Replacement document must not contain update operators such as $set` |
| `arrayFilters` option | rejected: `Unsupported replace option: arrayFilters` |
| previous jar, as control | `Unknown method: replace_document`; does not advertise the capability |

One pre-existing frontend flake (`DocumentBrowserFieldSearch › shows the value type selector…`) and six pre-existing `src-tauri` `commands::mcp` tests that parse this machine's pnpm shims fail identically on `main` and are unrelated.


t8y2 (approved):
Thanks for the thorough layering here — the capability handshake for older agent jars and the operator-document rejection on every path are exactly right. Merging as-is; one tiny follow-up note for a future PR: `MongoReplaceOptions` uses `deny_unknown_fields` while `MongoUpdateOptions` silently ignores unknown options, so tightening the update side to match would make the two verbs symmetric.

t8y2:
Thanks for the contribution! Merged in 9eafac706, will be released in the next version.

PR #9284, opened 2026-09-16
fix(mongodb): reject unknown update options instead of dropping them

thailoc-dev:
Follow-up to #9257, from the review note there.

## Problem

`MongoReplaceOptions` uses `deny_unknown_fields`; `MongoUpdateOptions` did not. So on the native driver an option `updateOne` does not apply was silently dropped:

```js
db.orders.updateOne({ sku: "A1" }, { $set: { stock: 0 } }, { collation: { locale: "en" } })
```

ran without the collation and reported success, while the same command on the legacy agent already failed with `Unsupported update option: collation`. The two verbs — and the two drivers — disagreed.

## Change

`MongoUpdateOptions` now rejects unknown fields too, so the native driver, the legacy agent, and `replaceOne` all apply the same rule: an option that would not take effect fails loudly rather than changing nothing.

```
Invalid update options: unknown field `collation`, expected `upsert` or `arrayFilters`
```

`findOneAndUpdate` already had this; no change there. The only callers pass the user's own options object through, so nothing internal relied on the leniency.

## Testing

- `cargo test -p dbx-core --lib` — 6415 passed
- New test: an unknown update option is rejected and named, the same input is rejected by the replace parser, and empty / absent options still parse
- `cargo clippy`, `cargo fmt` clean


t8y2 (approved):
Approved. `deny_unknown_fields` on `MongoUpdateOptions` matches the legacy Java agent whitelist exactly (`upsert`, `arrayFilters`) and closes the native-vs-legacy divergence noted in the #9257 review. Verified the shell layer consumes `multi` itself in `legacy_update_options`, so the legacy `update()` path is unaffected, and the error message names the rejected field. Tests and rust CI green. Merging as-is; supporting `collation`/`hint`/`let` can be a follow-up feature.

t8y2:
Thanks for the contribution! Merged in 91b8fa0aa8, will be released in the next version.

PR #9287, opened 2026-09-16
fix(editor): hide commit/rollback for clean mysql/pg manual transactions (#9018)

Abeautifulsnow:
## Problem

In manual transaction mode, MySQL and PostgreSQL query tabs required
Commit/Rollback even after read-only SELECTs (#9018). The sticky clean-state
toolbar introduced for Oracle (#7122) did not cover them.

## Root cause

`manual_transaction_proven_read_only` markers were emitted only for the
Oracle family (`is_oracle_proven_read_only_statement`), so every MySQL/PG
manual execution returned results without the marker and the frontend
monotonically dirtied the session.

## Fix

- `sql_risk.rs`: add `prove_read_only_for_database` — a strict, fail-closed
  read-only proof for MySQL/PostgreSQL. The statement must parse into exactly
  one AST; writable CTEs, DML set branches, locking clauses (`FOR UPDATE` /
  `FOR SHARE` / `LOCK IN SHARE MODE`), SELECT INTO, MySQL session writes
  (`:=`, `INTO @var`, executable comments, OUTFILE/DUMPFILE), window/table
  functions and every non-allowlisted function reject the proof. This is a UX
  heuristic only — never a security boundary.
- `query.rs`: generalize the manual-transaction classification pairing to
  MySQL/PostgreSQL (count/position pairing stays fail-closed on mismatch).
- Frontend: rename `oracleTxnPossiblyDirty` → `txnPossiblyDirty` (with a
  migration for detached-tab handoffs persisted by older builds) and gate the
  toolbar via `usesProvenReadOnlyStickyTransactionState`; Oracle keeps its
  own lexical classifier and ALTER SESSION compensation.

## Compatibility

jdbc / doris / kingbase and other dialects keep the legacy always-visible
Commit/Rollback rule (covered by tests). Cursor-page fetches and core no-op
scripts still neither set nor clear the sticky bit. Sticky members are
guaranteed to support manual transactions (test-guarded).

## Validation

- `cargo test --no-default-features --features sqlite-bundled -p dbx-core --lib sql_risk` — 27 passed
- `cargo test … --lib manual_transaction` — 7 passed
- `pnpm typecheck` — passed
- targeted vitest suites (toolbar, handoff migration, feature support,
  queryStore sticky state, hidden primary key, grid editor, txn expiry) — 210 passed

## Known limitations

Not yet verified against a live MySQL/PostgreSQL instance (Docker
unavailable during development). Generic `jdbc` connections keep the legacy
toolbar behavior and will join in a later iteration.


t8y2 (approved):
Approved. The fail-closed design checks out end to end: parse-strict proof with function allowlists, count/position pairing against the same dialect splitter, monotonic dirty aggregation, cursor fetches excluded, grid mutations and error paths dirty, and expiry recovery resetting only with a fresh replacement session. Legacy always-visible behavior for jdbc/doris/kingbase is test-guarded, and the detached-tab handoff migration preserves Oracle's path.

Two conservative limitations are fine as documented: window/table functions (e.g. `FROM generate_series(...)`) currently keep the buttons visible on clean sessions, and a shadowed allowlisted function name can produce a false clean — acceptable within the UX-heuristic boundary. Follow-up: a quick MySQL/PostgreSQL smoke after merge (SELECT-only hides buttons, DML/DDL/`BEGIN` shows them, auto-commit toggle cleans up).

t8y2:
Thanks for the contribution! Merged in e9be6d0419, will be released in the next version.

PR #9361, opened 2026-09-17
fix(plugins): show plugin context-menu entries in the connection sidebar menu

### LINKED ISSUE #9360: Plugin `context-menu` contributions never render in the saved-connection sidebar menu
### Summary

A plugin that declares a `context-menu` contribution is installed correctly and reported as
compatible, but its entry never appears in the saved-connection sidebar menu. The call site that
appends plugin entries to that menu is unreachable.

### Environment

- DBX 0.6.14 (also present on `main` at the time of writing)
- Windows 11 x64, desktop build

### Steps to reproduce

1. Install any plugin that declares a context-menu contribution, e.g.

   ```json
   {
     "type": "context-menu",
     "id": "com.example.plugin.info",
     "label": "Connection info",
     "menu": "connection"
   }
   ```

   A backend entrypoint is also required, since `validate_contributions` rejects a context-menu
   contribution without one. The contribution therefore needs at least a minimal sidecar.

2. Open the plugin center → Installed. The plugin is listed with the **compatible** badge, so
   `PluginManifest::compatibility()` passes and the manifest is not the problem.

3. Right-click any saved connection in the sidebar, and read the menu from top to bottom.

### Expected

The plugin's label appears at the bottom of the menu, below the built-in "copy name" entry.

### Actual

No plugin entry is rendered. The menu is identical to a run with no plugins installed. No error
is surfaced anywhere — not in the UI, not in the plugin center, not in the logs.

### Root cause

`apps/desktop/src/components/sidebar/SidebarTreeRuntimeHost.vue`

Plugin entries are appended by `appendPluginConnectionMenuItems()`, but `treeItemMenuItems()`
returns as soon as any menu factory claims the node:

```js
const sidebarMenuFactories = [buildConnectionSidebarMenu, buildDatabaseSidebarMenu, ...];

function treeItemMenuItems() {
  ...
  for (const factory of sidebarMenuFactories) {
    if (factory(factoryContext)) return items;   // returns here for connections
  }
  ...
  appendPluginConnectionMenuItems(items, node);  // never reached
  return items;
}
```

`buildConnectionSidebarMenu` is the **first** factory and returns `true` unconditionally for every
connection node:

```js
function buildConnectionSidebarMenu(context) {
  if (node.type === "connection") {
    ...
    return true;        // all saved connections end up here
  }
  ...
}
```

So control never reaches the call after the loop.

What makes this complete dead code rather than a partial failure is that the target function is
itself connection-only:

```js
function appendPluginConnectionMenuItems(items, node) {
  if (node.type !== "connection") return;
  const pluginItems = sidebarPluginRegistry.value.listContextMenuItems("connection");
  ...
}
```

and `listContextMenuItems("connection")` has no other call site anywhere in the frontend. The two
together mean plugin context-menu contributions cannot render on any surface, for any node type.

Note that `menu` is constrained to `"connection"` by the manifest schema (`const: "connection"`),
so the saved-connection sidebar menu is the only menu surface a plugin can target in v1 — and it is
exactly the one that is unreachable.

### Suggested fix

Call `appendPluginConnectionMenuItems(items, node)` inside the `node.type === "connection"` branch
of `buildConnectionSidebarMenu`, immediately before it returns, and drop the now-dead call after the
factory loop. Function declarations hoist, so the call may precede the definition.

Something equivalent at the `treeItemMenuItems` level would also work — for example letting the
connection branch fall through to the shared tail instead of returning early — but moving the call
is the smaller change.

### Notes

- This is purely a wiring bug. Manifest validation, signature/install handling, the plugin
  registry, and the plugin backend all behave correctly; the plugin center showing "compatible"
  while no menu entry appears is the tell.
- There is currently no UI that surfaces `context-menu` contributions at all:
  `PluginContributionsPanel.vue` renders connection providers, workbenches, and filesystem
  providers, but not context menus. That is why this went unnoticed — nothing in the app
  confirms the contribution was registered.

A PR with the fix is linked below.

ccyylhj:
Fixes #9360

## What changed

`appendPluginConnectionMenuItems()` was called after the `sidebarMenuFactories` loop in
`treeItemMenuItems()`:

```js
for (const factory of sidebarMenuFactories) {
  if (factory(factoryContext)) return items;   // buildConnectionSidebarMenu returns true here
}
...
appendPluginConnectionMenuItems(items, node);  // never reached for connection nodes
```

`buildConnectionSidebarMenu` is the first factory and unconditionally returns `true` for
`node.type === "connection"`, so the loop returns before the plugin call. Since
`appendPluginConnectionMenuItems()` in turn returns early unless the node is a connection, and
`listContextMenuItems("connection")` has no other call site, plugin `context-menu` contributions
were unreachable everywhere.

This moves the call into the connection branch, immediately before it returns, and removes the
call after the loop. Function declarations hoist, so no reordering is needed.

## Why here and not after the loop

The alternative — letting the connection branch fall through to the shared tail instead of
returning early — would change the control flow for every connection node and re-run the shared
tail (pin separator, universal copy-name) that the connection branch already handles itself.
Appending the plugin entries inside the branch keeps the change to one call and leaves the
existing menu structure untouched.

## Testing

- `oxfmt --check` and `oxlint` clean on the changed file; the `lint-staged` pre-commit hook
  (oxfmt) passes.
- Verified end to end with a minimal test plugin that declares a single `context-menu`
  contribution (`menu: "connection"`) plus a sidecar backend:
  - before: the plugin installs, the plugin center lists it with the **compatible** badge, and
    the connection menu shows no plugin entry;
  - after: the entry renders in the connection menu and clicking it dispatches
    `contextMenu/<contribution-id>` to the plugin backend, which returns the expected payload.
- No Rust changes, so `cargo-check-fast` / `cargo-test-fast` are unaffected.

I did not add a regression test. The behaviour lives in a large sidebar component whose menu
construction is driven by many stores, and per `CONTRIBUTING.md` the project prefers behaviour
tests over source-wiring snapshots. Happy to add a mounted-component test if you want one —
guidance on where it should live would be welcome.

## Notes

`PluginContributionsPanel.vue` does not surface `context-menu` contributions at all (it renders
connection providers, workbenches, and filesystem providers). That is likely why this went
unnoticed: nothing in the app indicated the contribution had been registered. Out of scope for
this fix, but mentioning it in case it is worth a follow-up.


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 48d0ebc90, will be released in the next version.

PR #9365, opened 2026-09-17
新增可选 CC-SWITCH 配置导入插件支持

ma123456mai:
修复 #8771

## 变更说明

- 保留 DBX 主机的 `ai-config-import` 能力、设置入口和配置导入流程。
- 只有在已安装、兼容并声明 `ai-config-import` 能力的插件时，CC-SWITCH 入口才可用。
- 未安装插件时，设置页只提供插件中心入口，不在 DBX 内下载、解压、安装或卸载插件。
- 删除 DBX 内置的 CC-SWITCH 数据库解析器；主机只向插件传递 `~/.cc-switch/cc-switch.db` 路径并调用 `import_ai_configs`。
- CC-SWITCH 导入器位于独立公开插件仓库，便于审查 API 密钥读取逻辑。

## 安全边界

- 插件声明能力：`ai-config-import`。
- 插件权限为空，不声明 `host.network`。
- 插件以只读方式读取用户的 `~/.cc-switch/cc-switch.db`，读取 Codex、Claude、Gemini 配置及 API 密钥，并返回规范化配置。
- 插件不写入 CC-SWITCH 数据库、不持久化凭据、不监听端口、不启动外部命令、不主动联网。
- `.dbxp` 仅作为插件仓库发布包，不提交到 DBX 或插件商店源码仓库。

## 支持平台

已生成以下平台的插件包：`darwin-arm64`、`darwin-x64`、`linux-arm64`、`linux-x64`、`windows-x64`。Windows ARM64 暂不在官方发布工作流支持范围内。

## 验证结果

- DBX 定向前端测试：通过。
- DBX `cargo check`：通过。
- 插件单元测试：通过。
- 插件核心测试：123 项全部通过。
- 插件持续集成和发布工作流：通过。

## 相关地址

- 插件源码：https://github.com/ma123456mai/dbx-cc-switch-plugin
- 插件发布包：https://github.com/ma123456mai/dbx-cc-switch-plugin/releases/tag/v0.1.2
- 插件商店提交：https://github.com/t8y2/dbx-store/pull/44

## 提交信息

- DBX 提交：`d394cad9258514798724a20299d06ef9ffb2e0cf`
- 插件源码提交：`a2311990c30689a2f26b80a4c0acf446ac1c5c61`


t8y2 (changes requested):
request changes — the direction we'd like to take here is the plugin-store route rather than an in-repo installer.

The host-side capability work (`ai-config-import` protocol, settings entry, import flow) is the right shape. What we can't merge is the installer: `CC_SWITCH_PLUGIN_DOWNLOAD_PREFIX` points at a release asset that doesn't exist yet and whose source/build isn't in this repo, so the button fails for every user — and the binary that actually reads `~/.cc-switch/cc-switch.db`, including API keys, would ship outside review.

Suggested restructure:

1. Keep the host capability and the settings entry, but gate the entry on the plugin being installed (the S3 sync tab from #9354 is the exact pattern) and drop the hardcoded download + install/uninstall sidecar logic. The entry can link to the plugin center or use the `dbx://plugins/install` deep link.
2. Package the importer itself as a `.dbxp` plugin and submit it to the official store — the guide covers structure, sidecar SDKs, packaging and the submission flow: https://dbxio.com/en/docs/plugin-development. Please keep the plugin source public, since it handles API keys. The SQLite parsing you already wrote in `load_ai_configs_from_path` moves there (today it's dead code — the command delegates to the plugin anyway), and the in-repo reader is then removed so the two parsers can't drift.

If you'd rather avoid a plugin entirely, the alternative is calling `load_ai_configs_from_path` directly in the command (the schema matches upstream cc-switch) — but the store route is the better fit here. Happy to help review the plugin package when it's ready.

ma123456mai:
已按照评审意见完成重构并推送中文提交：

- 删除 DBX 内置下载、压缩包安装、本地安装和卸载逻辑。
- 保留主机侧 `ai-config-import` 能力、设置入口和导入流程；未安装兼容插件时只跳转插件中心。
- CC-SWITCH 数据库解析器已移动到独立公开插件仓库，源码公开；插件以只读方式读取 `~/.cc-switch/cc-switch.db`，不声明网络权限且不主动联网。
- 插件版本 `v0.1.2` 已完成五个平台构建。
- 官方插件商店候选 PR：https://github.com/t8y2/dbx-store/pull/44

DBX 中文提交：
`d394cad9258514798724a20299d06ef9ffb2e0cf`

插件源码提交：
`a2311990c30689a2f26b80a4c0acf446ac1c5c61`

插件商店 PR 按要求只提交发布者信息和候选包清单，不提交 `.dbxp`、插件源码或目录文件。

t8y2 (changes requested):
Thanks for the restructure — this is exactly the shape we asked for. The earlier review items are all addressed: the installer is gone, the entry is gated on an installed, compatible plugin with the `ai-config-import` capability (same pattern as the S3 sync tab), the in-repo database reader is fully removed, and the importer now lives in your public plugin repo with the store submission. The security posture looks right: the host only passes the `~/.cc-switch/cc-switch.db` path through, keys are never logged or shown in plaintext, and web mode is properly gated.

Two things still block merge, both mechanical:

1. **CI formatting:** `rust-fmt-clippy` fails because `pub mod cc_switch;` sits out of alphabetical order in `crates/dbx-core/src/lib.rs` (rustfmt wants it after `backend_error`), and the `frontend` job's format check flags `apps/desktop/src/lib/ai/__tests__/aiConfigList.spec.ts`. Running `cargo fmt` and the frontend format command fixes both.
2. **Rebase:** the branch is 371 commits behind `main` and now conflicts. The important part: `main` moved `crates/dbx-core/src/plugins.rs` and `plugins/` into the new `crates/dbx-plugin-runtime` crate. Your `find_capability`/`invoke_capability` additions and `PluginCapabilityManifest` need to be re-applied to `crates/dbx-plugin-runtime/src/plugins.rs` and `crates/dbx-plugin-runtime/src/plugins/manifest.rs` — a plain merge would silently drop them and break the build. The API there is unchanged, and `dbx_core` re-exports the plugin runtime, so the port is a straightforward move; `crate::ai::AiConfigItem` still resolves in `dbx-core` after the crate split.

One smaller suggestion while you're in there: `load_cc_switch_ai_configs` invokes any compatible plugin declaring `ai-config-import`, while the UI gates on plugin id `cc-switch`. Pinning the command to `CC_SWITCH_PLUGIN_ID` (or checking the id in `find_capability`'s caller) keeps the two gates consistent.

Once rebased and formatted, this should be in good shape — happy to take another look.

ma123456mai:
已按本次评审意见完成更新：

- 已将分支 rebase 到官方 `main` 最新提交 `5d5a12ec`。
- 已将 `PluginCapabilityManifest`、`find_capability`、`invoke_capability` 迁移并保留在 `crates/dbx-plugin-runtime`。
- 已在 `dbx-core` 正确声明 `cc_switch` 模块。
- `load_cc_switch_ai_configs` 现在会校验 `cc-switch` 插件 ID、兼容性以及 `ai-config-import` 能力，和前端门控保持一致。
- 已完成 Rust 和前端格式化。

本地验证：
- `cargo check --locked -p dbx-core -p dbx-plugin-runtime` 通过。
- CC-SWITCH 相关前端测试 4 个文件、157 项测试全部通过。
- Rust 定向 `rustfmt --check` 和前端 `oxfmt --check` 通过。

最新提交：`08e48666df4485efc7b2d2cb42849597cc6f6994`。

PR #9375, opened 2026-09-17
feat(mongodb): support bulkWrite

thailoc-dev:
## What

`db.collection.bulkWrite([...])` with all six operation kinds and the `ordered` option:

```js
db.orders.bulkWrite([
  { insertOne:  { document: { sku: "D4", stock: 2 } } },
  { updateOne:  { filter: { sku: "A1" }, update: { $inc: { stock: 1 } } } },
  { updateOne:  { filter: { sku: "ZZ" }, update: { $set: { stock: 0 } }, upsert: true } },
  { updateMany: { filter: { archived: true }, update: { $set: { stock: 0 } } } },
  { replaceOne: { filter: { sku: "B2" }, replacement: { sku: "B2", stock: 9 } } },
  { deleteMany: { filter: { stock: { $lt: 0 } } } }
], { ordered: false })
```

The result is one row of counts, the way the shell prints a `BulkWriteResult`: `insertedCount`, `matchedCount`, `modifiedCount`, `deletedCount`, `upsertedCount`.

## How

Same layering as `replaceOne` (#9257): a `BulkWrite` command threaded through parser, driver, agent protocol, Java agent, Tauri, web, MCP and the frontend.

**Operations are validated before anything reaches a server**, identically in the Rust and TypeScript parsers and again in the agent: each entry must be exactly one `{ <op>: { … } }`; each op accepts only the fields it takes (`collation`, `hint` and other unsupported fields are rejected by name rather than dropped); update documents must use operators; replacements must not. Errors name the operation position: `bulkWrite() operation 3 (replaceOne) replacement must not contain update operators such as $set; use updateOne to modify fields.`

**Safety follows the least-bounded operation.** If any update/replace/delete in the batch has an effectively unbounded filter, the whole batch is treated like an unbounded `updateMany`, on every path (editor, MCP bridge, web policy).

**Two execution paths, with a deliberate difference.**
- The **Java agent** uses the driver's real `collection.bulkWrite(models, options)` — one round trip, native on MongoDB 3.6.
- The **native Rust driver** has no collection-level `bulkWrite`; its `Client::bulk_write` needs MongoDB 8.0+. So the native path runs the operations one at a time in order. Collection-level `bulkWrite` is not transactional, so the semantics are the same; only the round-trip count differs. `ordered: true` (the default) stops at the first failure; `ordered: false` attempts every operation and reports all failures. Using `Client::bulk_write` on 8.0+ servers would be a follow-up.

**Partial failures report what completed**, on both paths and in the same shape:
`bulkWrite stopped at operation 2: E11000 duplicate key … Completed: inserted 0, matched 0, modified 0, deleted 1, upserted 0`

**Older agents fail cleanly** via the `mongo_bulk_write` capability in the handshake, as with `replaceOne`.

Not included: autocomplete *inside* the operations array (the `{ <op>: { filter, update, … } }` shape) — the method itself is suggested, and the argument is treated as a no-suggestion position rather than a wrong one.

## Testing

- `cargo test -p dbx-core --lib` — 6457 passed; `dbx-web` + `dbx-mcp` — 239 passed
- `./gradlew :common:test :mongodb:test` — 206 + 107 passed, both protocol-contract tests (v1, v2) included
- `pnpm test` — 14782 passed
- New tests: Rust parser (every op kind, safety, 16 malformed forms), Java (models for every kind, 7 malformed forms, options, RPC recognition), frontend (parser, 15 malformed forms with their messages, safety, result rendering), completion guard updated

**Verified live against MongoDB 3.6** (throwaway `mongo:3.6` container) through the agent with the built jar:

| case | result |
| --- | --- |
| six-kind batch on seeded data | counts `1/2/2/1/1`; A1 incremented, B2 replaced with its old field gone, C3 deleted, D4 inserted, ZZ upserted |
| `ordered: true` with a duplicate key at op 2 | error; the later `deleteOne` did **not** run |
| `ordered: false`, same batch | error; the later `deleteOne` **did** run |
| partial failure message | names the failing operation and the completed counts |
| operator-less update / `writeConcern` option | rejected before the server |
| previous jar, as control | `Unknown method: bulk_write` |

Along the way a Java unit test caught a real bug in the first draft — a pipeline update was re-serialised via `List.toString()`, which is not JSON — fixed before this was pushed.

One frontend test (`shortcutRegistry › reports the intentional cross-scope overlaps`) fails identically on `main` and is unrelated.


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in f78dc1ba1, will be released in the next version.

PR #9380, opened 2026-09-17
feat(ai): auto intent routing for the assistant action menu (#9118)

Abeautifulsnow:
### Summary

Adds an **Auto** entry to the AI assistant's action menus. When selected, the
assistant detects the user's intent at send time and routes the message to the
right concrete action (explain, optimize, fix, explore schema, …) instead of
requiring a manual pick. A new Settings > AI toggle makes Auto the default for
new conversations.

### Changes

**Auto intent routing (#9118)**
- `Auto` heads both the Ask and Agent action menus (`AiActionSelection = AiAction | "auto"`); the transport-level `AiAction` union is unchanged.
- Routing resolves **before** any request is assembled: a zero-latency rules layer first, then a single tiny LLM classifier fallback (bounded timeout, never throws, falls back to the mode default) — `taskContract.action` can never carry `"auto"` over the wire.
- Routed messages show an `Auto · <action>` chip; clicking it switches the picker to the routed action (cross-mode routes switch mode first via the existing suppression handshake).
- A "Detecting intent…" indicator covers the classifier slow path; it is cleared on both the normal and the abandon path.
- Confirmed-write continuation turns bypass routing (pre-Auto confirmation contract preserved); external explicit triggers (`triggerAction`) keep the direct path; vector-DB connections keep the hidden-picker forced-`generate` branch.

**Default auto routing setting (follow-up)**
- New Settings > AI toggle "Auto Intent Routing by Default" (default **off**, opt-in): when enabled, new conversations land on `Auto`; mode switches and panel mount follow the same resolver, current conversations and restored ones are untouched.
- Persisted in `AiChatSelectionState` end-to-end: TS type, Pinia store lifecycle, and the Rust serde struct (`default_auto_routing`, camelCase wire key pinned by tests).
- Settings-search entry + locale copy for all 10 locales, guarded by an i18n parity spec.

**Chore**
- `dev-fast` recipe exports `RUST_MIN_STACK=16777216` (Windows dev environments hit default-stack overflows in deep rustc/test stacks).

### Testing

- New: `aiIntentRouter.spec.ts`, `AiAssistant.autoRouting.spec.ts` (wiring/wire-contract assertions), `aiAutoRoutingParity.spec.ts` (10-locale parity), settingsStore lifecycle tests, settingsSearch inventory test.
- AI-surface batch: 43 spec files / 545 tests green; full settings + i18n suites green; `vue-tsc` and `oxlint` clean.
- Rust: `cargo test -p dbx-core --lib ai_chat_selection` (serde round-trip, legacy payload compat, wire-key pinning) green; `cargo fmt --check` clean.

### Screenshots

a0f5d904-bb13-4e08-aee8-401ec0f6da5c

4db242e8-f00c-4019-bc92-31078ac33754


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 3f3c21aae, will be released in the next version.

PR #9391, opened 2026-09-17
导出 SQL 脚本选择不含主键但仍包含主键#9297

HighBigSky:
## 变更说明 / Change Description

解决9297的bug
## 变更类型 / Change Type

- [ ] 新功能 / New feature
- [x] Bug 修复 / Bug fix
- [ ] 代码重构 / Code refactoring
- [ ] 文档更新 / Documentation
- [ ] CI / 构建 / CI or build

## 涉及前端 / Frontend Changes

- [x] 本 PR 涉及前端改动，已附截图/录屏（见下方） / This PR includes frontend changes and screenshots or a recording are attached below.

image
image

## 验证 / Validation

后续pr建议：
1：导出菜单仍然没有可见开关：用户还是得去"复制 → 数据提取配置"里勾"排除主键"，导出才会跟随。建议按之前说的，在导出子菜单给 SQL 加一对"含主键/不含主键"（i18n key 现成）。
2:：查询结果导出依赖表元数据：自由 SQL、JOIN、聚合结果没有主键信息，勾了也不会排除（后端不猜）。要覆盖可以用 build_keyset_plan 那套单表识别补上。

- [x] `make check` 通过 / `make check` passes

## 关联 Issue / Related Issue

Related #9297


t8y2 (commented):
Maintainer rebase + fix on top of 519c731f8 — pushed as abfb14918 (old head 519c731f8 → rebased onto current main 137a1a416, branch rewritten with --force-with-lease; original head preserved in our backup ref).

Conflict cause: the branch was based on 2026-09-30's main (9190554e5), and main since refactored the same SQL INSERT writer call sites (dd84e6c06, e255c49e7d identity INSERT wrapping, b70d6da227 streaming JSON exports). Resolution preserved both sides: every INSERT-writer call keeps main's `insert_mode` batch sizing, `column_extras`, spatial columns and csv quote mode, and routes through your `_excluding` wrapper with the excluded columns. The 5 conflicting files were exactly the INSERT-writer call sites, imports, and test imports; `git range-diff` maps every PR behavior onto the resolved head.

Two semantic adjustments made while resolving, to align this feature with the exclude-PK semantics main settled in e3d5d1418 (DBeaver's rule: only auto-generated columns are dropped, manual keys keep their values):

1. "Exclude primary keys" now drops only auto-increment/identity PKs, on every path — the frontend picker filters with the same predicate copy-as-INSERT uses (`isAutoGeneratedColumn` over `tableMeta.columns[].extra`), and the backend `sql_export_excluded_columns` applies the same rule via the shared `extra_is_auto_generated` helper so sidebar-tree exports (which send raw PK lists) are covered too. Without this, exported INSERTs for natural-key tables (UUID/char PKs) would replay with missing NOT NULL values or silently re-key rows, and the same toggle would behave differently between copy-as-INSERT and export.
2. The "No insertable columns remain" error now fires only when the exclusion itself removed the last insertable column; emptiness caused by other omission rules (e.g. generated columns) keeps the silent empty result.

Also updated the live-test struct literals for the new request fields, and extended the app-tests: the three exclusion tests now model an `auto_increment` PK, plus a new regression test asserting manually-assigned PKs are kept.

Checks run: `cargo check -p dbx-core --no-default-features --tests` clean; `cargo test` for `table_export::` (42), `query_result_export::` (48), `database_export::` (104) all passing (RUST_MIN_STACK=16M — the default 8MB stack overflows locally on unrelated heavy tests); `pnpm vitest run packages/app-tests/useDataGridExport.test.ts` 43 passed; the grid/sidebar export specs 89 passed; `cargo fmt` applied.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in d17ce4662, will be released in the next version.

PR #9409, opened 2026-09-17
fix(redis): prevent key browser toolbar overlap at narrow widths (#9356)

### LINKED ISSUE #9356: [Bug] 建议redis左侧导航栏设置最小宽度，避免按钮挤压在一起会重叠

REDIS

Image

Image

DBX Version: v0.6.15
Runtime: Desktop
Operating System: macOS 15.7.7
Architecture: x86_64

_No response_

Abeautifulsnow:
#### Problem

In the Redis key browser, narrowing the left key pane (roughly 320–730px) with keys selected
made the toolbar action buttons paint over the search-mode segmented control (键|值|全部), and
the key-count text disappeared entirely. Reported as #9356 (v0.6.15, macOS, with screenshots).

#### Root cause

The Aug toolbar rewrite (#328a2aa10) replaced the wrapping flex toolbar with a fixed grid
(`auto minmax(0,1fr) auto`) and gave both side groups `flex-wrap: nowrap; min-width: 0`.
With buttons `shrink-0` and the actions group `justify-end`, any overflow spilled **left**,
painting 全选/清除 on top of the segmented control. The container queries only covered
≤320px while the toolbar's max-content is ~730px, and the pane's `min-width: min(256px, 64%)`
floor could not help at realistic widths.

#### Changes

- Actions group uses `flex-wrap: wrap` (keeps `justify-end`): wrapping can never overlap, and
  it holds in every layout branch (base, 340px, 240px tiers).
- New `@container (max-width: 740px)` tier collapses the widest label (设置过期时间) to an
  icon-only button (`aria-label` + `title` preserved) so the common band stays single-row.
- Key pane floor raised `min(256px, 64%)` → `min(360px, 64%)` — the single-row toolbar
  minimum. At the floor the toolbar is always one row; the count column yields first
  (full text stays on `title`).
- ≤340px tier (reachable only via the 64% guard on very small windows) shows a tidy two-row
  fallback: [mode | count] and the action buttons on their own full-width rows.
- Source-level guard tests (jsdom cannot evaluate container queries/flex layout) lock the
  wrap rule, both container tiers, the pane floor, and the template hooks against regression.

#### Testing

- `pnpm vitest run apps/desktop/src/components/redis/` — 14 files / 217 tests pass
  (5 new guard assertions in `RedisKeyBrowser.toolbarLayout.spec.ts`)
- `oxlint`, `oxfmt --check`, `vue-tsc` all clean
- Manually verified: narrowing the pane with keys checked no longer overlaps at any width;
  right-panel open + minimum width shows the single-row toolbar

#### Screenshots
e8051991-8bf5-40f5-aa37-c9ca96c1ff28

Fixes #9356


t8y2:
Thanks for the contribution! Merged in df8a659cc, will be released in the next version.

PR #9413, opened 2026-09-17
fix(mongodb): reject unsupported find() chain methods instead of dropping them

thailoc-dev:
## Problem

The editor's `find()` parser honours `sort`, `skip`, `limit`, `collation` and `count`, and **silently drops any other chained method**. So each of these runs, succeeds, and does something other than what was written:

```js
db.c.find({}).hint({ a: 1 })          // runs without the hint
db.c.find({}).explain("executionStats") // returns documents, not a plan
db.c.find({}).itcount()               // returns documents, not a number
db.c.find({}).forEach(d => print(d))  // plain find; the callback never runs
db.c.find({}).batchSize(10)           // ignored
```

Every other unsupported command in the editor fails with an error. These are the only ones that succeed with the wrong behaviour, which is worse: there is nothing to tell the user the query they ran is not the query they wrote.

The Rust parser (CLI, MCP) already rejects unknown chain methods, so the two sides also disagreed.

## Change

The editor now tokenises the chain after `find()` and rejects any call outside the set it executes — `sort`, `skip`, `limit`, `collation`, `count`, plus the no-ops `toArray` and `pretty` — matching the Rust parser. A property access without parentheses (`.count`) or trailing text is rejected the same way.

The message names the method and, where it helps, the alternative:

| chain | message |
| --- | --- |
| `.hint()`, `.batchSize()`, `.maxTimeMS()`, … | `find().hint() is not supported yet. Supported after find(): sort, skip, limit, collation, count, toArray, pretty.` |
| `.explain()` | `find().explain() is not supported in the editor yet; aggregate([...], { explain: true }) is.` |
| `.itcount()`, `.size()` | `… use find().count() or countDocuments() to count results.` |
| `.forEach()`, `.map()` | `… runs JavaScript, which the editor does not execute; run the find and read the results instead.` |
| `.toArray(1)` | `find().toArray() takes no arguments.` |
| `findOne({}).limit(1)` | `findOne() returns a result, not a cursor, so nothing can be chained after it.` |

**This is a behaviour change**: commands that previously "worked" by ignoring part of themselves now fail. That is the point, but worth stating.

`find().explain()` deserves a note. The backend supports it (`mongo_explain_find_core`, and `dbx-web` has `/mongo/explain-find`), but the desktop has no IPC for it, so the editor was returning documents. Rejecting it with a pointer is the honest interim; wiring explain into the editor is a natural follow-up.

## Testing

- `pnpm test` — 14947 passed
- New tests: ten rejected chains with their messages, the non-cursor methods, and the chain forms that must keep working (multi-line chains, `.count()`, no-ops, `collation`), plus the tokenizer itself


t8y2 (approved):
Thanks for this PR — the analysis is accurate and the direction is right. Verified against `main`: the TS `find` parser indeed accepts any dot-chain and silently drops everything outside `sort/skip/limit/collation`, while the authoritative Rust parser already rejects unknown chain methods, so this aligns the editor with it. Nice job matching the Rust set exactly, including the `toArray()/pretty()` no-ops, empty-arg-only `count()`, and rejecting property access without parentheses; the new tests cover both rejected forms and forms that must keep working (multi-line chains included).

Two non-blocking observations from my testing:

1. In a multi-statement editor buffer, one rejected chain fails the whole batch and the diagnostic points at the first command's tail, so users see the generic "Unexpected text after find(...)" message rather than your specific `.hint()` message (e.g. `db.a.find({}).limit(5); db.b.find({}).hint({x:1})`). This is a pre-existing batch-diagnostic limitation — your change still turns silent wrong execution into an error — but per-segment diagnostics would be a good follow-up.
2. One body nit: on current `main`, desktop execution re-parses each command via the Rust side, which returns `findExplain` — a case the desktop switch never handles — so the editor currently produces *no result* for `find().explain()` ("returns documents" was only true for older versions). Your rejection with a pointer to `aggregate([...], { explain: true })` is still the right interim behavior.

Approved — frontend CI green, 115/115 app-tests pass locally on the PR head.

t8y2:
Thanks for the contribution! Merged in 58200ce5a, will be released in the next version.

PR #9434, opened 2026-09-17
fix(mysql): clear stale-snapshot transactions and trim refresh metadata path

t8y2:
## Summary

Two fixes for recent MySQL regressions reported around v0.6.14–v0.6.15: stale rows in auto-commit tabs, and slower table loading/refresh.

### 1. Stale reads in auto-commit (A) tabs — `fix(mysql): clear leftover transactions that pin stale read snapshots`

Symptom: users must disconnect/reconnect (or switch a tab to manual mode) to see newly committed rows.

Root cause chain:
- Query/data tabs execute on tab-scoped **single-connection pools** that deliberately disable `COM_RESET_CONNECTION` on return (to preserve temporary tables and other session state).
- If such a connection is returned with an **open transaction** — a user-typed `BEGIN`/`START TRANSACTION` never committed, or an aborted/canceled batch whose cleanup was skipped — the MySQL REPEATABLE READ read view stays pinned on that connection.
- Every later auto-commit query on that tab reads the old snapshot forever. Manual-transaction mode uses a different dedicated connection with a fresh `BEGIN`, which is why switching to M appeared to "fix" reads, and disconnect destroyed the poisoned pool.

Fix:
- `execute_multi_mysql` now issues a `ROLLBACK` before returning the connection to the pool (server no-op when no transaction is open; on failure the connection is discarded and the pool rebuilt). This restores the auto-commit contract for tab pools regardless of how the transaction was left open.
- Disconnecting a connection now **rolls back its manual-transaction sessions** before draining pools (previously the session map was cleared without rolling back the dedicated connections).
- Fully proven read-only manual-transaction batches on native MySQL/PG connections now **rotate the snapshot** (ROLLBACK + fresh BEGIN, PG re-applies `search_path`) after the batch, so a clean read-only manual session no longer pins its first SELECT's snapshot — which also unblocks the v0.6.15 (#9287) hidden Commit/Rollback toolbar for MySQL. Rotation runs while the session is still marked busy; a rotation failure tears the session down and relies on the existing frontend recovery (next execution transparently opens a new session). Agent/external-driver sessions keep current semantics (their rollback path closes the session, so in-place reopen is not possible).

### 2. Table refresh performance — `perf(grid): trim metadata work off the table refresh critical path`

Since #9156 (v0.6.14) a toolbar refresh serially awaited a strict persisted-cache prefix deletion plus a forced getColumns **and** listIndexes round trip before executing the SELECT. The persisted deletion is now fire-and-forget (in-memory invalidation still guarantees fresh columns for the grid), the awaited segment fetches columns only (getColumns; surviving primary keys are intersected with the fresh columns — PK names never enter the SQL text, they only feed the large-value preview protection set), and index/PK metadata is refreshed in the background after the query (the columns facet is already cached, so that costs one listIndexes round trip off the critical path).

## Validation

- `cargo test -p dbx-core --no-default-features --features sqlite-bundled --lib` — 6256 passed, 0 failed (run with `RUST_MIN_STACK=32MB`; on default stack several unrelated pre-existing tests stack-overflow on clean `main` too — `mongodb_agent_registers_shell_query_tool`, `external_driver_table_export_*` — verified via stash that this branch does not change that).
- New unit test: `manual_txn_batch_proven_read_only_requires_non_empty_all_proven_results`.
- `pnpm typecheck` — clean.
- `vitest` `useDataGridActions.spec.ts` (42) + metadata cache suites (36) — all green; refresh-path specs updated to the new non-blocking semantics (deletion no longer gates metadata/SQL; index discovery asserted off the awaited path; stale-PK intersection covered).

## Notes

- Manual transaction semantics for write sessions are unchanged: once a session executes an unproven statement, Commit/Rollback behave exactly as before.
- The rotate cost is one ROLLBACK + one BEGIN per fully read-only manual batch on native MySQL/PG — both metadata-only operations on the held connection.


t8y2 (commented):
已完成维护者修补，提交 24ffa47bb（基于 160ada25a）。

修复：表格刷新：元数据未完整校验前保持只读，保留 Oracle 表的 ROWID；快照刷新不回滚历史写入，也不刷新备份一致性快照。

验证：前端 226；Rust 11；真实数据库 4；格式与 diff 检查通过。
验证边界：未运行完整桌面 GUI；已在远端 MySQL 8.0.46 / PostgreSQL 17.11 验证事务行为。

未新增依赖，保留原 PR 的功能范围。后续合并将以此 head 的状态为准。

t8y2:
Thanks for the contribution! Merged in fac270fc8, will be released in the next version.

PR #9450, opened 2026-09-17
perf(layout): keep panel divider drags smooth with stepped resize tracking

zhangdengjie:
Dragging a divider resized the CodeMirror editor and the DataGrid canvas on every pointer event, forcing per-frame re-measurement: profiling a 3s drag with a 100+ line script recorded 2k+ CodeMirror measure() calls dominated by getBoundingClientRect, plus per-frame clientWidth/scrollWidth reads from the result tabs, toolbars and grid topbar stacking synchronous relayouts on top (layout thrashing).

Introduce a shared drag-state module (panelResizeState): usePanelResize and all Splitpanes dividers report drag begin/end, heavy measurement chains (result tab scroll, toolbar overflow tiers, grid topbar, grid canvas) defer to a single re-sync when the drag ends, and globals.css isolates the editor and result subtrees while resizing.

QueryEditor pins the editor root to an explicit pixel size during a drag and only re-measures every N animation frames; DataGrid applies the same stepped re-measure/re-draw to its canvas runtime. The step is tunable via ~/.dbx/ui-tuning.json (panelResizeTrackEveryFrames, clamped 1..10, defaults to 5), read once at startup in desktop builds via the fs plugin.


t8y2 (commented):
Applied the maintainer review fixes in 06ef1f58.

Workspace resize state follows native mouse/touch start and release/cancel/blur/unmount rather than delayed Splitpanes resize callbacks. This prevents a queued frame from re-locking the editor after release while preserving size persistence and stepped tracking. Filesystem access adds only command-specific exists/readTextFile scopes for the exact $HOME/.dbx/ui-tuning.json file.

Validation: regression failures reproduced before the fix; 43/43 targeted tests now pass. Full frontend vue-tsc, changed-file formatting/lint, and git diff --check pass. The drag regressions use the installed Splitpanes with controlled animation frames. No live Tauri permission or GUI/E2E validation was performed.

The patch is pushed and the remote SHA is verified. Merge remains gated on CI and merge readiness for this exact head; this PR has not been merged.

t8y2 (commented):
CI follow-up applied in c4ec92a8.

The PR moves the topbar overflow measurement into an animation-frame callback and re-reads the element as `currentTopbar`. The existing source-contract test still required the old local variable name `topbar`, so CI failed even though the overflow threshold and progressive compaction logic remained present.

The assertion now examines only the measurement function and requires scrollWidth and clientWidth to come from the same element, retaining the one-pixel threshold without depending on a local name. No production behavior is changed in this follow-up.

Reproduced 17 passed / 1 failed before the test fix; the expanded toolbar/layout regression matrix now passes 61/61 on Node 22. Full frontend typecheck, changed-file formatting/lint and git diff --check pass. This is separate from the shared baseline useExportTracker totalRows failure, which remains outside this PR. Merge is still gated on the new exact-head CI.

t8y2:
Thanks for the contribution! Merged in 8eba12ea2, will be released in the next version.

PR #9454, opened 2026-09-17
feat(mongodb): run find().explain() from the editor

thailoc-dev:
## What

`find(...).explain([verbosity])` now runs from the query editor and shows the plan:

```js
db.orders.find({ status: "paid" }).sort({ createdAt: -1 }).limit(20).explain("executionStats")
```

The result is the explain document as a single row, the same way `aggregate([...], { explain: true })` is shown.

## Why now

#9413 made the editor reject chained methods it could not execute. `explain()` was one of them — until then the editor had been *silently returning documents* for it — and the rejection message said "not supported in the editor yet". This finishes that.

The backend already had everything: `mongo_explain_find_core`, the `FindExplain` command in the Rust parser, agent support (`explain_find`), and a web route. The desktop simply never got the IPC.

## Change

- **Tauri:** `mongo_explain_find` command, mirroring `mongo_find_documents` and the existing web route's parameters.
- **Frontend:** `mongoExplainFind` in the three backend layers; a `findExplain` command kind in the shell parser, wrapping the find it explains; a `queryStore` case that renders the plan; `explain` added to the cursor-method completion.
- **Parser rules:** `explain` must be the final chain call, and the verbosity must be one of `queryPlanner` (default), `executionStats`, `allPlansExecution`. Otherwise the diagnostics say which:
  - `find().explain() must be the final chain method.`
  - `find().explain() verbosity must be "queryPlanner", "executionStats" or "allPlansExecution", got 'verbose'.`

The command shape matches the Rust parser's `findExplain` exactly, so CLI/MCP and the editor agree.

## Testing

- `pnpm test` — 14992 passed
- `cargo clippy` on `src-tauri`, `pnpm typecheck`, `oxlint`, `oxfmt` clean
- New tests: explain with every find option, default verbosity, multi-line and bracket-accessor forms, plain find vs explain not confused, the misplaced/invalid cases with their messages; cursor completion updated
- Confirmed the agent's `explain_find` returns a real plan on a throwaway MongoDB 3.6 (`queryPlanner` / `executionStats`, winning stage `SORT`) — the path the legacy connections use

Two frontend tests fail identically on `main` and are unrelated: `shortcutRegistry › reports the intentional cross-scope overlaps` and `useExportTracker › tracks SQL file progress and clears terminal tasks`.


t8y2 (commented):
Applied the maintainer review fixes in 79514646.

findExplain now forwards the active optional executionId through queryStore and both adapters; the Tauri command uses the existing cancellable wrapper. Explain commands are classified as reads without masking writes in mixed batches, and calls that omit executionId remain compatible.

Validation: regression failures reproduced before the fix; 148/148 targeted tests now pass. Full frontend vue-tsc, changed-file formatting/lint, and git diff --check pass. Rust formatting also passes. Cancellation transport coverage uses mocked backends; no live database cancellation test was performed. cargo check --offline --locked -p dbx --no-default-features passes.

The patch is pushed and the remote SHA is verified. Merge remains gated on CI and merge readiness for this exact head; this PR has not been merged.

t8y2:
Thanks for the contribution! Merged in a78fc48a7, will be released in the next version.

PR #9455, opened 2026-09-17
feat(mongodb): support db.getSiblingDB()

### LINKED ISSUE #3936: [Bug] mongo筛选无效。脚本查询失败
### 数据库类型和版本

mongo

### 问题描述

1、直接输入筛选条件，{id:NumberLong('144115205316939462')}，没有任何响应
2、点击筛选，输入条件，点击应用，无任何反应

### 支持信息

当前版本为 0.5.52。
windows

### 优先级（是否紧急）

🔴 P0 紧急 · 下个版本必须修

### 补充信息

Image

Image

Image
db.getSiblingDB("iam_account").getCollection("user")
    .find({_id: NumberLong('144115205316939462')})
    .sort({phone: 1})
    .limit(21);

thailoc-dev:
## What

`db.getSiblingDB("name").<command>` runs the command against that database, for that command only — the shell's way of touching another database without switching to it.

```js
db.getSiblingDB("iam_account").getCollection("user")
  .find({ _id: NumberLong("144115205316939462") })
  .sort({ phone: 1 })
  .limit(21)
```

That is the exact command from #3936. Its other half, `NumberLong` in a filter, has worked since #8694; `getSiblingDB` was the remaining blocker, so this should close that issue.

Any collection or database command can follow the prefix: `db.getSiblingDB("x").orders.updateOne(...)`, `db.getSiblingDB("x").stats()`, `db.getSiblingDB("x").runCommand({...})`.

## How

Rather than adding a `database` field to every command variant, the parser wraps the inner command once:

```rust
MongoCommand::InDatabase { database: String, command: Box<MongoCommand> }
```

serialised as `{ kind: "inDatabase", database, command }`. The inner command is parsed exactly as it would be without the prefix — the prefix is stripped and the rest re-parsed as a normal `db.` command — so every existing parser, validator and error message applies unchanged.

The executors unwrap it:
- `execute_mongo_command_core` (CLI, MCP local) and the MCP remote backend recurse with the sibling database.
- The editor's `queryStore` swaps the target database for that one command and restores the session database afterwards, so unlike `use` nothing persists.

**Safety follows the wrapped command**: `is_mutating`, `is_dangerous` and the unbounded-filter checks delegate through the wrapper, so `db.getSiblingDB("x").c.deleteMany({})` is guarded exactly like `db.c.deleteMany({})`. The MCP server's production-database check now judges the *sibling* database (`MongoCommand::target_database()`), since that is the one being written to.

Malformed forms are rejected by name: a missing or non-string database, chaining two `getSiblingDB()` calls, or `use` after the prefix. Errors inside the wrapped command read as they would without it.

`getSiblingDB` is offered in completion after `db.`. Completing collection names *after* `getSiblingDB("x").` is not included — that needs the sibling database's collection list, which the completion does not load today.

## Testing

- `cargo test -p dbx-core --lib` — 6550 passed; `dbx-mcp` — 113 passed
- `pnpm test` — 15007 passed
- New tests both sides: the #3936 command verbatim, database-level commands, odd spacing, safety delegation, serialisation shape, a sibling command inside a multi-statement batch, and six malformed forms with their messages

Two frontend tests (`shortcutRegistry`, `useExportTracker`) fail identically on `main` and are unrelated.

Closes #3936


t8y2 (commented):
Maintainer patch — branch updated (old head `e6939ec37` → new head `f46e9f277`, rebased onto current `main` first).

Great feature overall, but as submitted `db.getSiblingDB("x")…` silently did nothing in the desktop editor: `queryStore.ts` unwrapped the wrapper on the frontend parse, then line ~6516 re-parsed the statement through the authoritative `mongoParseShellCommand`, which returns the still-wrapped `inDatabase` command, and the dispatch `switch` has no case for it — no result, no error. Two related gaps rode along the same root cause, so all three are fixed in the patch:

1. **Execution dispatch** (`queryStore.ts`): re-apply the unwrap to the authoritative parse result before dispatch, so the wrapped command executes against the sibling database and the `finally` restore keeps the session database unchanged (same semantics as the frontend path).
2. **Pagination planning**: `planMongoFindPagination` parses `db.<collection>.find(...)` directly and returned `null` for wrapper sources; the find case now plans against the prefix-stripped source, so chained `.skip()/.limit()` inside the wrapper still apply.
3. **Streaming export**: the export path re-parses the full statement too; it now unwraps and fetches pages against the wrapper's database instead of rejecting with `QUERY_RESULT_EXPORT_UNSUPPORTED_ERROR`.
4. **Rust guard parity** (`mongo_shell.rs`): the malformed-`getSiblingDB` diagnostic no longer fires for methods that merely start with the name (`db.getSiblingDBX(...)`), matching the TS parser; it now requires the `(` call form.

Checks run locally: `vitest run queryStore.mongoExecutionSummary.spec.ts` 5/5 (new regression test pins wrapper execution + session-database restore), `cargo test -p dbx-drivers mongo_shell` 43/43 (new negative test for the prefix guard), `cargo fmt --check` clean, `oxlint` clean on the touched TS files. CI will re-run on this head.

t8y2 (commented):
Thanks for rebasing onto the post-`renameCollection` main — that kept the maintainer patch intact and made this easy. One more rebase was needed on top (old head `f4e32563f` → new head `23030d811`): your rebase landed just before `db.createCollection()`/`db.dropDatabase()` support merged, so the three shared mongo files conflicted again. The resolution is a pure union — your getSiblingDB content is untouched, and the createCollection/dropDatabase parsing/completions from main are added alongside. Two stale assertions in `mongoShellCommand.test.ts` were dropped because both commands are now supported (their "not supported" messages no longer exist).

All local checks re-verified on the final head: `mongo_shell` Rust tests 44/44, mongo TS suites 125+/5+ green, `oxfmt`/`cargo fmt` clean. CI re-runs on `23030d811`; merge follows once green.

t8y2:
Thanks for the contribution! Merged in f1e0f54a2, will be released in the next version.

t8y2:
Post-merge maintainer follow-up: 23790bd5e is now on main.

The MCP validation path now unwraps the sibling-database command before applying target-database scope/policy checks, runCommand restrictions, aggregate output scope and production checks. This prevents the getSiblingDB wrapper from skipping checks that apply to the equivalent unwrapped command.

Validation: all three new security regressions failed before the fix and pass after it; the related Mongo tests pass 10/10, with cargo check, clippy and formatting also passing. No live database or desktop E2E was run. The exact-head CI run was cancelled by later main activity; it is not being reported as green or rerun.

PR #9461, opened 2026-09-17
fix(editor): keep SELECT * intact when a column completion inserts nothing after it

q396921921:
## What

Accepting a column completion — either a single item or the batch checkbox "insert selected fields" action — with the cursor positioned right before an existing `*` left the wildcard stranded in the query.

Repro:
1. Type `SELECT * FROM users`
2. Move the cursor to just before `*` (nothing typed there)
3. Trigger completion (Ctrl+Space) and accept one or more columns

Before: `SELECT id, name* FROM users` (invalid SQL, fails to execute)
After: `SELECT id, name FROM users`

This surfaced while investigating #9433 (a different, already-fixed bug in the same batch-selection checkbox feature) — it's a separate root cause in the shared completion-accept code, not the checkbox UI itself.

## Why

All three completion-accept code paths (single-item accept, snippet-apply accept, and the batch "insert selected fields" accept via `applyBatchColumnSelection`/`batchColumnSelectionReplaceTo`) compute their replacement `to` boundary by only checking for a matching closing quote or, in INSERT mode, a stray `)`. None of them accounted for a `*` sitting immediately after the cursor, so it was never consumed by the edit.

## Fix

Added `shouldSwallowSelectStar(from, to, nextCharacter)` in `batchColumnSelection.ts`: it extends the replacement range past the `*` only when the completion replaced nothing (`from === to`). That's the same condition that reliably distinguishes the SELECT wildcard from a real multiplication operator — `SELECT price*qty` has a typed prefix before the cursor (`from < to`), so it's left untouched. Wired into all three accept paths.

## Testing

- Added unit tests for `shouldSwallowSelectStar` and the new `batchColumnSelectionReplaceTo` case (empty-prefix `*` vs. multiplication), plus updated the existing tests for the new `from` parameter.
- Full suite: `pnpm --filter dbx vitest run` (via `apps/desktop`) — 11756/11758 passed; the 2 failures are pre-existing and unrelated (a date-relative backup-filename test and a shortcut-conflict test), confirmed unaffected by this change.
- `vue-tsc --noEmit` passes.
- Manually verified in a local dev build (isolated SQLite connection):
  - Single-accept path: `SELECT * FROM users` → accept `id` → `SELECT id FROM users`
  - Batch checkbox path: check `age` + `city` → insert → `SELECT age, city FROM users`, executes successfully
  - Safety check: `SELECT age*qty FROM users`, accept completion on `age` → `SELECT age *qty FROM users` (multiplication operator preserved)


q396921921:
## 复现与修复截图

**复现（修复前）**：在 `SELECT * FROM users`，光标停在 `*` 前触发补全，勾选 `email` + `name` 后点"插入已选字段"：

勾选 email/name

结果残留了原来的 `*`：

残留星号

执行直接语法报错：

执行报错

**修复后（本地隔离环境验证）**：

单条补全接受路径 → `SELECT id FROM users`，干净：

单条补全修复后

多选框路径：勾选 age + city：

多选框勾选

插入后 → `SELECT age, city FROM users`，干净：

多选框修复后

执行成功，返回列结构正确：

执行成功

**安全边界验证**：`SELECT age*qty FROM users`，光标停在 `age` 后接受补全 → `SELECT age *qty FROM users`，乘法运算符没有被误删：

乘法安全

t8y2 (commented):
Patched in 20ad60723.

Completion acceptance now consumes an existing asterisk only when SQL token context explicitly marks it as a standalone SELECT projection wildcard. Normal, snippet and batch acceptance preserve multiplication such as SELECT |*qty FROM users, while SELECT |* and SELECT DISTINCT |* retain the intended replacement behavior.

Validation: The new regression reproduced the unsafe inference before the fix. SQL-context and completion acceptance suites pass 172/172; vue-tsc, formatting, lint and git diff --check pass. No database or desktop E2E was run.

t8y2:
Thanks for the contribution! Merged in ca8bc9bf1, will be released in the next version.

PR #9462, opened 2026-09-17
fix(sql): stop quoting all-uppercase Oracle identifiers in star expansion

### LINKED ISSUE #9163: [Bug] 关闭双引号标识符后查询窗口选择*双引号依然存在

通用

1、复现步骤 / Steps: 关闭双引号标识开关，选中表右键新建查询，使用别名.*查询所有；
2、期望结果 / Expected: sql语句中无双引号；
3、实际结果 / Actual: sql语句中依然存在双引号；

Image

Image

DBX版本：0.6.13
运行环境：desktop
操作系统：window 11
系统架构：Inter i7

_No response_

q396921921:
## Summary
- On an Oracle connection, expanding `alias.*` in the SQL editor (autocomplete, or the "expand * to columns" action) wrapped every non-lowercase column name in double quotes, even plain all-uppercase ones like `AGENT_NAME` — e.g. `SELECT t."AGENT_NAME", t."PROTOCOL", t."SPARE1" FROM ...` instead of `SELECT t.AGENT_NAME, t.PROTOCOL, t.SPARE1 FROM ...`.
- Root cause: `quoteSelectStarColumnIdentifier()` decided whether a column needed quoting via `requiresPostgresIdentifierQuote()`, whose rule ("not all-lowercase needs quotes") is correct for PostgreSQL but wrong for Oracle, where unquoted identifiers fold to *uppercase* — so all-uppercase names never need quoting.
- Fix: detect Oracle explicitly and delegate to `quoteSqlIdentifier(identifier, "oracle")`, which already implements the correct rule and is already trusted by two sibling call sites in the same file. One line added, no other logic touched.
- Also corrected two existing tests in `semantic/completion.spec.ts` that had encoded this same bug as expected behavior (an Oracle `"ID"` column expected to come back quoted).

Fixes #9163

## Test plan
- [x] `pnpm vitest run apps/desktop/src/lib/__tests__/sql/sqlCompletion.context.spec.ts apps/desktop/src/lib/__tests__/sql/semantic/completion.spec.ts` — all pass, including 2 new regression tests for the reported scenario
- [x] `pnpm check` (connection-types / format / lint / typecheck / full test suite) — all green except one pre-existing, unrelated failing test (`shortcutRegistry.spec.ts`), confirmed to fail identically in isolation on unmodified code
- [x] Manual before/after verification against a real Oracle 21c instance, on both macOS (Chromium) and Windows 11 (the platform reported in the issue) — screenshots to follow in a PR comment


q396921921:
## Before / after evidence

Reproduced against a real Oracle 21c instance, same connection and same steps in both shots: place the cursor after `SELECT `, type `t.`, accept the `t.*` completion.

**Before (macOS, web build):**
16-after-star-expansion

**Before (Windows 11 — the platform reported in the issue, real browser):**
05-after-expansion-windows

**After (fix applied):**
after-fix-03-expanded

`SELECT t."AGENT_NAME", t."PROTOCOL", t."SPARE1" FROM dbx_test.t9163 t` → `SELECT t.AGENT_NAME, t.PROTOCOL, t.SPARE1 FROM dbx_test.t9163 t`

t8y2 (commented):
Patched in a2e71edb5.

Oracle star expansion now reuses the canonical Oracle-aware conditional quoting helper. Ordinary uppercase identifiers remain unquoted, while mixed/lowercase, reserved and non-bare names retain required quotes across menu, qualified-star and all-column completion paths. Other database quoting paths are unchanged.

Validation: Six new assertions failed before the fix. Three focused suites pass 293/293; vue-tsc, formatting, lint and git diff --check pass. Cross-checked the existing Oracle quoting path and DBeaver. No Oracle database or desktop E2E was run.

t8y2:
Thanks for the contribution! Merged in aa4ed91c4, will be released in the next version.

PR #9470, opened 2026-09-18
fix(mcp): hide the dead next button on the last MCP authorization step

### LINKED ISSUE #9309: [Feature] MCP权限管理最后一步按钮文案建议改为“确定”
### 来源
QQ 群反馈 · **反馈人**：Body

### AI 摘要
用户建议在MCP权限管理流程的最后一步，将按钮文案从“下一步”改为“确定”，以更明确地表达操作完成。

### 原始描述
mcp权限管理最后一步，下一步是不是改成确定？

Abeautifulsnow:
Fixes #9309.

## Summary

The MCP authorization stepper (Settings → MCP → Permissions) ends on the "Summary and
review" step with a footer button that still reads **Next** and is hard-disabled — a dead
control pointing at a page that does not exist.

This removes it from the last step. The button stays in the DOM as a same-size `invisible`
placeholder so the centred step-progress label does not shift between steps.

## Why not "change Next to OK"

The issue asks for the label to become 确定 (OK). This deliberately does not do that:

1. **Nothing to confirm.** Every permission control saves as it is changed
   (`saveMcpPolicy`), so the last step holds no pending draft. The step is titled
   "Summary and review", but the "review" is decorative.
2. **It would duplicate a button that already exists.** The MCP tab has its own dialog
   footer whose first button is Close (`EditorSettingsDialog.vue:10067`). An OK that also
   closes would be a second button with the same meaning in the same dialog.
3. **The usual precedent does not apply.** `DataGenerateDialog.vue:1282-1298` is often cited
   for "last step renders OK", but its OK appears only *after* a real action has run
   (`startInsert`) — "an action happened, confirm the result". Nothing happens here.

The root of the reporter's confusion: this UI looks like a wizard but behaves as section
navigation. All four steps are directly reachable from the clickable step strip at the top,
and the footer's Previous/Next are only a convenience — so dropping the terminal Next costs
no navigation.

## Changes

| File | Change |
|---|---|
| `apps/desktop/src/components/settings/McpAuthorizationStepper.vue` | New `isLastStep` computed; footer button is now `:class="{ invisible: isLastStep }"` + `:disabled="isLastStep"` (was `:disabled="currentIndex === steps.length - 1"`) |
| `apps/desktop/src/components/settings/__tests__/McpAuthorizationStepper.spec.ts` | New — 7 cases; the component previously had no test coverage |

No i18n key is added or changed and no locale file is touched — `settings.mcpAuthNextStep`
is still used by steps 1–3.

### Why `invisible` rather than `v-if`

The footer is a `flex ... justify-between` row (`[Previous] [progress] [Next]`). Removing the
third child shifts the progress label sideways, and a fixed-width spacer cannot compensate
because its width would have to track a label that changes per locale. Keeping the same
element with the same text and changing only its visibility makes the right-hand column
pixel-identical across all four steps.

## Verification

- `oxlint --vue-plugin` — clean
- `oxfmt --check` — clean
- `vue-tsc --noEmit --project apps/desktop/tsconfig.json` — exit 0
- `apps/desktop/src/components/settings/__tests__/` — 4 files / 24 tests pass
- The new spec is a real regression guard: with the component reverted, exactly the two
  last-step assertions fail
- Manual check in the running app: step 4 shows only Previous at the bottom-left and no
  button at the bottom-right, and the progress label does not shift when switching between steps

## Notes

- The last-step button was already `disabled` before this change, so the spec's focus
  assertion passes both before and after — it guards against a future removal of `disabled`
  rather than against #9309 itself.


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 745c201f0, will be released in the next version.

PR #9490, opened 2026-09-18
feat(mongodb): show shell command diagnostics in the editor as you type

thailoc-dev:
## What

MongoDB shell commands are now checked in the editor as you type, the way Redis and Oracle already are: a command that would fail gets a red underline with the same message Run would produce, and a risky one gets a yellow one.

```js
db.reports.find({ id: Foo("x") })
                      ~~~~~~~~   Unsupported value Foo(...) in the filter argument of find(). Supported value constructors: …

db.reports.replaceOne({ a: 1 }, { $set: { b: 2 } })
                                  ~~~~   replaceOne() replaces the whole document, so it must not contain update operators such as $set; use updateOne() to modify fields.

db.reports.deleteMany({})
~~~~~~~~~~~~~~~~~~~~~   ⚠ This delete has no effective filter and will delete every document in reports.
```

Until now every error appeared only after pressing Run — and on a legacy-agent connection that meant a round trip through the SSH tunnel and the Java agent before finding out.

## How

The editor's diagnostics pipeline (`StateField` + `Decoration.mark` in `QueryEditor.vue`) already renders `SqlSemanticDiagnostic`s and has an engine-specific slot; MongoDB's was an explicit `setSemanticDiagnostics([])` no-op. This adds `mongoSyntaxDiagnostics.ts`, mirroring `redisSyntaxDiagnostics.ts`, and plugs it in.

- Statements are split with `splitMongoCommandTextRanges` — the same splitter the executor uses — so an underline always corresponds to a statement that would fail when run.
- **Errors** come from `describeMongoCommandParseFailure` (#9155), so the underline says exactly what Run would.
- **Warnings** for commands that parse but deserve a second look: `updateOne`/`updateMany`/`deleteOne`/`deleteMany`/`replaceOne`/`findOneAndUpdate`/`findOneAndReplace`/`findOneAndDelete` with no effective filter (the same test the MCP safety policy uses), a `bulkWrite` containing such an operation, and `drop()`. A `db.getSiblingDB("x").…` wrapper (#9455) is looked through.
- **Underline placement:** the offending token where it can be found — a value constructor, the `$set` in a replacement, trailing text — otherwise the `db.collection.method` head, as Redis underlines the command name.
- **No noise while typing:** a command with unclosed delimiters under the cursor is in progress, not an error; it becomes one once the cursor leaves it. Diagnostics for the other, completed statements stay visible meanwhile.

Like the Redis diagnostics, this runs regardless of the "SQL semantic diagnostics" setting, since it is not SQL.

Known limitation, conservative rather than wrong: an unclosed statement *above* the one being typed stays silent until the cursor moves off it, because the open brace swallows what follows into one block. The executor has the same blind spot.

Because the diagnostics call the parser, they track every parser change automatically — the `bulkWrite` and `explain` support merged since this was written needed no changes here.

## Testing

- `pnpm test` — 15391 passed
- New `mongoSyntaxDiagnostics.test.ts`, 9 tests asserting the **underlined text** rather than line numbers: constructor and operator placement across lines, head placement for unsupported methods, argument shapes and `getCollection`/`getSiblingDB` call forms, every warning kind with a bounded counter-example, per-statement reporting, and typing behaviour


t8y2 (commented):
Patched in 95886d762.

Added missing unbounded-filter warnings for findOneAndUpdate, findOneAndReplace and findOneAndDelete. Mongo diagnostics now use the existing per-statement incomplete-input handling instead of suppressing every statement while the current statement is unfinished. Runtime execution safety checks are unchanged.

Validation: The new regressions reproduced both issues before the fix. The diagnostics and shell-command suites pass 126/126; vue-tsc, formatting, lint and git diff --check pass. No MongoDB or desktop E2E was run.

t8y2 (approved):
Re-reviewed d0df32b31 after the rebase and follow-up fixes.

The range-diff confirms the original contribution and maintainer fix were preserved. The unsupported-method fixture now avoids renameCollection, and the follow-up correctly handles getSiblingDB write warnings and call-form command-head underlining.

Validation: the syntax diagnostics, shell-command and completion suites pass 182/182 on this exact head. No database or desktop E2E was run; this approval does not claim pending GitHub checks have passed.

t8y2:
Thanks for the contribution! Merged in 28c219ff0, will be released in the next version.

PR #9494, opened 2026-09-18
feat(mongodb): support db.createCollection() and db.dropDatabase()

thailoc-dev:
## What

Two database-level shell helpers that were rejected:

```js
db.createCollection("events")
db.createCollection("logs", { capped: true, size: 1048576, max: 1000 })
db.createCollection("v", { validator: { $jsonSchema: { required: ["a"] } } })
db.dropDatabase()
```

## How

Both are shorthand for a server command, so — as with `db.stats()` in #9233 — the parser maps them onto `runCommand`, which is fully supported on every path including the legacy agent:

| shell | command |
| --- | --- |
| `db.createCollection(name, options?)` | `{ create: name, ...options }` |
| `db.dropDatabase()` | `{ dropDatabase: 1 }` |

No driver, agent or IPC changes. `createCollection` options (`capped`, `size`, `max`, `validator`, and the rest) pass straight through for the server to validate, which keeps the parser out of the business of tracking which options each MongoDB version accepts. The name must be a non-empty string without `$`, options must be a document, and an options key of `create` is rejected so it cannot shadow the collection name.

`dropDatabase` is classed **dangerous**, which it should be, because `runCommand` is. That has a side effect worth stating: `createCollection` is classed dangerous too, so under the MCP/CLI safety policy it needs "dangerous" writes enabled rather than plain writes. That is conservative for an operation that only creates an empty collection; if it turns out to matter, a dedicated command kind would separate the two.

Also updated: completion after `db.`, and the parse-failure tables so neither is reported as unsupported.

## Testing

- `cargo test -p dbx-core --lib mongo_shell` — 41 passed
- `pnpm test` — 15385 passed
- New tests both sides: both commands, capped and validator options, dangerous classification, and the rejected forms with their messages (arguments to `dropDatabase`, missing/non-string/empty/`$` names, non-document options, a `create` key, trailing text)


t8y2:
Thanks for the contribution! Merged in 7c51454bf, will be released in the next version.

PR #9496, opened 2026-09-18
feat(mongodb): support db.collection.renameCollection()

thailoc-dev:
## What

```js
db.orders.renameCollection("orders_2024")
```

Works on both the native driver and the MongoDB Legacy Agent, so it runs on MongoDB 3.6 as well as current servers.

## How

The native backend already had `mongo_rename_collection_core`, the Tauri command, the web route and the frontend IPC — the sidebar's rename action uses them. What was missing was the shell command, the executor arms, and agent support (the core returned *"legacy agent does not support rename collection"*).

- **Parser (Rust + TS):** a `RenameCollection { collection, newName }` command. The name must be a non-empty string without `$` and must differ from the current name.
- **`dropTarget` is rejected, not ignored.** The shell's second argument, `renameCollection("x", true)`, silently deletes an existing collection named `x`. Accepting the call and dropping the flag would rename *without* that and fail confusingly if `x` exists; ignoring the destructive intent is worse. So it fails with *"dropTarget is not supported; drop the target collection first."* — explicit about what to do.
- **Java agent:** `rename_collection` via `collection.renameCollection(new MongoNamespace(...))`, with the same name checks as the native driver (`system.*` refused, same-name refused) so both paths reject the same input before asking the server. Capability `mongo_rename_collection` in both contract versions, so an older jar fails with an upgrade hint.
- **Executor arms** in `mongo_ops` and the MCP remote backend; `queryStore` dispatch; completion after a collection; parse-failure tables updated.

Classed mutating but not dangerous, like the other schema operations (`createIndex`, `dropIndex`).

## Testing

- `cargo test -p dbx-core --lib` — 6555 passed; `dbx-web` + `dbx-mcp` — 239 passed
- `./gradlew :common:test :mongodb:test` — 207 + 109 passed, both protocol-contract tests included
- `pnpm test` — 15385 passed
- New tests: Rust parser (happy path, serialisation, 7 rejected forms), Java (RPC recognition, 5 invalid name pairs), frontend (bracket/getCollection targets, 7 rejected forms with messages), completion guard updated

**Verified live on a throwaway MongoDB 3.6** through the agent with the built jar:

| case | result |
| --- | --- |
| rename `orders` → `orders_2024` with 2 documents | `ok`; `orders_2024` holds both documents, `orders` gone |
| same name / `system.*` target | rejected by the agent before the server, same messages as the native driver |
| source collection missing | server error `NamespaceNotFound`, passed through |
| previous jar, as control | `Unknown method: rename_collection` |

Two frontend tests (`shortcutRegistry`, `DocumentBrowserFieldSearch`) fail on `main` as well and are unrelated.


t8y2:
Thanks for the contribution! Merged in 2db28fc23, will be released in the next version.

PR #8877, opened 2026-09-11
feat(sql-file): import split SQL ZIP packages

hanhvs:
## Follow-up to split ZIP export

This is the import-side follow-up to **PR #8808**, which exports large SQL datasets as a ZIP package containing statement-safe `part-N.sql` files and `manifest.json`.

PR #8808 intentionally made each part independently valid SQL and preserved strict manifest ordering. This PR consumes that exact package format so users can round-trip a large export without manually unzipping, ordering, and executing each part.

## Memory-pressure protection

Large one-file imports can create sustained database pressure, particularly for SQL Server containers with limited memory. This PR does **not** claim to eliminate database-side out-of-memory errors: a single oversized statement, index rebuild, or undersized DB instance can still fail.

It reduces the common large-import risk by:

- Never loading the entire package or all parts into DBX memory.
- Streaming only the current SQL part through the existing statement executor.
- Executing `manifest.parts` strictly one at a time, never concurrently.
- Waiting 500ms after a part reaches terminal completion before beginning the next part, giving the DB connection/server a recovery window.
- Stopping immediately on terminal error or cancellation, so no later part adds more load.

The ZIP validator also caps extracted package size and individual part size, rejecting unsafe archives before they can consume unbounded local memory or disk.

## What this adds

- Validates DBX split SQL ZIP packages before use:
  - requires `manifest.json`
  - validates `dbx-sql-export-parts-v1`
  - preserves manifest order, never ZIP physical or lexical order
  - rejects unsafe/traversal paths, duplicate parts, missing entries, invalid manifests, oversized parts, and oversized total decompressed content
- Safely extracts parts to trusted temporary directories.
- Supports ZIP package preview/upload in desktop and web UI.
- Cleans extracted package directories after execution.
- Shows package part count in the SQL file UI, localized in every supported locale.

## Import semantics

- No `GO` is generated or injected.
- Each part is already statement-safe from PR #8808.
- Existing `Continue on error` semantics remain within each individual part.

## Relational constraints

A temporary relational-constraint bypass option is planned, but is intentionally **not exposed yet**. It must disable and restore constraints in the same pinned session, including error and cancellation paths, and must be scoped to engines where this is safe. A UI toggle without guaranteed restoration would be unsafe.

## Validation

- `cargo check --workspace --lib --tests`
- `cargo test -p dbx-core --lib sql_file_zip_package`
- `pnpm typecheck`
- ZIP validation tests cover manifest ordering and unsafe traversal paths.

PR #9509, opened 2026-09-18
fix(sql-file): extend relational constraint bypass to Postgres and SQL Server

hanhvs:
## Hotfix for #8877

Follow-up hotfix to the import split-ZIP PR (#8877). The "Temporarily disable relational constraints" toggle added there only implemented the bypass for MySQL-compatible connections (`SET FOREIGN_KEY_CHECKS`). Reported after merge: a user importing into **SQL Server** saw the toggle have zero effect — FK/duplicate-key errors during import were not bypassed, because the backend never had a SQL Server (or PostgreSQL-family) implementation.

## What changed

- **PostgreSQL-family** (Postgres, GaussDB, openGauss): disables/re-enables via `ALTER TABLE ... DISABLE TRIGGER ALL` / `ENABLE TRIGGER ALL` across every table in the database (system schemas excluded; per-table insufficient-privilege errors are swallowed instead of aborting the whole bypass for multi-owner databases).
- **SQL Server**: disables/re-enables via `ALTER TABLE ... NOCHECK CONSTRAINT ALL` / `WITH NOCHECK CHECK CONSTRAINT ALL` across every table the importing login can alter (pre-filtered with `HAS_PERMS_BY_NAME`).
- Both wired into `execute_sql_file_content` and `execute_sql_file_paths`, with restore guaranteed on success, error, and cancellation — same guarantee the existing MySQL path already has.
- MySQL behavior/implementation is unchanged.
- Frontend (`SqlFileExecutionDialog.vue`): the toggle now also shows for PostgreSQL-family and SQL Server connections, not just MySQL-compatible ones.

## Known limitation (by design, same as MySQL)

This does **not** bypass unique/primary-key violations on any engine — only foreign-key (and on PostgreSQL, other trigger-enforced) constraints. Duplicate-key errors are still skipped/logged via the existing `continueOnError` flag, not suppressed at the constraint level.

## Testing

- `cargo check --workspace --lib --tests` — clean
- `cargo test -p dbx-core --lib sql_file_import` — 50/50 passed
- `cargo fmt --check` / `cargo clippy -p dbx-core --lib` — clean (no new warnings)
- `pnpm typecheck` / `pnpm lint` — clean (no new warnings)

Marked draft pending review of the SQL Server/Postgres SQL against a live instance.


t8y2 (changes requested):
Thanks for extending the constraint-bypass toggle to PostgreSQL-family and SQL Server — the per-engine kind detection and catalog-level ALTER approach mirroring DBeaver's structure is sound, and CI is green. Two blocking issues before this can merge:

1. **Restore-after-cancel never runs** (`crates/dbx-core/src/data/sql_file_import.rs:633`, `:937`): the re-enable call passes the already-cancelled import `CancellationToken` into `execute_sql_statement_with_options`, whose pre-dispatch cancellation check (`crates/dbx-core/src/query/mod.rs:2714`) returns early — so after a user cancels, FK/trigger enforcement stays disabled database-wide (catalog-level, affects all sessions), and the `let _ =` discards the error silently. This contradicts the PR body's "restore guaranteed on cancellation". Please pass a fresh token (or `None`) for the enable calls. The existing MySQL path doesn't have this hole because it uses `conn.query_drop` on the pinned connection.

2. **PostgreSQL toggle is a silent no-op for regular owners**: `ALTER TABLE ... DISABLE TRIGGER ALL` hits internal FK/RI triggers that only superusers may disable (non-superuser owners get SQLSTATE 42501), and the per-table `EXCEPTION WHEN insufficient_privilege THEN NULL` swallows every one — so for a non-superuser importing role the toggle silently changes nothing and FK errors continue. Since the PR is draft pending live verification anyway, please verify against a live PostgreSQL instance, and consider surfacing a warning when the disable pass altered zero tables.

Also, there are no new tests for the bypass on either engine (live tests still construct `skip_relational_constraints: false`) — please add coverage for both the restore path and the insufficient-privilege path, or at minimum the live-instance verification notes.

hanhvs:
Thanks for the thorough review — both issues fixed in c9f9a7475:

1. **Restore-after-cancel**: both restore call sites (`execute_sql_file_content` and `execute_sql_file_paths`) now pass a fresh `CancellationToken::new()` instead of the import's token, so the restore SQL is actually sent even after a user cancels. The previously-silent `let _ =` now logs the error via `log::error!` if the restore itself fails.

2. **PostgreSQL silent no-op**: `POSTGRES_DISABLE_ALL_RELATIONAL_TRIGGERS_SQL` now counts attempted vs. successfully-disabled tables and `RAISE EXCEPTION`s when every table was skipped for `insufficient_privilege`, so a non-superuser importing role gets a clear failure instead of a toggle that silently did nothing.

Added unit tests for both (source-level assertions since there's no live PostgreSQL/SQL Server instance in this test module) plus coverage for `relational_constraint_bypass_kind`'s per-engine dispatch. `cargo test -p dbx-core --lib data::sql_file_import` — 54/54 passing, `cargo fmt`/`cargo clippy` clean.

Still marked draft-turned-ready pending your live-instance verification note from the original review — happy to add live-instance test scaffolding (following the existing `#[ignore]`d `live_sqlserver_*`/`live_postgres_*` pattern) if that's preferred over the source-level tests for merge.

t8y2 (approved):
Thanks for the quick turnaround — both blockers are convincingly resolved: the restore calls now use a fresh CancellationToken (verified the pre-dispatch cancellation check in query/mod.rs no longer blocks the re-enable, and failures are logged instead of swallowed), and the PostgreSQL disable block now raises loudly when every table was skipped for insufficient_privilege rather than silently no-op'ing. CI is fully green on this head. Live verification against a real non-superuser PostgreSQL role on our test instance will follow around release; the #[ignore]d live-test scaffolding you offered would be a welcome follow-up.

t8y2:
Thanks for the contribution! Merged in 30fe61d4, will be released in the next version.

PR #9535, opened 2026-09-18
feat(mcp): record SQL executions in query history

caichangqing1120:
## Summary

- persist SQL dispatched by `dbx_execute_query` and `dbx_execute_batch` into the existing query history for local and web-backed MCP runtimes
- mark MCP-originated entries through `details_json` and surface an `MCP` badge in the query history UI
- record backend execution failures accurately while excluding policy/preflight rejections and keeping history persistence failures non-fatal
- keep multi-statement batch execution as one history entry and derive operation/activity metadata from the SQL

Closes #9527

## Verification

- `cargo test -p dbx-mcp --lib --no-default-features --features sqlite-bundled` (`88 passed`)
- `pnpm exec vitest run apps/desktop/src/lib/history/__tests__/historyEntrySource.spec.ts` (`2 passed`)
- `pnpm typecheck`
- `cargo fmt --check -p dbx-mcp`
- focused `oxlint` and `oxfmt --check` for the changed frontend files
- `git diff --check`

## Known unrelated baseline

The full frontend suite previously completed with `15484 passed` and one unrelated existing failure in `apps/desktop/src/lib/__tests__/editor/shortcutRegistry.spec.ts:447`. A targeted rerun on the final rebased branch still reports `44 passed, 1 failed` because `conflicts.viewTableDdl` is undefined. This PR does not modify the shortcut registry or that test.

PR #9540, opened 2026-09-18
feat(mcp): make dbx_execute_query row limit configurable

Abeautifulsnow:
## Problem

Refs https://github.com/t8y2/dbx/issues/9482

External AI apps calling the DBX MCP server could not get more than 100 rows out of
`dbx_execute_query`. The count was hardcoded in two independent places, so changing either
one alone did nothing:

- `crates/dbx-mcp/src/server.rs` built the tool arguments as `json!({ "sql": ..., "limit": 100 })`,
  so the model could never ask for a different number.
- `crates/dbx-core/src/ai/agent_tools.rs` re-clamped whatever arrived with
  `.min(MAX_ALLOWED_ROWS)`, where `MAX_ALLOWED_ROWS = 100`.

Deep-analysis queries that legitimately need a few hundred rows had no way to get them, and
nothing in the result hinted that a larger request was possible.

## Solution

`dbx_execute_query` gains an optional `max_rows` parameter:

| Input | Result |
| --- | --- |
| omitted | 100 rows — identical to today |
| `500` | 500 rows |
| `0` | clamped to 1 |
| `100000` | clamped to 1000 — clamped, never an error |

`MAX_EXECUTE_QUERY_ROWS = 1_000` in `crates/dbx-core/src/ai/agent_too
definition of the ceiling; the MCP layer imports it instead of declaring its own. It sits
deliberately below the driver fetch ceiling (`dbx_drivers::execution:
`const _: () = assert!(MAX_EXECUTE_QUERY_ROWS < dbx_drivers::execution::MAX_ROWS);` enforces that
ordering at compile time, so the published "up to N rows" contract ca

### What changed, by layer

- **dbx-core** — new `MAX_EXECUTE_QUERY_ROWS` ceiling plus an
  `execute_query_row_limit(requested, db_type)` resolver. Splitting by `db_type` keeps the
  MongoDB shell ceiling at 100 while raising the SQL one, so the two
  true. The result-truncation notice now names the effective cap, and only claims the row cap when
  it was actually reached (a driver-side size budget can set `truncat
- **dbx-mcp server** — `max_rows` on `ExecuteQueryRequest`, published with a single concrete
  `integer` type (per the #6344 union-type fix), clamped to the share
  value used to be, plus a rewritten tool description.
- **dbx-mcp Web backend** — the POST body now carries `maxRows` and c
  this transport never reaches the in-process core clamp.
- **Docs** — `mcp.mdx`, `mcp.cn.mdx` and the npm README (both languag

## Behaviour changes worth knowing

1. **Web mode now fetches what it renders.** The POST used to omit `m
   back to `MAX_ROWS = 10_000` — it fetched 10,000 rows in order to render 100. For the default
   path the rendered output is byte-identical (`format_query_result`
   For the `cell_char_offset`/`cell_char_limit` path it is **not**: that renderer printed every
   fetched row, i.e. up to 10,000, contradicting the documented "at m
   branch is now correctly bounded — a fix, not a no-op.
2. **The in-app AI assistant's SQL ceiling rises from 100 to 1000**,
   `execute_query` tool. Its MongoDB ceiling is unchanged at 100, thanks to the `db_type` split.

## Deliberately out of scope

- **MongoDB over MCP stays at 100.** The Mongo branch returns before the tool arguments are
  built, and the databases in the report do not include Mongo. Disclo
- **Multi-statement scripts routed to `dbx_execute_batch` stay at 100 rows per statement.** A
  script multiplies its row budget by the statement count, so the bat
  `execute_query` default. Disclosed in the tool description and in the corrected comment.

## Testing

| Command | Result |
| --- | --- |
| `cargo test -p dbx-core --lib ai::agent_tools` | 59 passed |
| `cargo test -p dbx-mcp --lib` | 92 passed |
| `cargo test -p dbx-mcp --test protocol` | 11 passed |
| `cargo test -p dbx-core --lib` (full) | 2584 passed, 5 failed |
| `cargo clippy --workspace --all-targets -- -D warnings` (CI's invocation) | no diagnostics in the changed files |

New coverage: `execute_query_row_limit_splits_sql_and_mongo_ceilings`,
`execute_query_tool_schema_publishes_the_sql_row_ceiling_only`,
`query_result_formatter_only_claims_the_row_cap_when_it_was_reached`,
`execute_query_forwards_and_clamps_max_rows`, and `execute_query_forw
last one is the only test that goes through rmcp JSON serialization, i.e. the only proof that the
published name `max_rows` maps to the internal `limit` argument.

The 5 failures in the full `dbx-core` lib run are pre-existing on a W
to this change — `connection::jdbc::tests::imports_offline_jdbc_plugin_and_rewrites_bundle_paths`
(path separators), `data::docs::export::tests::docs_export_bundle_is_t
fixtures), the two `legacy_xls_*` cases (CFB fixture), and
`persistence::storage::tests::import_user_data_db_replaces_empty_targ
control run at the parent revision in a temporary worktree produced the identical failure set.


t8y2 (commented):
维护者冲突代解 9c2adbdbc（原 head 7576fea5，base 为合并 #9535 后的 main 0f6db4b2c）：#9535 在 `dbx_execute_query` 的同一语句块新增了 `history_sql` 克隆（用于 MCP 查询历史记录），与本 PR 的 `max_rows` 可配置化发生单文件单块冲突。解法为意图并集——保留可配置的 `max_rows`（clamp 1..MAX_EXECUTE_QUERY_ROWS，默认 100）并保留 `history_sql = request.sql.clone()`，`arguments` 的 limit 采用 `max_rows`；`history_sql` 仍按 #9535 语义传给 save_mcp_sql_history。已跑 `cargo check --no-default-features -p dbx-mcp` 与 `cargo fmt`，均通过；`git diff --check` 干净。无依赖变更。

t8y2 (commented):
Follow-up to the conflict resolution: the merged history test helper (33770c11d) also constructs `ExecuteQueryRequest`, which now requires the new `max_rows` field — initialized to `None` so the test keeps exercising the default limit. `cargo check --no-default-features -p dbx-mcp --all-targets` and `cargo fmt` pass.

t8y2:
Thanks for the contribution! Merged in 5278594e, will be released in the next version.

PR #9561, opened 2026-09-18
fix(xugu): expose identity columns to data generation

mapan0424:
## Summary

This pull request fixes Xugu identity-column metadata exposed to DBX data generation.

Xugu represents `IDENTITY` / `AUTO_INCREMENT` columns through its serial metadata catalogs instead of returning the generation expression through the normal column-default field. The shared data generator therefore treated an identity primary key as an ordinary required column and could generate duplicate values, resulting in a unique-constraint error.

The change:

- enriches Xugu column metadata with `extra = "auto_increment"` when the column is backed by Xugu identity/serial metadata;
- reuses the existing desktop data-generator contract, so no frontend or shared database-driver behavior changes;
- keeps ordinary sequence defaults such as `DEFAULT sequence.NEXTVAL` represented as column defaults;
- makes identity enrichment optional, preserving normal column metadata when the account cannot read the sequence catalog;
- avoids ambiguous case-insensitive catalog matches and closes the first metadata result set before issuing the enrichment query.

This is intentionally a separate PR from the table-import change so each behavior can be reviewed and reverted independently.

## Validation

- `GONOSUMDB=gitee.com/XuguDB/go-xugu-driver go test ./...`
- `GONOSUMDB=gitee.com/XuguDB/go-xugu-driver CGO_ENABLED=0 go build -trimpath -ldflags='-s -w'`
- Live Xugu validation with an identity table confirmed:
  - `get_columns` reports only the identity column as `auto_increment`;
  - generated inserts omit the identity column;
  - generated values follow the configured start/increment sequence;
  - ordinary columns remain unchanged.
- A separate `DEFAULT sequence.NEXTVAL` table remained represented by `column_default` and continued to generate sequential values.
- The diff is limited to `agents/drivers/xugu`; no other database driver or frontend files are changed.


t8y2:
Thanks for the contribution! Merged in 5a597772, will be released in the next version.

PR #9562, opened 2026-09-18
feat(plugins): hand SOCKS5 transport proxy routes to connection providers

jinpy666:
## Problem

Plugin connections cannot safely combine transport layers with multi-endpoint protocols.

The desktop tunnel path forwards exactly one remote endpoint taken from the connection's standard host/port fields. Providers whose endpoints live in provider-defined fields have empty host/port - the Kafka Studio plugin keeps its bootstrap list in a provider textarea and declares no host/port binding. When a user attaches an SSH/proxy transport layer, the tunnel forwards to an **empty target**, the plugin dials into it and times out with no hint about the real cause (reported as jinpy666/dbx-plugin-kafka#28).

Even with host/port filled, a static tunnel reaches only one broker: Kafka test connections can succeed over the bootstrap endpoint while topic browsing fails against unreachable advertised.listeners addresses.

## Solution

Connection providers can now opt in with `proxy_route` on the connection-provider contribution. For those, DBX resolves the transport chain into a host-managed SOCKS5 route and delivers it in the lifecycle payload as `runtime.proxy`, keeping `runtime.host`/`runtime.port` at the connection's logical endpoint:

- SSH as the final layer exposes the hop's dynamic SOCKS5 endpoint (`start_transport_layers_with_final_ssh_socks5`, previously gated to the RocketMQ path), so every advertised listener is reachable through one tunnel
- a SOCKS5 proxy layer is used directly, tunneled through any preceding layers
- HTTP-tunnel chains fall back to the static path

This mirrors the existing RocketMQ SOCKS handling (`rocketmq_socks_proxy_for_transport_layers`) at the plugin-contribution level.

Providers without the flag keep today's static-tunnel behavior (fine for single-endpoint providers such as SSH/LDAP, which declare binding host/port fields). A plugin connection that would tunnel to an empty endpoint now fails with an actionable error instead of hanging until timeout.

The Kafka Studio plugin already implements the client side of this contract (dialing its full bootstrap and metadata-resolved broker set through `runtime.proxy`), so Kafka-over-SSH/SOCKS works end to end once its manifest opts in.

## Changes

- `plugins/manifest.schema.json`: new optional `proxy_route` contribution flag
- `crates/dbx-plugin-runtime`: parse the flag (`deny_unknown_fields` stays strict), `PluginRuntimeProxy` type, `runtime.proxy` in the lifecycle payload, `wants_transport_proxy_route` resolver; `test_connection`/`connect_connection`/`invoke_connection_action` thread the route
- `crates/dbx-core/src/connection/mod.rs`: `connection_endpoint` refactor + plugin SOCKS route branch + empty-endpoint guard; `plugin_connection_endpoint` for plugin call sites
- `crates/dbx-drivers`: un-gate `start_transport_layers_with_final_ssh_socks5` (generic transport capability)
- `crates/dbx-web/src/routes/connection.rs`, pool creation path, plugin action invocation: pass the route through
- docs: `plugins/README.md` and the plugin-development docs pages (en/cn) document the flag and the `runtime.proxy` payload; desktop TS types updated

## Testing

- `cargo test -p dbx-core -p dbx-plugin-runtime -p dbx-drivers -p dbx-web`: 69 suites, 0 failures (new tests: flag parsing defaults/opt-in, `runtime.proxy` payload shape)
- `cargo clippy` clean for touched crates; `cargo fmt` applied
- frontend: `vue-tsc` typecheck, oxfmt, oxlint pass (`node scripts/run-check.mjs` connection-types/format/lint/typecheck green; the vitest `shortcutRegistry` failure reproduces on pristine main under macOS and is unrelated)


t8y2:
Thanks for the contribution! Merged in 2b15327d, will be released in the next version.

PR #9602, opened 2026-09-19
feat(mcp): add native MySQL session transactions

xcstatus:
## 变更说明 / Change Description

Pinned MCP sessions retain connection state, but cannot keep an explicit transaction open across calls. This change adds `enable_transactions: true` to `dbx_open_session` and explicit begin, commit and rollback tools for native local MySQL. Each opted-in session owns one physical connection, serializes operations and returns per-call transaction state and outcome.

A recoverable statement error such as MySQL 1062 can leave the transaction active. Uncertain outcomes, including a lost COMMIT acknowledgement, become terminal `unknown` without reconnecting or replaying SQL. A typed MySQL path reads fresh transaction status from the same connection.

Scope, read-only, production and exact-confirmed-SQL checks remain enforced and are rechecked after queueing. A parsed safe SQL subset, shared lifecycle invalidation, an aggregate 32-owner budget and bounded cleanup protect the owned connection. Existing stateless queries, legacy sessions and atomic batches retain their APIs. Rollback guarantees require transactional tables such as InnoDB; unsupported backends fail explicitly.

为原生 MySQL 的 MCP 固定会话增加跨调用事务控制和逐次状态返回，保留权限检查与旧接口。重复键等可恢复错误不自动结束事务；提交确认丢失时永久返回未知结果，不重连、不重放。中英文文档说明使用方式与限制。

## 变更类型 / Change Type

- [x] 新功能 / New feature
- [ ] Bug 修复 / Bug fix
- [ ] 代码重构 / Code refactoring
- [x] 文档更新 / Documentation
- [ ] CI / 构建 / CI or build

## 涉及前端 / Frontend Changes

The desktop MCP policy tool selector now includes begin, commit and rollback, with labels in all 10 supported locales. English and Chinese MCP documentation explain the new transaction tools and limitations.

## 验证 / Validation

- [ ] `make check` 通过 / `make check` passes

Candidate `652fa934` merges main at `d7e1b47a` into `e76b4117`, preserving the author's rustls test bootstrap fix. The backend conflict retains both transaction lifecycle tests and the upstream standalone plugin engine-gate fix. Core transaction, HTTP, session and native test files are unchanged from `e76b4117`.

Fixed-candidate validation on Rust 1.97.1 with the current CI application/fast feature union:

- Targeted nextest lifecycle/owner/cap/policy/engine-gate regressions: 11 passed.
- HTTP regressions: 4 passed.
- Strict workspace clippy, MCP package formatting and documentation export freshness check passed.
- Separate native MySQL 5.7 / InnoDB / REPEATABLE READ gate on `652fa934`: 1 passed, 0 ignored, 13.13s; exact disposable fixture cleanup succeeded.

The native gate verifies stable physical connections, real INSERT contention, 1062 active continuation, current reads, permissions, close/cancel/TTL cleanup, and terminal unknown after connection loss. The fault proxy observes a successful server COMMIT while withholding its acknowledgement, verifies committed data, and confirms exactly one COMMIT without replay. Credentials and environment identifiers remain local.

Validation limits: the complete local application nextest run, with local stack/temp-path/proxy adjustments, finished with 3525 passed, 2 failed and 193 skipped. The Consul and custom-CA tests passed in isolated reruns; the two full-suite failures remain unresolved and their root cause is not established. The broad Makefile checks were not run. Whole-repository formatting/whitespace checks also report inherited vendor issues; the MCP package and PR delta against main pass their checks. This is not a claim that the full local suite passed.

## 关联 Issue / Related Issue

Refs #9546.


t8y2 (commented):
The `rust-test (workspace)` failure on the four new `http.rs` tests is a rustls provider bootstrap race, not a logic bug in the transaction handling. Those tests build reqwest-backed clients — the direct `reqwest::Client::new()` calls and the `StreamableHttpClientTransport` created inside `open_active_transaction` — before any server-side session has installed a rustls `CryptoProvider`. In workspace builds multiple rustls crypto features are enabled, so rustls cannot automatically determine a process-level provider and every client construction panics with "No rustls crypto provider is configured". The suite only appeared green in runs where some other test happened to install the provider first, which is exactly the kind of ordering dependence CI exposes.

Pushed e76b4117 to fix it: the top of the `start_http_test_server` helper in `crates/dbx-mcp/src/http.rs` now installs `rustls::crypto::aws_lc_rs::default_provider().install_default()` before anything builds a client. All four failing tests route through that helper before their first client construction, and this is the same idempotent pattern the crate already uses in `backend.rs:3911` and `server.rs:422` (install is a one-time process-level operation; subsequent calls are no-ops). Verified locally with `cargo fmt --check` and `cargo check -p dbx-mcp --no-default-features`, both clean.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 1294708ee, will be released in the next version.

PR #9607, opened 2026-09-19
fix(redis): skip comment lines when executing editor commands

thailoc-dev:
## What

A `#` or `--` note in the Redis query editor is no longer sent to the server as a command.

```
# warm the cache
GET user:1
```

Before, running this produced two results: `Redis command is blocked for safety: #` for the first line (or, with the safety check off, Redis's own `ERR unknown command '#'`), then the `GET`. The editor's live diagnostics already treated the line as a comment and showed nothing, so the first sign of a problem was the failed run.

## How

Three places decided what a Redis "line" is, each with its own rule:

| | `#` line | `--` line |
|---|---|---|
| `buildRedisSyntaxDiagnostics` | comment | comment |
| `executableStatementRanges(sql, "redis")` | comment | command |
| `queryStore` execution | command | command |

They now share one predicate, `isRedisCommentLine` in `redisCommandTokenizer.ts`, and the executor takes its command list from `executableStatementRanges` — the same split that gives each result its `sourceFrom`/`sourceTo` — instead of splitting the text a second time. That also removes the possibility of the two lists drifting apart, which they did whenever a comment line was present (results after a comment were annotated with the range of the wrong line).

`--` follows the diagnostics' existing rule; Redis itself has no comment syntax, so the editor's convention is the only one that matters here.

## Testing

- New `queryStore` test: comment lines are not sent, and each result's source range points at its own line. Fails on `main` (the `#` and `--` lines are sent).
- `sqlStatementRanges` Redis cases extended with a `--` line for both the range list and the cursor lookup.
- `pnpm vitest` on the redis, statement-range and store suites: 2702 passed. `vue-tsc`, `oxfmt`, `oxlint` show no new findings.
- Checked against a throwaway `redis:7-alpine` container that `#` and `--` really are rejected by the server (`ERR unknown command '#'`).


t8y2 (commented):
Nice catch on the executor side — I pushed a follow-up in d48f296e1 because two upstream gates still raw-split the batch and would have kept the bug alive for default settings. The Redis safety pre-scan in useSqlExecution.ts (both the editor path and executeTargetSql) fed every non-empty line to classifyRedisCommandSafety, which fails closed on unknown tokens, so a `# comment` line classified "blocked" and the whole batch aborted with the blocked toast before your fixed executor ever ran. sqlLooksLikeMutation in readOnlyWriteAccess.ts had the same raw split, so a comment line above a read-only command still triggered a spurious read-only write-unlock prompt.

The patch filters comment lines with your own isRedisCommentLine predicate at all three sites (the predicate was already exported, so no signature changes) — everything that is not a comment line keeps the exact same fail-closed behavior. Checks: oxlint clean on both changed files, useSqlExecution.spec.ts 67 passed (added one case: default-settings Redis batch with `#`/`--` lines reaches the executor instead of being blocked), readOnlyWriteAccess.spec.ts 13 passed (extended the classification test: comment lines over GET stay non-mutating, over SET still mutate), and the redis tests in packages/app-tests/queryStore.test.ts still pass.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in c5312553f, will be released in the next version.

PR #9613, opened 2026-09-19
build(win7): make the WebView2 loader reproducible

luojiyin1987:
## Summary

- Vendor `webview2-com-sys` 0.38.2 with its x86, x64, and arm64 loaders.
- Add the pinned WebView2 1.0.902.49 loader under `win7/x64`.
- Select the legacy loader only for `x86_64-win7-windows-msvc`.
- Verify the static loader and probe DLL SHA256 values before use.
- Emit `rerun-if-changed` for every selected loader file.
- Stop modifying files inside the Cargo registry.

## Correctness change

The old script replaced a library inside the Cargo registry. The replacement was not an explicit Cargo input. A compile cache could therefore return an artifact linked against different loader bytes.

The patched crate now owns every loader input. Cargo tracks the selected files through `build.rs`. The build fails when the Win7 loader hashes differ.

## Scope

This PR does not change CI jobs, sccache, LTO, job splitting, or gate policy. CI performance work remains separate.

## Validation

- `node --test .github/scripts/webview2-win7-loader.test.mjs`
- `pnpm exec oxfmt --check .github/scripts/webview2-win7-loader.test.mjs`
- `cargo metadata --locked --format-version 1 --no-deps`
- `cargo check --locked -p webview2-com-sys --target x86_64-pc-windows-gnu`
- `git diff --check`


luojiyin1987:
Addressed the review findings in `dc90ebcde`:

- Added `vendor/webview2-com-sys/**` to `windows_win7_bundle` routing.
- Added a workflow regression assertion for that path.
- Reworded the stale cache-policy test. The PR still leaves caching disabled.
- Removed the stale registry-patch text from the loader audit error.
- Documented the upstream x86 loader and its SHA256.
- Added the upstream MIT LICENSE from commit `b74dc5e2b394044bea5191052868ce7a106c202c`.

Validation passed:

- 14 Node workflow and loader tests
- 17 Windows installer tests
- `actionlint`
- formatter checks for changed JavaScript and TypeScript files
- `git diff --check`

luojiyin1987:
The failed Rust job was unrelated to the vendored WebView2 crate.

`upstream/main` changed `apps/desktop/src/styles/tokens.css` without refreshing the committed docs export. The workspace test reported that exact drift.

Commit `85d36006f` regenerates only the docs export CSS and manifest. The targeted Rust regression test now passes.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in ead307dba, will be released in the next version.

PR #9626, opened 2026-09-19
fix(objects): double-click a view to open its data tab, fix source objects toggled off by double-click

yaoxinghuo:
## Problem

In the object browser (the main-area list view), double-clicking a **view**
does nothing visible, while double-clicking a **table** correctly opens its
data tab. Double-clicking the same view from the left sidebar works fine.
Reproduced on both PostgreSQL and Oracle — it is database-independent.

Two underlying issues:

1. **Inconsistent action mapping.** The sidebar treats `table`, `view` and
   `materialized_view` as data nodes (`open-data`), but the object browser's
   `doubleClickRowAction` maps `VIEW`/`MATERIALIZED_VIEW` to `open-source`
   (DDL side panel). There is no way to open a view's data from the object
   browser via double-click, unlike the sidebar.

2. **Double-click is cancelled out for "same-action" rows.** For rows whose
   single-click and double-click actions are identical (`VIEW`, `PROCEDURE`,
   `FUNCTION`, `TRIGGER`, `SEQUENCE`, `PACKAGE`, ... → `open-source`), the
   first click of a double-click gesture runs `openSource()` which opens the
   side panel; the second click (`event.detail === 2`) runs it again and hits
   the "click the same row to toggle off" guard in `openSource()`, so the
   just-opened panel is immediately closed — the net effect is a no-op.
   Tables are unaffected because their single (`table-info`) and double
   (`open-table`) actions differ and the single action is deferred/cancelled.
   This is a regression introduced by `5bcd24d` ("feat(objects): open database
   objects in side panel") which added the toggle behavior. On MongoDB the
   same mechanism also causes `open-table` to be emitted twice per
   double-click.

## Changes

- `lib/table/objectBrowserRowAction.ts`: `doubleClickRowAction` now maps
  `TABLE`/`VIEW`/`MATERIALIZED_VIEW` to `open-table`, matching the sidebar's
  `dataNodeTypes` so double-clicking a view opens its data tab. The existing
  `emit("openTable", { tableType: "VIEW" })` path (already used by the
  "View Data" context menu) handles it.
- `components/objects/ObjectBrowser.vue`: in the `isDouble` branch of
  `onRowClick`, skip executing the resolved action when it is identical to
  the row's single-click action — the gesture's first click already ran it.
  This keeps `open-source` rows (procedures, functions, ...) opening their
  source panel instead of toggling it back off, and avoids duplicate
  `open-table` emissions for MongoDB.
- Updated `objectBrowserRowAction.spec.ts` expectations and added a
  `MATERIALIZED_VIEW` case plus a "same-action rows are not deferred" case
  (PROCEDURE).

## Resulting behavior

| Interaction | Before | After |
|---|---|---|
| Single-click view | Source side panel | Unchanged (deferred 250 ms like tables) |
| Double-click view | Panel opens then closes (no-op) | **View data tab** (same as sidebar) |
| Double-click procedure/function/etc. | No-op | Source side panel opens |
| Double-click MongoDB table/view | `openTable` emitted twice | Emitted once |

## Testing

- `pnpm vitest run apps/desktop/src/lib/__tests__/table/objectBrowserRowAction.spec.ts` — 38/38 pass
- `pnpm typecheck` (vue-tsc) — clean
- `pnpm lint` (oxlint) — clean
- Manually verified in `tauri dev`: double-clicking a view in the object
  browser now opens its data tab on both Oracle and PostgreSQL connections;
  tables and source objects behave as described above.


t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 853a408d3, will be released in the next version.

PR #9633, opened 2026-09-19
feat(mongodb): support database dump and restore over the legacy agent

thailoc-dev:
## What

**Back up database** and **Restore database** now work on connections that use the `mongodb-legacy` Agent. Before, both menu items opened normally and then failed with:

> Database dump/restore requires the native MongoDB driver

That covers every server the native driver refuses (it requires MongoDB ≥ 4.2), which is exactly the population the Agent exists for.

Collection-level **BSON export** gains the same Agent support, since a database dump is built on it.

## How

`mongodb_dump.rs` was hard-wired to `mongodb::Client`. It now goes through a small `DumpClient` enum with a native arm and an Agent arm. The operations a dump/restore needs are all plain server commands, so the Agent arm runs them through the Agent's existing generic `runCommand` — **no Agent (Java) change and no new capability**:

| Operation | Native | Agent |
| --- | --- | --- |
| list collections with options | `list_collections()` | `runCommand({listCollections:1})`, drained via `getMore` |
| list indexes | `list_indexes()` | `runCommand({listIndexes:…})`, drained via `getMore` |
| create / drop / createIndexes / buildInfo | `run_command` | `runCommand` |
| insert restore batches | `insert_bson_documents` | `insertMany` (canonical Extended JSON, the same path collection BSON import already uses) |
| read documents for the dump | `for_each_find_document` | the Agent's find cursor, which streams canonical Extended JSON |

Type fidelity: canonical Extended JSON is lossless, so `Int32`/`Int64`, `Date`, `Binary`, `ObjectId` etc. survive both directions. The Agent's older page-based fallback returns *relaxed* JSON (Int32/Int64 collapse), so BSON export over an Agent without find-cursor support is refused with an "upgrade the Agent" message rather than silently producing a lossy dump. The dump/restore entry point requires `runCommand` + `insertMany` capabilities the same way.

The native path is unchanged in behaviour: each native arm is the code that was there before, just moved.

## Testing

- New opt-in integration test `legacy_agent_database_round_trip_through_dbx` (`DBX_MONGO_LEGACY_DUMP_TEST_HOST`). Everything in it — seeding, dump, restore, verification — goes through the Agent, because the native driver cannot talk to the server it targets. Run against a throwaway `mongo:3.6` container with the stock installed Agent jar: directory/archive × plain/gzip all pass — 6 collections (validator + collation, capped, empty, `odd/name` with Binary and `i64::MAX`, two chained views), documents byte-identical, options and indexes identical, selective `dropExisting` restore, duplicate-`_id` reporting.
- The Agent-made directory dump was also restored with the official `mongorestore` and inspected in `mongo`: `NumberLong`, `ISODate`, capped/validator/collation options, unique and TTL indexes, and both views come back correctly.
- `cargo test -p dbx-core --test mongodb_dump_restore` (6 non-ignored), `--lib mongodb_dump` (8), `--lib mongodb_import_export` (53) pass. `cargo fmt`/`clippy` clean for the touched files.
- Docs: the "native driver only" sentence in `mongodb.mdx` / `mongodb.cn.mdx` updated.


t8y2 (approved):
Reviewed in detail — this is a clean piece of work. The `DumpClient` enum keeps the native arms as moved-unchanged code, the agent arms reuse the canonical Extended JSON helpers already proven by collection-level BSON import, and the agent's `runCommand` surfaces ok:0 failures (no silent create/index misses) while `fetchFindCursor` streams canonical (lossless) Extended JSON — refusing the relaxed page fallback for BSON export is the right call. The `getMore` namespace derivation for `listCollections`/`listIndexes` cursors matches the driver convention, and the new opt-in round-trip test coverage is thorough.

Two optional nits, neither blocking: (1) `mongodb_dump.rs:266` — `cursor.get_i64("id").unwrap_or(0)` silently ends iteration on an unparseable id; erroring would be strictly safer than silent truncation. (2) The dump entry gate checks `MongoRunCommand`+`MongoInsertDocuments` but not `MongoFindCursor`, so an agent with only the first two starts a dump and fails later with the "BSON export … needs find-cursor" message — still actionable, just late. Feel free to address in a follow-up.

t8y2:
Thanks for the contribution! Merged in a67bbb4ed5, will be released in the next version.

PR #9636, opened 2026-09-20
ci(win7): enable safe Rust compile caching

luojiyin1987:
## Summary

- Enable Rust sccache for `windows-win7-bundle`.
- Enable the same cache for `build-windows-7-offline`.
- Use `win7-webview2-1.0.902.49-v1` as the cache namespace.
- Disable the sccache idle timeout during long final links.

## Safety

The WebView2 loader is now a tracked Cargo input. Cargo invalidates cached compilation when the loader changes.

This PR does not wrap `CC` or `CXX`. OpenSSL and NMAKE keep their existing compiler path.

This PR does not change LTO, codegen units, crate types, linker settings, job structure, or release profiles.

## Measurement

The cold run used the new cache namespace. The warm run included later project changes.

| Metric | Cold run | Warm run |
| --- | ---: | ---: |
| `Build DBX for Windows 7` | 74m52s | 53m11s |
| Cache hits | 0 | 759 |
| Cache misses | 844 | 85 |
| Cache hit rate | 0% | 89.93% |
| Cache write errors | 127 | 0 |

The warm build saved **21m41s**, or about **29%** of the build step.

The final LTO and MSVC link remain outside the sccache scope. These stages limit further savings.

- Cold run: https://github.com/t8y2/dbx/actions/runs/35479186470/job/105993794701
- Warm run: https://github.com/t8y2/dbx/actions/runs/35486034601/job/106012715971

## Validation

- 11 workflow structure tests
- 17 Windows installer tests
- `actionlint` with existing shellcheck findings ignored
- formatter check
- `git diff --check`


luojiyin1987:
The failed `packages` job is unrelated to Win7 sccache. Its package scripts call `cargo nextest`, but the job does not install that command.

The focused fix is #9637. After it merges, I will update this branch and rerun the cold measurement.

t8y2 (commented):
Maintainer patch: verified the vendored-loader precondition for re-enabling sccache on the Win7 jobs holds on current main (`[patch.crates-io] webview2-com-sys` + verify-only loader script + build.rs SHA hard-fails + loader version baked into the cache namespace), so this PR is the planned follow-up rather than a reintroduction of the known-bad pattern. Toolchain and CC/CXX wrapping are untouched; cache backend is GH Actions cache with graceful degradation.

Patch adds the missing observability the PR body's Measurement section promises: a `sccache --show-stats` step (`if: always()` + `continue-on-error: true`, same as `windows-standard-check`) to both `windows-win7-bundle` (ci.yml) and `build-windows-7-offline` (release.yml), so warm-run hit rates are collectable from CI logs. No other changes.

Note: the `packages`/`ci` failures on the previous head were the pre-existing missing-nextest issue, fixed separately by #9637.

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in 01bd07bb9, will be released in the next version.

PR #9637, opened 2026-09-20
fix(ci): install nextest for package tests

luojiyin1987:
## Summary

- Install pinned `cargo-nextest` in the `packages` job.
- Add `packages` to the workflow test for nextest consumers.

## Cause

The package scripts now execute `cargo nextest run`. The `packages` job did not install that command. It failed before compiling any package.

This failure also appears in #9636, but it is unrelated to Win7 sccache. This separate PR keeps both changes focused.

## Validation

- 11 workflow structure tests
- `actionlint`
- `git diff --check`


t8y2 (approved):
Verified the root cause on main: `8b757b97b` switched `@dbx-app/cli`'s test to `cargo nextest run` but the `packages` job never installed nextest. The pinned `taiki-e/install-action` step matches the convention used by `rust-test`/`agent-rust`/`agent-integration` exactly, and the `packages` job ran green at this head with nextest installing in under a second. Structure tests pass (11/11).

One follow-up worth noting (not blocking): `mcp-release.yml`'s `prepare` job also runs `pnpm test:packages` without nextest installed, so the next manually-dispatched package release will hit the same failure. A follow-up PR adding the same install step there would close that out.

t8y2:
Thanks for the contribution! Merged in db5e6f106b, will be released in the next version.

PR #9695, opened 2026-09-20
feat(ai): show rotating status ring on scroll-to-bottom button while generating

Abeautifulsnow:
### What

The AI panel's scroll-to-bottom button now doubles as an AI busy indicator:
while a run is active for the visible conversation (`isGenerating`), the button
stays visible (even when the view is already pinned to the bottom) and renders
a decorative clockwise-rotating arc ring around the arrow, in the style of
reference apps like Si-Octo. Clicking still force-scrolls to the bottom.

### Why

Previously the button only appeared when the user scrolled away from the
bottom, so during a long generation there was no persistent cue that the
assistant was still working once auto-scroll kept the view at the bottom.

### Changes

- `AiAssistant.vue` — button visibility is now `showScrollToBottom || isGenerating`
  (the scroll hysteresis state machine and all `isGenerating` lifecycle paths are
  untouched); added an `aria-hidden` SVG arc ring (`data-ai-scroll-ring`,
  ~288° open arc, `animate-spin` + `motion-reduce:animate-none`,
  `text-primary`/`currentColor` so all themes come free)
- `AiAssistant.scrollRing.spec.ts` (new) — source-assertion tests pinning the
  wiring: busy-visible OR-condition, ring gating/animation/reduced-motion/
  a11y attributes, `pointer-events-none` and `fill="none"`, force-scroll click,
  and the 120/48px hysteresis thresholds pinned by exact declaration text

### Testing

- `pnpm vitest run` scoped to all `AiAssistant*.spec.ts` + `packages/app-tests/ai*`:
  26 files / 214 tests passed
- `pnpm lint` (0 new warnings in changed files), `pnpm typecheck` (exit 0),
  `oxfmt --check` clean

### Notes

- No new i18n keys (the ring is decorative/`aria-hidden`; the button reuses the
  existing `ai.scrollToBottom` copy present in all 10 locales)
- Known edge: for a brand-new empty conversation, the ring is not shown during
  the brief window before the first message is pushed (the whole message area,
  including the button, renders only when `messages.length > 0`); bounded and
  queued runs are unaffected


t8y2 (approved):
Thanks for the clean, well-scoped patch — approved.

What I verified:
- The OR-condition (`showScrollToBottom || isGenerating`) is purely additive at the template level; the scroll hysteresis state machine (120/48px thresholds, auto-scroll pause/resume) and every `isGenerating` lifecycle path are untouched, so all terminal states converge correctly (generation ends at bottom → button unmounts; ends while scrolled up → plain scroll affordance remains).
- The ring is a pure CSS animation (`animate-spin`, transform-only) with `motion-reduce:animate-none`, torn down automatically by `v-if` — no timers or listeners to leak, and it lives in the AI panel rather than on any editor/high-frequency repaint path, so no perf concerns.
- `aria-hidden` + `pointer-events-none` + `currentColor` all check out; the arc geometry matches the existing repo spinner idiom, and the source-assertion spec follows the established house style for this suite.

One non-blocking observation: the ring is gated on `isGenerating` only. With background runs enabled, a visible-conversation run in the `queued`/`preparing` phases shows no busy cue on the button even though the composer already switches to queue-send (`hasActiveRunForCurrentConversation`). That's a defensible scoping choice (and matches the generation-status line's gating), so feel free to leave as-is — just flagging it in case you want the two busy indicators to agree. The disclosed empty-conversation edge (ring absent before the first message push) is fine — bounded and documented.

t8y2:
Thanks for the contribution! Merged in 89d29b28bc, will be released in the next version.

PR #9728, opened 2026-09-20
fix(test): decouple partner adapter test from model version

luojiyin1987:
## Summary

Remove the fixed Jalapeno model version from the runtime adapter test.

The test now verifies two stable contracts:

- The preset has a default model.
- The supported model list contains that default model.

A separate preset test verifies the current partner model values. Model version updates no longer break this adapter contract test.

## Validation

- `pnpm exec vitest run packages/app-tests/settingsStore.test.ts`
- 75 tests passed
- `pnpm exec oxfmt --check packages/app-tests/settingsStore.test.ts`
- `git diff --check`


t8y2 (approved):
Thanks — verified the failure reproduces on current `main` (`packages/app-tests/settingsStore.test.ts:1014`, GLM-5.2 vs GLM-5.3 after eebb08227 missed this file), and the follow-up commit making the assertion invariant-based (default model must be non-empty and present in `models`) is the better fix — it keeps the adapter/preset wiring checks while staying robust to future default-model updates. LGTM.

t8y2:
Thanks for the contribution! Merged in 9ff574923, will be released in the next version.

PR #9763, opened 2026-09-20
fix(plugins): check and update installed plugins with provenance-guarded sources (#9641)

### LINKED ISSUE #9641: [Bug] 已安装的插件不支持新版本检测和更新

不涉及

Image

实际插件商店支持，可以考虑两个入口都可以：

Image

最近的版本

_No response_

Abeautifulsnow:
Fixes t8y2/dbx#9641

## Problem

The Plugin Store already shows when an installed plugin has a newer version and offers
one-click update — but the **Installed** tab of the Plugin Center offers no update
indication and no update action at all (only Rollback / Uninstall), even though it is the
primary place users manage installed plugins.

## Root cause

The update signal only existed inside the marketplace listing join
(`buildMarketplacePluginListings`): the installed tab's data model
(`InstalledPlugin` / `FrontendPluginDefinition`) carries no `latestVersion`, and the tab
never consumed the `marketplaceListings` computed that the same component already holds.
No backend work was required for the fix itself — updating is a normal marketplace
install of a newer version, and the installer already supports versioned directories,
atomic swap, one-step rollback, signature verification and the connection/operation
lifecycle guard.

### Frontend (installed tab)

- **Update affordances in both entries** (the issue's request): amber per-row `↑ v{latest}`
  pill, a primary "Update to v{version}" button in the detail pane (reusing the exact
  marketplace install path), an amber "N plugin updates available" banner with
  **Update all** (sequential batch with `(current/total)` progress), and an amber count
  pill on the **Installed** tab trigger.
- **Honest states** — an incomplete check can never read as "up to date":
  - total catalog fetch failure → "cannot check" + retry;
  - **partial failure** (some enabled repositories error while others succeed) → a
    partial-failure note plus per-repository error rows; the "all up to date" note and
    the "not in repositories" badge are suppressed in this case;
  - no enabled repositories; plugin absent from every catalog (local `.dbxp` install);
    catalog entry with no artifact for the current target (`unsupportedTarget` badge);
  - "checking…" state while the catalog loads.
- **Source-change confirmation**: when the recorded install provenance differs from the
  offering repository / publisher / signing key, an update requires an explicit
  confirmation dialog (compare current vs candidate) and retries with
  `allowSourceChange: true`. The store-tab batch and the automatic updater skip such
  listings with a report instead of failing them one by one.
- Update success now dispatches `notifyComponentPluginsUpdated()` (App-level badge
  refreshes immediately) and clears the plugin icon cache (previously only
  rollback/uninstall did, leaving stale icons after an update).
- Fixed a real comparator bug while touching this code: `compareVersions` ranked
  releases *below* their own prereleases (`1.0.0 < 1.0.0-beta`) because it
  `localeCompare`d the whole prerelease string; replaced with identifier-wise semver
  prerelease ordering, keeping the numeric-aware fallback for legacy non-semver
  versions.
- i18n: 24 new `pluginPlatform` keys synced across the 8 locales that define the
  namespace (az/tr intentionally fall back to en, matching current state).

### Backend (Rust, update-source hardening)

- `PluginActivationRecord` now persists a `PluginInstallProvenance`
  (`repositoryId` / `publisher` / `signingKeyId` / `source`) with every activation
  (`#[serde(default)]`, backward compatible; records without it stay unconstrained
  until the next install records fresh provenance).
- New `ensure_update_continuity` guard: a marketplace update whose repository,
  publisher, or signing key differs from the recorded provenance — or that would
  downgrade below the active version — is rejected unless
  `PluginMarketplaceInstallRequest.allowSourceChange` is set. It runs as a **read-only
  pre-flight before any download** (no sidecar teardown on rejection) and again
  authoritatively under the store lock.
- Rollback carries the target version's own provenance forward; file/URL installs carry
  the recorded `repositoryId` forward (a non-marketplace overwrite cannot silently erase
  a marketplace install's source identity).
- `InstalledPluginInfo.provenance` is exposed to the frontend (additive, optional).

## Deliberate decisions

- **Multi-repository dedup**: when several enabled repositories offer the same plugin
  id, the installed tab resolves one source deterministically — a repository allowed to
  verify the plugin (official/enterprise) first, then the higher version — shared by
  store cards, installed tab, and batch paths.
- **File/URL installs bypass the continuity guard** by design: they are explicit user
  actions with no confirmation channel, and guarding them would hard-block legitimate
  reinstall flows. Provenance erasure is still prevented (carry-forward + actual
  publisher/key recorded).
- Update badges use amber (the app's "update" color in the driver store banner/pills)
  rather than red.

## Testing

- Frontend: plugin lib/components/integration suites (253), i18n parity (455),
  `useComponentUpdates` (32) — including new tests for the installed update index,
  multi-repo dedup, source-change confirm-then-retry (`allowSourceChange: true` on the
  wire), honest states, and the partial-failure regression.
- Rust: `cargo test -p dbx-plugin-runtime` (118) — including source-change
  reject/confirm, rollback provenance carry-forward, pre-provenance records staying
  unconstrained, and `plugin-data/` surviving update + rollback byte-identical.
- `pnpm typecheck`, `cargo fmt --check`, `cargo check --workspace --tests` clean.

## Checklist

- [x] Both entries (store tab and installed tab) can detect and apply plugin updates
- [x] Updates reuse the marketplace install path (signature/checksum verification,
      atomic swap, one-step rollback); plugin data dirs are untouched
- [x] Every "cannot check" case is visible instead of silently reading as up to date
- [x] New user-facing strings translated in all 8 `pluginPlatform` locales

## Screenshots
f41b343f-99df-4f99-9bee-09863713da8b


t8y2 (commented):
维护者代为解决了与 main(4d150fc90,同步更新中心状态)的冲突并做了语义去重。

- 旧 head `48b507c2a0` → 新 head `291e4e91c9`,基线 main `f5d5ae3a5`(rebase,第二个提交逐字节等价,仅第一个提交的 App.vue 部分被去重改写)。
- 冲突原因:两边在同一批安装/更新路径上各自加了平行的通知机制——你的 `COMPONENT_PLUGINS_UPDATED_EVENT` App 级徽章监听器 vs main 在每个变更路径上的 `notifyComponentUpdatesChanged()`(强制刷新+工具栏同步)。
- 去重方式(保留双方意图):面板保留你的 `notifyComponentPluginsUpdated()` 调用(面板自身的既有监听器做图标缓存+已装列表刷新,`installMarketplaceListing` 里不再需要 main 的直连 `listPlugins()` 二次拉取);删除 App.vue 里冗余的徽章监听器;`installMarketplaceListing` 与 `runUpdateAllInstalled` 的成功路径补上 `notifyComponentUpdatesChanged()`,让工具栏/更新中心徽章走 main 的机制。这样每次安装/更新只触发一次 `componentUpdates.refresh()`,两个表面都刷新。
- 依赖标志无变化(无新增依赖)。
- 已跑检查:改动文件 oxlint 0 告警、oxfmt 全部通过;`UpdateDialog.spec`/`PluginContributionsPanelBatch.spec`/`useComponentUpdates.spec`/`backendErrors.spec`/`pluginMarketplace.spec` 共 602 用例全过;range-diff 审计确认除 App.vue 去重外无行为差异。等新 head 的 CI 绿了即可合并。

t8y2 (approved):

t8y2:
Thanks for the contribution! Merged in cab674299, will be released in the next version.

PR #9785, opened 2026-09-20
fix(mongodb): fail early on unreadable cursor ids and outdated agents in dump/restore

thailoc-dev:
## What

Follow-up to #9633, addressing both review nits.

1. **Unreadable cursor id is an error, not "done".** `command_cursor_id` now accepts only `Int64`/`Int32` and returns an error for anything else (missing, string, double). Previously `get_i64("id").unwrap_or(0)` would have read a malformed id as "cursor exhausted" and ended the `listCollections`/`listIndexes` drain early — a dump silently missing collections or indexes.
2. **Outdated agents are refused before any work starts.** The dump/restore entry gate now also requires `MongoFindCursor`, alongside `MongoRunCommand` and `MongoInsertDocuments`. An agent without it used to get past the gate and fail at the first collection with the BSON-export message; it now gets the same "upgrade or reinstall the MongoDB Legacy driver" message up front.

## Testing

- New unit test `command_cursor_id_accepts_integers_and_refuses_anything_else`.
- `cargo test -p dbx-core --lib mongodb_dump`: 9 passed.
- Live: `legacy_agent_database_round_trip_through_dbx` against a throwaway `mongo:3.6` with the stock installed agent still passes (directory/archive × plain/gzip).
- `cargo fmt` / `clippy` clean for the touched file.


t8y2 (commented):
Maintainer patch (`924ad2daf` → `b0b0e2f8`), implementing both review notes:

1. **Find-cursor gate moved to the dump path only.** `dump_client` is shared by dump, restore, and the catalog preview, but only the dump's BSON export reads through a find cursor. The `MongoFindCursor` requirement now lives in a new `DumpClient::require_find_cursor()` called at the top of `dump_mongodb_database`; restore and preview are gated on `mongo_run_command` + `mongo_insert_documents` only, so agents from the v0.6.12–v0.6.14 capability window regain working restore. The deep fallback in `mongodb_import_export.rs` stays as is.
2. **`command_cursor_id` diagnostics**: a missing id now reports `Cursor id missing from command response` instead of rendering `None` through the catch-all arm.

Checks run: `cargo check -p dbx-core --no-default-features` clean, `cargo fmt -p dbx-core` no changes, `git diff --check` clean. (The inline `mongodb_dump` nextest run hit a local disk-full error, not a test failure — the suite is covered by CI on this head.)

t8y2:
Thanks for the contribution! Merged in 9b82ce86f9, will be released in the next version.

PR #9838, opened 2026-09-21
feat(mongodb): support findOneAndUpdate, findOneAndReplace and findOneAndDelete over the legacy agent

thailoc-dev:
## What

`db.coll.findOneAndUpdate()`, `findOneAndReplace()` and `findOneAndDelete()` now run on connections that use the `mongodb-legacy` Agent. The query editor already parsed all three; on the Agent they failed at execution with `MongoDB legacy agent does not support findOneAndUpdate`.

## How

All three are the `findAndModify` server command, which the Agent's existing generic `runCommand` already reaches — so **no Agent (Java) change and no new capability**.

- `mongo_driver.rs` gains three pure builders, `find_one_and_update_command` / `find_one_and_replace_command` / `find_one_and_delete_command`, that produce the command document from the same inputs and option structs the native helpers use (`returnDocument`/`new` → `new`, `projection` → `fields`, `sort`, `upsert`, `arrayFilters`, pipeline updates as an array). Option validation is shared, so an unknown option is rejected identically on both drivers.
- `find_one_and_replace_command` refuses a replacement containing update operators, as the native path does — the server would otherwise silently run it as an update.
- `find_and_modify_response_result` maps the reply's `value` (document or `null`) to the same zero-or-one result the native helpers return.
- `mongo_ops.rs`: the three `PoolKind::Agent` arms call a small `agent_run_command_document` helper (capability-checked) instead of returning the "not supported" error.

Known limitation, same as every other `runCommand` reply on the Agent today: the returned document comes back as relaxed Extended JSON, so `NumberLong(15)` displays as `15` in the result grid. The database itself is unaffected.

## Testing

- Unit: `find_and_modify_commands_mirror_the_native_options` (every option, defaults, pipeline update, operator refusal, delete shape) and `find_and_modify_response_maps_value_to_zero_or_one_document`.
- Live, opt-in: new `live_mongodb_legacy_agent_find_and_modify.rs` drives real shell text through `mongo_shell::parse` → `execute_mongo_command_core` — the query tab's path — against a throwaway `mongo:3.6` with the stock installed Agent. Covers before/after return, sort, projection, upsert, no-match, `arrayFilters`, replace (and its refusal), delete with sort, and that a server-side error (`The path 'tags' must exist…`) surfaces with the server's own message.
- `cargo test -p dbx-drivers --lib mongo_driver` (119), `-p dbx-core --lib mongo_ops` (30) pass; `fmt`/`clippy` clean for the touched files.


t8y2 (commented):
Maintainer patch pushed in 1f27674aa: the native `find_one_and_replace` now refuses replacement documents containing update operators via the same `reject_update_operator_replacement` helper the new agent command builder uses, so both drivers behave identically (the official MongoDB shell also refuses this client-side). Previously the native path sent the operator document as `update`, letting the server silently run it as an update — diverging from the agent path this PR adds and from the PR description's parity claim, which now holds.

Checks run locally: find_and_modify unit tests 5/5 (`--no-default-features --features sqlite-sqlcipher`), `cargo fmt --check` clean.

t8y2:
Thanks for the contribution! Merged in 98d3209026, will be released in the next version.

PR #9884, opened 2026-09-21
feat(mongodb): support distinct over the legacy agent

thailoc-dev:
## What

`db.coll.distinct(field, filter)` now runs on connections that use the `mongodb-legacy` Agent. The query editor already parsed it; on the Agent it failed with `MongoDB legacy agent does not support distinct`.

## How

Same approach as #9838: `distinct` is a server command (`{ distinct, key, query }`) that the Agent's existing generic `runCommand` already reaches — **no Agent (Java) change and no new capability**. The comment on the old error arm ("no read that could stand in for it") predates `runCommand` on the Agent.

- `mongo_driver.rs`: the native `distinct` helper's field/filter validation and result shaping are factored into `distinct_filter_document` / `distinct_values_result`, shared by the new `distinct_command` builder and `distinct_response_result` (maps the reply's `values` array to the same `MongoDocumentResult`). The native path's behaviour is unchanged.
- `mongo_ops.rs`: the `PoolKind::Agent` arm uses the `agent_run_command_document` helper introduced in #9838.
- The live test file from #9838 is renamed to `live_mongodb_legacy_agent_commands.rs` since it now covers more than `findAndModify`.

## Testing

- Unit: `distinct_command_and_response_mirror_the_native_helper` (command shape, empty filter, field/filter validation, values mapping incl. `null`).
- Live, opt-in: `distinct_runs_over_the_legacy_agent` drives shell text through `mongo_shell::parse` → `execute_mongo_command_core` against a throwaway `mongo:3.6` with the stock installed Agent: plain field, array field flattened with a filter, dotted nested path, and no-match → empty result. Both live tests in the file pass.
- `cargo test -p dbx-core --lib mongo_ops` (30) and the driver `distinct` tests (3) pass; `fmt`/`clippy` clean for the touched files.


t8y2 (approved):
Thanks for this — it's a clean follow-up to #9838 and follows the established pattern exactly: `distinct` goes through the shared, capability-gated `agent_run_command_document`, and the factored helpers guarantee native/agent parity for filter parsing, result shape, empty results and the values-array error path. The live test rename and new opt-in cases (array flattening, dotted paths, no-match) are nice touches. Two entirely optional nits for whenever you touch this next: `distinct_response_result` could take the response by value and move the array out instead of cloning, and the field-required check is now duplicated between the two layers. LGTM.

t8y2:
Thanks for the contribution! Merged in 87d1e114c4, will be released in the next version.

PR #9893, opened 2026-09-21
feat(mongodb): support collection stats and create database over the legacy agent

thailoc-dev:
## What

Two more sidebar features now work on connections that use the `mongodb-legacy` Agent:

- **Collection statistics** (document count, sizes, index count) — previously `MongoDB legacy agent does not support collection stats helpers`.
- **Create database** — previously refused by the backend *and* hidden by the sidebar for legacy connections. Both are lifted.

## How

Same approach as #9838 / #9884: both are server commands the Agent's generic `runCommand` already reaches — **no Agent (Java) change and no new capability**.

- `collStats`: the native helper already ran it as a command; its builder is now `pub` as `collection_stats_command` (with the collection-name check moved into it) and reused by the Agent arm together with `collection_stats_result_from_document`. The per-database object-statistics path that used the same builder is updated.
- Create database: MongoDB has no such command — a database exists once it holds a collection — so the native path creates a placeholder `dbx_init` collection. The name is now a shared constant and the Agent arm creates the same collection via `{ create: "dbx_init" }`; the trim/required check is shared through `validate_create_database_name`.
- Frontend: `connectionNamespaceCreationTarget` no longer returns `null` for legacy MongoDB connections, so **Create database** appears in the menu; the read-only check still applies. The existing `databaseFeatureSupport` expectations are updated accordingly.
- Also folds in the two optional nits from #9884: `distinct_response_result` takes the response by value (no clone), and the field-required check lives in one place (`distinct_inputs`) used by both the native helper and the command builder.

## Testing

- Unit: `collection_stats_command_serializes_scale` extended (no scale → no field; blank name refused), new `create_database_name_is_trimmed_and_required`; `distinct` tests updated for the by-value signature. Driver `mongo_driver` suite: 125 passed; `mongo_ops`: 30 passed.
- Live, opt-in: new `collection_stats_and_create_database_run_over_the_legacy_agent` in `live_mongodb_legacy_agent_commands.rs` against a throwaway `mongo:3.6` with the stock installed Agent — count/nindexes/size with and without `scale`, error for a missing collection, created database visible in `listDatabases` with its `dbx_init` collection, blank name refused. All three live tests in the file pass.
- Frontend: `lib/__tests__/database` + `mongoCapabilities` specs: 518 passed; `vue-tsc`, `oxfmt`, `oxlint` show no new findings.

Not in this PR: the sidebar still hides **Rename collection** on legacy connections (`usesNativeMongoDriver` gate) even though the Agent gained `renameCollection` in #9496 — a one-line follow-up.


t8y2 (approved):
Thanks for this — reviewed and it looks good. Both features follow the established legacy-agent pattern exactly: capability-gated runCommand with shared builders, native/agent parity on scale handling, view behavior, and the placeholder-collection semantics, with unit coverage plus an opt-in live test. CI green at the head. For the record: this covers the db.<collection>.stats() shell path, not the database view's object-statistics columns — that gap remains open but your now-public helpers make the follow-up straightforward. The .ok()? on the new builder validation is an optional nit. LGTM.

t8y2:
Thanks for the contribution! Merged in c30f6821d3, will be released in the next version.

PR #9899, opened 2026-09-21
fix(dameng): paste/copy table with data fails on identity columns (#9714)

### LINKED ISSUE #9714: [Bug] 达梦数据库粘贴表报错了。

通用

Image
注Navacat一次搞定达梦的管理工具SQLark百灵连接和DBX报错一样。

v0.6.17

克隆为新表，和粘贴表功能重叠感觉点击克隆为新表也可以使用粘贴表的功能，现在想克隆一个带数据的表操作太麻烦了
1、搜索找到这个表。
2、点击复制表。
3、取消搜索
4、粘贴表
5、点击确认。

如果克隆为新表和粘贴表一样只需要点一次确认。

Navacat 只需要点复制结构和数据没弹出窗口直接就成功了

yiqiui:
## Summary
Closes #9714.

Pasting/copying a Dameng table **with data** fails whenever the table has an IDENTITY (自增) column. The generated copy SQL was `INSERT INTO target (cols) SELECT cols FROM source` with no identity handling, and DM rejects assigning to an identity column unless a column list is given **and** `SET IDENTITY_INSERT` is ON:

> dm.jdbc.driver.DMException: 仅当指定列列表，且SET IDENTITY_INSERT为ON时，才能对自增列赋值

`build_copy_table_data_sql` already wraps **SQL Server** copies with `SET IDENTITY_INSERT <t> ON/OFF`. Dameng uses the same syntax but wasn't wired, so the paste errored while Navicat / SQLark succeed.

## Behavior
- `crates/dbx-sql/src/db_admin_sql.rs`: added a `dameng_identity_insert` option (mirroring `sqlserver_identity_insert`) and generalized the wrap so Dameng copies with identity columns are emitted as `SET IDENTITY_INSERT <t> ON;\nINSERT …;\nSET IDENTITY_INSERT <t> OFF;`.
- `apps/desktop/src/lib/table/tableClipboard.ts` + `dbAdminSql.ts`: set `damengIdentityInsert` when the DB is Dameng and the writable columns include an identity column; it flows through the existing `...dataCopyColumnOptions` spread at both call sites (ObjectBrowser + SidebarTreeRuntimeHost).
- Unaffected: tables without identity columns (no wrap), SQL Server / Postgres behavior unchanged, non-Dameng types unchanged.

## Test plan
- [x] `cargo test -p dbx-sql --lib db_admin_sql` → 41 passed (added a Dameng identity-copy regression test).
- [x] Verified against a **live DM8 instance**: created a source+target table with `ID INT IDENTITY(1,1)`, seeded identity values.
  - OLD form (`INSERT … (ID,NAME) SELECT …` without IDENTITY_INSERT) → reproduces the exact reported error.
  - NEW form (`SET IDENTITY_INSERT … ON; INSERT …; OFF`) → succeeds, 3 rows copied, identity values preserved (max id 103).
- [ ] Frontend `typecheck`/vitest (change is additive: one new optional field + one new returned property matching the updated interface).

Note: I could not exercise the full packaged desktop app end-to-end (build/run of the Tauri app), but the fix is at the SQL-generation layer and is proven against the real database with the exact statement the app now emits.


t8y2 (commented):
Maintainer patch pushed in cebef1868, closing both CI blockers and adding the missing coverage: the six `tableClipboard.test.ts` expected literals now include `damengIdentityInsert: false`, a new `dameng` case asserts `damengIdentityInsert: true` with the same writable-column projection as SQL Server (the new branch previously had zero frontend coverage), and `cargo fmt` fixes the indentation violations in `db_admin_sql.rs` (the struct-literal keys and the over-long condition/asserts).

Checks run locally: `tableClipboard.test.ts` 6/6, `builds_copy_table_data_sql` pass, `cargo fmt --check` clean.

t8y2:
Thanks for the contribution! Merged in 0c6fc610cd, will be released in the next version.

PR #9907, opened 2026-09-21
fix(editor): respect MyBatis placeholders in MySQL highlighting

### LINKED ISSUE #9878: [Bug] # 占位符渲染问题

Mysql

Image
渲染成注释样式了

DBX 版本: v0.6.18
运行环境: 桌面端
操作系统: Windows 10.0.26200.9457
系统架构: x86_64
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/153.0.0.0 Safari/537.36 Edg/153.0.0.0
数据库类型: mysql, nacos, plugin, redis, sqlserver
本地驱动版本: rocketmq 0.1.13

_No response_

tdog54646:
## Changes

- Recognize enabled MyBatis `#{name}` placeholders in MySQL-family editor parsing, semantic highlighting and completion, including Chinese and astral names.
- Normalize only the reference-analysis input so placeholders do not produce false diagnostics on following clauses. The editor document and SQL sent for execution retain their original text.
- Reconfigure parsing when parameter syntax settings change, preserve real comments and literals, and cache backward scans across chunks for long SQL input.

Component screenshot (production QueryEditor mounted locally without a database connection):

MyBatis placeholders and real MySQL comments

## Why

Fixes #9878. MySQL's hash-comment parsing currently consumes enabled MyBatis placeholders and the SQL that follows them, suppressing highlighting/completion and causing misleading LIMIT diagnostics.

## Validation

- Seven relevant Vitest suites: **399 tests passed**, including mounted QueryEditor settings toggles, parser/semantic/completion behavior, Unicode chunk boundaries and long-input read budgets. New regressions were verified failing before their fixes.
- A temporary probe linked against the production `dbx-sql` library: **3 tests passed**, covering the original LIMIT-clause error, normalized analysis and preserved Unicode diagnostic offsets.
- `pnpm exec vue-tsc --noEmit --project apps/desktop/tsconfig.json` passed.
- Formatting and lint checks on changed frontend files passed; one pre-existing lint warning remains in an unchanged section of `sqlParameters.ts`.
- Edge UI verification confirmed enabled/disabled MyBatis behavior and preservation of real hash comments.
- Full repository `make check` and a live MySQL connection test were not run.


t8y2 (changes requested):
Thanks for the thorough work — the input-masking approach via `configureLanguage({ wrap })` is right, and the offset-preserving `' inner'` normalization for reference analysis is nicely done. Verified edge handling (`#{p(jdbcType=...)}` rejected consistently with the substitution system, plain `#` comments / `@var` / T-SQL `#temp` untouched, execution path unmodified).

One small fix blocks merge: CI fails on `apps/desktop/src/lib/__tests__/editor/queryEditorSqlSignature.spec.ts:11`. That spec matches the exact old watcher string `watch([() => props.databaseType, () => props.dialect, () => props.syntaxDialect, sqlDriverProfile]`, which no longer matches after the new `enabledSyntaxes.join(",")` watch source was appended to `QueryEditor.vue`. Please update the expected substring there (or match a stable prefix) so `frontend-test`/`frontend`/`ci` go green.

One scope note worth a sentence in the PR body: the new `#{`/`${` branch in `tokenizeSqlSemantic` (tokens.ts:132) applies to every dialect and defaults to enabled for callers that don't pass options, so features like copy-without-comments, DDL display, and insert-value hints now see `#{name}` as a parameter token rather than a comment. That looks consistent with the substitution system's defaults and the suite passes — just state the wider tokenizer effect explicitly since the title says "MySQL highlight".

PR #9924, opened 2026-09-21
fix(mongodb): offer rename collection on legacy agent connections

thailoc-dev:
## What

**Rename collection** now appears in the sidebar context menu on connections that use the `mongodb-legacy` Agent. The Agent has supported `renameCollection` since #9496 (capability-gated in the backend), but the menu was still hiding the action behind a "native driver only" check from before that.

## How

- `useSidebarDatabaseSpecificMutationRuntime.ts`: `canRenameMongoCollectionNode` drops the `usesNativeMongoDriver` condition. Read-only connections are still excluded through `supportsMongoAllDriverMutations`, and an Agent too old for `renameCollection` gets the backend's existing "upgrade or reinstall the MongoDB Legacy driver" error rather than a hidden menu item.
- Rename was the last user of `supportsNativeMongoDriverMutations`, so that helper and its tests are removed instead of left as dead code.

## Testing

- New spec case: rename is offered for `undefined`, `mongodb-native` and `mongodb-legacy` profiles, and still not on a read-only connection. Fails on `main` for the legacy profile.
- `useSidebarDatabaseSpecificMutationRuntime.rename`, `mongoCapabilities`, `tabMenu` specs: 20 passed. `vue-tsc`, `oxfmt`, `oxlint` show no new findings.


t8y2 (approved):
Thanks for the follow-up! Verified that this is exactly the missing piece after #9496: the backend already routes legacy-agent renames through the `mongo_rename_collection` capability gate (`mongo_ops.rs`) with a clear upgrade error, and the agent handler mirrors the native driver's validation and namespace semantics (same-db rename, no dropTarget). Read-only exclusion is preserved via `supportsMongoAllDriverMutations`, and removing `supportsNativeMongoDriverMutations` is clean — rename was its last caller. Frontend/typecheck/test CI all green. Merging as-is.

t8y2:
Thanks for the contribution! Merged in 808317e0e7, will be released in the next version.

PR #9925, opened 2026-09-21
feat(mongodb): fill database view object statistics over the legacy agent

thailoc-dev:
## What

The database view's per-collection **rows** / **size** columns now fill in on connections that use the `mongodb-legacy` Agent. Previously the Agent pool fell through to the empty default in `list_object_statistics_once`, so those columns stayed blank. This is the gap noted in the #9893 review.

## How

`mongo_agent_list_object_statistics` (in `mongo_ops.rs`) mirrors `mongo_driver::list_object_statistics`: list the collections through the Agent, skip views, and run one `collStats` per collection through the Agent's `runCommand` — reusing `collection_stats_command` (made public in #9893) and `object_statistics_from_collection_stats` (now public). A collection whose `collStats` fails is left out so the view keeps that row blank instead of failing the listing, as on the native path. The Agent connection is a single RPC channel, so the calls run sequentially rather than with the native path's bounded concurrency. `schema/mod.rs` dispatches MongoDB Agent pools to it, alongside the existing Oracle/Dameng/Kingbase/GBase Agent branches.

No Agent (Java) change and no new capability; an Agent without `runCommand` returns an empty list as before.

## Testing

- Live, opt-in: `collection_stats_and_create_database_run_over_the_legacy_agent` now also calls `list_object_statistics_core` against a throwaway `mongo:3.6` with the stock installed Agent: a populated collection (3 rows, non-zero bytes, `schema` = database), an empty collection (0 rows), and a view that must be absent. `system.views` is listed, as on the native path (no layer filters `system.*`). All three live tests in the file pass.
- `cargo test -p dbx-core --lib mongo_ops` (30) passes; `fmt`/`clippy` clean for the touched files.


t8y2 (approved):
Verified the head at 6ea2179676. The dispatch placement, capability gating, and fail-soft semantics match the native path and the Oracle/Dameng agent branches exactly, and the live-test assertions cover rows/size/schema plus the view skip. All checks pass at the exact head and there's no overlap with the recent main changes. Merging as-is.

One optional observation for a follow-up (not blocking): the sibling agent statistics branches pass `agent_metadata_timeout(db_config.as_ref())`, while this path relies on the per-RPC agent timeout — worth considering if a database with hundreds of collections ever makes the sequential `collStats` loop feel slow.

t8y2:
Thanks for the contribution! Merged in def42987a4, will be released in the next version.

PR #9944, opened 2026-09-22
fix(ci): invoke Win7 sccache stats on PowerShell

luojiyin1987:
## Summary

- Run sccache from PATH in the Windows 7 CI job.
- Add a workflow test for the command.

PowerShell cannot execute the previous environment variable syntax.
The stats step now completes and prints cache counters.

## Validation

- actionlint .github/workflows/ci.yml
- node --test .github/scripts/ci-workflow.test.mjs
- YAML parse check
- git diff --check


t8y2 (approved):
Thanks! Verified end to end: the old step failed pwsh parsing (`Unexpected token 'show-stats'`), and with this change the win7 job's "Show sccache stats" step now prints real counters (1041 compile requests, 55.15% hit rate). The fix also matches the sibling windows-standard-check job, and the new workflow-test assertion guards it. All checks pass, including windows-win7-bundle.

t8y2:
Thanks for the contribution! Merged in f8d7b57c19, will be released in the next version.

PR #9947, opened 2026-09-22
fix(mqtt): localize max packet size label in connection dialog

### LINKED ISSUE #9931: [Bug] MQTT connection settings option not translated

MQTT

One of the MQTT options text is not translated from chinese.

Image

DBX Version: v0.6.19
Runtime: Desktop
Operating System: macOS 13.7.8
Architecture: x86_64
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko)
Database Types: mysql

🟢 P3 低优 · 有空再修 / Low · Fix when possible

_No response_

Abeautifulsnow:
## Problem

Fixes: #9931 

The MQTT form in the New/Edit Connection dialog hardcoded its max packet size
label in Chinese (`最大报文（字节）`) directly in the template, so every locale
— English, Turkish, Azerbaijani, etc. — rendered the label in Chinese
regardless of the app language.

## Fix

- Bind the label to vue-i18n: `{{ t("connection.mqttMaxPacketSize") }}`
  (`ConnectionDialog.vue`).
- Declare `connection.mqttMaxPacketSize` in **all 10 locale files**, anchored
  next to the `connection.mqtt*` cluster (after `mqttConnectTimeout`):
  - en / zh-TW / ja / it / es / pt-BR: `Max Packet Size (bytes)`
    (zh-TW/ja/it/es/pt-BR keep English, matching their untranslated mqtt form
    cluster in-file)
  - zh-CN: `最大报文（字节）` (original text, now translatable)
  - tr: `Maksimum Paket Boyutu (bayt)`
  - az: `Maksimal paket ölçüsü (bayt)`
  - ko: `최대 패킷 크기(바이트)` — single-key precedent: ko owns none of the
    mqtt form cluster, and the spec pins that it must not grow it.

## Guardrails (`mqttConnectionLabelParity.spec.ts`)

New parity spec with two layers:

- **Source-level**: reads each locale file's source text and asserts the key is
  declared. Asserting on imported locale modules is vacuous here —
  `withEnglishFallback` deep-merges English under every locale at import time,
  so a dropped source key would be silently covered by the merge (same trap
  documented in `docsNamespaceParity.spec.ts`).
- **Runtime resolution**: per-locale `createI18n` must resolve the label to a
  non-empty string that is not the key path.
- Plus: cluster-anchoring checks, the ko exception pinned both ways, and a
  call-site regression assertion (the component must render via `t()` and must
  not contain `最大报文`).

## Verification

- `pnpm vitest run apps/desktop/src/i18n` — 24 files / 783 tests passed
  (new spec: 32 tests)
- `pnpm exec oxlint --vue-plugin`, targeted `oxfmt`, `vue-tsc --noEmit` — clean

## Commits

- `492e56239` fix(mqtt): localize max packet size label across all locales
- `a2a61303b` test(mqtt): resolve max packet size label via vue-i18n in parity spec


t8y2 (approved):
Reviewed — looks good. Verified `connection.mqttMaxPacketSize` is declared exactly once in all 10 locale files, anchored after `mqttConnectTimeout` in the nine locales that own the mqtt cluster, with the ko single-key exception pinned both ways. No hardcoded `最大报文` remains in `ConnectionDialog.vue`, and asserting on locale sources rather than imported modules correctly sidesteps the `withEnglishFallback` deep-merge trap (same rationale as `docsNamespaceParity.spec.ts`). One non-blocking nit: the spec comments reference `i18n-locales.md` twice, but no such file exists in the repo — consider pointing at the actual convention doc or dropping the reference.

t8y2:
Thanks for the contribution! Merged in b7152a29cd, will be released in the next version.

PR #7127, opened 2026-08-25
fix(grid): work around WKWebView stale canvas layer on right-click

klinge1011:
## 变更说明

修复 macOS（WKWebView）下查询结果表格右键行出现"蓝色全选遮罩"的问题：右键任一行后，整个数据区被选区蓝覆盖（被右键的一行为更深蓝），表头/工具栏正常；网页端与 Windows 版均无此现象。

  根因：WKWebView 会为 <canvas> 元素滞留一个幽灵合成层——屏幕显示该层旧的（全选）帧位图，而 canvas 2D 缓冲早已重绘为最新内容；整帧重绘、清屏甚至修改 canvas 尺寸都无法清除该幽灵层。仅 macOS 的 WKWebView
  触发（网页端 Chromium / Windows WebView2 均不出现），排查通过"移动 canvas 后遮罩跟随 / 隐藏 canvas 后遮罩消失 / 缓冲与屏幕色彩不一致"等对照实验逐步排除逻辑层与 DOM 层，最终定位到该复合层驻留行为。

  修复：双 canvas 翻转。网格数据面改为两枚同尺寸 canvas 交替绘制：每帧绘制进隐藏的一枚，画完立即交换可见性；幽灵层依附于旧元素、随 display:none
  一并废弃，新显示的画布无历史可滞留。命中测试、悬停光标、ResizeObserver 测量等交互路径同步切换到当前活动画布，滚动定位与绘制逻辑不变。

## 变更类型

- [ ] 新功能
- [X] Bug 修复
- [ ] 性能优化
- [ ] 代码重构
- [ ] 文档更新
- [ ] CI / 构建

## 涉及前端

- [X] 本 PR 涉及前端改动，已附截图/录屏（见下方）
修复前：
image
修复后：
PixPin_2026-08-25_12-01-01

## 验证

  手动验证（macOS WKWebView 实测）：
  1. 查询结果视图右键任一行 —— 无遮罩
  2. Ctrl+A 全选后右键某行 —— 无遮罩
  3. 滚轮上下/横向滚动 —— 行号与数据对位、无残影
  4. 滚动后点选/右键 —— 命中正确

- [X] `make check` 通过
- [ ] `make cargo-check-fast` 通过
- [X] 相关测试通过

## 关联 Issue

Close #7059

PR #9979, opened 2026-09-22
perf(grid): eliminate layout thrashing and surface flip jank during canvas scroll

zhangdengjie:
### Summary

Profiling data grid scrolling on macOS (WKWebView) revealed that 76.3% of main-thread time was spent in **Layout** (107~112ms per frame, resulting in ~8 FPS), while actual Canvas 2D cell drawing took only 4~9ms.

This PR fixes the scrolling jank by addressing the two layout triggers:

1. **Remove double-canvas `display: 'none'` surface flip**:
   Toggling CSS `display` between two sticky canvas elements on every frame forced WebKit to recalculate styles and re-layout the containing element with huge scroll extents (up to 1M+ px for large datasets). Replaced with single-canvas in-place drawing.

2. **Batch scrollbar layout queries into `requestAnimationFrame`**:
   In `onCanvasScroll`, synchronous reads of `scrollHeight`, `clientWidth`, and `offsetWidth` combined with custom scrollbar style updates created layout thrashing on every scroll event. These measurements and style applications are now deferred and coalesced via a single rAF frame (`flushCanvasScrollSync`), while immediate `scrollLeft` CSS updates and repaint scheduling remain instant.

### Verification

- All 63 test files and 735 tests pass (`npx vitest run`).
- TypeScript verification passes with zero errors (`vue-tsc --noEmit`).
- Verified real-world smooth scrolling in `make dev`.


t8y2 (changes requested):
I'm requesting changes because this removes the workaround introduced in #7127 for #7059. That fix was added after macOS WKWebView kept showing a stale canvas compositing layer after selection/right-click repaints; normal redraws and canvas resizing did not clear it, and the dual-surface flip was the mechanism that did.

Please preserve the dual-surface fallback outside the hot scrolling path—for example, flip only after selection/context-menu invalidations—or provide targeted macOS WKWebView validation against the original reproduction before removing it. The updated tests only check source structure and cannot catch this compositor regression.

PR #9983, opened 2026-09-22
fix(plugins): carry plugin file handle ids as uuid strings end to end

jinpy666:
## Problem

Plugin workbench fileTransfer uploads via the native bridge (file picker and OS drag-drop) fail almost always on desktop with `unknown plugin file handle`; the plugin then cancels the transfer as `local-read-error`. User reports: jinpy666/dbx-plugin-ssh#83, jinpy666/dbx-plugin-ssh#79.

Root cause: `plugin_file.rs` keyed handles by a `u64` cut from the first 8 uuid bytes. Serialized as a JSON number, that value exceeds `Number.MAX_SAFE_INTEGER` ~99.95% of the time (byte 6 of a uuid v4 is `0x4?`, and the leading bytes dominate the magnitude), so every JS layer rounded it into a double. The host then looked the handle up under a different key and rejected the read/write. The Rust registry was fine — the wire format was not.

## Fix

Handle ids are opaque tokens, not numbers. Align them with the existing `downloadId` convention (uuid string end to end):

- `plugin_file.rs`: registry keys `u64 → String`; ids generated as `Uuid::new_v4().to_string()`; `plugin_file_read/write/close` take `handle_id: String`; `PluginFileHandle.handle_id` documented as a JSON string, never a number.
- `tauri.ts` / `http.ts`: `PluginLocalFileHandle.handleId` and read/write/close signatures `number → string`.
- `PluginWorkbenchHost.vue`: removed the `Number()` conversions in `parseHandleId`; `openTauriHandles` is now `Set<string>`; comment states the rule (never run handle ids through `Number()`).
- Owner mismatches in `owned_handle` now `log::warn!` (external response unchanged — still indistinguishable from a missing handle to the plugin), so the two "unknown handle" cases can be told apart in host logs.

The web (`w<n>`) handle path was already string-based and is untouched. Plugins see only the opaque prefixed string, so no plugin-side change is required.

## Verification

- `cargo test --lib commands::plugin_file`: 8 passed, incl. a regression test asserting the wire id serializes as a JSON string, parses as a uuid, and the exact sent-back string hits the same registry key
- `cargo clippy --lib`: no new warnings
- `pnpm typecheck` and `pnpm vitest run` (PluginWorkbenchHost + pluginHostBridge specs, 57 tests): passed
- End-to-end on macOS: built the app from this branch, installed the SSH plugin, drag-drop upload that previously failed now completes; host log showed zero `unknown plugin file handle` events during the session

## Risks / Notes

- Wire-format change for a Rust-facing command pair; tolerated by design since the numeric path never worked (failure rate ≈99.98%) and the plugin contract already types `handleId` as a string.
- `git status` of this branch contains no lockfile or generated-file churn.


t8y2 (approved):
Excellent diagnosis and fix. Verified the precision math (u64 from uuid bytes exceeds 2^53 with ~99.95% probability), traced the full round-trip (registry key -> `t<uuid>` prefix -> `parseHandleId` -> Tauri command -> registry lookup), and confirmed the plugin-side contract was already string-typed, so the no-plugin-change claim holds. The regression test asserting string serialization + exact key round-trip is exactly right. One tiny nit for a future pass: the owner-mismatch `log::warn!` could include the offending handle id to aid debugging. All checks green. Approving.

t8y2:
Thanks for the contribution! Merged in 4f0f69131b, will be released in the next version.

PR #10023, opened 2026-09-22
fix(ai): sync chat panel header title with conversation rename (#9904)

### LINKED ISSUE #9904: [Bug] ai对话会话面板左上角的对话title与历史消息中面板的title不一致

通用

通过历史会话修改会话标题，但是面板左上角的信息没发生变化，这里理应发生改变，因为面板关闭的时候我需要看到当前会话的主题是什么

DBX 版本: v0.6.18
运行环境: 桌面端
操作系统: Windows 10.0.26200.9457
系统架构: x86_64
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/153.0.0.0 Safari/537.36 Edg/153.0.0.0
数据库类型: chromadb, clickhouse, dameng/dm, duckdb, elasticsearch, etcd, gaussdb, influxdb, kingbase, milvus, mongodb, mq/kafka, mq/pulsar, mysql, mysql/mariadb, nacos, neo4j, oracle, plugin, postgres, postgres/cockroachdb, qdrant, redis, saphana, sqlite, tdengine, trino, zookeeper
本地驱动版本: dameng 0.1.72, db2 0.1.64, duckdb 0.1.20, kingbase 0.1.58, neo4j 0.1.48, oracle 0.1.63, saphana 0.1.59, sqlserver-legacy 0.1.36, tdengine 0.1.49

_No response_

Abeautifulsnow:
## Summary

Fixes #9904 — renaming the current conversation from the AI chat history list left the panel's top-left title stale.

## Root cause

The panel header renders `{{ chatTitle }}`, a computed that derived its text **only** from the first user message of the panel-local `messages` snapshot:

```js
const first = messages.value.find((m) => m.role === "user" && m.kind !== "contextSummary");
return first ? messageTitle(first).slice(0, 30) : t("ai.newChat");
```

Renaming (`commitRenameConversation`) writes `renamedConversationTitles` and `conversations.value[i]` and persists — none of which `chatTitle` read, so the computed had no reactive dependency on the rename and never re-evaluated. The history row (which reads `conversations`) updated correctly, matching the report. Reopening the panel did not help either: `selectConversation` only refills `messages`.

## Changes

**`25bfa3afe` — header title follows the conversation title**
- `chatTitle` now uses the same precedence as `buildConversationSnapshot`: rename claim → stored title → first-message excerpt → `ai.newChat`.
- A failed rename save now retracts the claim (`renamedConversationTitles.delete`), so the header, the history row and the persisted record stay on the old title together.

**`83c2c6eb3` — recovered runs no longer overwrite a renamed title**
- Found during self-review: `persistPendingInputRecovery` persisted `messageTitle(first)` outright, so restarting with a recovered queued run overwrote a renamed title on disk, in the list, and therefore in the header. It now reuses the same precedence.

## Testing

- New `apps/desktop/src/components/editor/__tests__/AiAssistant.chatTitle.spec.ts` — 3 source-assertion cases pinning the computed precedence, the rename-commit ordering (claim set before the async save), the claim retract on a failed save, and the recovery path.
- `npx vitest run apps/desktop/src/components/editor/__tests__/` plus the 3 `aiAssistant*` app-tests → **46 files / 304 tests pass**.
- `oxlint` and `oxfmt --check` clean on both changed files.

## Notes

- Intentional, documented behaviour change: a saved chat's header now follows the stored title exactly like the history row, so editing the first message no longer retitles the header. This is the semantics the issue asks for ("I need to see what the current session's topic is").
- Out of scope / pre-existing: `renamedConversationTitles` entries are never evicted when a conversation is deleted.


t8y2 (commented):
Maintainer conflict resolution pushed: 83c2c6eb → 6d741eac (merge commit on top of main e1fc28ab8; no history rewritten).

Conflict cause: main's 1566bb8a9 (plugin AI panel support) added `conversation.pluginContext?.title ||` precedence plus a `pluginContext` field to `persistPendingInputRecovery` after this branch diverged (merge base 9bd9b9326), colliding with this PR's rename-claim title chain in the same function.

Resolution (both behaviors preserved): the title chain is now the union —
`conversation.pluginContext?.title || renamedConversationTitles.get(conversation.id) || conversation.title || (first ? messageTitle(first).slice(0, 50) : "Untitled")`
— so plugin-scoped chats keep the plugin-provided title (main's intent) while a recovered run still never overwrites a user-renamed or stored title (this PR's intent). Main's `pluginContext: conversation.pluginContext,` field line is kept. The `chatTitle` computed and `buildConversationSnapshot` hunks auto-merged cleanly (plugin fallback intact as last resort).

Checks run at 6d741eac: `vitest run` on `AiAssistant.chatTitle.spec.ts` + `AiAssistant.scrollRing.spec.ts` + `AiAssistant.templateDefaults.spec.ts` — 15/15 passed. `git diff --check` clean; resolved-tree diff vs main is exactly this PR's original 2-file scope (+91/−1). Backup ref of the original head kept locally.

One pre-existing, non-blocking tension (unchanged by this merge): header precedence (rename claim → stored → plugin) is the inverse of `buildConversationSnapshot`'s plugin-first order for plugin-scoped chats — only observable if a user renames a plugin conversation, a case already divergent on main. Happy to align in a follow-up.

t8y2:
Thanks for the contribution! Merged in 6a417b9c90, will be released in the next version.

